PR #8, opened 2022-09-08
Bug 1786777 - Fix level and repoFullName for cron tasks

gbrownmozilla:


gbrownmozilla:
It looks to me like the Decision task completed normally here -- github just doesn't realize that?

.taskcluster.yml
```diff
@@ -29,9 +30,13 @@ tasks:
...
                   then: '${event.pull_request.base.repo.full_name}'
                   else:
-                      $if: 'tasks_for in ["cron", "action"]'
+                      $if: 'tasks_for == "action"'
                       # Trim git@github.com:
                       then: '${repository.url[15:-4]}'
+                      else:
+                          $if: 'tasks_for == "cron"'
+                          # Trim https://github.com/
+                          then: '${repository.url[19:]}'
```
escapewindow, line 39: We use this type of construct all the time.
I wonder if it would be good to propose a `urlparse` or similar in json-e, so we might be able to use `urlparse(${repository.url})["path"]` or similar.

.taskcluster.yml
```diff
@@ -7,6 +7,7 @@ tasks:
     - $let:
           trustDomain: mobile
           canonicalRepo: git@github.com:mozilla-mobile/firefox-android.git
+          canonicalHttpsRepo: https://github.com/mozilla-mobile/firefox-android
```
JohanLorenzo, line 10: Follow-up: Damn! I'm sad we now have to support both SSH and HTTP urls in `.taskcluster.yml`. It seems We're still getting SSH urls from the Github Webhook. I'd like switch back to HTTP, just like the rest of our Github repos. I didn't see any settings I can use to make the switch.

@csadilek, do you have the rights to revoke the firefoxci-taskcluster Github App[1] then to install it again? I wonder if that resets the URL to HTTP now that the repo is public. 🤔 

[1] https://github.com/mozilla-mobile/firefox-android/settings/installations
[2] https://github.com/apps/firefoxci-taskcluster

JohanLorenzo (approved):
Agreed to merge this PR to get cron jobs working. I left a note about using HTTP everywhere but I'm certain this note doesn't block this PR. Let's go 👍

.taskcluster.yml
JohanLorenzo: I agree this is a great proposal. I'd propose we formalize the API of this `urlparse()` function in https://github.com/mozilla-releng/mozilla-repo-urls, first. Once we're happy about the data it returns, then we should be in a good position to make this behavior part of json-e[1]

[1] https://github.com/json-e/json-e/

.taskcluster.yml
gbrownmozilla: +1 : It would be tidier and less error prone if we used HTTP urls everywhere.

csadilek: @JohanLorenzo Yes, I do. I can grant you that permission as well, but otherwise, happy to try.

PR #9, opened 2022-09-13
sync-ac: Use full URLs in filtered commit messages 

JohanLorenzo:
After discussing with @csadilek, we still want the imported commits to reference existing pull requests or issues. For instance https://github.com/mozilla-mobile/android-components/commit/faa52da0ff0a71d0031acb9e894bec177c171edb should still point to https://github.com/mozilla-mobile/android-components/issues/12766. More precisely, we want full URLs to be injected the commit messages. This way, we don't lose the link if we ever wanted to migrate the monorepo elsewhere later. 

There's one thing: a `#number` may either reference an issue or a pull request, meaning we have to support 2 types of URLs[1]. I changed the logic to ask Github the list of all PRs and all issues so that we know which URL to use. Then, the regexes are automatically now generated. This makes the regexes simpler and more effective than the previous implementation. 

I scanned the generated commit messages and saw very few places where a link couldn't be generated. I'll provide the list of known exception in the next comment.

[1] As a matter of fact `/issues` URLs automatically redirect to `/pull` on Github. That said, we want the commit messages not to surprise readers in case they were expecting an issue and end up with a PR instead.


JohanLorenzo:
Known exceptions:
 * On https://github.com/mozilla-mobile/android-components/commit/043db4c1cafeebf2c5bd5a61f606e1f9b28aede4 and https://github.com/mozilla-mobile/android-components/commit/8d02e5944fe0f0b5edc49308515b3954474be192, the number actually points to a Fenix PR. We have no way of programmatically know that. This message was busted when it landed anyway. I don't think it's worth fixing it.
  * https://github.com/mozilla-mobile/android-components/commit/c661f572d902803f994abf282c33451e69c9d2f0 and https://github.com/mozilla-mobile/android-components/commit/7aebe6205da1683db7fab76d26c6508fcd3e38b6 have 2 `#` instead of one. The URL could be generated although an extra `#` remains. Not worth fixing either, IMO.
  * https://github.com/mozilla-mobile/android-components/commit/fc71520bb0ad4d4bce66ed591d901ff966d2ab8e got its issue moved to https://github.com/mozilla-mobile/fenix/issues/10246. 

These are the only exceptions I found on 5 years of commits 🙂

csadilek (approved):
Wow, nice work! Tested and works great!

gbrownmozilla (approved):
lgtm!

PR #10, opened 2022-09-15
Stop cloning repo through SSH

JohanLorenzo:
Now that the repo is public, we don't have to deal with SSH anymore and we can resume using HTTP. This PR just reverts a couple commits we made last month.

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -6,8 +6,7 @@ policy:
 tasks:
     - $let:
           trustDomain: mobile
-          canonicalRepo: git@github.com:mozilla-mobile/firefox-android.git
-          canonicalHttpsRepo: https://github.com/mozilla-mobile/firefox-android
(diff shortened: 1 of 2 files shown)


.taskcluster.yml
```diff
@@ -277,14 +274,9 @@ tasks:
                                       $if: 'tasks_for == "action"'
                                       then: >
                                           cd /builds/worker/checkouts/vcs &&
-                                          git submodule init &&
-                                          git submodule update &&
                                           ln -s /builds/worker/artifacts artifacts &&
                                           ~/.local/bin/taskgraph action-callback
                                       else: >
-                                          cd /builds/worker/checkouts/vcs &&
```
gbrownmozilla, line 285: Are you sure this isn't needed?

gbrownmozilla (approved):

.taskcluster.yml
JohanLorenzo: Oops, good catch! 👍  Fixed.

PR #12, opened 2022-09-15
Bug 1791084 - Add "AC-" to treeherder platform names

gbrownmozilla:
Here's the simplest approach: Just add AC- to the existing treeherder "platform" names. I'm thinking that keeps things looking very similar to the way they do today for the android-components repo. When the other projects are added we can have "Focus-ui-tests" and "Fenix-ui-tests", etc, keeping each project's tests grouped on one line ("platform").

Since these are all in task defaults, the tricky part will come when incorporating focus and fenix. I don't have a clear vision for how that will work out, but I'm sure we can sort something out. Keep it simple for now?

I don't have strong feelings here, as long as we don't end up with a mish-mash of a-c/focus/fenix tasks on treeherder -- happy to hear other suggestions.

### DIFF
--- a/taskcluster/ci/build-samples-browser/kind.yml
+++ b/taskcluster/ci/build-samples-browser/kind.yml
@@ -33,7 +33,7 @@ task-defaults:
     run-on-tasks-for: [github-pull-request, github-push]
     treeherder:
         kind: build
-        platform: android-all/opt
+        platform: AC-android-all/opt
         tier: 1
     worker-type: b-android
     worker:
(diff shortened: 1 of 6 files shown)


JohanLorenzo (approved):
I'm not sure what direction to follow too. For sure, we need a way to differentiate AC/Focus/Fenix jobs. Using the platform sounds like a good idea. Plus, the changes are not too complex for now. Agreed to land this as is and iterate once people voice their opinion later on 👍

PR #14, opened 2022-09-15
Bug 1788606 - Run gradle just once to determine component dependencies

gbrownmozilla:
Instead of running gradle once for each configuration, run just once for all configurations, then filter out data for all the other configurations.

In local tests this only resulted in a slight (5%?) performance improvement.

### DIFF
--- a/taskcluster/ac_taskgraph/loader/build_config.py
+++ b/taskcluster/ac_taskgraph/loader/build_config.py
@@ -46,27 +46,32 @@ def get_upstream_deps_for_components(components):
     """Return the full list of local upstream dependencies of a component."""
     component_dependencies = defaultdict(set)
 
-    for configuration in CONFIGURATIONS_WITH_DEPENDENCIES:
-        logger.info("Looking for dependencies in '%s' configuration" % configuration)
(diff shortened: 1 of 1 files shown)


gbrownmozilla:
In https://firefox-ci-tc.services.mozilla.com/tasks/J-APWIPzS7GsFaV11j9Lag/runs/0/logs/public/logs/live.log, this still took over 4 minutes -- no better than the original code.
```
[task 2022-09-15T23:11:07.864Z] 2022-09-15 23:11:07,864 - INFO - Generated 2 tasks for kind toolchain
[task 2022-09-15T23:15:20.926Z] 2022-09-15 23:15:20,926 - INFO - Found direct upstream dependencies for component 'browser-domains': ['support-test']
```
I'm requesting review for your consideration, but I am perfectly happy to discard this PR.

csadilek (commented):
Thanks for giving this a try! If we're not better off with this change after all, I would vote for skipping it. Then we don't need to retest that we're detecting the correct components to build.

gbrownmozilla:
Yeah, I think that's the sensible thing to do. Closing without merging.

PR #19, opened 2022-10-13
Bug 1786780 - Migrate release android-components jobs

JohanLorenzo:
Tested on staging at https://firefox-ci-tc.services.mozilla.com/tasks/groups/T1cHe7oqROKKeKlhJbq3jg

### DIFF
--- a/taskcluster/ci/build/kind.yml
+++ b/taskcluster/ci/build/kind.yml
@@ -31,15 +31,15 @@ task-defaults:
     artifact-template:
         type: file
         name: public/build/{artifact_file_name}
-        path: '/builds/worker/checkouts/vcs/{component_path}/build/maven/org/mozilla/components/{component}/{version}/{artifact_file_name}'
+        path: '/builds/worker/checkouts/vcs/android-components/{component_path}/build/maven/org/mozilla/components/{component}/{version}/{artifact_file_name}'
     tests-artifact-template:
         type: directory
         name: public/reports/tests
-        path: '/builds/worker/checkouts/vcs/{component_path}/build/reports/tests'
+        path: '/builds/worker/checkouts/vcs/android-components/{component_path}/build/reports/tests'
     lint-artifact-template:
         type: file
         name: public/reports/lint-results-release.html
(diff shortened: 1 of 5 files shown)


taskcluster/ci/build/kind.yml
```diff
@@ -31,23 +31,23 @@ task-defaults:
     artifact-template:
         type: file
         name: public/build/{artifact_file_name}
-        path: '/builds/worker/checkouts/vcs/{component_path}/build/maven/org/mozilla/components/{component}/{version}/{artifact_file_name}'
+        path: '/builds/worker/checkouts/vcs/android-components/{component_path}/build/maven/org/mozilla/components/{component}/{version}/{artifact_file_name}'
```
gbrownmozilla, line 34: I'm not sure I understand how these paths are used. What do we expect for {component_path}? How will we allow for the addition of fenix/focus?

gbrownmozilla (approved):

taskcluster/ci/build/kind.yml
JohanLorenzo: These paths get transformed[1] into a regular `payload.artifacts` definition which is understood by docker worker. It enables docker-worker to publish artifacts which are then used by signing tasks. Here's an example of a generated task[2].

`{component_path}` is due to the way gradle generates builds. Gradle outputs them in the folder of the component. That value is evaluated by the aforementioned transform.

I confess I haven't thought of fenix and focus builds yet. That's a great question! Now that I think of it, I believe focus and fenix should be in a separate taskgraph `kind.yml`. The main driver is: we want to build apps after components are available on `maven.mozilla.org`. Thus, fenix and focus can't belong to the same `kind.yml`. In order words, the path logic can be totally different in fenix and focus since they don't need to use the same transform.

Let me know if you'd like more details. I'm happy to provide them 🙂 

[1] https://github.com/mozilla-mobile/firefox-android/blob/0de9cea3e72def77920bf254c9d31d06c6db7252/taskcluster/ac_taskgraph/transforms/build.py#L139-L141
[2] https://firefox-ci-tc.services.mozilla.com/tasks/bShC3qhKSjW0Grk9vgM_ug

PR #22, opened 2022-10-17
Bug 1789931 - part 5: Fix l10n.toml location in `sync-ac.sh`

JohanLorenzo:
This fixes the regression I introduced in https://github.com/mozilla-mobile/firefox-android/commit/b1a420d45690890257a87320cebe44f2bae9ac96. This file should be under the `android-components` folder but this folder doesn't (and mustn't) exist until the repo is migrated over. This patch modifies the script to move this file to the right folder.

### DIFF
--- a/monorepo-migration/sync-ac.sh
+++ b/monorepo-migration/sync-ac.sh
@@ -85,7 +85,7 @@ function _merge_histories() {
 function _move_files_into_subfolder() {
     cd "$CURRENT_REPO_PATH"
     git mv --force 'l10n.toml' "$REPO_NAME_TO_SYNC/l10n.toml"
-    git commit --message 'Fix l10n.toml file location'
+    git commit --message 'Move l10n.toml into android-components subfolder'
 }
 
 function _clean_up_temporary_repo() {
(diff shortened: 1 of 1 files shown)


monorepo-migration/sync-ac.sh
```diff
@@ -77,12 +78,18 @@ function _remove_old_tags() {
...
     cd "$SCRIPT_DIR"
-    git pull --no-edit --tags --allow-unrelated-histories --no-rebase --force "$REPO_PATH"
+    git pull --no-edit --tags --allow-unrelated-histories --no-rebase --force "$TMP_REPO_PATH"
     git commit --amend --message "$MERGE_COMMIT_MESSAGE"
 }
 
+function _move_files_into_subfolder() {
+    cd "$CURRENT_REPO_PATH"
+    git mv --force 'l10n.toml' "$REPO_NAME_TO_SYNC/l10n.toml"
+    git commit --message 'Fix l10n.toml file location'
```
gabrielluong, line 88:
```suggestion
    git commit --message 'Move l10n.toml into android-components subfolder'
```
Maybe? I feel like we could appreciate the added clarity given our general concerns about the growing complexity of this script.

gabrielluong (approved):

monorepo-migration/sync-ac.sh
JohanLorenzo: Good call! Suggestion applied 🙂

PR #35, opened 2022-10-31
Post migration cleanup

csadilek:
- .github is only needed in the root directory
- .cron.yml, .mergify.yml, .cron.yml  are also used from root only
- LICENCE is alreay present in root
- CODE_OF_CONDUCT is moved to root

`.git-blame-ignore-revs` will be addressed separately.


JohanLorenzo (approved):
LGTM! 👍

PR #38, opened 2022-10-31
Bug 1796482 - Allow for notifying consumers on unsuccessful Save to PDF requests

Amejia481:
Fenix PR https://github.com/mozilla-mobile/fenix/pull/27525

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
+++ b/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
@@ -1017,6 +1017,14 @@ sealed class EngineAction : BrowserAction() {
     ) : EngineAction(), ActionWithTab
 
     /**
+     * Indicates the given [tabId] was unable to generate a requested save to PDF page.
+     */
+    data class SaveToPdfExceptionAction(
+        override val tabId: String,
+        val throwable: Throwable,
+    ) : EngineAction(), ActionWithTab
+
+    /**
      * Clears browsing data for the tab with the given [tabId].
      */
     data class ClearDataAction(
--- a/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
+++ b/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
@@ -45,6 +45,11 @@ internal object EngineStateReducer {
         is EngineAction.OptimizedLoadUrlTriggeredAction -> {
             state
         }
+        is EngineAction.SaveToPdfExceptionAction -> {
+            throw IllegalStateException(
+                "You need to add a middleware to handle this action in your BrowserStore. ($action)",
+            )
+        }
         is EngineAction.SuspendEngineSessionAction,
         is EngineAction.CreateEngineSessionAction,
         is EngineAction.LoadDataAction,
--- a/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -262,6 +262,9 @@ class GeckoEngineSession(
             { throwable ->
                 // Log the error. There is nothing we can do otherwise.
                 logger.error("Save to PDF failed.", throwable)
+                notifyObservers {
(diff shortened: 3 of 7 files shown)


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -1016,6 +1016,14 @@ sealed class EngineAction : BrowserAction() {
         override val tabId: String,
     ) : EngineAction(), ActionWithTab
 
+    /**
+     * Indicates the given [tabId] was unable to generate a requested save to PDF page.
+     */
+    data class SaveToPdfErrorAction(
```
jonalmeida, line 1022: Could we rename this? Actions are usually some form of directive, past or present, (e.g. "toggle desktop mode", "exit fullscreen mode"), but here we are saying "save to pdf error" as a statement. If we use `SaveToPdfException`, it works as a past version of a directive "a save to pdf threw an exception". If we stick with `Error` it should really be `Errored` which is an odd past tense word to use, hence why `Exception` might be more fitting.

If you have other ideas too, that could work as well!

```suggestion
    data class SaveToPdfExceptionAction(
```

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
```diff
@@ -45,6 +45,7 @@ internal object EngineStateReducer {
         is EngineAction.OptimizedLoadUrlTriggeredAction -> {
             state
         }
+        is EngineAction.SaveToPdfErrorAction,
```
jonalmeida, line 48: Adding it to this list means that it could throw the error saying, "You need to add _EngineMiddleware_ to your BrowserStore." which isn't true for this action.

I think we ought to be using our own exception here instead:

```suggestion
        is EngineAction.SaveToPdfErrorAction -> {
            throw IllegalStateException("You need to add a middleware to handle this action in your BrowserStore. ($action)")
        }
```

android-components/docs/changelog.md
```diff
@@ -10,6 +10,9 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/android-components/blob/main/buildSrc/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/android-components/blob/main/.config.yml)
 
+* **browser-engine-gecko**
+  * 🆕 A new action `SaveToPdfErrorAction` was added to the `EngineAction` to allow for notifying consumers on unsuccessful Save to PDF requests. For more references see [bug #1796482](https://bugzilla.mozilla.org/show_bug.cgi?id=1796482).
```
jonalmeida, line 14: See comment above.

```suggestion
  * 🆕 A new action `SaveToPdfExceptionAction` was added to the `EngineAction` to allow for notifying consumers on unsuccessful Save to PDF requests. For more references see [bug #1796482](https://bugzilla.mozilla.org/show_bug.cgi?id=1796482).
```

jonalmeida (commented):
Thanks, I left two comments that I think we should address if we're going down this path.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/EngineStateReducer.kt
Amejia481: 👍🏽

Amejia481:
Thanks for the suggestions, I updated the PR with the comments :)

jonalmeida (approved):
Looks good!

When this lands, it'll throw an exception if we don't have the middleware to handle the error.

Amejia481:
Perfect, we will need to coordinate between landing this change and the Fenix one!

csadilek:
@Amejia481 just a heads-up that the source branch of this PR is not on your fork, but on the "upstream" directly. Doesn't matter too much (as long as the branch is deleted again), but likely unintentional.

Amejia481:
Thanks, I'll do!

gabrielluong:
Hello, we recently ran into a hiccup during the monorepo migration where our CI tasks weren't checking the right files and running the appropriate build and test runs for any of the changed files and components. Unfortunately, this means your PR will need to rebase in order for CI to run appropriately to ensure all the appropriate checks are passing prior to landing. Sorry for the inconvenience!

PR #59, opened 2022-11-01
Bug 1795903 - Introduce StorageMaintenanceWorker in A-C that runs periodically to prune places storage.

kycn:
WorkManager's Worker is created for periodically running storage maintenance operation on places database. DB size limit is exposed from A-S that indicates the maximum DB size to aim for, in bytes. If the database exceeds this size, a small number of visits will be pruned. This operation is done periodically by a periodic work request that, optimally, will run once a day as long as the worker constraints are met.

### How to test
WorkManager's periodic worker is running in A-C. Testing the feature is a bit tricky: see [pruning logic in A-S](https://mozilla.github.io/application-services/book/rust-docs/src/places/storage/history.rs.html#415-428)
As seen from the code snippet, we are pruning the elements starting from earliest 7 days ago.

Worker initialization is not enabled in SampleApplication in this PR as I was not sure whether it would be feasible to have this feature on the SampleApplication or not.
To test this feature on SampleApplication please follow the instructions:

1. Add `GlobalPlacesDependencyProvider.initialize(components.historyStorage)` and  `components.historyStorage.registerStorageMaintenanceWorker()` one after another in `SampleApplication`'s `onCreate` method. 
2. You can change the `dBSizeLimit` locally to set a smaller size in order to test in case your DB (`places.sqlite`) size is not as big as `75MiB`. In order to meet 7-days condition in the pruning process, you can change the date/time of your emulator/device (go back in time for at least 7 days), populate your db and then rollback to the current time.
3. For triggering the worker periodically, you can change the date/time of your emulator/device to a future date.

_**Note:** WorkManager does not ensure that the Worker will run exactly at the same time after the period delay passes, so the periodic waiting times may change while testing because of the work request constraints and OS operations. In addition, one of the work request's constraints is `setRequiresDeviceIdle(true)` and it is hard to meet immediately. Tried forcing the idle mode but could not trigger the worker right away. (My guess is the retry time intervals that are bound to `BackoffPolicy` which is, by default, `EXPONENTIAL`. Though, it is not quite clear from the docs whether OS retries the work request to run right after the constraints are met or it aligns with the `BackoffPolicy`.) I suspect that if the user is using the device actively, meeting the same 24h execution window for the work request is going be very hard._ 

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt
+++ b/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt
@@ -6,7 +6,10 @@ package mozilla.components.browser.storage.sync
 
 import android.content.Context
 import android.net.Uri
+import android.os.Build
 import androidx.annotation.VisibleForTesting
+import androidx.work.ExistingPeriodicWorkPolicy
+import androidx.work.WorkManager
 import kotlinx.coroutines.withContext
 import mozilla.appservices.places.PlacesApi
 import mozilla.appservices.places.PlacesReaderConnection
@@ -43,7 +46,7 @@ const val AUTOCOMPLETE_SOURCE_NAME = "placesHistory"
  */
 @Suppress("TooManyFunctions", "LargeClass")
 open class PlacesHistoryStorage(
-    context: Context,
+    private val context: Context,
     crashReporter: CrashReporting? = null,
 ) : PlacesStorage(context, crashReporter), HistoryStorage, HistoryMetadataStorage, SyncableStore {
     /**
@@ -245,6 +248,27 @@ open class PlacesHistoryStorage(
                 places.writer().pruneDestructively()
             }
         }
+    }
+
+    /**
+     * Enqueues a periodic storage maintenance worker to WorkManager that prunes database entries
+     * when it exceeds [PlacesHistoryStorageWorker.DB_SIZE_LIMIT_IN_BYTES].
+     */
+    override fun registerStorageMaintenanceWorker() {
+        WorkManager.getInstance(context).enqueueUniquePeriodicWork(
+            PlacesHistoryStorageWorker.UNIQUE_NAME,
+            ExistingPeriodicWorkPolicy.KEEP,
+            periodicStorageWorkRequest<PlacesHistoryStorageWorker>(
+                tag = PlacesHistoryStorageWorker.UNIQUE_NAME,
+            ) {
+                constraints {
+                    setRequiresBatteryNotLow(true)
+                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+                        setRequiresDeviceIdle(true)
+                    }
+                }
+            },
+        )
     }
 
     /**
--- a/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
+++ b/android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
@@ -16,6 +16,7 @@ import mozilla.appservices.places.PlacesWriterConnection
 import mozilla.appservices.places.uniffi.PlacesException
 import mozilla.components.concept.base.crash.CrashReporting
 import mozilla.components.concept.storage.Storage
+import mozilla.components.concept.storage.StorageMaintenanceRegistry
 import mozilla.components.concept.sync.SyncStatus
 import mozilla.components.concept.sync.SyncableStore
 import mozilla.components.support.base.log.logger.Logger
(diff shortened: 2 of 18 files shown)


android-components/components/concept/storage/src/main/java/mozilla/components/concept/storage/StorageMaintenanceRegistry.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Unregisters the storage maintenance worker that is registered
+     * by [StorageMaintenanceRegistry.registerStorageMaintenanceWorker].
+     * See also [Storage.runMaintenance].
+     *
+     * @param uniqueWorkName Unique name of the work request that needs to be unregistered.
+     * */
+    fun unregisterStorageMaintenanceWorker(uniqueWorkName: String)
+}
```
kycn, line 27: Unregister is added intentionally although there is no "unregistering" (which may refer to cancelling the existing workers via using their unique names) logic existing for the workers in the code for now.

android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesStorage.kt
```diff
@@ -183,4 +187,21 @@ abstract class PlacesStorage(
...
+
+    /**
+     * Unregisters the storage maintenance worker that is registered
+     * by [PlacesStorage.registerStorageMaintenanceWorker].
+     *
+     * @param uniqueWorkName Unique name of the work request that needs to be unregistered
+     * */
+    override fun unregisterStorageMaintenanceWorker(uniqueWorkName: String) {
+        // See child classes for implementation details, it is not implemented by default
+    }
```
kycn, line 206: Why i did do this way? 
To ensure that all the children do not need to know about the interface (compilation safety) and make it flexible for any usecase which may include either, both or none of the implementations of these two methods. 
Ex: We have maintenance task only for `PlacesHistoryStorage` and we only have the registering logic. Therefore, only `registerStorageMaintenanceWorker` method is implemented in `PlacesHistoryStorage` class. 
Future Ref: In case `RemoteTabsStorage` is required to have a maintenance task, then it would only require to implement `StorageMaintenanceRegistry` and provide custom implementations for these methods.

gabrielluong:
Hello, we recently ran into a hiccup during the monorepo migration where our CI tasks weren't checking the right files and running the appropriate build and test runs for any of the changed files and components. Unfortunately, this means your PR will need to rebase in order for CI to run appropriately to ensure all the appropriate checks are passing prior to landing. Sorry for the inconvenience!

android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt
```diff
@@ -247,6 +250,27 @@ open class PlacesHistoryStorage(
         }
     }
 
+    /**
+     * Enqueues a periodic storage maintenance worker to WorkManager that prunes database entries
+     * when it exceeds [PlacesHistoryStorageWorker.DB_SIZE_LIMIT_IN_BYTES].
+     */
+    override fun registerStorageMaintenanceWorker() {
+        WorkManager.getInstance(context).enqueueUniquePeriodicWork(
+            PlacesHistoryStorageWorker.UNIQUE_NAME,
```
jonalmeida, line 259: Self-note: we're using the same `PlacesHistoryStorageWorker.UNIQUE_NAME` here and for the work request tag. This is okay because we only have one unique work request so using the same tag has no unintended side-effects if we were to query for work by either.

android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/RemoteTabsStorage.kt
```diff
@@ -83,8 +83,8 @@ open class RemoteTabsStorage(
         }
     }
 
-    override suspend fun runMaintenance() {
-        // There's no such thing as maintenance for remote tabs, as it is a in-memory store.
+    override suspend fun runMaintenance(dbSizeLimit: UInt) {
+        // Storage maintenance workflow for remote tabs is not implemented yet.
```
jonalmeida, line 87: Comment updating. ❤️

jonalmeida (approved):
This looks great, I think you can land this when ready. 🙂 

> Worker initialization is not enabled in SampleApplication in this PR as I was not sure whether it would be feasible to have this feature on the SampleApplication or not.

This part of the comment sparked a thought I should have had earlier - we should enable this call from happening in the `sample-browser` as well. It's a good place to have code run in that too. The only value is for developer ergonomics; having a sample browser that works similar to how Fenix/Focus would.

Before landing, it would be good to add a line to the [`changelog.md`](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) around the new `runMaintenance` parameter and `StorageMaintenanceRegistry` APIs. Just a short one-line for each should suffice. 👍 

:shipit:

android-components/components/browser/storage-sync/src/main/java/mozilla/components/browser/storage/sync/PlacesHistoryStorage.kt
kycn: Yes, i had exactly the same note to myself while coding :)

PR #82, opened 2022-11-02
Bug 1795105: Don't store thumbnails in BrowserState

jonalmeida:
Prior to this patch, we would fetch a HQ thumbnail of the visible viewport for each session in the BrowserStore when the page was loaded. This system worked well at the time because we stored only one thumbnail for each session and removed the thumbnails for memory when received a low memory callback from the system.

We have learnt recently that these callbacks happen far too late for us to react to them; at that point we are already going to be killed by the OS. Also, since these thumbnails are HQ bitmaps, they can take a significant amount of space which we rarely need at that resolution to always be in memory.

In our use cases of thumbnails, we currently use them for two major features: tabs tray images and gesture tab switching. Thumbnails only use a partial amount of the thumbnail that is visible in the tabs tray,. The gesture tab switching feature only has to load one thumbnail at a time when using that feature and has already been retrieving the thumbnails straight from disk, which makes it less of a performance hit on IO.

For these reasons, we are removing our Bitmap instances from living in memory so that middlewares can observe the action, store the thumbnails and retrieve them from that storage as well.

Will update the changelog before landing.

cc: @kycn @ohall-m who might be interested in reviewing this patch too.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


csadilek (approved):
One nit to address before landing, but looks great to me!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
```diff
@@ -30,7 +30,7 @@ internal object ContentStateReducer {
                 it.copy(icon = null)
             }
             is ContentAction.RemoveThumbnailAction -> updateContentState(state, action.sessionId) {
```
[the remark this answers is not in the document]
jonalmeida: Thanks, that's a good catch! Fixed.

jonalmeida:
Update with the review comments and a changelog entry.

android-components/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabViewHolder.kt
```diff
@@ -108,14 +108,12 @@ class DefaultTabViewHolder(
...
         // In the final else case, we have no cache or fresh screenshot; do nothing instead of clearing the image.
-        if (thumbnailLoader != null && tab.content.thumbnail == null) {
+        if (thumbnailLoader != null) {
             val thumbnailSize = THUMBNAIL_SIZE.dpToPx(thumbnailView.context.resources.displayMetrics)
             thumbnailLoader.loadIntoView(
                 thumbnailView,
                 ImageLoadRequest(id = tab.id, size = thumbnailSize),
             )
-        } else if (tab.content.thumbnail != null) {
-            thumbnailView.setImageBitmap(tab.content.thumbnail)
```
kycn, line 118: This check corresponds to the case when `thumbnailLoader` is `null`. What is the UI/UX state there when this case happens if we remove this check and do nothing for `(thumbnailLoader == null)` case?

kycn (approved):
Great! Thanks for notifying me as well. Some parts of this PR are actually interrelated with the memory profiling task that I am focusing nowadays. So it double helped :)) I only had one question about a use case while reviewing, you can see that above.

android-components/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabViewHolder.kt
jonalmeida: Good question! In this case, the `thumbnailLoader` wouldn't do anything and the tabs tray item would render nothing just a square. UX-wise this isn't great, however, the `TabViewHolder` implementation is no longer used in anywhere exception the sample-browser and reference-browser.

Fenix uses a custom implementation of the `TabViewHolder` that fallsback to rendering the favicon if a thumbnail is unavailable.

Once Fenix moves to a Jetpack Compose tabs tray and no longer depends on the API hierarchy of the `TabViewHolder` we can remove this code/component entirely and the sample/reference browsers can also move to a Compose UI. There were some thoughts about unifying the implementations under a common tabs tray, but with Compose, that seems to be a little less valuable.

You can read a bit about this under the Compose section of the Android Components blog post: https://mozac.org/2021/07/05/whats-next.html

kycn: Got it! Thanks for the details and the link. I will be digging into that. 😄

PR #92, opened 2022-11-03
Bug 1799230 - Use Non-Transitive R Classes

fice-t:
According to [Android documentation](https://developer.android.com/studio/build/optimize-your-build.html#use-non-transitive-r-classes):

>You should use non-transitive [R classes](https://developer.android.com/reference/android/R) to have faster builds for applications with multiple modules. Doing so helps prevent resource duplication by ensuring that each module's R class only contains references to its own resources, without pulling references from its dependencies. This leads to faster builds and the corresponding benefits of compilation avoidance.

Full clean build before (`./gradlew --profile --rerun-tasks assembleDebug`): ~14m 50s
Full clean build after (`./gradlew --profile --rerun-tasks assembleDebug`): ~12m 30s

I didn't test incremental compilation, but it should show stronger results on changes to modules that have resources that aren't used by (many) other modules.

This also appears to decrease the apk size of Fenix by ~1MB.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/android-components/components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuHighlightTest.kt
+++ b/android-components/components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuHighlightTest.kt
@@ -13,6 +13,7 @@ import mozilla.components.support.test.robolectric.testContext
 import org.junit.Assert.assertEquals
 import org.junit.Test
 import org.junit.runner.RunWith
+import mozilla.components.ui.colors.R as colorsR
 
 @RunWith(AndroidJUnit4::class)
 class BrowserMenuHighlightTest {
@@ -36,12 +37,13 @@ class BrowserMenuHighlightTest {
     @Suppress("Deprecation")
     @Test
     fun `classic highlight effect converts background tint`() {
+        val colorId = colorsR.color.photonRed50
         val highlight = BrowserMenuHighlight.ClassicHighlight(
             startImageResource = 0,
             endImageResource = 0,
             backgroundResource = 0,
-            colorResource = R.color.photonRed50,
+            colorResource = colorId,
         )
-        assertEquals(HighPriorityHighlightEffect(testContext.getColor(R.color.photonRed50)), highlight.asEffect(testContext))
+        assertEquals(HighPriorityHighlightEffect(testContext.getColor(colorId)), highlight.asEffect(testContext))
     }
 }
--- a/android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonsPermissionsAdapterTest.kt
+++ b/android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonsPermissionsAdapterTest.kt
@@ -9,12 +9,12 @@ import android.widget.TextView
 import androidx.core.content.ContextCompat
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import mozilla.components.feature.addons.R
 import mozilla.components.feature.addons.ui.AddonPermissionsAdapter.PermissionViewHolder
 import mozilla.components.feature.addons.ui.AddonPermissionsAdapter.Style
 import mozilla.components.support.test.mock
 import mozilla.components.support.test.robolectric.testContext
 import mozilla.components.support.test.whenever
+import mozilla.components.ui.colors.R
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.mockito.Mockito.verify
--- a/android-components/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
+++ b/android-components/components/browser/errorpages/src/main/java/mozilla/components/browser/errorpages/ErrorPages.kt
@@ -9,6 +9,7 @@ import android.content.Context
 import androidx.annotation.StringRes
 import mozilla.components.support.ktx.android.content.appName
 import mozilla.components.support.ktx.kotlin.urlEncode
+import mozilla.components.ui.icons.R as iconsR
 
 object ErrorPages {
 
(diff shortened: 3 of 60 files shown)


jonalmeida:
This is really cool. 🙂 

Did you use the "Migrate to Non-Transitive R Classes" to generate these? I suspect not or partially.

fice-t:
I used the tool initially and then did some manual conversions afterwards where the inlined namespaces resulted in long lines or where there was duplication. The tool also missed a couple dependencies which I added.

android-components/components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuHighlightTest.kt
```diff
@@ -36,12 +36,13 @@ class BrowserMenuHighlightTest {
     @Suppress("Deprecation")
     @Test
     fun `classic highlight effect converts background tint`() {
+        val colorId = mozilla.components.ui.colors.R.color.photonRed50
```
gabrielluong, line 39: I think we would prefer to import path and use an appropriate alias instead of inlining the full path.

gabrielluong:
@fice-t could you file a bug for this in https://bugzilla.mozilla.org/enter_bug.cgi?product=Fenix? Ideally, we should have a way to trace back the PR to a bug/issue with some description of the task.

fice-t:
OK, I created https://bugzilla.mozilla.org/show_bug.cgi?id=1799230.

android-components/components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuHighlightTest.kt
fice-t: Dealer's choice, but I think the utility of an alias makes more sense the more times it's used in a file. Using an alias just for one or two instances with no line length issue could be argued in the name of consistency, though.

gabrielluong: I think I feel strongly enough about this that I would suggest adding the imports. I would like to avoid inadvertently encouraging others to think it's okay to use the full namespace.

gabrielluong: Before proceeding with my suggested, I will bring this change up with the team just for general awareness. I don't foresee any concerns coming out of it, but I just want to avoid encouraging any additional changes before knowing for sure if we'll accept the proposed changes.

onemahon: I'd agree, @gabrielluong - inlining the path seems unnecessarily messy. I'd personally prefer the style of having paths imported (and, optionally, include an alias for the import - particularly if there are name collisions that require that).

onemahon: Actually, I think I'd phrase my preference as even more strict: always import the path (up to but not including `.R`, so every resource can start with the standard `R.`- prefix), and allow optional aliases but _only_ when doing so adds some useful clarity to which resource you're referencing.

Am I correct in believing that Android Studio adds the import statements for you automatically with the default text completion? If so, that seems like it would be the option that requires the least amount of thought while developing.

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonsPermissionsAdapterTest.kt
```diff
@@ -9,12 +9,12 @@ import android.widget.TextView
...
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import kotlinx.coroutines.ExperimentalCoroutinesApi
-import mozilla.components.feature.addons.R
 import mozilla.components.feature.addons.ui.AddonPermissionsAdapter.PermissionViewHolder
 import mozilla.components.feature.addons.ui.AddonPermissionsAdapter.Style
 import mozilla.components.support.test.mock
 import mozilla.components.support.test.robolectric.testContext
 import mozilla.components.support.test.whenever
+import mozilla.components.ui.colors.R
 import org.junit.Test
```
onemahon, line 18: Was this part of the automatic conversion tool? There's a whole bunch of similar instances throughout the PR. Any idea why these got changed?

(Correct me if I'm wrong, but I thought both of these represent non-transitive resource path references, so I wouldn't expect there to be a performance difference. If that's the case, I tend to get a little nervous about automated changes whose purpose we don't have a clear justification for)

So, essentially:

1. Does this risk changing any runtime behavior? (e.g. referencing the wrong resource)
2. Does this actually add any value, or was it just automatically updated? (And, if so, do we know why the tool picked these changes?)

onemahon:
Just left a couple comments, but I'm planning on leaving full review up to the experts - this is a really interesting idea, and a nice catch on the potential performance boost! Thanks for bringing this up!

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonsPermissionsAdapterTest.kt
fice-t: The parts where the import name changes are manual changes that the automatic tool actually missed; these changes point the `R` reference to the proper class whereas before they were pointing to a different class that just so happened to pull in the correct `R` resource(s) upon dependency resolution (which no longer occurs for non-transitive `R` classes).

> Correct me if I'm wrong, but I thought both of these represent non-transitive resource path references

The removed import is actually a transitive reference as the `addons` module's `R` doesn't directly contain the resource (in this case the color `photonBlue40`), and instead depends on the `ui-colors` module that contains the `R` class that contains the color resource.

Before: `AddonsPermissionsAdapterTest` -> `addons R` -> `ui-colors R` -> `photonBlue40`
After: `AddonsPermissionsAdapterTest` -> `ui-colors R` -> `photonBlue40`

> Does this risk changing any runtime behavior? (e.g. referencing the wrong resource)

No, the changes were required to compile successfully with `android.nonTransitiveRClass=true`. I also checked that there were no other possible resources that the statements could have been referencing instead.

onemahon: Ahh, that all makes sense, I was indeed misinterpreting the context of non-transitive R classes. Appreciate the thorough rundown.

android-components/components/browser/menu/src/test/java/mozilla/components/browser/menu/BrowserMenuHighlightTest.kt
fice-t: > Actually, I think I'd phrase my preference as even more strict: always import the path (up to but not including .R, so every resource can start with the standard R.- prefix)

I don't think that Kotlin allows this. When attempting to import e.g. `mozilla.components.ui.colors` instead of `mozilla.components.ui.colors.R` I get: `Packages cannot be imported`. According to the [docs](https://kotlinlang.org/docs/packages.html#imports) it seems like you can't import packages like you can in e.g. Python. One could use a wildcard, but that unnecessarily brings along all other references as well (which may result in more conflicts).

> Am I correct in believing that Android Studio adds the import statements for you automatically with the default text completion?

I had to enable the option, but the option does exist. I don't think text completion for `R` classes works without first getting the import correct. though.

kycn: I agree with using alias for the consistency. Though we have files with only one or two changes now, and seems reasonable to skip using alias for them; I believe it is possible in the future that people will refer to the resources more and may populate the files with more `full/path` resource files.

gabrielluong:
@fice-t I have already asked the team to weigh in and their feedback is already present here so I think we're okay with proceeding. In terms of full name vs alias, I am still in favour of going with alias for everything. Even if there is currently only 1 or 2 usages where it might make sense to just allow the full namespace path in a file, I think it would be better to future-proof ourselves by using the alias from the get-go since this will mitigate future refactoring if we suddenly had more usages of the R class in the file.

gabrielluong (approved):

gabrielluong:
From my own testing:

Full clean build before (./gradlew --profile --rerun-tasks assembleDebug): ~4m 31s
Full clean build after (./gradlew --profile --rerun-tasks assembleDebug): ~3m 18s

gabrielluong:
@fice-t Thank you again for the patch! I landed this https://github.com/mozilla-mobile/firefox-android/pull/170

PR #105, opened 2022-11-05
Bug 1795105 - Don't store thumbnails in BrowserState 

jonalmeida:
The difference from this patch and [the original](https://github.com/mozilla-mobile/firefox-android/pull/82) that was backed is the `ThumbnailMiddleware`.

The middleware was written to consume the thumbnail actions and not let them pass further down to the reducer which would throw an exception (intentionally). However, this middleware also observes actions under `TabListAction` which were being consumed as well unintentionally.

This patch updates the middleware to separate allow the actions to continue if they are non-thumbnail ones.

Also see the `thumbnail actions are the only ones consumed by the middleware` test added to cover this behaviour.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


csadilek (approved):
👍 ah great find! That's tricky to read in the middleware. The two `when` blocks seem to be the clearest way, but I am curious if there's a potentially better API here. 🤔 

Anyway, 🚢 it.

jonalmeida:
> 👍 ah great find! That's tricky to read in the middleware. The two `when` blocks seem to be the clearest way, but I am curious if there's a potentially better API here. 🤔
> 
> Anyway, 🚢 it.

Yeah, definitely. I went through a few iterations to make it more readable, and most readable but verbose way was to have some nested if-else statements.

Looking at the other examples, I realized that there aren't any other examples of middlewares that do both: consume actions and observe others, so this was a bit tricky to write. Open to suggestions though.

Will merge this by EOD to leave time for any more feedback (maybe from kaya as well).

android-components/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsMiddleware.kt
```diff
@@ -56,6 +51,19 @@ class ThumbnailsMiddleware(
...
+                thumbnailStorage.saveThumbnail(action.sessionId, action.thumbnail)
+            }
+            is ContentAction.RemoveThumbnailAction -> {
+                thumbnailStorage.deleteThumbnail(action.sessionId)
+            }
+            else -> {
+                // Do not let the thumbnail actions continue through to the reducer.
+                next(action)
+            }
+        }
```
kycn, line 67: nit: Could we have used `return` for these two actions and merged two `when`s? Not sure which looks more readable though 😄 (seems almost similar to me)

kycn (approved):
Looks great! Aaaand... tricky... 😂

android-components/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/ThumbnailsMiddleware.kt
jonalmeida: Oh yes, that is a lot nicer to read actually! The tests pass too - thanks!

PR #110, opened 2022-11-06
Bug 1811751 - fix estimated reading time disappear, related to android-repository PR 11951.

ryg-git:
This PR handles the case in which if user opens reader mode for articles which are in a different language, the time estimate disappears.

ex.

time estimate disappears when reader mode is enabled and user refreshes the page.

after the changes are applied,

when user enables the reader mode and refreshes the page, the time estimate is displayed in article's language.

 
This PR is related to android-components repository PR [11951](https://github.com/mozilla-mobile/android-components/pull/11951 =100x)

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1811751


ryg-git:
Hello @csadilek,

I have made changes to the code according to our discussion in [PR](https://github.com/mozilla-mobile/android-components/pull/11951) I made in android-components repository,

The changes take care of the case in which for some languages the reading time estimate does not have three parts,
so the added `else` block shows reading time in article's language.

before changes the reading time goes away after refresh,

https://user-images.githubusercontent.com/69393483/212741735-13cf3bbf-35ff-4a52-98dd-f68c2f2a6023.mp4

after changes the reading time is shown in article's native language,

https://user-images.githubusercontent.com/69393483/212742805-738f9652-5d67-4fb2-b41b-611d19025f70.mp4

please let me know if you want me make any changes.

Thanks!

android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview.js
```diff
@@ -179,6 +179,10 @@ class ReaderView {
           readingTimeString = `${readingTimeMinsFast} - ${readingTimeString}`;
         }
         return readingTimeString;
+      } else {
+        // This block handles the case when `parts` array does not have length 3,
+        // For some languages parts may not have length 3.
+        return parts.map(e => e.value).join(" ");
```
csadilek, line 185: I've reviewed this in the original PR here: https://github.com/mozilla-mobile/android-components/pull/11951#issuecomment-1282791718

The main problem is that we do not have reading time estimates for this language, and therefore even the parts you do get back here are incorrect. This is because we fall back to estimating the reading time in English, and incorrectly try to present it as reading time in the article's language.

So, you're right that there are languages for which we won't have all three parts, but there's a bigger problem, because the reading time estimates for these languages are also not correct.

On Firefox Desktop, we also fall back to reading time estimates in English, but then also display it in English if the language is not supported. So, I recommend we do the same here, see the fix in my [comment](https://github.com/mozilla-mobile/android-components/pull/11951#issuecomment-1282791718) above.

csadilek (changes requested):

android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview.js
csadilek: Alternatively, instead of displaying an incorrect estimate, maybe we shouldn't display one at all?

ryg-git: Thank you for the review @csadilek, I will update the PR according to your comment which will also make the app consistent with the desktop behavior.

csadilek: OK, thanks, I think we can land this next week. Can you rebase and squash your commits into one?

ryg-git: Thank you for the review, I will make the needful changes as soon as I can, thanks.

csadilek (approved):
Thanks!

csadilek:
Tested this manually and run UI tests too.

PR #124, opened 2022-11-08
Bug 1799679 - Clear search terms on url load

Alexandru2909:
For https://github.com/mozilla-mobile/fenix/issues/25812, we should clear the search terms only when loading a new URL. Currently we dismiss the search terms whenever we receive a redirected `LoadRequest`. As explained in [this comment](https://github.com/mozilla-mobile/fenix/issues/25812#issuecomment-1307157758), multiple search engines redirect the user to the search results, clearing the search terms.

This behavior was initially introduced to fix a search term [issue](https://github.com/mozilla-mobile/fenix/issues/7632). This is not reproducible after the changes in this PR.

https://user-images.githubusercontent.com/35462038/200655876-43db56c8-ab67-4226-a1fb-cf93445cdf81.mp4

The previosuly reproducible behavior of saving the search term when navigating through history is not reproducible:

https://user-images.githubusercontent.com/35462038/202745603-b41de905-9a0f-4df4-b0e9-814853621e84.mp4

The behavior of saving the search term for sites that use location change to navigate to different pages is still reproducible:

https://user-images.githubusercontent.com/35462038/202744904-9b757b67-5e1c-4a36-8543-7ad64ac57b6a.mp4

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/EngineObserverTest.kt
```diff
@@ -1307,26 +1307,6 @@ class EngineObserverTest {
         }
     }
 
-    @Test
```
rocketsroger, line 1310: This test can be put back in until unified search is enabled by default on release.

rocketsroger (commented):
Looks good but let's land this next week after the cut.  Since this will affect all users, not just unified search users.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -165,13 +166,18 @@ class GeckoEngineSession(
...
+        shouldKeepSearchTerm: Boolean,
     ) {
         val scheme = Uri.parse(url).normalizeScheme().scheme
         if (BLOCKED_SCHEMES.contains(scheme) && !shouldLoadJSSchemes(scheme, flags)) {
             logger.error("URL scheme not allowed. Aborting load.")
             return
         }
 
+        if (!shouldKeepSearchTerm) {
+            notifyObservers { onSearchTermDismiss() }
```
csadilek, line 178: I don't think it's right for the browser engine implementation to control dismissing search terms. It has otherwise no notion of search terms so this isn't quite the right abstraction. 

I think this should be handled outside the engine by the caller and/or the observers of these events. I would propose adding additional sate to the load request (in the browser store) for this. The state you added here is whether or not to keep the search terms, which can properly be reduced further to whether or not the current/last load is a search. Once in the store you can access it everywhere and dismiss as needed.

csadilek (changes requested):

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
csadilek: In other words, I think we can introduce more generic events from the engine if needed, add some state to the store and handle it in the caller / observers. Happy to talk through options! Sorry that this is a bit of a painful patch to go through.

csadilek: A perhaps even simpler solution is to dispatch an action to clear the search terms  (`UpdateSearchTermsAction`) in the use cases directly. Then you also don't need to loop through this additional parameter. In the use cases, it's implicitly clear if the request is a search or not so it's easy to decide there and they already dispatch actions to the store.

We can then keep `onNavigateBack` and a add a new `onNavigateForward` and dispatch the action as well in `EngineObserver`. This makes the diff much smaller.

Then the only case missing is line 219 below where you dispatch as part of loadData, but we can introduce a new event for that so it's not tied to search terms.

csadilek (commented):
Thanks, @Alexandru2909. This looks great now! Let me know when you're ready and I will make a final pass.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/EngineObserver.kt
```diff
@@ -46,6 +47,26 @@ internal class EngineObserver(
...
+    override fun onGotoHistoryIndex() {
+        store.dispatch(ContentAction.UpdateSearchTermsAction(tabId, ""))
+    }
+
+    override fun onLoadData() {
+        store.dispatch(ContentAction.UpdateSearchTermsAction(tabId, ""))
+    }
+
+    override fun onLoadUrl() {
+        if (store.state.findTabOrCustomTab(tabId)?.content?.isSearch == true) {
```
csadilek, line 63: I was wondering where you decide to reset `isSearch`. I think this is fine for the use case here, but worth noting that if the user decides to refine the search in content (e.g., via typing and changing the search directly in the web page's search box) then we will not capture this next load as a search. 

However,that's already the case and the awesomebar won't have those search terms anyway. We do have logic to detect the search terms in this case and could address that in the future, if we ever wanted to: https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/feature/search/src/main/java/mozilla/components/feature/search/ext/SearchEngine.kt#L101

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/SearchUseCases.kt
```diff
@@ -77,16 +77,19 @@ class SearchUseCases(
...
                 // If no `sessionId` was passed in then create a new tab
-                tabsUseCases.addTab(searchUrl)
+                tabsUseCases.addTab(searchUrl, isSearch = true)
             } else {
                 // If we got a `sessionId` then try to find the tab and load the search URL in it
                 val existingTab = store.state.findTabOrCustomTab(sessionId)
                 if (existingTab != null) {
-                    sessionUseCases.loadUrl(searchUrl, sessionId = existingTab.id)
+                    store.dispatch(ContentAction.UpdateIsSearchAction(existingTab.id, true))
+                        .invokeOnCompletion {
```
csadilek, line 86: Why do we use `invokeOnCompletion` here, but not in the `else` case where a new tab is added? It's unlikely to be needed here, but if it is, we should also add it to the other case as well?

From the docs of `invokeOnCompletion`: ` There is no guarantee on the execution context in which the handler is invoked.` -> Since these use cases are expected to be invoked on the main thread, I think we will need to wrap the handler in a `launch(Main)`.

csadilek: Ah I see, in the `else` case, the tab use case dispatches an action to the store anyway (to load the URL) and all actions are executed in order so that's why you don't need it there. In `SessionUseCase.loadUrl` on the other hand, you may hit the optimized path that calls `engineSession.loadUrl` directly, which is why you added the `invokeOnCompletion` to make sure the new flag is set before the URL starts loading.

So, yeah, making sure we switch to main for the load inside `invokeOnCompletion` should be good enough then.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/EngineObserver.kt
Alexandru2909: I believe this is the behavior described in the [first comment](https://github.com/mozilla-mobile/firefox-android/pull/124#issue-1440774722), in the last video. Should we file a follow-up for this? I think it would be worth investigating for search engine result pages which update the location based on the content displayed (see Reddit search engine), especially in the context of the Unified Search [tickets](https://mozilla-hub.atlassian.net/browse/FNXV2-22090).

Alexandru2909: Filed [1804636](https://bugzilla.mozilla.org/show_bug.cgi?id=1804636) in this scope.

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/SearchUseCases.kt
```diff
@@ -77,16 +81,21 @@ class SearchUseCases(
...
                 // If no `sessionId` was passed in then create a new tab
-                tabsUseCases.addTab(searchUrl)
+                tabsUseCases.addTab(searchUrl, isSearch = true)
             } else {
                 // If we got a `sessionId` then try to find the tab and load the search URL in it
                 val existingTab = store.state.findTabOrCustomTab(sessionId)
                 if (existingTab != null) {
-                    sessionUseCases.loadUrl(searchUrl, sessionId = existingTab.id)
+                    store.dispatch(ContentAction.UpdateIsSearchAction(existingTab.id, true))
+                        .invokeOnCompletion @OptIn(DelicateCoroutinesApi::class) {
```
csadilek, line 90: What's interesting/challenging here is that we do this to prevent a race condition for the optimized case in `SessionUseCases`, which loads the URL directly on an existing engine session to prevent a context switch. However, with this, and I haven't considered this before, we now introduce another context switch, which will make search slower again in practice.

Did you actually reproduce a problem here in practice? Did you see a case where the state in the store wasn't yet updated without `invokeOnCompletion`? A bit unfortunate, that we have this context switch here now. If this is only theoretical, I could also file a follow and we remove the `invokeOnCompletion` for now. I need to think about this some more.

csadilek: Also, thanks @Alexandru2909 for working through this. It's a surprisingly tricky problem, but we're close to landing it.

Alexandru2909: In the previous iteration this case was unintentionally taken care of, but after changing the implementation and a bit of testing I found the problem is reproducible and decided to add `invokeOnCompletion`.

csadilek: @Alexandru2909 OK, then I propose the following simplification and we can land it: https://github.com/csadilek/firefox-android/commit/71b4a5d4cc0d80a839783800f5a9d7063446ff1c

Actions are processed in order anyway, and this way we don't need the callback and context switch. This still means that this search use case will be slightly slower, but there' no way around that and still fixing this bug. So, I'd be OK with it.

Wdyt?

Alexandru2909: Sounds good to me, added the change now

csadilek: 👍 OK, let's land after the release branch is cut. There's one nit below about the ktlint-baseline.

csadilek (approved):
👍 This is good to go now, after we cut the 109 branch.

csadilek:
@Alexandru2909 Ok, landing this now. Can you update the corresponding tickets tomorrow, please? Thanks again.

csadilek:
@Alexandru2909 Let's re-create the PR. I don't think it will be hard to figure out the Focus problem, we just didn't have the time today so decided to back out instead to unblock folks :).

PR #138, opened 2022-11-11
Bug 1800268 - Support external url filtering in some Awesomebar providers

Mugurell:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1800268

### DIFF
--- a/android-components/components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/AutocompleteProvider.kt
+++ b/android-components/components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/AutocompleteProvider.kt
@@ -1,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.toolbar
+
+/**
+ * Object providing autocomplete suggestions for the toolbar.
+ * More such objects can be set for the same toolbar with each getting results from a different source.
+ * If more providers are used the [autocompletePriority] property allows to easily set an order
+ * for the results and the suggestion of which provider should be tried to be applied first.
+ */
+interface AutocompleteProvider : Comparable<AutocompleteProvider> {
+    /**
+     * Retrieves an autocomplete suggestion which best matches [query].
+     *
+     * @param query Segment of text to be autocompleted.
+     *
+     * @return Optional domain URL which best matches the query.
+     */
+    suspend fun getAutocompleteSuggestion(query: String): AutocompleteResult?
+
+    /**
+     * Order in which this provider will be queried for autocomplete suggestions in relation ot others.
+     *  - a lower priority means that this provider must be called before others with a higher priority.
+     *  - an equal priority offers no ordering guarantees.
+     *
+     * Defaults to `0`.
+     */
+    val autocompletePriority: Int
+        get() = 0
+
+    override fun compareTo(other: AutocompleteProvider): Int {
+        return autocompletePriority - other.autocompletePriority
+    }
+}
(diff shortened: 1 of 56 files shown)


Mugurell:
After syncing with Verdi this will need some changes and is depending on other tickets also.
-> Draft for now.

android-components/components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/AutocompleteProvider.kt
```diff
@@ -0,0 +1,36 @@
...
+    suspend fun getAutocompleteSuggestion(query: String): AutocompleteResult?
+
+    /**
+     * Order in which this provider will be queried for autocomplete suggestions in relation ot others.
+     *  - a lower priority means that this provider must be called before others with a higher priority.
+     *  - an equal priority offers no ordering guarantees.
+     *
+     * Defaults to `0`.
+     */
+    val autocompletePriority: Int
```
rocketsroger, line 30: ❓ does this only guarantee the order of the query but it doesn't guarantee the order of the suggestions?

Mugurell: This will only apply to autocomplete.

Currently we have only two autocomplete providers - history and domains with [the results from history being hardcoded as prioritary](https://searchfox.org/mozilla-mobile/rev/7c70c3df303e19acc69877821c3770fe85393005/firefox-android/android-components/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt#37-42).
For unified search we need to support other providers also - tabs, history, bookmarks and the `autocompletePriority` property will allow to avoid a hardcoded order between them all (issue that we still currently have with the awesomebar suggestions) and give control to the client.
Keeping the previous functionality: showing first history autocomplete if available and fallback on domains suggestions would be controlled in Fenix from [this change](https://github.com/mozilla-mobile/fenix/pull/27810/files#diff-989cb00682a18f94bf85f95b196445939a33ddf903cce43eb1a319cc13ae10e0R363-R365) (`ShippedDomainsProvider(1)`) where we initialize the provider with a lower priority than the default `0` for the history autocomplete provider.

Regarding the functionality from A-C [providers are kept in a `SortedSet`](https://github.com/mozilla-mobile/firefox-android/blob/34f103aeaabe07397a3367846f7bfe3df2b43916/android-components/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt#L30) - sorted by their `autocompletePriority` such that the providers with the lowest priority will be first and then [they will be queried in that order for autocomplete suggestions](https://github.com/mozilla-mobile/firefox-android/blob/34f103aeaabe07397a3367846f7bfe3df2b43916/android-components/components/feature/toolbar/src/main/java/mozilla/components/feature/toolbar/ToolbarAutocompleteFeature.kt#L38).

Looks like I don't have a test for this scenario, will add it, thank you!

rocketsroger: I see.  Thanks for the clarification.

Mugurell:
@rocketsroger updated the patch now with the updated filtering. Tested in Fenix with an updated patch also (https://github.com/mozilla-mobile/fenix/pull/27810) and everything seems to work as intended.

Mugurell:
`lint-buildconfig-android-components`:
> ERROR - .buildconfig.yml file changed! Please update it by running:
>     curl --location https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/YbFP4-epSPKoRMt-H020_Q/artifacts/public%2Fgit%2Fbuildconfig.diff | gunzip | git apply

rocketsroger:
> @rocketsroger updated the patch now with the updated filtering. Tested in Fenix with an updated patch also ([mozilla-mobile/fenix#27810](https://github.com/mozilla-mobile/fenix/pull/27810)) and everything seems to work as intended.

Thanks, I'll do another round of review on this and the Fenix PR today.  Thanks

rocketsroger (approved):
Looks good.  Thanks

PR #168, opened 2022-11-17
Don't try and record history for URLs with an invalid scheme.

mhammond:
history.canAddUri() previously treated a URL with an empty scheme as valid, but application-services does not.

While I can't work out an STR to make Fenix actually try and add history using such a URL, I have verified that if it did, we'd end up with an error as reported in https://github.com/mozilla/application-services/issues/5235.


jonalmeida (approved):
Interesting - this looks good to land.

I filed BZ-1801164 so we can link this change against a bug. Will update this commit to add it.

jonalmeida:
Didn't get a change to do this on Thursday, so I'll merge it in on Monday with the other PRs from Ben.

PR #181, opened 2022-11-18
For #25810 - Unified search 'search popup menu' should give hint to the user that there are more menu items available by scrolling

t-p-white:
Fenix changes here: https://github.com/mozilla-mobile/fenix/pull/27895

The unified search 'search popup menu' should give hint to the user that there are more menu items available by scrolling. This is indicated by displaying a half menu item if the available height will not fit the full list of items.

Existing behaviour:

https://user-images.githubusercontent.com/1833156/202675300-ab202c8f-5d3c-4fa9-aed5-bca265d872d7.mov

New behaviour LTR: 

https://user-images.githubusercontent.com/1833156/210393163-c7e3aece-c3da-4ccd-8eef-b221d7821ba3.mov

New behaviour RTL: 

https://user-images.githubusercontent.com/1833156/210393280-c810b7a4-b8e9-4d80-988c-150c4c69e01c.mov

### UI states:
* Does menu fit or does menu have to scroll
* Search bar Top/Bottom
  * Portrait keyboard open
  * Portrait keyboard closed
  * Landscape keyboard open
  * Landscape keyboard closed

**Note**: This issue is covering the direction and popup height only, another issue is in progress for positioning https://github.com/mozilla-mobile/fenix/issues/27998.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/BrowserMenuController.kt
```diff
@@ -55,11 +52,20 @@ class BrowserMenuController(
             style?.let { setStyle(it) }
         }
 
+        // Monitor for changes to the parent layout and dismiss pop up if displayed, else the menu
+        // could be displayed in the wrong position. For example, if the menu is displayed and the
+        // device orientation changes from portrait to landscape or vice versa.
```
t-p-white, line 57: Note: This is also consistent with the 'settings' menu popup

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -9,96 +9,192 @@ import android.view.Gravity
...
+                    reversed,
+                    upMenuViewContainerHeight,
+                    downMenuViewContainerHeight,
+                    fitsUp,
+                    fitsDown,
+                )
+            }
         }
+    } else {
+        Log.log(
```
t-p-white, line 80: Adding logging as a starter - should this do something else to capture the error e.g. included in telemetry data?

mavduevskiy:
@t-p-white 
just a heads up: for a successful build `BrowserMenuControllerTest` needs to be updated🙃

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/BrowserMenuController.kt
mavduevskiy: @rocketsroger 
I found a solid [reasoning](https://github.com/mozilla-mobile/fenix/pull/17682#issuecomment-769883653) why the hosts (fragments, controllers) handle the dismissal of the menus (like [here](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#1475)).  

I was wondering why the same approach hasn't been used towards bookmarks fragment or tabs tray? They keep the `BrowserMenu` at the same position, and Bookmarks fragment actually chops a good chunk from the menu list. Could it be a universal approach that we want to hide menus after the configuration change because of misalignment?

rocketsroger: As discussed with @mavduevskiy, I think in this case we're not using `OnConfigurationChanged` but using layout to determine if we should dismiss the menu.  However, we could add a parameter in `show()` to allow the user to determine that it should "auto dismiss"?

android-components/components/browser/menu2/src/test/java/mozilla/components/browser/menu2/ext/PopupWindowTest.kt
```diff
@@ -126,35 +137,200 @@ class PopupWindowTest {
...
+        doReturn(mockRecyclerView).`when`(menuContentView)
+            .findViewById<RecyclerView>(R.id.mozac_browser_menu_recyclerView)
+
+        val x = 0
+        anchor = createMockViewWith(x, 40)
+        val containerHeight = mockRecyclerView.measuredHeight.plus(1)
+        doReturn(containerHeight).`when`(menuContentView).measuredHeight
+
+        popupWindow.displayPopup(menuContentView, anchor, Orientation.UP)
+
```
mavduevskiy, line 240: magic number with no comments :) 
50 / 8 = 6 – item height
40 anchor height, so 6 items could fit. we crop the last one, so it's 33, plus 1 padding == 34!

android-components/components/browser/menu2/src/test/java/mozilla/components/browser/menu2/ext/PopupWindowTest.kt
```diff
@@ -126,35 +137,200 @@ class PopupWindowTest {
...
+        val expectedMenuContainerHeight = 33
```
mavduevskiy, line 259: I wonder if there is a way not to hardcore the expected height? At least an explanation of the math, it required debugging to understand where the expected value comes from.
The calculation is something like this
`val expectedMenuContainerHeight = (itemHeight * ((viewHeight / itemHeight) - 0.5)).toInt() + viewPadding`

Verdi:
After discussing this we want the menu to:

- Take up as much vertical space as it needs to show the entire list.
  - If we can't display the whole list it should be truncated in the middle of an item to indicate that it's scrollable
- If the toolbar is on bottom the menu should be aligned with the bottom of the viewport.
- If the toolbar is on top the menu should be aligned with the top of the viewport.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -9,96 +9,192 @@ import android.view.Gravity
...
 import mozilla.components.browser.menu2.R
 import mozilla.components.concept.menu.Orientation
+import mozilla.components.support.base.log.Log
+
+const val MIN_NUMBER_MENU_ITEMS = 6
+const val HALF_MENU_ITEM = 0.5
 
 internal fun PopupWindow.displayPopup(
-    containerView: View,
+    menuViewContainer: View,
```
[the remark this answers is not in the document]
t-p-white: I just found it a more usable name while doing dev but happy to revert as it was 👍

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -9,96 +9,192 @@ import android.view.Gravity
...
+    if (recyclerViewAdapter != null) {
+        val recyclerViewItemCount = recyclerViewAdapter.itemCount
+
+        val upMenuViewContainerHeight = getMenuViewContainerHeight(
+            recyclerViewHeight,
+            recyclerViewItemCount,
+            availableHeightToTop,
+            menuViewContainer.measuredHeight,
+        )
+        val downMenuViewContainerHeight = getMenuViewContainerHeight(
```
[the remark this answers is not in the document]
t-p-white: Good observation! It was intentional, however 6 may be more optimal

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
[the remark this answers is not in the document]
mavduevskiy: Personally, I might be finding it more usable as well, but when introducing changes to the already excising code we should consider if we follow (or break) the pattern that is already existing and if there is a reason for that and if the change if really necessary. Here, the same object is being passed down and it didn't change its parameters or behavior, so changing its name seems not necessary (though I do agree that `menuViewContainer` is a more accurate naming!)

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -9,96 +9,192 @@ import android.view.Gravity
...
+
+const val MIN_NUMBER_MENU_ITEMS = 6
+const val HALF_MENU_ITEM = 0.5
 
 internal fun PopupWindow.displayPopup(
-    containerView: View,
+    menuViewContainer: View,
     anchor: View,
     preferredOrientation: Orientation? = null,
-    forceOrientation: Boolean = false,
```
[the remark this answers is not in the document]
mavduevskiy: It might be a good idea to mention such "behind the scene" discussions briefly in PR or add a comment around that change, to prevent any questions

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/BrowserMenuController.kt
```diff
@@ -37,29 +37,38 @@ class BrowserMenuController(
...
-     * @param anchor The view on which to pin the popup window.
-     * @param orientation The preferred orientation to show the popup window.
-     * @param forceOrientation When set to true, the orientation will be respected even when the
-     * menu doesn't fully fit.
-     */
+    private var orientation: Orientation? = null
+
     override fun show(
         anchor: View,
         orientation: Orientation?,
```
t-p-white, line 44: Is there a reason for orientation being nullable? To me, it seems like this could cause unpredictable behaviour or a bug.

mavduevskiy:
For the future reference: one of the approaches we were trying here was to stop relying on the `showAsDropDown` method and use only `showAtLocation`, but decided against it. 

Going through the source code one could find that the main difference between the two is that `showAsDropDown` will reposition itself after layout change while `showAtLocation` won't. Also, `showAsDropDown` is aligned to the anchor, while `showAtLocation` will be displayed at give `x, y` coordinates, which does mean completely manual position calculation.

mavduevskiy:
Heading into right direction!

A few notes:
1. `numberOfItemsFitExactly` could be improved to account for rounding – if we can fit 6.9 items, we will display just 5.5. Displaying 6.5 sounds more reasonable. Not critical.
3. The case with the keyboard and showing the menu from the bottom could be handled by something like this:
```
orientation == Orientation.UP && containerViewHeight > availableHeightToTop ->
    showPopupWhereBestFits(
        anchor,
        fitsUp,
        fitsDown,
        isRtl,
        containerHeight,
    )
```
`showAtLocation` in this case should switch to using `anchor.getLocationInWindow(anchorPosition)` for getting the correct coordinates in the landscape mode of the `searchFragment`.
4. `showAtAnchorLocation` should also apply the possible shortened height of the view.
5. ktlint fails!

t-p-white:
Thank you @mavduevskiy, really useful feedback. The above comments should now be resolved. As discussed, we'll hold off landing this until @rocketsroger has had a chance to also take a loo .

android-components/components/browser/menu2/src/test/java/mozilla/components/browser/menu2/ext/PopupWindowTest.kt
```diff
@@ -4,180 +4,388 @@
...
+        verify(popupWindow, never()).showAsDropDown(any())
+        verify(popupWindow, never()).showAsDropDown(any(), anyInt(), anyInt())
+        verify(popupWindow, never()).showAtLocation(any(), anyInt(), anyInt(), anyInt())
     }
 
     @Test
-    fun `WHEN displaying prefer up popup from bottom right THEN show popup up and to the left`() {
-        anchor = createMockViewWith(x = 90, y = 190, false)
-        popupWindow.displayPopup(menuContentView, anchor, Orientation.UP)
+    fun `WHEN orientation up THEN showAsDropDown is called with expected config`() {
```
mavduevskiy, line 130: Maybe we could change the ambiguous `called with expected config` THEN condition to something more specific? Like 
`WHEN orientation up & does fit available height THEN the popup keeps its height & is positioned from the bottom of the anchor with leftBottomAnimation`

It is a mouthful, but it describes the business requirement we have. It might come in handy if we change the behavior: tests will start to fail, and the name of the test would tell us exactly what is the use case we should fix.

android-components/components/browser/menu2/src/test/java/mozilla/components/browser/menu2/ext/PopupWindowTest.kt
```diff
@@ -4,180 +4,388 @@
...
+
+        popupWindow.displayPopup(containerView, anchor, Orientation.UP)
+
+        val availableHeight = anchor.rootView.measuredHeight
+        val maxAvailableHeightForRecyclerView = availableHeight - CONTAINER_PADDING
+        val numberOfItemsFitExactly =
+            (maxAvailableHeightForRecyclerView.toFloat() / ITEM_HEIGHT.toFloat()).roundToInt()
+        val numberOfItemsFitWithOverFlow = numberOfItemsFitExactly - HALF_MENU_ITEM
+        val updatedRecyclerViewHeight = (numberOfItemsFitWithOverFlow * ITEM_HEIGHT).toInt()
+        val calculatedHeight = updatedRecyclerViewHeight + CONTAINER_PADDING
```
mavduevskiy, line 280: easy to read, nice!

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/BrowserMenuController.kt
mavduevskiy: Looking at [the patch](https://github.com/rocketsroger/android-components/commit/87c168eab46ea1570aeabe029e48af3b18331e39) I don't see a reason why it should be null. I think we could make it not nullable as a follow up. Pinging the author, maybe there was a not-so-obvious reason
 @rocketsroger

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -8,109 +8,280 @@ import android.graphics.Rect
...
+    val numberOfItemsFitWithOverflow = numberOfItemsFitExactly.numberOfItemsFitWithOverflow()
+    val updatedRecyclerViewHeight = (numberOfItemsFitWithOverflow * menuItemHeight).toInt()
+
+    return updatedRecyclerViewHeight + containerPadding
+}
+
+/**
+ * The number of menu items that fit exactly, minus a half menu item (indicates more menu items exist).
+ */
+private fun Int.numberOfItemsFitWithOverflow() = this - HALF_MENU_ITEM
```
mavduevskiy, line 211: Nit: same comment to a single line function as before, I find it to be an overkill for what it hides :)

 ```
   val numberOfItemsFitWithOverflow = numberOfItemsFitExactly - HALF_MENU_ITEM
```

File: android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt [hunk 12]
```diff
@@ -8,109 +8,280 @@ import android.graphics.Rect
...
+
 private fun PopupWindow.showPopupWhereBestFits(
     anchor: View,
-    fitsUp: Boolean,
-    fitsDown: Boolean,
-    reversed: Boolean,
+    isRtl: Boolean,
     containerHeight: Int,
 ) {
+    val (availableHeightToTop, availableHeightToBottom) = getMaxAvailableHeightToTopAndBottom(
```
mavduevskiy, line 218: single line

mavduevskiy (changes requested):
Looking good! Few small nits here and there.

The main things are: isRTL vs reversed, improving the naming of tests to explain clearly the condition we are trying to match, reducing extra lines for improved readability.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -8,109 +8,280 @@ import android.graphics.Rect
...
-@Suppress("LongParameterList")
+/**
+ * Determine whether the container view can display all menu items (without scrolling) within
+ * the available height.
+ *
+ * @return The original container height if the container view can display all menu items
+ * (without scrolling), else calculate the maximum available container height for a scrollable
+ * view with a half menu item.
+ */
+private fun getContainerHeight(
```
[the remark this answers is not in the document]
t-p-white: Thanks!

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -8,109 +8,280 @@ import android.graphics.Rect
...
-        else -> showPopupWithUpOrientation(anchor, containerHeight, reversed)
+        else -> showPopupWithUpOrientation(anchor, containerHeight, isRtl)
     }
 }
 
 private fun PopupWindow.showPopupWithUpOrientation(
     anchor: View,
     containerHeight: Int,
-    reversed: Boolean,
+    isRtl: Boolean,
```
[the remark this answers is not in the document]
t-p-white: Good spot

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
t-p-white: That's fair. It was doing more prior to the latest change, but agree that it is overkill now

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -8,109 +8,280 @@ import android.graphics.Rect
...
 ) {
-    animationStyle = if (reversed) {
```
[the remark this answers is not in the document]
t-p-white: My thoughts are that it must not have been the correct to begin with. Unless I’ve misunderstood, the original `val reversed = availableWidthToLeft < availableWidthToRight`  would have actually been true if it was LTR?
image

android-components/components/browser/menu2/src/test/java/mozilla/components/browser/menu2/ext/PopupWindowTest.kt
```diff
@@ -4,180 +4,388 @@
...
-
-        // Makes the availableHeightToBottom bigger than the menuContentView
-        setScreenHeightAndWidth()
     }
 
     @Test
-    fun `WHEN displaying prefer down popup from top left THEN show popup down and to the right`() {
-        anchor = createMockViewWith(x = 0, y = 0, false)
-        popupWindow.displayPopup(menuContentView, anchor, Orientation.DOWN)
+    fun `WHEN recycler view has no adapter THEN show is never called`() {
```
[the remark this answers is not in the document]
t-p-white: Good suggestion. I've updated to boolean as a starter - this may change based on feedback from on @rocketsroger around the displayPopup function error handling

mavduevskiy (changes requested):
Looks good to me! One thing before approval, maybe we should change the way RTL is determined? Currently it relies heavily on the implied knowledge about the anchor position (at start), and won't work correctly for the icon positioned at the end. We could improve that by getting orientation from the anchor itself.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -8,109 +8,237 @@ import android.graphics.Rect
...
+
+            // This could include cases where the keyboard is being displayed.
+            if (containerHeight > availableHeightToTop) {
+                showPopupWhereBestFits(anchor, isRTL, containerHeight)
+            } else {
+                showPopupWithUpOrientation(anchor, containerHeight, isRTL)
+            }
+        }
+
         else -> {
```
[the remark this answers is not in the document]
t-p-white: > Maybe using view.isRTL inside PopupWindow.displayPopup would be a better approach? It would actually tell the direction of the layout without relying on the start/end position of the anchor

Good suggestion, I don't see any issue with this (@rocketsroger thoughts?) I'll update to use the view.

t-p-white:
@mavduevskiy, @rocketsroger following demo of the updated behaviour, @Verdi approves 👍

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -8,109 +8,229 @@ import android.graphics.Rect
...
+ * Show the [PopupWindow] with the required configuration.
+ */
+private fun PopupWindow.showPopup(
+    orientation: Orientation?,
+    anchor: View,
+    recyclerViewHeight: Int,
+    recyclerViewItemCount: Int,
+    containerViewHeight: Int,
+) {
+    val containerHeight =
```
mavduevskiy, line 84: Looking at the name of the parameter and the name of the function `getContainerHeight` and the name of passed parameter `containerViewHeight` to be a bit confusing.  `containerViewHeight` is the height of the container if it's drawn as it is, but then the function is calculating the new height. Maybe changing the name from `getContainerHeight` (since we already have it) to `getNewContainerHeight` or `calculateContainerHeight` would be a more clear way to tell what's happening between a passed height and the new calculated height.

mavduevskiy (approved):
One naming nit (optional, change if you like the suggestion) and let's ship it! I really like the feature, good job! 

I would prefer @rocketsroger to take a final look before we ship it.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
rocketsroger: Is this something that you expect might happen in a corner case? Or is this just something you want to record in case that it does happen?

rocketsroger (approved):
Looks good to me.  Only question left is the error when view is initialized incorrectly and the comment about naming @mavduevskiy asked.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
t-p-white: Definitely a corner case - I can't think of a situation where this could happen as we always have the default menu items (which can not be removed by the user). If this does somehow happen then something very unexpected has occurred. I think by checking the conditions, we can at least provide somewhat of a graceful experience - the logging may be superfluous if it is not accessible (would we get sight of errors or would this be local to the device?)

PR #234, opened 2022-11-29
Bug 1800700 - Add default search suggest header

sarah541:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

Added a priority parameter to each of the SuggestionProviders and ordering of group of suggestions will be done according to that.

### DIFF
--- a/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionFetcher.kt
+++ b/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionFetcher.kt
@@ -98,7 +98,8 @@ internal class SuggestionFetcher(
      * Updates [state] to include the [suggestions] from [provider].
      */
     @Synchronized
-    private fun processResultFrom(
+    @VisibleForTesting
+    internal fun processResultFrom(
         group: AwesomeBar.SuggestionProviderGroup,
         provider: AwesomeBar.SuggestionProvider,
         suggestions: List<AwesomeBar.Suggestion>,
@@ -113,14 +114,25 @@ internal class SuggestionFetcher(
         updatedSuggestions.addAll(suggestions)
         updatedSuggestions.sortByDescending { suggestion -> suggestion.score }
 
+        if (updatedSuggestions.isNotEmpty()) {
+            group.priority = updatedSuggestions[0].score
+        }
+
         val updatedSuggestionMap = suggestionMap.toMutableMap()
         updatedSuggestionMap[group] = updatedSuggestions
         state.value = updatedSuggestionMap
         val profilerEndTime = profiler?.getProfilerTime() // THIS MUST OCCUR RIGHT AFTER STATE UPDATE.
 
         // Markers can only be added on the main thread right now.
-        ThreadUtils.postToMainThread {
-            profiler?.addMarker("Suggestion update", profilerStartTime, profilerEndTime, provider::class.simpleName)
+        profiler?.let {
+            ThreadUtils.postToMainThread {
+                profiler.addMarker(
+                    "Suggestion update",
+                    profilerStartTime,
+                    profilerEndTime,
+                    provider::class.simpleName,
+                )
+            }
         }
     }
 
--- a/android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
+++ b/android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
@@ -190,6 +190,8 @@ interface AwesomeBar {
      * A group of [SuggestionProvider]s.
      *
      * @property providers The list of [SuggestionProvider]s in this group.
+     * @property priority An optional priority for this group. Decides the order of this group
+     * in the AwesomeBar suggestions. Group having the highest integer value will have the highest priority.
      * @property title An optional title for this group. The title may be rendered by an AwesomeBar
      * implementation.
      * @property limit The maximum number of suggestions that will be shown in this group.
@@ -197,6 +199,7 @@ interface AwesomeBar {
      */
     data class SuggestionProviderGroup(
         val providers: List<SuggestionProvider>,
+        var priority: Int = 0,
         val title: String? = null,
         val limit: Int = Integer.MAX_VALUE,
         val id: String = UUID.randomUUID().toString(),
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -15,6 +15,9 @@ permalink: /changelog/
 * **feature-prompts**:
   * Added permission requests for accessing media files (`READ_MEDIA_AUDIO`, `READ_MEDIA_AUDIO`, `READ_MEDIA_AUDIO`) when uploading files on devices with Android 13 and later.
 
+* **feature-awesomebar**
+  * `SuggestionProviderGroup` now has a new parameter `priority` that decides the order of this group in the AwesomeBar suggestions. Priority is same as the `score` of `AwesomeBar.Suggestions`. Group having the highest integer value will have the highest priority.
+  
 # 109.0.0
 * [Commits](https://github.com/mozilla-mobile/firefox-android/compare/v108.0.0...v109.0.0)
 * [Dependencies](https://github.com/mozilla-mobile/firefox-android/blob/v109.0.0/android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt)
--- a/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/AwesomeBar.kt
+++ b/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/AwesomeBar.kt
@@ -96,7 +96,7 @@ fun AwesomeBar(
 
         // This state does not need to be remembered, because it can change if the providers list changes.
         @SuppressLint("UnrememberedMutableState")
-        val suggestions = derivedStateOf { fetcher.state.value }.value.toSortedMap(compareBy { it.title })
+        val suggestions = derivedStateOf { fetcher.state.value }.value.toSortedMap(compareByDescending { it.priority })
 
         LaunchedEffect(text, fetcher) {
             fetcher.fetch(text)
--- a/android-components/components/compose/awesomebar/src/test/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionFetcherTest.kt
+++ b/android-components/components/compose/awesomebar/src/test/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionFetcherTest.kt
@@ -6,12 +6,14 @@ package mozilla.components.compose.browser.awesomebar.internal
 
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
+import mozilla.components.concept.awesomebar.AwesomeBar
 import mozilla.components.concept.awesomebar.AwesomeBar.SuggestionProvider
 import mozilla.components.concept.awesomebar.AwesomeBar.SuggestionProviderGroup
 import mozilla.components.support.test.any
 import mozilla.components.support.test.mock
 import mozilla.components.support.test.rule.MainCoroutineRule
 import mozilla.components.support.test.rule.runTestOnMain
+import org.junit.Assert.assertEquals
 import org.junit.Rule
 import org.junit.Test
 import org.mockito.Mockito.doAnswer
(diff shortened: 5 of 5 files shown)


android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -157,6 +157,11 @@ interface AwesomeBar {
          **/
         fun groupTitle(): String? = null
 
+        /**
+         * A priority number for ordering the suggestions.
+         **/
```
gabrielluong, line 162:
```suggestion
         */
        fun groupTitle(): String? = null

        /**
         * A priority number for ordering the suggestions.
         */
```
Drive by

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -157,6 +157,11 @@ interface AwesomeBar {
...
+        fun priority(): Int? = null
```
kycn, line 163: When it comes to assigning priorities to suggestion providers, it maybe a good idea to hold an int-valued enum class instead of hardcoding int values to every single class. So returning an enum like the following (instead of an int):
``` kotlin
enum class ProviderPriority(val priority: Int) {
    BOOKMARKS(5),
    COMBINED_HISTORY(4),
    HISTORY_STORAGE(4),
    SEARCH_ACTION(1),
    SEARCH_ENGINE(0),
    SEARCH(3),
    SYNC_TABS(6)
}
```
and assigning the priority values in only one class would make its maintenance easier as we could see the priority order from a single source. The priority comparison, then, can be made via the values of enums. For the null (or default) case, I do not have the context but, adding a dummy enum such as NONE(-1) would work I guess.

sarah541: Hello @kycn , that was exactly mine and @gabrielluong's and @rocketsroger's discussion yesterday. I will be implementing it the enum way. Thank you for another confirmation! Appreciate the feedback

gabrielluong: @kycn Definitely agree on this and it's great to also see this brought up independent of our discussion. I added some thoughts on the Fenix discussion (see https://github.com/mozilla-mobile/fenix/pull/28024#pullrequestreview-1201297534) because I would expect the consumer of the components to manage their app's provider priority order. We basically want to avoid hardcoding the `priority` in AC.

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -145,6 +145,22 @@ interface AwesomeBar {
...
+            SEARCH_ACTION,
+            SEARCH_SUGGESTION,
+            COMBINED_HISTORY,
+            HISTORY,
+            SYNCED_TAB,
+            SESSION,
+            BOOKMARKS,
+            SHORTCUTS,
+            CLIPBOARD,
+        }
```
gabrielluong, line 162: I am open to further discussion here. I had imagined that this `enum` with the int priority would exist in the consumer (Fenix). By defining these `Type`s within the interface, we are basically saying these are the only available suggestion types, however, I don't think that would be our intention since this is just a concept that could be implemented by the consumer depending on whatever they want to do.

gabrielluong: @kycn I would love to further validate these thoughts if you have any input as well.

Mugurell: Agree that this enum (which isn't extendable) shouldn't stay in `concept`.
An interface should not know who implements it and as you've said clients are free to come with their own implementations on top of what the interface knows at this time.
If the intent is to force developers think about choosing the right priority in relation to others - and not use just any `Int` this approach helps though. Using a wrapper over the `Int` goes half way to solve developer's haste.

I am not a fan of another `enum` since it introduces a new type for developers to reason about and choose which could lead to miss matching.
I see myself having to read quite a bit on the providers to know how to choose the right priority to use between `SEARCH_ENGINE`, `SEARCH_ACTION`, `SEARCH_SUGGESTION`.
For clearer mappings, since these priorities would be used when a new provider instance is constructed we could use a simple map of each provider class and it's priority and avoid having to reason about the new enum types.
In Fenix a top level const:
```kotlin
val awesomeBarSuggestionsPriorities: Map<KClass<out AwesomeBar.SuggestionProvider>, Int> = mapOf(
    SearchActionProvider::class to 1,
    SearchEngineSuggestionProvider::class to 2
    ....
)
```
(with the provider class's default being `0`)

from which we can easily read and use the priority with 
```
SearchActionProvider(
    store = components.core.store,
    searchUseCase = searchUseCase,
    icon = searchBitmap,
    showDescription = false,
    priority = awesomeBarSuggestionsPriorities[SearchActionProvider::class],
)
```

(As an alternative to the above map we could also declare in Fenix extension `priority` properties for each provider type. This would also ensure no involuntary miss matching.)

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -157,6 +173,11 @@ interface AwesomeBar {
          **/
         fun groupTitle(): String? = null
 
+        /**
+         * A priority number for ordering the suggestions.
+         */
+        fun priority(): Type
```
Mugurell, line 179: nit: documentation says "a priority number" but this returns `Type`.

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
kycn: Hey @gabrielluong thanks for notifying! After reading [your comments in Fenix side](https://github.com/mozilla-mobile/fenix/pull/28024#pullrequestreview-1201297534), I strongly agree that the current approach is not as flexible as it should be. The maintenance is hard and as you mentioned is error-prone for future additions. In the simplest use case, where we would want to introduce a new provider (with a new type), this approach would require us to publish a new a-c release with the new type. This should be avoided for the reasons you described in your comment I've just referenced. The concept of a "concept" in a-c is conflicting with the idea of providing a list of order for suggestion providers. That is another reason why enum should be avoided here The concept modules are public apis that in general should behave quite generic and should not provide any implementation details. 
Addressing to the flexibility you mentioned in Fenix pr, I unfortunately do not have a very strong opinion that solves all the "what if" questions you asked reasonably. 
When I was going over the code I realized maybe we can introduce a new ProviderSorter interface/api in `AwesomeBar` feature module and implement it on Fenix end (could be in `AwesomeBarWrapper)` by injecting any sort of sorting mechanism that we like. This sorter then maybe issued to Awesomebar (in a-c) over the `addProvider` and `removeProvider` apis via additional (null-default) parameters. This is not my final suggestion, i would need to think more about this as it definitely needs more investigation/refinement and discussion I believe.

gabrielluong: I am not against having a `ProviderSorter` interface. I should also point out some of my suggested use cases are hypothetical at the moment (eg, the requirement to have a dynamic sorting depending on which suggestions are shown). I am thankful that we're considering potential use cases because flexibility is always nice.

However, I don't want us to block the current issues at hand for too long. So, I am looking to suggest a simpler path forward in the meantime and we can file a followup about introducing a more throughout solution. If we went back to the original commit https://github.com/mozilla-mobile/firefox-android/commit/b2053b2b6e029110085ec31a302caf966bd42fae that introduces the `priority` as Int, remove the hardcoded `priority`, add a breaking change note in the changelog, and added the following enum in Fenix, and use the `ProviderPriority` to set the `priority` when the suggestion provider classes are initialized in Fenix.

```
enum class ProviderPriority(val priority: Int) {
    BOOKMARKS(5),
    COMBINED_HISTORY(4),
    HISTORY_STORAGE(4),
    SEARCH_ACTION(1),
    SEARCH_ENGINE(0),
    SEARCH(3),
    SYNC_TABS(6)
}
```

How does that sound?

kycn: Sounds good to me! My only question would be, then, what about when all the providers get null values as priority? Or will we make the priority fileds of the providers non-null? 
We were sorting the providers according to their titles previously, could it be a default fallback in case we use nullable priority fields for providers and the consumer of a-c would not want to provide any sort of sorting logic with the integers?

sarah541: Just wanted to mention out here that @csadilek and @rocketsroger suggested that it would be better if we could sort the groups using already defined suggestion scores (as mentioned in my above comment) for most of the providers. Will be moving forward with this and see if we could have the results this way .

gabrielluong:
@sarah541 I added @csadilek as a reviewer since I have a bit less context around the scores, but I can still review this if Christian can't get to it. I am also wondering if you could add any tests for this?

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -192,11 +192,14 @@ interface AwesomeBar {
...
      * @property title An optional title for this group. The title may be rendered by an AwesomeBar
      * implementation.
+     * @property priority An optional priority for this group. Decides the order of this group
+     * in the AwesomeBar suggestions.
      * @property limit The maximum number of suggestions that will be shown in this group.
      * @property id A unique ID for this group (uses a generated UUID by default)
      */
     data class SuggestionProviderGroup(
         val providers: List<SuggestionProvider>,
+        var priority: Int? = 0,
```
Amejia481, line 202: As we have a default value, is there any specific reason why this property should be null-able?

android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionFetcher.kt
```diff
@@ -113,6 +113,10 @@ internal class SuggestionFetcher(
         updatedSuggestions.addAll(suggestions)
         updatedSuggestions.sortByDescending { suggestion -> suggestion.score }
 
+        if (updatedSuggestions.isNotEmpty()) {
+            group.priority = updatedSuggestions[0].score
```
Amejia481, line 117: Should be check that this index is accessible before we try too access it. For example maybe this list could be empty, and we could cause an exception by accessing it.

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
Amejia481: Should we expand on how the priorities work? :)

csadilek (commented):
This looks good to me @sarah541! I think the highest suggestion's score also being the groups score addresses the concerns. Is it possible to add a test to `SuggestionFetcherTest` for this?

I'd also appreciate another look from @Mugurell who knows this code well. :)

docs/changelog.md
```diff
@@ -22,6 +22,7 @@ permalink: /changelog/
   * ⚠️Increased `compileSdkVersion` to 33 (Android 13)
 * **feature-awesomebar**
   * `SearchSuggestionProvider` and `SearchActionProvider` now have a new parameter `suggestionsHeader`, to add title to suggestions.
+  * `SuggestionProviderGroup` now has a new parameter `priority` that decides the order of this group in the AwesomeBar suggestions.
```
gabrielluong, line 25: You will want to update this into the new in-development section for 110

Mugurell (approved):
LGTM in the context of Fenix and https://github.com/mozilla-mobile/fenix/pull/28024 but the API does seem a bit brittle and yet restrictive with the functionality working in Fenix just because of the exact configuration of groups and providers used and the default `score` providers set for their suggestions.
Maybe a note would help in the changelog about how the priority is calculated as a headsup for clients since they cannot really control it.

PR #242, opened 2022-12-01
Bug 1642977 - Let `complete` task depend on tasks are targeted and not optimized away

JohanLorenzo:
The current implementation of the complete task causes multiple issues: 
 1. It pulls optimized away tasks back in because it doesn't properly use taskgraph's `soft-dependencies` 
 2. It led to https://github.com/mozilla-mobile/android-components/pull/7617 which optimizes away tasks during the generation of the full graph (instead of during the optimization phase)
 3. It doesn't respect `tasks_for` and pulls tasks that shouldn't run on untrusted PRs (more details in [bug 1642977 comment 0](https://bugzilla.mozilla.org/show_bug.cgi?id=1642977#c0))
 4. `soft-dependencies` actually don't work (I tried earlier today) because we pass all tasks ID to `are_dependencies_completed.py` and this command cannot be reevaluated after during the optimization phase.
 5. We need to optimize away A-C tasks when only focus changes and https://github.com/mozilla-mobile/android-components/pull/7617 doesn't cover all type of tasks (just the build ones). 

That's why I rewrote these tasks as a morph one. The original idea is Tom Prince's. Left more details and pointed in [bug 1642977 comment 3](https://bugzilla.mozilla.org/show_bug.cgi?id=1642977#c3). 

I think most of this code should eventually be upstreamed to standalone taskgraph because there is close to no bits specific to Mobile (except the email notification). I'm implementing it here to see how it behaves in Mobile, then upstream it later.

Let me know how that looks 🙂


ahal (changes requested):
See previous comment :)

PR #253, opened 2022-12-02
Bug 1803141 - part 1-6: Build/test components only when they change

JohanLorenzo:
Depends on https://github.com/mozilla-mobile/firefox-android/pull/243

I've had this PR at the back of my mind for many months. We optimized the number of build jobs we spawn in https://github.com/mozilla-mobile/android-components/pull/7617. That said, at that time we knew we implemented a hack. I left more details in [bug 1803141 comment 0](https://bugzilla.mozilla.org/show_bug.cgi?id=1803141#c0). 

Basically, this PR removes the hack and makes the logic more generic. This enable UI tests as well as `samples-browser` to be scheduled only when necessary. Before then, they were run for any type of change. In addition to that, we **won't schedule all A-C jobs on pushes to `main` anymore**. The reason is: with Focus coming to the monorepo soon, it doesn't make sense to run all these jobs even though the underlying code hasn't changed. 

I also took the liberty to factorize the configuration of the UI tests and I realized a couple of jobs weren't testing the right thing nor they were even scheduled. I re-enabled them and they're luckily green. 

Suggested reviews: 
 * @AaronMT: part 4 to 6. They're about UI tests and when we should schedule them
 * @csadilek: part 3, 6, and 7. They change when tasks are scheduled based on what changed file
 * @jcristau: all commits but part 6 and 7 maybe?

### DIFF
--- a/taskcluster/ac_taskgraph/transforms/gradle_optimization.py
+++ b/taskcluster/ac_taskgraph/transforms/gradle_optimization.py
@@ -1,0 +1,28 @@
+from taskgraph.transforms.base import TransformSequence
+
+from ..build_config import get_path, ANDROID_COMPONENTS_DIR
+from ..gradle import get_upstream_deps_for_gradle_projects
+
+
+transforms = TransformSequence()
+
+
+@transforms.add
+def extend_optimization_if_one_already_exists(config, tasks):
+    deps_per_component = get_upstream_deps_for_gradle_projects(ANDROID_COMPONENTS_DIR)
+
+    for task in tasks:
+        optimization = task.get("optimization")
+        if optimization:
+            skip_unless_changed = optimization["skip-unless-changed"]
+
+            component = task["attributes"]["component"]
+            dependencies = deps_per_component[component]
+            component_and_deps = [component] + dependencies
+
+            skip_unless_changed.extend([
+                f"android-components/{get_path(component)}/**"
+                for component in component_and_deps
+            ])
+
+        yield task
--- a/taskcluster/ac_taskgraph/transforms/gradle_optimization.py
+++ b/taskcluster/ac_taskgraph/transforms/gradle_optimization.py
@@ -17,6 +17,10 @@ def extend_optimization_if_one_already_exists(config, tasks):
             skip_unless_changed = optimization["skip-unless-changed"]
 
             component = task["attributes"]["component"]
+            # TODO Remove this special case when ui-test.sh is able to accept "browser-engine-gecko"
+            if component == "browser":
+                component = "browser-engine-gecko"
+
             dependencies = deps_per_component[component]
             component_and_deps = [component] + dependencies
 
--- a/taskcluster/ac_taskgraph/transforms/ui_tests.py
+++ b/taskcluster/ac_taskgraph/transforms/ui_tests.py
@@ -1,0 +1,52 @@
+from taskgraph.transforms.base import TransformSequence
+
+
+transforms = TransformSequence()
+
+
+_ANDROID_TASK_NAME_PREFIX = "android-"
+
+
+@transforms.add
+def set_component_attribute(config, tasks):
+    for task in tasks:
+        component_name = task.pop("component", None)
+        if not component_name:
+            task_name = task["name"]
+            if task_name.startswith(_ANDROID_TASK_NAME_PREFIX):
+                component_name = task_name[len(_ANDROID_TASK_NAME_PREFIX):]
+            else:
+                raise NotImplementedError(f"Cannot determine component name from task {task_name}")
+
+        attributes = task.setdefault("attributes", {})
+        attributes["component"] = component_name
+
+        yield task
+
+
+@transforms.add
+def define_ui_test_command_line(config, tasks):
+    for task in tasks:
+        run = task.setdefault("run", {})
+        post_gradlew = run.setdefault("post-gradlew", [])
+        post_gradlew.append(
+            ['automation/taskcluster/androidTest/ui-test.sh', task["attributes"]["component"], 'arm', '1']
+        )
+
+        yield task
+
+
+@transforms.add
+def define_treeherder_symbol(config, tasks):
+    for task in tasks:
+        treeherder = task.setdefault("treeherder")
+        treeherder.setdefault("symbol", f"{task['attributes']['component']}(unit)")
+
+        yield task
+
+
+@transforms.add
+def define_description(config, tasks):
+    for task in tasks:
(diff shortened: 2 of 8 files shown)


taskcluster/ac_taskgraph/transforms/ui_tests.py
```diff
@@ -0,0 +1,52 @@
...
+        yield task
+
+
+@transforms.add
+def define_ui_test_command_line(config, tasks):
+    for task in tasks:
+        run = task.setdefault("run", {})
+        post_gradlew = run.setdefault("post-gradlew", [])
+        post_gradlew.append(
+            ['automation/taskcluster/androidTest/ui-test.sh', task["attributes"]["component"], 'arm', '1']
```
AaronMT, line 33: We have something in our backlog for a while to make this more robust without needing additional parameters like the arch and shard count here (legacy decision 😄 ) but for the time-being this is fine.

AaronMT (approved):

taskcluster/ac_taskgraph/transforms/gradle_optimization.py
```diff
@@ -0,0 +1,32 @@
...
+
+    for task in tasks:
+        optimization = task.get("optimization")
+        if optimization:
+            skip_unless_changed = optimization["skip-unless-changed"]
+
+            component = task["attributes"]["component"]
+            # TODO Remove this special case when ui-test.sh is able to accept "browser-engine-gecko"
+            if component == "browser":
+                component = "browser-engine-gecko"
```
csadilek, line 22: I don't understand this workaround. Can you elaborate? What is the browser component and where does it get added? :)

csadilek (approved):
This looks great extending skip_unless_changed! I have one question below about a workaround I don't understand. ✅

taskcluster/ac_taskgraph/transforms/ui_tests.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+
+        yield task
+
+
+@transforms.add
+def define_treeherder_symbol(config, tasks):
+    for task in tasks:
+        treeherder = task.setdefault("treeherder")
+        treeherder.setdefault("symbol", f"{task['attributes']['component']}(unit)")
```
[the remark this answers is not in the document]
JohanLorenzo: Great question! In this case, the component (e.g.: `samples-glean`) is the group symbol. `unit` is the job name. The idea is to regroup the test next to the build:

image

I took this liberty because I get the feeling it's better to bring related things closer together. That said, I'm happy to get back to the existing behavior 🙂

taskcluster/ac_taskgraph/transforms/ui_tests.py
```diff
@@ -0,0 +1,52 @@
...
+
+        yield task
+
+
+@transforms.add
+def define_description(config, tasks):
+    for task in tasks:
+        task.setdefault("description", f"Run unit tests on device for {task['attributes']['component']}")
```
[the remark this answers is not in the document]
JohanLorenzo: Just to make sure I understand: should I put `Run unit/ui tests`? Or do you want me to replace `unit` by `ui` (as in `s/unit/ui`)?

taskcluster/ac_taskgraph/transforms/ui_tests.py
JohanLorenzo: Sounds good! I'm happy to help make this code more generics once you all feel ready 🙂

taskcluster/ac_taskgraph/transforms/gradle_optimization.py
JohanLorenzo: I implemented this workaround to keep this existing test working: 

https://github.com/mozilla-mobile/firefox-android/blob/54c1653a88edb76472c9579ed7834ae61b6e98e0/taskcluster/ci/test/kind.yml#L56-L62

Unlike other `ui-test.sh` calls, this time we don't pass an actual Android Component, we just pass `browser`. I might have misunderstood. I based this implementation on the config just above and on this piece of shell script:

https://github.com/mozilla-mobile/firefox-android/blob/54c1653a88edb76472c9579ed7834ae61b6e98e0/android-components/automation/taskcluster/androidTest/ui-test.sh#L85-L97

@AaronMT when we call `ui-test.sh browser`, do we test `browser-engine-gecko`?

taskcluster/ac_taskgraph/transforms/ui_tests.py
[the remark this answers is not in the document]
AaronMT: If this is a description for everything I would use `unit/ui`

taskcluster/ac_taskgraph/transforms/gradle_optimization.py
AaronMT: It was previously `unit-browser-engine-gecko-nightly`, and `ui-browser` `was samples-browser`

taskcluster/ac_taskgraph/transforms/gradle_optimization.py
```diff
@@ -17,6 +17,10 @@ def extend_optimization_if_one_already_exists(config, tasks):
             skip_unless_changed = optimization["skip-unless-changed"]
 
             component = task["attributes"]["component"]
+            # TODO Remove this special case when ui-test.sh is able to accept "browser-engine-gecko"
```
jcristau, line 20: nit: if the special-case is needed, could we put it in the yaml (e.g. have 2 different variables for the ui-test.sh argument and component name, that default to the same thing but can be overridden)?

jcristau (approved):
Looks like a nice improvement, thanks.

taskcluster/ac_taskgraph/transforms/ui_tests.py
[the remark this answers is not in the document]
JohanLorenzo: Thanks, done!

taskcluster/ac_taskgraph/transforms/gradle_optimization.py
JohanLorenzo: Okay, I'm keeping the workaround as is then. I'm happy to change it if I misunderstood something.

taskcluster/ac_taskgraph/transforms/gradle_optimization.py
JohanLorenzo: I think we should get rid of the special-case by changing `ui-test.sh` and then removing this if statement. If we create 2 variables in the YAML, it's less obvious that this special case is not here to stay. I'm going to land this code as is. I'm happy to change it if you feel I should.

PR #262, opened 2022-12-06
Bug 1796301 - Add paddingEnd in OriginView for RTL config

rahulsainani:
### What
• Add `paddingEnd` for RTL layouts, we could add padding for LTR as well and set it directly in the layout, but the idea was to preserve the space in `OriginView` for the title/url.
• No tests since only RTL layout was affected.

### Screenshots

| Hebrew as RTL example | English as LTR example - unchanged | Desktop RTL |
|------|-----|----|
|Screenshot_20221206_141521|Screenshot_20221206_132722|Screen Shot 2022-12-07 at 10 45 34 AM|

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
+++ b/android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- DisplayToolbar -->
+    <dimen name="mozac_browser_toolbar_origin_padding_end">16dp</dimen>
+</resources>
--- a/android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
+++ b/android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
@@ -1,4 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources>
     <!-- DisplayToolbar -->
     <dimen name="mozac_browser_toolbar_origin_padding_end">16dp</dimen>
--- a/android-components/components/browser/toolbar/src/main/res/layout/mozac_browser_toolbar_displaytoolbar.xml
+++ b/android-components/components/browser/toolbar/src/main/res/layout/mozac_browser_toolbar_displaytoolbar.xml
@@ -104,6 +104,7 @@
         android:layout_width="0dp"
         android:layout_height="40dp"
         android:layout_marginTop="8dp"
+        android:paddingEnd="@dimen/mozac_browser_toolbar_origin_padding_end"
         app:layout_constraintEnd_toStartOf="@+id/mozac_browser_toolbar_page_actions"
         app:layout_constraintStart_toEndOf="@+id/mozac_browser_toolbar_security_indicator"
         app:layout_constraintTop_toTopOf="parent"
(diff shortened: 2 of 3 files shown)


android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
```
gabrielluong, line 2: Could you add the Mozilla license here? You can see an example with the `dimens.xml` file below.

android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
```diff
@@ -0,0 +1,5 @@
...
+    <!-- DisplayToolbar -->
+    <dimen name="mozac_browser_toolbar_origin_padding_end">16dp</dimen>
```
gabrielluong, line 4: Just for my own curiosity, I am wondering how we came up with 16dp?

rahulsainani: Sure, it looked consistent with the other/icon paddings and at the start/end. Happy to update if there's a specific grid we adhere to.

android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
rahulsainani: Good catch, just added!

android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
```diff
@@ -0,0 +1,5 @@
...
+</resources>
```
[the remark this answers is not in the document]
rahulsainani: Just added 👍

android-components/components/browser/toolbar/src/main/res/values-ldrtl/dimens.xml
gabrielluong: No worries, I think I just need a moment later today to pull the down the change and try to rationalize the padding of everything and see if the value fits with the existing layout. Normally, we would ask UX to double check if the screenshots look right, that said visually it looks right to me.

rahulsainani: Sounds good! 👍

gabrielluong:
@rahulsainani Just a quick status update, I am still digging through the Display Toolbar code, it's just been so long since I last looked at this and what I am kinda looking to piece together is how things are laid out and how do some of the views get their padding. It's not quite all in the XML and I am actually getting the sense that the approach might not necessarily be correct due to the lack of horizontal padding/margin in the XML. 

However, I also took a look at Desktop to see how does the RTL toolbar look, and I believe there's still more work to do if we actually wanted to implement RTL for the OriginView correctly. I am gonna ping some folks to weigh in.
Screen Shot 2022-12-07 at 10 45 34 AM

gabrielluong (changes requested):
I am gonna request changes since I believe the correct solution to this is align the text to the right for RTL configurations.

rahulsainani:
Thanks again for checking the desktop and mobile in RTL configs. Will wait for the UX feedback on whether to do it the fennick/chrome that you shared on slack (similar to this PR) or whether to make the text "aligned" to the right, for the lack of the better term.

gabrielluong (approved):
Spoke with @jonalmeida and @Amejia481 who're okay with the current approach and we'll make the best use of the situation to improve things in the meantime. We will look to follow up about seeing what the right look for RTL configuration should be after hearing back from Product/UX.

Amejia481 (approved):

PR #264, opened 2022-12-06
Bug 1800146 - Set tab's GeckoSession priority based on form data

calumozilla:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1800146


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
```diff
@@ -54,27 +66,59 @@ class SessionPrioritizationMiddleware : Middleware<BrowserState, BrowserAction>
...
 
-        // We need to make sure, we alter the previousHighestPriorityTabId, after the session is linked.
+        // We need to make sure we alter the previousHighestPriorityTabId, after the session is linked.
         // So we update the priority on the engine session, as we could get actions where the tab
         // is selected but not linked yet, causing out sync issues,
         // when previousHighestPriorityTabId didn't call updateSessionPriority()
         if (currentEngineSession != null) {
-            previousSelectedTab?.engineState?.engineSession?.updateSessionPriority(DEFAULT)
+            // check for existing form data here and only set DEFAULT if there is
+            previousSelectedTab?.engineState?.engineSession?.checkForFormData()
```
jonalmeida, line 104: I think only this line needs to be wrapped in the `mainScope` call. We can avoid doing excessive work on the UI thread that way (e.g. finding the previous tab) I think, unless there is some reason we are trying to avoid some async situation of finding and reacting to the previous tab?

jonalmeida (commented):
Thanks for the draft - left a comment about the line we were discussing offline. Haven't reviewed the rest.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/EngineObserver.kt
```diff
@@ -404,4 +404,13 @@ internal class EngineObserver(
...
         store.dispatch(EngineAction.SaveToPdfExceptionAction(tabId, throwable))
     }
+
+    override fun onCheckForFormData(containsFormData: Boolean) {
+        store.dispatch(ContentAction.CheckForFormDataAction(tabId, containsFormData))
+    }
+
+    override fun onCheckForFormDataException(throwable: Throwable) {
+        // fix to dispatch CheckForFormDataException
+        store.dispatch(EngineAction.SaveToPdfExceptionAction(tabId, throwable))
```
Amejia481, line 414: nit: We will need a new action :)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
```diff
@@ -299,6 +299,14 @@ internal object ContentStateReducer {
...
                 it.copy(showToolbarAsExpanded = action.expanded)
             }
+            is ContentAction.CheckForFormDataAction -> updateContentState(state, action.tabId) {
+                // todocathy there is no store parameter to update, but this cannot be removed
+                it.copy(desktopMode = action.containsFormData)
+            }
+            is ContentAction.ClearHighPrioritySessionAction -> updateContentState(state, action.tabId) {
+                // todocathy there is no store parameter to update, but this cannot be removed
+                it.copy(desktopMode = action.tabId == "0")
+            }
```
Amejia481, line 309: Should we create a new property on [ContentState](https://github.com/mozilla-mobile/firefox-android/blob/e1ccf1865c6f8485801c05a544d80f52032006a4/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/ContentState.kt#L64) to represent `containsFormData`?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
```diff
@@ -54,27 +66,59 @@ class SessionPrioritizationMiddleware : Middleware<BrowserState, BrowserAction>
                     updatePriorityIfNeeded(state)
                 }
             }
+            is ContentAction.CheckForFormDataAction -> {
+                val tab = context.state.findTab(action.tabId)
+                if (action.containsFormData) {
+                    tab?.engineState?.engineSession?.updateSessionPriority(HIGH)
+                    logger.info("Update the tab ${tab?.id} priority to ${HIGH.name}")
+                    clearHighPriority(context, tab!!.id)
```
Amejia481, line 74: Maybe we could use `.let` to avoid the use of `!!`
```kotlin
                    tab?.let {
                        clearHighPriority(context, tab.id)
                    }
```

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
calumozilla: Yes, I should've followed up on the design doc about the change. I had asked you on there if there was a specific action to tell us when a tab is un-selected, because the initial idea I had about using Unlink won't work. Unlink only occurs when a tab is deleted. We need to use the combined action of TabListAction and Link to modify the previous selected tab.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
calumozilla: I'm working on updating this to throw an exception that should direct users to use the middleware instead. 
`throw IllegalStateException("You need to add SessionPrioritizationMiddleware to your BrowserStore. ($action)")`

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
calumozilla: Good idea, I'll update this.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
Amejia481: This should  achieve it, I was just curios 😅

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
```diff
@@ -302,6 +302,12 @@ internal object ContentStateReducer {
             is ContentAction.UpdateExpandedToolbarStateAction -> updateContentState(state, action.sessionId) {
                 it.copy(showToolbarAsExpanded = action.expanded)
             }
+            is ContentAction.CheckForFormDataAction,
+            is ContentAction.ClearHighPrioritySessionAction,
+            is ContentAction.CheckForFormDataExceptionAction,
```
Amejia481, line 307: How we are going handle this action? Is this going to be on the app?

calumozilla: Ya Fenix will need to address the action, similar to SaveToPDF: https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt#34

Amejia481: Ok, in that case ,what are we going to show to users?  As it's not completely clear how they are going to benefit from know that the exception happened? :)

Amejia481: Are they going to have a way to recover? As in the case of the `SaveToPDF` one we are notifying users because they will be waiting for the PDF to be generated, in this cases it looks like this is something that is going to happen without users know it.

calumozilla: That's a good question. There wouldn't be a need to notify users that the exception happened. It'd be only helpful for the developer to know if for some unforeseen reason the GV API failed. Do you have a recommendation for how an app or Fenix should handle this exception, or maybe this exception is not necessary?

Amejia481: >  Do you have a recommendation for how an app or Fenix should handle this exception, or maybe this exception is not necessary?

Maybe we could just log the info, but if we would like to keep track of how many times this happens, we could send a glean event from the app.

calumozilla: Ya that sounds like a good plan. Are there examples of logging to glean in Fenix?

Amejia481: Yes, we did on the Save to PDF feature https://github.com/mozilla-mobile/fenix/pull/27666/

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
```diff
@@ -60,21 +98,34 @@ class SessionPrioritizationMiddleware : Middleware<BrowserState, BrowserAction>
...
+            }
+
             currentEngineSession.updateSessionPriority(HIGH)
             logger.info("Update the currentSelectedTab ${currentSelectedTab.id} priority to ${HIGH.name}")
-            logger.info("Update the previousSelectedTab ${previousSelectedTab?.id} priority to ${DEFAULT.name}")
             previousHighestPriorityTabId = currentSelectedTab.id
         }
     }
+
+    private fun clearHighPriority(context: MiddlewareContext<BrowserState, BrowserAction>, tabId: String) {
```
jonalmeida, line 122: We should probably add some tests for this to know that we are dispatching the action to clear high priority.

There is a `CaptureActionsMiddleware` you can add to your test `BrowserStore` to catch the actions that go through the store and make assertions with it.

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddlewareTest.kt
```diff
@@ -12,13 +13,21 @@ import mozilla.components.browser.state.store.BrowserStore
...
 import org.junit.Assert.assertEquals
+import org.junit.Rule
 import org.junit.Test
 import org.mockito.Mockito.verify
 
 class SessionPrioritizationMiddlewareTest {
+    @get:Rule
+    val coroutinesTestRule = MainCoroutineRule()
+    private val dispatcher = coroutinesTestRule.testDispatcher
+    private val logger = Logger("SessionPrioritizationMiddlewareTest")
```
jonalmeida, line 30: nit: unused logger and dispatcher.

jonalmeida (commented):
This looks really good - just some minor questions and nits.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt
calumozilla: Sounds good, I created a separate task for this on Fenix: https://github.com/mozilla-mobile/fenix/issues/28280. Once the other comments are resolved, this patch can be merged first, without the telemetry in place, correct?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
```diff
@@ -49,7 +83,11 @@ class SessionPrioritizationMiddleware : Middleware<BrowserState, BrowserAction>
             is TabListAction,
```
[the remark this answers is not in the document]
calumozilla: We should cancel the job if restoring means bringing that tab back to the selected tab. Same for `MoveTabsAction`, if any of the tabs are moved into the active/selected tab. If that's not the case, then we could specify the actions like:
```
when (action) {
    is TabListAction.AddTabAction,
    is TabListAction.AddMultipleTabsAction,
    is TabListAction.SelectTabAction,
    is TabListAction.RemoveTabAction,
    is TabListAction.RemoveTabsAction,
    is EngineAction.LinkEngineSessionAction,
  ```

File: android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt (hunk 6 above)
Comment (calumozilla) [reply android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt:122]: So when I was adding this test, `CaptureActionsMiddleware` errored out with `no action found` even though I can see printouts from the clear high priority action. I did notice it was able to capture the action when it passes through the `ContentStateReducer` instead of throwing the exception `throw IllegalStateException("You need to add SessionPrioritizationMiddleware. ($action)")`. I observed the same thing with other actions that also throw exceptions `RemoveThumbnailAction, UpdateThumbnailAction`. I'm wondering if this is a possible limitation to `CaptureActionsMiddleware`.

Let me know if it's not clear, we can talk about it

File: android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt (hunk 5 above)
Comment (calumozilla) [reply android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/ContentStateReducer.kt:307]: Fenix telemetry issue and patch, once this AC patch is approved and merged: https://github.com/mozilla-mobile/fenix/pull/28444

File: android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt [hunk 9]
```diff
@@ -790,6 +790,22 @@ sealed class ContentAction : BrowserAction() {
...
+    /**
+     * Updates the [ContentState] with the provided [tabId] to the appropriate priority based on any
+     * existing form data.
+     */
+    data class CheckForFormDataAction(val tabId: String, val containsFormData: Boolean) : ContentAction()
+
+    /**
+     * Lowers priority of the [tabId] after certain period of time
+     */
+    data class ClearHighPrioritySessionAction(val tabId: String) : ContentAction()
```
Amejia481, line 803: nit: Not sure if from the name is easy to determine that the priority will be lowered after certain period of time.

Amejia481: From the name it sounds a it will clear the high priority for the given session, `SetDefaultPriorityAfterTime` or something similar 😅

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
```diff
@@ -16,13 +22,19 @@ import mozilla.components.concept.engine.EngineSession.SessionPriority.HIGH
...
 import mozilla.components.lib.state.MiddlewareContext
 import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.base.coroutines.Dispatchers as MozillaDispatchers
 
 /**
  * [Middleware] implementation responsible for updating the priority of the selected [EngineSession]
  * to [HIGH] and the rest to [DEFAULT].
  */
-class SessionPrioritizationMiddleware : Middleware<BrowserState, BrowserAction> {
+class SessionPrioritizationMiddleware(
```
Amejia481, line 31: It will be nice to have a way to test this in nightly, without the pressure of having to revert in case something goes wrong or we don't want it ride the trains until collecting enough info. Are we going to duplicate this middleware or do we have another way to test the new behaviour and keeping the old one as fallback ?

calumozilla: If we create a new middleware with the new implementation, how would we distinguish between using the new vs the old? A flag somewhere?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
calumozilla: Sounds good, will update the name

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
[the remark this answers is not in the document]
jonalmeida: Right, thanks for reminding me that actions which affect the selected tab are the ones we care about here. I don't think `MoveTabsAction` will affect the selected tab, but that might be okay to have it included.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
jonalmeida: Are you adding the `CaptureActionsMiddleware` as the last middleware in the list?

When I tried to replicate what you described above with the `ThumbnailsMiddlewareTest`, I can comment out the `UpdateThumbnailAction` action in the test to see the middleware fail correctly.

<details>

<summary>Stack trace</summary>

```
Action of type class mozilla.components.browser.state.action.ContentAction$UpdateThumbnailAction was dispatched: UpdateThumbnailAction(sessionId=test-tab1, thumbnail=Mock for Bitmap, hashCode: 1365863352)
java.lang.AssertionError: Action of type class mozilla.components.browser.state.action.ContentAction$UpdateThumbnailAction was dispatched: UpdateThumbnailAction(sessionId=test-tab1, thumbnail=Mock for Bitmap, hashCode: 1365863352)
	at mozilla.components.support.test.middleware.CaptureActionsMiddleware.assertNotDispatched(CaptureActionsMiddleware.kt:75)
	at mozilla.components.browser.thumbnails.ThumbnailsMiddlewareTest.thumbnail actions are the only ones consumed by the middleware(ThumbnailsMiddlewareTest.kt:171)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
```

</details>

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
Amejia481: A flag in the constructor could be an alternative too, @jonalmeida do you have any ideas :) ?

jonalmeida: We can do this in Fenix with a feature flag around the middleware in the list.

Amejia481: Should we go then with the flag in the constructor? :)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
calumozilla: Ya I added it as the last middleware.

If you still have it, would you be able to paste the modified test?  When I have just this in the test, it should work but won't capture the action.

```
store.dispatch(ContentAction.UpdateThumbnailAction("test-tab1", mock())).joinBlocking()
capture.assertLastAction(ContentAction.UpdateThumbnailAction::class) {}
```

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
jonalmeida: Sorry, when I sent my comment, I hadn't seen any of the other previous messages. Adding a flag in the constructor seems overly complicated to do since we're modifying the existing middleware and a flag around the middleware also doesn't make sense since we need the existing behaviour to continue (I confused this feature with the `trimMemoryAutomatically` flag).

@calumozilla has a good point about distinguishing the old with the new behaviour. Maybe the best way forward is to land this after the next release and monitor crash rates + ask users that had the bug with form data disappearing to reproduce again. We will have time to back out if needed then.

What do you both think?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
calumozilla: Talked to Jon and resolved this by specifying the order of middleware in the unit test as `middleware = listOf(capture, middleware)`. This allows the capture middleware to be invoked first. That way the action is added to the list of `capturedActions` before it reaches the SessionPrioritizationMiddleware and is returned without reaching the `ContentStateReducer`.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
```diff
@@ -16,13 +22,21 @@ import mozilla.components.concept.engine.EngineSession.SessionPriority.HIGH
...
 
 /**
  * [Middleware] implementation responsible for updating the priority of the selected [EngineSession]
  * to [HIGH] and the rest to [DEFAULT].
+ *
+ * @property updatePriorityAfterMillis Update priority to default after timeout.
  */
-class SessionPrioritizationMiddleware : Middleware<BrowserState, BrowserAction> {
+class SessionPrioritizationMiddleware(
+    private val updatePriorityAfterMillis: Long = 300000,
```
kycn, line 34: nit: maybe we could create a constant for this value instead of a magic number and explain what it does and its default value's meaning with a comment

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddlewareTest.kt
```diff
@@ -64,7 +97,7 @@ class SessionPrioritizationMiddlewareTest {
     }
 
     @Test
-    fun `GIVEN a previous selected tab with priority DEFAULT WHEN selecting and linking a new tab THEN update the previous tab to DEFAULT and the new one to HIGH`() {
+    fun `GIVEN a previous selected tab with priority DEFAULT WHEN selecting and linking a new tab THEN update the new one to HIGH and the previous tab based on if it contains form data`() = runTestOnMain {
```
kycn, line 100: nit/out of curiosity: does this mean that we can have 2 HIGH prio tabs/sessions at the same time?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
```diff
@@ -60,21 +100,38 @@ class SessionPrioritizationMiddleware : Middleware<BrowserState, BrowserAction>
...
+
+    private fun updatePriorityToDefault(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        tabId: String,
+        updatePriorityAfterMillis: Long,
+    ) {
+        // store and launch the new job related to the tabId
+        var updateJob: Job = waitScope.launch {
+            delay(updatePriorityAfterMillis)
+            context.store.dispatch(ContentAction.UpdatePriorityToDefaultAfterTimeoutAction(tabId))
```
kycn, line 132: nit/out of curiosity: is this a common approach to dispatch an action inside an action that are privately interrelated to each other, i.e. `UpdatePriorityToDefaultAfterTimeoutAction` is (as far as i can imagine) internal and only dispatched whenever `CheckForFormDataAction` is triggered. As there is no external dispatching logic to this action, could we have just include the `UpdatePriorityToDefaultAfterTimeoutAction`'s lines inside middleware in  the `CheckForFormDataAction` case? No strong opinion here, and I may be wrong with my assumptions (i.e. the `UpdatePriorityToDefaultAfterTimeoutAction` is only internally handled by `CheckForFormDataAction`).

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -521,6 +521,28 @@ class GeckoEngineSession(
...
     }
 
+    /**
+     * See [EngineSession.checkForFormData].
+     */
+    override fun checkForFormData() {
+        geckoSession.containsFormData().then(
+            { result ->
+                if (result == null) {
+                    logger.error("No result from GeckoView containsFormData.")
```
kycn, line 531: is this an error that we would want to measure the telemetries as well? What does `null` indicate here about the containsFormData bool?

kycn (commented):
Great job 🚀  just left a couple of comments which, i think, one of them maybe an important one (the one related to the `result==null` case for the containsFormData bool). When it catches the attention and resolved, I would be happy to approve. 
💯 💯

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
calumozilla: Good point, I added a comment for the number

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddlewareTest.kt
calumozilla: Hey Kaya, yes it would be possible to have multiple high priority tabs. Any background tab with form data may have a high priority for a certain time.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
calumozilla: It's true we could probably measure this event as well. To be honest I don't think the telemetry will be all that useful, it seems unlikely to me that GeckoView will return null or even throw. We could remove this, but this form data feature is just to test and see if it will have a significant impact on users who notice their tabs being killed more often. Perhaps we could add more telemetry in the future.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
calumozilla: I think it's an okay approach, it also helps with unit tests to be able to verify any action occurred. Since in either case, eventually the priority will result in `DEFAULT`.

calumozilla:
> Great job 🚀 just left a couple of comments which, i think, one of them maybe an important one (the one related to the `result==null` case for the containsFormData bool). When it catches the attention and resolved, I would be happy to approve. 💯 💯

Thanks for taking a look and reviewing!

jonalmeida (approved):
This looks good to me - thanks for pushing this through!

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
kycn: yes agreed, this should be a next step in case it is worth it :)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/SessionPrioritizationMiddleware.kt
kycn: yes; makes sense. Just wondered whether it was smth we were avoiding or not :)

Amejia481 (approved):
🚀

Amejia481:
@calumozilla awesome work 💪🏽 !!

Feel free to add the label **"🛬 needs landing"** when you are ready to land.

PR #269, opened 2022-12-07
Bug 1804157 - Revert breadcrumb coroutine scope changes

bendk:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
+++ b/android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
@@ -181,13 +181,11 @@ class CrashReporter(
      * ```
      */
     override fun recordCrashBreadcrumb(breadcrumb: Breadcrumb) {
-        scope.launch {
-            if (crashBreadcrumbs.size >= maxBreadCrumbs) {
-                crashBreadcrumbs.removeAt(0)
-            }
-
-            crashBreadcrumbs.add(breadcrumb)
+        if (crashBreadcrumbs.size >= maxBreadCrumbs) {
+            crashBreadcrumbs.removeAt(0)
         }
+
+        crashBreadcrumbs.add(breadcrumb)
     }
 
     internal fun onCrash(context: Context, crash: Crash) {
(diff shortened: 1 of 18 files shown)


Amejia481 (approved):

android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
```diff
@@ -181,13 +181,11 @@ class CrashReporter(
...
-            if (crashBreadcrumbs.size >= maxBreadCrumbs) {
-                crashBreadcrumbs.removeAt(0)
-            }
-
-            crashBreadcrumbs.add(breadcrumb)
+        if (crashBreadcrumbs.size >= maxBreadCrumbs) {
+            crashBreadcrumbs.removeAt(0)
         }
+
+        crashBreadcrumbs.add(breadcrumb)
```
jonalmeida, line 188: I think I just figured out the bug: crash reporting is setup in _all processes_ so every tab, media, and main process have access to the singleton and to adding breadcrumbs. This is where multiple accessors are probably coming from. 🙂 

We could fix this by adding a `@Synchronized` annotation on the `crashBreadcrumbs` to cover this.

We can still back this out and land the fix after the 109 cut if that feels safer.

cc @csadilek who has seen bugs like this and might have other suggestions.

jonalmeida (approved):

android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt
Mugurell: Not sure just adding `@Synchronized` would solve the concurrent access issue.
There's some documentation about handling this in [the documentation for the `synchronizedList` method](https://cs.android.com/android/platform/superproject/+/master:libcore/ojluni/src/main/java/java/util/Collections.java;l=2500-2512).

jonalmeida: That's a much better suggestion than mine. :)

Since there are more details to discuss, I think we should back this out then since any direction will require some bake time in nightly and we're about the cut beta on Monday.

PR #273, opened 2022-12-08
Bug 1804594 - Differentiate search engines between "general" and "topic specific" when they are loaded

gabrielluong:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/search/SearchEngine.kt
+++ b/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/search/SearchEngine.kt
@@ -19,6 +19,7 @@ const val OS_SEARCH_ENGINE_TERMS_PARAM = "{" + "searchTerms" + "}"
  * @property type the type of this search engine.
  * @property resultUrls the list of the queried suggestions result urls.
  * @property suggestUrl the search suggestion url.
+ * @property isGeneral whether the search engine is a general search engine.
  */
 data class SearchEngine(
     val id: String,
@@ -27,6 +28,7 @@ data class SearchEngine(
     val type: Type,
     val resultUrls: List<String> = emptyList(),
     val suggestUrl: String? = null,
+    val isGeneral: Boolean = false,
 ) {
     /**
      * A enum class representing a search engine type.
--- a/android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/SearchEngineReader.kt
+++ b/android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/SearchEngineReader.kt
@@ -22,6 +22,23 @@ internal const val URL_TYPE_SUGGEST_JSON = "application/x-suggestions+json"
 internal const val URL_TYPE_SEARCH_HTML = "text/html"
 internal const val URL_REL_MOBILE = "mobile"
 internal const val IMAGE_URI_PREFIX = "data:image/png;base64,"
+internal const val GOOGLE_ID = "google"
+
+// List of general search engine ids, taken from
+// https://searchfox.org/mozilla-central/rev/ef0aa879e94534ffd067a3748d034540a9fc10b0/toolkit/components/search/SearchUtils.sys.mjs#200
+internal val GENERAL_SEARCH_ENGINE_IDS = setOf(
+    GOOGLE_ID,
+    "ddg",
+    "bing",
+    "baidu",
+    "ecosia",
+    "qwant",
+    "yahoo-jp",
+    "seznam-cz",
+    "leit-is",
+    "coccoc",
+    "baidu",
+)
 
 /**
  * A simple XML reader for search engine plugins.
@@ -47,7 +64,17 @@ internal class SearchEngineReader(
             type = type,
             resultUrls = resultsUrls,
             suggestUrl = suggestUrl,
+            isGeneral = isGeneralSearchEngine(identifier, type),
         )
+
+        /**
+         * Returns true if the provided [type] is a custom search engine or the [identifier] is
+         * included in [GENERAL_SEARCH_ENGINE_IDS].
+         */
+        private fun isGeneralSearchEngine(identifier: String, type: SearchEngine.Type): Boolean =
+            type == SearchEngine.Type.CUSTOM ||
+                identifier.startsWith(GOOGLE_ID) ||
+                GENERAL_SEARCH_ENGINE_IDS.contains(identifier)
     }
 
     /**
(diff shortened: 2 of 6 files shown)


android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/SearchEngineReader.kt
```diff
@@ -22,6 +23,19 @@ internal const val URL_TYPE_SUGGEST_JSON = "application/x-suggestions+json"
 internal const val URL_TYPE_SEARCH_HTML = "text/html"
 internal const val URL_REL_MOBILE = "mobile"
 internal const val IMAGE_URI_PREFIX = "data:image/png;base64,"
+internal const val GOOGLE_ID = "google"
+
+// List of general search engine ids, taken from
+// https://searchfox.org/mozilla-central/rev/ef0aa879e94534ffd067a3748d034540a9fc10b0/toolkit/components/search/SearchUtils.sys.mjs#200
+internal val GENERAL_SEARCH_ENGINE_IDS = setOf(
+    GOOGLE_ID,
```
gabrielluong, line 31: I think we could've also just put `google-b-m` and `google-com-nocodes` instead of doing a `startsWith` as well. No strong opinions here. I went with `startsWith` because I saw there was `google-b-amzes` for Echo Show in my search https://searchfox.org/mozilla-mobile/search?q=google-&path=&case=false&regexp=false

android-components/components/feature/search/src/test/java/mozilla/components/feature/search/storage/SearchEngineReaderTest.kt
```diff
@@ -33,10 +35,11 @@ class SearchEngineReaderTest {
...
         val readSearchEngine = reader.loadFile(searchEngine.id, file)
-        Assert.assertEquals(searchEngine.id, readSearchEngine.id)
-        Assert.assertEquals(searchEngine.name, readSearchEngine.name)
-        Assert.assertEquals(searchEngine.type, readSearchEngine.type)
-        Assert.assertEquals(searchEngine.resultUrls, readSearchEngine.resultUrls)
+        assertEquals(searchEngine.id, readSearchEngine.id)
+        assertEquals(searchEngine.name, readSearchEngine.name)
+        assertEquals(searchEngine.type, readSearchEngine.type)
+        assertEquals(searchEngine.resultUrls, readSearchEngine.resultUrls)
+        assertFalse(readSearchEngine.isGeneral)
```
gabrielluong, line 42: Snuck the false check here

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/SearchEngineReader.kt
```diff
@@ -22,6 +22,19 @@ internal const val URL_TYPE_SUGGEST_JSON = "application/x-suggestions+json"
 internal const val URL_TYPE_SEARCH_HTML = "text/html"
 internal const val URL_REL_MOBILE = "mobile"
 internal const val IMAGE_URI_PREFIX = "data:image/png;base64,"
+internal const val GOOGLE_ID = "google"
+
+// List of general search engine ids, taken from
+// https://searchfox.org/mozilla-central/rev/ef0aa879e94534ffd067a3748d034540a9fc10b0/toolkit/components/search/SearchUtils.sys.mjs#200
+internal val GENERAL_SEARCH_ENGINE_IDS = setOf(
```
Mugurell, line 29: It seems there are other "general" search engines in `list.json`.
At a quick look there's at least Cốc Cốc[ for vietnamese](https://github.com/mozilla-mobile/firefox-android/blob/7fe93f305b499e01e749748fdc27981bff770187/android-components/components/feature/search/src/main/assets/search/list.json#L880) - https://coccoc.com/search.
I think we should "manually" go through the current search engines from `list.json` and ensure all general ones are added to this list.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/search/SearchEngine.kt
```diff
@@ -27,6 +28,7 @@ data class SearchEngine(
     val type: Type,
     val resultUrls: List<String> = emptyList(),
     val suggestUrl: String? = null,
+    val isGeneral: Boolean = false,
```
Mugurell, line 31: Wanted to ask about the direction to take in regards to user added search engines.
Because based on https://mozilla-hub.atlassian.net/browse/FNXV2-22038 only "general" search engines can be set as the default:
- should the default here be `true` or
- should we amend Fenix to go with "only general _or user added_ search engines can be set as the default"

cc @DreVla

gabrielluong: So, I thinking we will just set `isGeneral` when we call `createSearchEngine` in Fenix https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/settings/search/AddSearchEngineFragment.kt#180

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/SearchEngineReader.kt
```diff
@@ -47,7 +64,11 @@ internal class SearchEngineReader(
             type = type,
             resultUrls = resultsUrls,
             suggestUrl = suggestUrl,
+            isGeneral = type == SearchEngine.Type.CUSTOM || isGeneralSearchEngine(identifier),
```
gabrielluong, line 67: I don't know if I am happy with this check all custom search engines to be isGeneral. I could also do something similar to above with the variables `name`, `suggestUrl`, etc which will be set when it's reading the file instead of deriving `isGeneral` from the type.

gabrielluong:
@Mugurell I am wondering if we even need this patch if we could just derive that the search engine is general when it's a custom type or if it's bundled and part of the `GENERAL_SEARCH_ENGINE_IDS`, which could go directly into Fenix as well.

Mugurell:
> @Mugurell I am wondering if we even need this patch if we could just derive that the search engine is general when it's a `CUSTOM` type or if it's a `BUNDLED` and part of the `GENERAL_SEARCH_ENGINE_IDS`, which could go directly into Fenix as well.

That would work also and is what @DreVla has been doing in his Fenix drafts.
It could be an extension property or we could promote this property in the SearchEngine API. By adding this differentiation in AC this patch helps with encapsulating the logic for what a general/topic specific search engine is (clients should not need to be responsable for this (checking against that list of general search ids)) and allows to easily adapt behavior anywhere in Fenix and AC.
Both approaches seem okay to me but I would incline a bit more to going with with patch with the difference between an extension property and a new internal property to me being related to:
- whether we want to give more visibility to this by promoting it in the SearchEngine API and as I understand based on the recent discussions this separation will indeed serve as the basis for other important functionalities in the future on all platforms so for me it would make sense to be part of the "official" API,
- whether we want one eager (this patch) vs multiple lazy type inferences (an extension property). It's probably hard to track the performance differences but based on the above point I would still prefer the property to already be available.

Mugurell (approved):
LGTM, thanks!

DreVla:
I have been following this patch and also tested it for this ticket: https://github.com/mozilla-mobile/fenix/pull/28112. I have to agree with what @Mugurell said, rather taking this one patch as basis for the many other tasks involving the separation between the two search engines categories in fenix.

gabrielluong:
Sounds good, I will land this after the cut. We'll need to update the changelog.

PR #292, opened 2022-12-11
Bug 1805070 - Update detekt to v1.22.0 in A-C

gabrielluong:
I compared various base configs ([1.22.0](https://github.com/detekt/detekt/blob/v1.22.0/detekt-core/src/main/resources/default-detekt-config.yml), [1.19.0](https://github.com/detekt/detekt/blob/v1.19.0/detekt-core/src/main/resources/default-detekt-config.yml) and our original [1.0.0RC6](https://github.com/detekt/detekt/blob/RC6-4/detekt-cli/src/main/resources/default-detekt-config.yml)) to note any changes. Most of the changes will have come from the latest default base config which has turned on a lot more rules.

 I also annotated rules and configurations where our team has manually checked the default settings and added a link to the original issue (if it exists) or the PR.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1805070


android-components/build.gradle
```diff
@@ -320,8 +320,6 @@ detekt {
     config = files("$projectDir/config/detekt.yml")
     baseline = file("$projectDir/config/detekt-baseline.xml")
 
-    failFast = false
```
gabrielluong, line 323: The --fail-fast CLI flag (and failFast Gradle property) has been removed. It was deprecated since 1.16.x - https://github.com/detekt/detekt/pull/5290

See https://github.com/detekt/detekt/releases/tag/v1.22.0

android-components/config/detekt.yml
```diff
@@ -7,10 +7,17 @@ build:
     # style: 1
     # comments: 1
 
+config:
+  validation: true
+  warningsAsErrors: false
+  checkExhaustiveness: false
+  # when writing own rules with new properties, exclude the property path e.g.: 'my_rule_set,.*>.*>[my_property]'
+  excludes: ''
```
gabrielluong, line 15: This is just to align with the base default config https://github.com/detekt/detekt/blob/v1.22.0/detekt-core/src/main/resources/default-detekt-config.yml

My understanding of this is that it just validates the config file (`detekt.yml`).

android-components/config/detekt.yml
```diff
@@ -32,7 +39,15 @@ console-reports:
   #  - 'NotificationReport'
   #  - 'FindingsReport'
   #  - 'FileBasedFindingsReport'
-    - 'LiteFindingsReport'
+  #  - 'LiteFindingsReport'
```
gabrielluong, line 42: Uncommented this since there's no reason to really exclude this or any particular reports.

android-components/config/detekt.yml
```diff
@@ -465,41 +544,35 @@ style:
...
     ignoreUsageInGenerics: false
   FunctionOnlyReturningConstant:
     active: true
     ignoreOverridableFunction: true
     ignoreActualFunction: true
-    excludedFunctions: ''
-  LibraryCodeMustSpecifyReturnType:
-    active: true
-    excludes: ['**']
-  LibraryEntitiesShouldNotBePublic:
```
gabrielluong, line 488: The library ruleset is now introduced via a new plugin, which I didn't import. See https://detekt.dev/changelog#migration

We can followup on this if you believe we want these library ruleset.

focus-android/quality/detekt.yml
```diff
@@ -596,15 +687,15 @@ style:
...
-    active: false
+    active: true
   UseArrayLiteralsInAnnotations:
-    active: false
+    active: true
   UseCheckNotNull:
     active: true
   UseCheckOrError:
-    active: false
+    active: true
```
jonalmeida, line 698: Why make this true? I see that [the detekt team](https://github.com/detekt/detekt/issues/4643) are enabling them default true as a way to promote, but I don't see the value it brings. Do you have a stronger opinion to wanting this updated?

focus-android/quality/detekt.yml
```diff
@@ -596,15 +687,15 @@ style:
   UnusedPrivateMember:
     active: true
     allowedNames: '(_|ignored|expected|serialVersionUID)'
-    ignoreAnnotated: ['Composable']
+    ignoreAnnotated: ['Preview'] # Configuration for Compose https://github.com/mozilla-mobile/android-components/issues/11866
   UseAnyOrNoneInsteadOfFind:
-    active: false
+    active: true
   UseArrayLiteralsInAnnotations:
```
jonalmeida, line 693: Same for this one, [the docs](https://detekt.dev/docs/rules/style/#usearrayliteralsinannotations) say it's more readable but I don't see any reason to make us change our default.

jonalmeida (approved):
Thanks, this looks good. The file is quite large to check every change, but maybe we can fixup the config as we go about it? I left some comments on the focus-android detekt that apply for the A-C one too.

focus-android/quality/detekt.yml
gabrielluong: This is enabled by default in detekt since v1.21.0 https://detekt.dev/docs/rules/style/#useanyornoneinsteadoffind. I wouldn't say we really have our own default except where we have a comment that stated that we enabled or changed a property and the PR/issue that was responsible for that decision. Our default is more based on what was the default of the previous base configs. 

I would lean towards following the base configuration because keeping this inactive means we have an intentional intervention and now requires another comment about why or when we made this decision. Looking at the code sample, I don't see any reason why we wouldn't want to accept more readable code. 

I am also thinking about this from an upgrade point of view where we are currently diff our current config and the new base config to sync any new rules that needs to be added to the file.

focus-android/quality/detekt.yml
gabrielluong: See https://github.com/mozilla-mobile/firefox-android/pull/292#discussion_r1054926875

focus-android/quality/detekt.yml
gabrielluong: I am also looking to turn on more rules after this. Especially rules that will make our code more readable and simplifies things for us. Based on our review guidelines, I want to front load and really empower our linters to do most of the heavy lifting because most of my review codes are based around simplifying and making code more readable.

jonalmeida: > This is enabled by default in detekt since v1.21.0 https://detekt.dev/docs/rules/style/#useanyornoneinsteadoffind. 

I was referring to the line I commented on, [`UseArrayLiteralsInAnnotations`](https://detekt.dev/docs/rules/style/#usearrayliteralsinannotations). Using `any` or `none` sounds fine to me.

> Our default is more based on what was the default of the previous base configs.

This is not true. We chose our defaults intentionally based on our needs. 

> Looking at the code sample, I don't see any reason why we wouldn't want to accept more readable code.

I am all for this, but using defaults aren't adding that value, it is just code churn. `UndocumentedPublicClass`, `UnreachableCatchBlock` and `CognitiveComplexMethod` are _valuable_ to the code base. The detekt defaults are made by the developers that want to switch some changes to true because they invested in building those but that does not mean those make sense for us.

> making code more readable.

Great! Some of these checks are not readable to me though (and maybe others). 😅

csadilek (commented):
@jonalmeida @gabrielluong I've looked through the rules now, and thought about your discussion: 

I like the adjustments of existing rules that have already been active, but recommend that we do **not** introduce new rules, as part of this PR. This is so the PR can be unblocked (it's been around for a while), and because I think some of these rules require more discussion e.g., `InjectDispatcher`, which we already violate in various places, and just ignoring those cases isn't good either.

I recommend landing this without introducing any new rules, but with the refinements, and putting up separate PRs for specific new rules. On those PRs, we can then also include the required refactorings to actually follow the new rules. This will allow us to properly judge if we want to introduce this a new rule or not.

gabrielluong:
This is just waiting for me to go through the review feedback and making the changes.

gabrielluong:
@jonalmeida This is ready for a review. There's no new rules that are turned on in this iteration. I added a comment next to each rule that differ from the default detekt configuration to indicate that these are turned off intentionally. That way when we need to bump the detekt version and update our config again we will hopefully not just take the defaults as is.

android-components/config/detekt.yml
```diff
@@ -32,7 +39,15 @@ console-reports:
   #  - 'NotificationReport'
   #  - 'FindingsReport'
   #  - 'FileBasedFindingsReport'
-    - 'LiteFindingsReport'
+  #  - 'LiteFindingsReport'
+
+output-reports:
+  active: true
+  exclude:
```
jonalmeida, line 46: Ignore if what you have is valid YAML; I thought it might cause parsing errors.

```suggestion
  exclude: []
```

jonalmeida (approved):
LGTM! I only reviewed the AC version on the understanding that the Fenix and Focus ones are identical.

android-components/config/detekt.yml
gabrielluong: This should be okay. It follows what was in their default config https://github.com/detekt/detekt/blob/v1.22.0/detekt-core/src/main/resources/default-detekt-config.yml#L46

PR #295, opened 2022-12-12
add karma search engine

lillabldv:
KARMA Search is the search engine that takes care of biodiversity and animal welfare. 

See more : https://karmasearch.org


kbrosnan:
Search engines need to have a search agreement with Mozilla. We collect suggested feature changes at https://connect.mozilla.org/t5/ideas/idb-p/ideas

PR #298, opened 2022-12-12
Bug 1804274: Adds entrypoints to firefox accounts interactions

tarikeshaq:
The changes needed for https://github.com/mozilla-mobile/fenix/pull/28176

I'll create a PR for the reference browser too

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1804274


tarikeshaq:
The breaking change in AC is fixed in the reference browser too, in https://github.com/mozilla-mobile/reference-browser/pull/2288

jonalmeida (commented):
Submitting my PR comments for now.

jonalmeida:
Spoke with Tarik offline about updating the PR review changes. Added @MatthewTighe as a reviewer since we reviewed and worked the changes together.

MatthewTighe (commented):
It looks like the CI failures are caused by not having a concrete implementation of `FxAEntryPoint` available in A-C tests. I think it would be best to create inline fakes in those samples and tests, something like `object : FxAEntryPoint { override val entryPointName = "this isn't used here" }`.

Alternatively, we could create some default and stick it in the companion of the interface. My biggest concern with that approach is that it has a greater potential for misuse.

Now that I'm thinking about it more though, I'm wondering if we even want the `entryPoint` parameters that are propagated through these functions changes to be mandatory parameters. It makes sense for our use case, but I'm not sure it would for any other consumer of A-C. Should we make them optional?

fenix/app/src/main/java/org/mozilla/fenix/components/accounts/FenixFxAEntryPoint.kt
```diff
@@ -0,0 +1,111 @@
...
+    companion object CREATOR : Parcelable.Creator<FenixFxAEntryPoint> {
+        override fun createFromParcel(parcel: Parcel): FenixFxAEntryPoint {
+            val parcelEntryName = parcel.readString() ?: Unknown
+            return FenixFxAEntryPoint.values().first { it.entryName == parcelEntryName }
+        }
+
+        override fun newArray(size: Int): Array<FenixFxAEntryPoint?> {
+            return arrayOfNulls(size)
+        }
+    }
```
jonalmeida, line 110: @tarikeshaq @MatthewTighe this is ready for a final review, but before landing, I'd like to draw your attention to this addition I had to make.

The nav graph needs the argument parameter to be `Parcelable` so this part was partially generated. The `createFromParcel` function introduced an `Unknown` entrypoint for compile-time safety.

jonalmeida:
Didn't see your messages earlier - sorry!

> Now that I'm thinking about it more though, I'm wondering if we even want the `entryPoint` parameters that are propagated through these functions changes to be mandatory parameters. It makes sense for our use case, but I'm not sure it would for any other consumer of A-C. Should we make them optional?

Tarik and I discussed this offline before introducing this change: entryPoint should be mandatory because we add new entry points to Sync over time, we want to know and accurately report that to fxa. If we stick with defaults, then chances are that we lose this data over time.

fenix/app/src/main/java/org/mozilla/fenix/components/accounts/FenixFxAEntryPoint.kt
MatthewTighe: Would this boilerplate be avoidable with the `@Parcelize` [annotation](https://developer.android.com/kotlin/parcelize)?

MatthewTighe (approved):
One optional suggestion but this otherwise LGTM

fenix/app/src/main/java/org/mozilla/fenix/components/accounts/FenixFxAEntryPoint.kt
jonalmeida: Yeah, I tried that first. It didn't work because it's an enum so it couldn't figure out how to return the default value for `createFromParcel`. It kept trying to return a `FenixFxAEntryPoint` and not a _value type_ of `FenixFxAEntryPoint`, so the override was needed.

Then the next problem: `parcel.readString()` was returning a nullable string which can occur from the OS crapping out during navigation too on some rare cases. Crashing because the entry point couldn't be serialized didn't see like a good enough reason.

android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
```diff
@@ -39,6 +39,19 @@ data class MigratingAccountInfo(
     val kXCS: String,
 )
 
+/**
+ * Enum representing all the possible entry points into FxA
```
tarikeshaq, line 43: nit: the "enum" part of this is an artifact of the old comment I wrote 😆

fenix/app/src/main/java/org/mozilla/fenix/components/accounts/FenixFxAEntryPoint.kt
```diff
@@ -0,0 +1,111 @@
...
+package org.mozilla.fenix.components.accounts
+
+import android.os.Parcel
+import android.os.Parcelable
+import mozilla.components.concept.sync.FxAEntryPoint
+
+/**
+ * Fenix implementation of [FxAEntryPoint].
+ */
+enum class FenixFxAEntryPoint(override val entryName: String) : FxAEntryPoint, Parcelable {
```
tarikeshaq, line 14: Woah the `override val entryName` awsome! TIL!

fenix/app/src/main/java/org/mozilla/fenix/components/accounts/FenixFxAEntryPoint.kt
tarikeshaq: Ah gosh, thank you for wrestling with this, I might have come across a form of this when doing the first version of this but I can't recall how it worked.

+1 from me - and the comment on the `Unknown` seems clear enough that consumers shouldn't set it manually

tarikeshaq (commented):
Thanks Jonathan, this looks great!! 

GitHub isn't letting me approve my own PR, but lets 🚢

fenix/app/src/main/java/org/mozilla/fenix/components/accounts/FenixFxAEntryPoint.kt
jonalmeida: I added some more comments in this area to explain the reasoning since it isn't obvious why we had to write some of this boilerplate.

Thanks for bringing it up!

jonalmeida:
@Mergifyio refresh

PR #304, opened 2022-12-13
Open Default browser OS settings for Huawei devices

mcarare:
### Pull Request checklist

**Before and after on a Huawei P10 with Android 9.**

  

**Note that from the previous screen it was impossible to reach the _after_ screen**

### After merge

### GitHub Automation

Used by GitHub Actions.


mcarare:
Note that this ( just like with the other devices' intent action) takes you to the default apps list on a Huawei device. 
I could not find if it is possible to open the default browser screen specifically and I would not complicate it further to avoid issues when the activity / fragment names differ from OS version to another.

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/ext/Context.kt
```diff
@@ -0,0 +1,47 @@
...
+import android.provider.Settings
+import androidx.annotation.RequiresApi
+import androidx.core.os.bundleOf
+import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.utils.ManufacturerCodes
+
+const val SETTINGS_SELECT_OPTION_KEY = ":settings:fragment_args_key"
+const val SETTINGS_SHOW_FRAGMENT_ARGS = ":settings:show_fragment_args"
+const val DEFAULT_BROWSER_APP_OPTION = "default_browser"
+const val ACTION_MANAGE_DEFAULT_APPS_SETTINGS_HUAWEI = "com.android.settings.PREFERRED_SETTINGS"
```
mcarare, line 20: We should test this on more Huawei devices other than Huawei P10 with Android 9 that was used for development. I added a try-catch block to avoid crashes in case they decided to change the action.

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/ext/Context.kt
```diff
@@ -0,0 +1,47 @@
...
+private val logger = Logger("navigateToDefaultBrowserAppsSettings")
+
+/**
+ * Open OS settings for default browser.
+ */
+@RequiresApi(Build.VERSION_CODES.N)
```
mcarare, line 26: Instead of adding a version check in code, I added the annotation that would make sure anyone who wants to use it does so in a proper manner. Using an "if" check would hide that.

mcarare:
Thanks also to @amocirean who tested who also tested before and after on a Huawei MediaPad M3 tablet with Android 7.

**Before:**


**After:**

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/ManufacturerCodesTest.kt
```diff
@@ -0,0 +1,45 @@
...
+import mozilla.components.support.utils.ManufacturerCodes.manufacturer
+import org.junit.Assert.assertFalse
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class ManufacturerCodesTest {
+
+    @Test
+    fun testIsHuawei() {
+        manufacturer = "Huawei" // expected value for Huawei devices
```
jonalmeida, line 16: What magic is this? How are these tests working?!

jonalmeida: Ah nevermind, it's importing the field from the singleton class. 😄

jonalmeida (approved):
Thanks for upstreaming this!

PR #312, opened 2022-12-13
Fix test search bar shows search term on edit

csadilek:
We need to revert the A-C change because it breaks the Focus UI test `testSearchBarShowsSearchTermOnEdit`, independent of the `isSearch` name conflict.


csadilek:
@Alexandru2909 Unfortunately, I have to revert the A-C fix here, because the Focus UI test is failing with it `testSearchBarShowsSearchTermOnEdit`. So, that's likely a real regression we have introduced. We can re-land it once fixed. Looks like behaviour in Focus is different so we need to look into that.

jonalmeida (approved):
> @Alexandru2909 Unfortunately, I have to revert the A-C fix here, because the Focus UI test is failing with it `testSearchBarShowsSearchTermOnEdit`. So, that's likely a real regression we have introduced. We can re-land it once fixed. Looks like behaviour in Focus is different so we need to look into that.

Thanks for catching this and reverting!

PR #314, opened 2022-12-13
Bug 1802577 - Target Android 13 (SDK 33) in Focus

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1802577


mcarare:
As a note: In Focus, having the notification every time a tab is opened has rendered the need for handling download notifications permission obsolete (because we would have already requested and received/denied the permission to show notifications.

This would not be the case in Fenix. We would need to handle it in AC probably, but for Focus it is good enough for now.

focus-android/app/src/main/java/org/mozilla/focus/session/PrivateNotificationFeature.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+class PrivateNotificationFeature(
+    context: Context,
+    private val browserStore: BrowserStore,
+    private val sessionNotificationServiceClass: KClass<SessionNotificationService>,
+    private val permissionRequestHandler: (() -> Unit),
+) : LifecycleAwareFeature {
+
+    private val applicationContext = context.applicationContext
+    private var scope: CoroutineScope? = null
```
rahulsainani, line 33: What's this used for? Seems it's only cancelled in `onStop()` 🤔

focus-android/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -71,6 +75,15 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
     private lateinit var startupTypeTelemetry: StartupTypeTelemetry
     private var _binding: ActivityMainBinding? = null
     private val binding get() = _binding!!
+    private lateinit var privateNotificationObserver: PrivateNotificationFeature
+    private val notificationPermission =
+        registerForActivityResult(ActivityResultContracts.RequestPermission()) { granted ->
+            when {
+                granted -> {
```
rahulsainani, line 82: For my understanding, we aren't handling `notGranted` and `shouldShowPermissionRationale()` cases, should we?

focus-android/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -71,6 +75,15 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
     private lateinit var startupTypeTelemetry: StartupTypeTelemetry
     private var _binding: ActivityMainBinding? = null
     private val binding get() = _binding!!
+    private lateinit var privateNotificationObserver: PrivateNotificationFeature
```
rahulsainani, line 78: Could we initialize this here and convert this to val? And call `start()` in `onCreate()`

focus-android/app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
```diff
@@ -33,13 +35,34 @@ class SessionNotificationService : Service() {
 
     private var shouldSendTaskRemovedTelemetry = true
 
+    private val notificationManager: NotificationManager by lazy {
```
rahulsainani, line 38: Should we use `[NotificationManagerCompat](https://developer.android.com/reference/androidx/core/app/NotificationManagerCompat)` ?

focus-android/app/src/main/java/org/mozilla/focus/session/PrivateNotificationFeature.kt
mcarare: Good question! I updated the code.

focus-android/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
mcarare: If UX / Product decides we should notify the user of the consequences of not granting the notification permission, we could do that in the `notGranted` case. Currently, we have no other reason to handle that.

focus-android/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
mcarare: We need context as a param here, so initializing it here would throw an NPE.

focus-android/app/src/main/java/org/mozilla/focus/session/SessionNotificationService.kt
mcarare: AFAIK `NotificationManagerCompat` only helps with support for side channeled notifications (only available before SDK 19, so it's not the case for us) and to internalize the need for checking if notification channels are available (SDK >= 26). We added the latter in `createNotificationChannelIfNeeded` which we also use as an early return. ( As opposed to relying on `NotificationManagerCompat` to later check for this.)

If there is any other use case I am not aware of, I am happy to change it (we use it in other parts of the app, so I guess it would be just the same).

rahulsainani: I see, if it's a conscious decision to use it and we are aware of the differences, it makes sense! 👍

focus-android/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
```diff
@@ -71,6 +74,15 @@ open class MainActivity : LocaleAwareAppCompatActivity() {
...
```
jonalmeida, line 77: Small nit: either name the variable a feature or name the class an observer. 😄

jonalmeida (approved):
Since we have cut the release, let's land this? 🚀

focus-android/app/src/main/java/org/mozilla/focus/activity/MainActivity.kt
mcarare: Yeah, we talked about this when I first submitted this, but I still haven't decided. 😅 I guess I have to now. 😄

mcarare: Renamed the variable to `privateNotificationFeature`.

PR #386, opened 2022-12-19
Bug 27998 – Add settings to BrowserMenu to overlap anchor view completely

mavduevskiy:
Adds settings to account for spacing between the menu view and the list (created by `elevation` + `cardUseCompatPadding`) and setting extra offset to shift the item even further. 

I followed the approach that was implemented in the `BrowserMenu.v1` with calculating parameters separately and storing them in `MenuPositioningData` class.

NB: in order it work, we have to pass the [search_selector](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/res/layout/search_selector.xml#12), not the view with the same icon from the `HomeFragment` as it has a different size (added margins).

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=27998

### DIFF
--- a/android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
+++ b/android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
@@ -1,0 +1,255 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.menu2.ext
+
+import android.graphics.Rect
+import android.view.View
+import androidx.annotation.Px
+import androidx.core.view.ViewCompat
+import androidx.core.view.WindowInsetsCompat
+import androidx.recyclerview.widget.RecyclerView
+import mozilla.components.browser.menu2.R
+import mozilla.components.concept.menu.MenuStyle
+import mozilla.components.concept.menu.Orientation
+import mozilla.components.support.ktx.android.view.isRTL
+import kotlin.math.roundToInt
+
+const val HALF_MENU_ITEM = 0.5
+
+@Suppress("ComplexMethod")
+internal fun inferMenuPositioningData(
+    containerView: View,
+    anchor: View,
+    style: MenuStyle? = null,
+    orientation: Orientation?,
+): MenuPositioningData? {
+    // Measure the menu allowing it to expand entirely.
+    val spec = View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED)
+    containerView.measure(spec, spec)
+
+    val recyclerView = containerView.findViewById<RecyclerView>(R.id.mozac_browser_menu_recyclerView)
+    val recyclerViewAdapter = recyclerView.adapter ?: run {
+        // We might want to track how often and in what circumstances the menu gets called without
+        // valid parameters once we have a system for that.
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=1814816
+        return null
+    }
+
+    val hasViewInitializedCorrectly = containerView.measuredHeight > 0 && containerView.measuredWidth > 0 &&
+        recyclerView.measuredHeight > 0 && recyclerViewAdapter.itemCount > 0
+    if (!hasViewInitializedCorrectly) {
+        // Same as above: https://bugzilla.mozilla.org/show_bug.cgi?id=1814816
+        return null
+    }
+
+    val menuHorizontalPadding = containerView.measuredWidth - recyclerView.measuredWidth
+    val menuVerticalPadding = containerView.measuredHeight - recyclerView.measuredHeight
+    var horizontalOffset = style?.horizontalOffset ?: 0
+    var verticalOffset = style?.verticalOffset ?: 0
+
+    // Elevation creates some padding between the menu and its container, so that the start corner
+    // of the menu doesn't match the corner of the anchor view. If the user wants the menu to hide
+    // the anchor completely, we have to adjust the position of the menu to compensate for the inner
+    // padding of the menu container.
+    if (style?.completelyOverlap == true) {
+        horizontalOffset -= menuHorizontalPadding / 2
+        verticalOffset -= menuVerticalPadding / 2
+    }
+
+    // The menu height might be adjusted: if there is not enough space to show all the items,
+    // it will crop the last visible item in half, to give the user a hint that it is scrollable.
+    val containerViewHeight = calculateContainerHeight(
+        recyclerView.measuredHeight,
+        recyclerViewAdapter.itemCount,
+        containerView.measuredHeight,
+        style?.verticalOffset ?: 0,
+        anchor,
+    )
+
+    val (availableHeightToTop, availableHeightToBottom) = getMaxAvailableHeightToTopAndBottom(anchor)
+    val (availableWidthToLeft, availableWidthToRight) = getMaxAvailableWidthToLeftAndRight(anchor)
+
+    val fitsUp = availableHeightToTop + anchor.height >= containerViewHeight
+    val fitsDown = availableHeightToBottom + anchor.height >= containerViewHeight
+    val fitsRight = availableWidthToRight + anchor.width >= containerView.measuredWidth
+    val fitsLeft = availableWidthToLeft + anchor.width >= containerView.measuredWidth
+
+    val notEnoughHorizontalSpace = !fitsRight && !fitsLeft
+    val fitsBothHorizontalDirections = fitsRight && fitsLeft
+    val drawingLeft = if (notEnoughHorizontalSpace || fitsBothHorizontalDirections) {
+        anchor.isRTL
+    } else {
+        !fitsRight
+    }
+
+    val anchorPosition = IntArray(2)
+    anchor.getLocationInWindow(anchorPosition)
+    var (anchorX, anchorY) = anchorPosition
+
+    // Position the menu above the anchor if the orientation is UP and there is enough space.
+    if (orientation == Orientation.UP && fitsUp) {
+        anchorY -= containerViewHeight - anchor.height
+        verticalOffset = -verticalOffset
+    }
+
+    if (drawingLeft) {
+        anchorX -= containerView.measuredWidth - anchor.width
+        horizontalOffset = -horizontalOffset
+    }
+
+    return MenuPositioningData(
+        x = anchorX + horizontalOffset,
+        y = anchorY + verticalOffset,
+        containerHeight = containerViewHeight,
+        animation = getAnimation(fitsUp, fitsDown, drawingLeft, orientation),
+    )
+}
+
+private fun getMaxAvailableHeightToTopAndBottom(anchor: View): Pair<Int, Int> {
+    val anchorPosition = IntArray(2)
+    val displayFrame = Rect()
+
+    val appView = anchor.rootView
+    appView.getWindowVisibleDisplayFrame(displayFrame)
+
+    anchor.getLocationOnScreen(anchorPosition)
+
+    val bottomEdge = displayFrame.bottom
+
+    val distanceToBottom = bottomEdge - (anchorPosition[1] + anchor.height)
+    val distanceToTop = anchorPosition[1] - displayFrame.top
+
+    return distanceToTop to distanceToBottom
+}
+
+private fun getMaxAvailableWidthToLeftAndRight(anchor: View): Pair<Int, Int> {
+    val anchorPosition = IntArray(2)
+    val displayFrame = Rect()
+
+    val appView = anchor.rootView
+    appView.getWindowVisibleDisplayFrame(displayFrame)
+
+    anchor.getLocationOnScreen(anchorPosition)
+
+    val distanceToLeft = anchorPosition[0] - displayFrame.left
+    val distanceToRight = displayFrame.right - (anchorPosition[0] + anchor.width)
+
+    return distanceToLeft to distanceToRight
+}
+
+/**
+ * Determine whether the container view can display all menu items (without scrolling) within
+ * the available height.
+ *
+ * @return The original container height if the container view can display all menu items
+ * (without scrolling), else calculate the maximum available container height for a scrollable
+ * view with a half menu item.
+ */
+private fun calculateContainerHeight(
+    recyclerViewHeight: Int,
+    recyclerViewItemCount: Int,
+    containerViewHeight: Int,
+    menuStylePadding: Int,
+    anchor: View,
+): Int {
+    // Get the total screen display height.
+    val totalHeight = anchor.rootView.measuredHeight
+
+    // Note: We cannot use getWindowVisibleDisplayFrame() as the height is dynamic based on whether
+    // the keyboard is open.
+    // Get any displayed system bars e.g. top status bar, bottom navigation bar or soft buttons bar.
+    val systemBars =
+        ViewCompat.getRootWindowInsets(anchor)?.getInsets(WindowInsetsCompat.Type.systemBars())
+    // Store the vertical status bars.
+    val topSystemBarHeight = systemBars?.top ?: 0
+    val bottomSystemBarHeight = systemBars?.bottom ?: 0
+
+    // Deduct any status bar heights from the total height.
+    val availableHeight = totalHeight - (bottomSystemBarHeight + topSystemBarHeight)
+
+    val menuItemHeight = recyclerViewHeight / recyclerViewItemCount
+
+    // We must take the menu container padding into account as this will be applied to the final height.
+    val containerPadding = containerViewHeight - recyclerViewHeight
+
+    val maxAvailableHeightForRecyclerView = availableHeight - containerPadding - menuStylePadding
+
+    // The number of menu items that can fit exactly (no cropping) within the max app height.
+    // Round the number of items to the closet Int value to ensure the max space available is utilized.
+    // E.g if 6.9 items fit, round to 7 so the calculation below will show 6.5 items instead of 5.5 .
+    val numberOfItemsFitExactly =
+        (maxAvailableHeightForRecyclerView.toFloat() / menuItemHeight.toFloat()).roundToInt()
+
+    val itemsAlreadyFitContainerHeight = recyclerViewItemCount <= numberOfItemsFitExactly
+
+    return if (itemsAlreadyFitContainerHeight) {
+        containerViewHeight
+    } else {
+        getCroppedMenuContainerHeight(numberOfItemsFitExactly, menuItemHeight, containerPadding)
+    }
+}
+
+private fun getAnimation(
+    fitsUp: Boolean,
+    fitsDown: Boolean,
+    drawingLeft: Boolean,
+    orientation: Orientation?,
+): Int {
+    val isUpOrientation = orientation == Orientation.UP
+    return when {
+        isUpOrientation && fitsUp -> if (drawingLeft) {
+            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuRightBottom
+        } else {
+            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuLeftBottom
+        }
+        fitsDown -> if (drawingLeft) {
+            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuRightTop
+        } else {
+            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuLeftTop
+        }
+        else -> if (drawingLeft) {
+            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuRight
+        } else {
+            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuLeft
+        }
+    }
+}
+
+private fun getCroppedMenuContainerHeight(
+    numberOfItemsFitExactly: Int,
+    menuItemHeight: Int,
+    containerPadding: Int,
+): Int {
+    // The number of menu items that fit exactly, minus a half menu item (indicates more menu items exist).
+    val numberOfItemsFitWithOverflow = numberOfItemsFitExactly - HALF_MENU_ITEM
+    val updatedRecyclerViewHeight = (numberOfItemsFitWithOverflow * menuItemHeight).toInt()
+
+    return updatedRecyclerViewHeight + containerPadding
+}
+
+/**
+ * Data needed for menu positioning.
+ */
+data class MenuPositioningData(
+    /**
+     * [WindowManager#LayoutParams#x] of params the menu will be added with.
+     */
+    @Px val x: Int = 0,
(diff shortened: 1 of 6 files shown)


mavduevskiy:
required for https://github.com/mozilla-mobile/fenix/issues/27998

mavduevskiy:
A quick update: run into an issue with [setting height](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt#197) manually,– `showAsDropDown` would try to [move the window up](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/PopupWindow.java#1859) if there is not enough space for it to be shown. Then, if would stick it to the top of the view, ignoring the offsets provided. 

`showAtLocation` is a working solution, though it requires a bit more work to calculate the position for `Orientation.Bottom`. It also worked nicely with show at bottom if there is keyboard case.

I also realized that there is a lot of things I would like to use from the original library in menu2, I talked about that with @jonalmeida , and he proposed moving some shared code to the concept folder, so we could also use it for the old menu if we choose. also, shared resources (animation) could be moved to the UI folders,- both UI and concept folders a dependences in menu and menu2. 

also, this positioning code might be used for some time even when we move to compose, as we could create a window where we want to and then draw a simple composable inside (it’s not clear when or if composable with allow such level of customization as we have with the android views and their positioning)

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,269 @@
...
+
+private fun getAnimation(
+    fitsUp: Boolean,
+    fitsDown: Boolean,
+    drawingLeft: Boolean,
+    orientation: Orientation?,
+): Int {
+    val isUpOrientation = orientation == Orientation.UP
+    return when {
+        isUpOrientation && fitsUp -> if (drawingLeft) {
```
mavduevskiy, line 216: This covers the edge case with the opened keyboard. UX team requested not to cut the menu in this case, but to show the menu from at the bottom of the screen.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt
```diff
@@ -29,226 +20,25 @@ internal fun PopupWindow.displayPopup(
...
-                "recyclerViewItemCount: $recyclerViewItemCount must be more than 0.",
-        )
-        return false
-    }
-
-    // Popup window does not need a input method. This avoids keyboard flicker when menu is opened.
     inputMethodMode = PopupWindow.INPUT_METHOD_NOT_NEEDED
 
-    showPopup(orientation, anchor, recyclerViewHeight, recyclerViewItemCount, containerViewHeight)
+    showAtAnchorLocation(anchor, positioningData)
```
mavduevskiy, line 30: The reason behind switching to `showAtAnchorLocation` is the manual setting of [height](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/PopupWindow.kt#197) for the popup. It makes the [PopupWindow](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/PopupWindow.java#1859) to move the menu up offscreen in there is keyboard opened, ignoring any provided offsets, making precise positioning impossible.

android-components/components/browser/menu2/src/main/res/values/dimens.xml
```diff
@@ -4,7 +4,7 @@
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
 <resources>
     <dimen name="mozac_browser_menu2_corner_radius">4dp</dimen>
-    <dimen name="mozac_browser_menu2_elevation">8dp</dimen>
+    <dimen name="mozac_browser_menu2_elevation">5dp</dimen>
```
mavduevskiy, line 7: [Link](https://www.figma.com/file/epu0Sj6mCPk7K282sNpHgI/Site?node-id=3850%3A59494&t=M9Wqnn9vei1aRxTI-0) to the design. Having less then 8 came in handy, because getting the [4dp offsets](https://www.figma.com/file/epu0Sj6mCPk7K282sNpHgI/Site?node-id=3850%3A60276&t=M9Wqnn9vei1aRxTI-4) with down orientation is impossible – 8 elevation means also 8dp spacing, and there is just 6dp towards the top edge of the rootview.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,269 @@
...
+import mozilla.components.concept.menu.MenuStyle
+import mozilla.components.concept.menu.Orientation
+import mozilla.components.support.base.log.Log
+import mozilla.components.support.ktx.android.view.isRTL
+import kotlin.math.roundToInt
+
+const val HALF_MENU_ITEM = 0.5
+
+@Suppress("ComplexMethod")
+internal fun inferMenuPositioningData(
```
mavduevskiy, line 23: It is a complex method and I decided against splitting it into multiple smaller ones. It is quite "deep" method, and I wanted to keep the order and all the logic to be visible in once place, trying to make it as easy to read as possible.

jonalmeida: Sometimes splitting into smaller ones causes more readable issues than it solves. Good thinking. 🙂

rocketsroger (commented):
Works great! Some questions.  Thanks

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,269 @@
...
+    style: MenuStyle? = null,
+    orientation: Orientation?,
+): MenuPositioningData? {
+    // Measure the menu allowing it to expand entirely.
+    val spec = View.MeasureSpec.makeMeasureSpec(0, View.MeasureSpec.UNSPECIFIED)
+    containerView.measure(spec, spec)
+
+    val recyclerView = containerView.findViewById<RecyclerView>(R.id.mozac_browser_menu_recyclerView)
+    val recyclerViewAdapter = recyclerView.adapter ?: run {
+        Log.log(
```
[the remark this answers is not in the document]
mavduevskiy: True about numbers, but the design definitely has 5 all over the menus. Should I talk to the UX team?👀

I still have [little context](https://github.com/mozilla-mobile/firefox-android/pull/181#discussion_r1056115520) on loggers, but you didn't mind it back when it was introduced :)

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,269 @@
...
+    var verticalOffset = style?.verticalOffset ?: 0
+
+    if (style?.completelyOverlap == true) {
+        horizontalOffset -= menuHorizontalPadding / 2
+        verticalOffset -= menuVerticalPadding / 2
+    }
+
+    // The menu height might be adjusted: if there is not enough space to show all the items,
+    // it will crop the last visible item in half, to give the user a hint that it is scrollable.
+    containerViewHeight = calculateContainerHeight(
```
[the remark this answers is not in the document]
mavduevskiy: I was not sure if I should change the old `calculatedContainerHeight` but the problem with it – the majority of the time the function would just return the same height. It doesn't technically calculate the new height, it adjusted in case there is not enough space to show the full menu.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,269 @@
...
+            message = "containerViewHeight: $containerViewHeight must be more than 0," +
+                "recyclerViewHeight: $recyclerViewHeight must be more than 0 &" +
+                "recyclerViewItemCount: $recyclerViewItemCount must be more than 0.",
+        )
+        return null
+    }
+
+    // Elevation creates some padding between the menu and its container, so that the start corner
+    // of the menu doesn't match the corner of the anchor view. If the user wants the menu to hide
+    // the anchor completely, we have to adjust the position of the menu to compensate for the inner
```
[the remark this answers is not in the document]
mavduevskiy: true! thanks, changing that

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,269 @@
...
+        return null
+    }
+
+    var containerViewHeight = containerView.measuredHeight
+    val containerViewWidth = containerView.measuredWidth
+    val recyclerViewHeight = recyclerView.measuredHeight
+    val recyclerViewItemCount = recyclerViewAdapter.itemCount
+
+    val hasViewInitializedCorrectly =
+        containerViewHeight > 0 && containerViewWidth > 0 && recyclerViewHeight > 0 && recyclerViewItemCount > 0
```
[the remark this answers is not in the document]
mavduevskiy: I don't really expect this check to be a meaningful defensive mechanism – we don't expect the view to have zero size in any realistic circumstances. It was more of a defense against possible division by zero.

mavduevskiy:
@rocketsroger 
Thanks for the review, Roger! Adjusted the PR, please let me know if you have any follow ups to the discussion. I am up for a quick call if discussing it in comments is getting too lengthy :)

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,270 @@
...
+    val (leftSpace, rightSpace) = getMaxAvailableWidthToLeftAndRight(anchor)
+
+    val fitsUp = topSpace + anchor.height >= containerViewHeight
+    val fitsDown = bottomSpace + anchor.height >= containerViewHeight
+    val fitsRight = rightSpace + anchor.width >= containerViewWidth
+    val fitsLeft = leftSpace + anchor.width >= containerViewWidth
+
+    val notEnoughHorizontalSpace = !fitsRight && !fitsLeft
+    val fitsBothHorizontalDirections = fitsRight && fitsLeft
+    val drawingLeft = if (notEnoughHorizontalSpace || fitsBothHorizontalDirections) {
```
t-p-white, line 96: nit: could this be more descriptive? Perhaps something like
```suggestion
    val menuIsLTR = if (notEnoughHorizontalSpace || fitsBothHorizontalDirections) {
```

t-p-white, line 96: I'm slightly confused on this, isn't `drawingLeft` isRtl? which would explain in `getAnimation`
```
isUpOrientation && fitsUp -> if (drawingLeft) {
            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuRightBottom
        } else {
            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuLeftBottom
        }
```

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,270 @@
...
+        verticalOffset = -verticalOffset
+    }
+
+    if (drawingLeft) {
+        x -= containerViewWidth - anchor.width
+        horizontalOffset = -horizontalOffset
+    }
+
+    return MenuPositioningData(
+        x = x + horizontalOffset,
```
t-p-white, line 118:
```suggestion
        x += + horizontalOffset,
```

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,270 @@
...
+        y = y + verticalOffset,
```
t-p-white, line 119:
```suggestion
        y =+ verticalOffset,
```

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,270 @@
...
+    anchor.getLocationInWindow(anchorPosition)
+    var (x, y) = anchorPosition
+
+    // Position the menu above the anchor if the orientation is UP and there is enough space.
+    if (orientation == Orientation.UP && fitsUp) {
+        y -= containerViewHeight - anchor.height
+        verticalOffset = -verticalOffset
+    }
+
+    if (drawingLeft) {
```
t-p-white, line 112: comment here similar to above may be useful

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
t-p-white: 👍  appreciate the use of comments for this func  too 🙂

t-p-white (changes requested):
Looking good @mavduevskiy 👍 The 'requested change' status is there for when you've added tests

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
rocketsroger: I was confused about this as well but I think Mike has this right though.  He's not updating `x` or `y` here.  It's just unfortunate that the parameter name is the same.

t-p-white: Ah good catch, so it is! 🙃

mavduevskiy:
@t-p-white 
Tests are in the `PopupWindowTest` file🙂

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
mavduevskiy: I myself don't like the naming, but couldn't find a better one. It's different from `isRTL`.

On a high level, the function decides where to position the menu regarding the anchor. It doesn't consider `isRTL` in regular circumstances, because the anchor might be at the start and at the end of the parent. Bookmarks screen has menu shown from the three dot menu of a bookmark item, and we have to draw it to the left of the anchor.

So, this direction will affect the horizontal shifting and the animation.

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
mavduevskiy: fair comment, it does look confusing! changed the naming

mavduevskiy:
I wonder why locally `detekt` doesn't complain, but the CI does🤔

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
t-p-white: Gotcha, makes sense 👍  how about something like `positionMenuToLeftOfAnchor`? Appreciate it's more verbose but may make it more self evident

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,255 @@
...
+    // it will crop the last visible item in half, to give the user a hint that it is scrollable.
+    val containerViewHeight = calculateContainerHeight(
+        recyclerView.measuredHeight,
+        recyclerViewAdapter.itemCount,
+        containerView.measuredHeight,
+        style?.verticalOffset ?: 0,
+        anchor,
+    )
+
+    val (topSpace, bottomSpace) = getMaxAvailableHeightToTopAndBottom(anchor)
```
t-p-white, line 71: nit: 
```suggestion
    val (availableHeightToTop, availableHeightToBottom) = getMaxAvailableHeightToTopAndBottom(anchor)
```

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
```diff
@@ -0,0 +1,255 @@
...
+    val (leftSpace, rightSpace) = getMaxAvailableWidthToLeftAndRight(anchor)
```
t-p-white, line 72: nit: 
```suggestion
    val (availableWidthToLeft, availableWidthToRight) = getMaxAvailableWidthToLeftAndRight(anchor)
```

android-components/components/browser/menu2/src/main/java/mozilla/components/browser/menu2/ext/MenuPositioningData.kt
t-p-white: Thanks for the context here, may be worth including this as a comment on the function too 🙂

android-components/components/browser/menu2/src/test/java/mozilla/components/browser/menu2/ext/PopupWindowTest.kt
```diff
@@ -40,293 +41,683 @@ private const val HALF_MENU_ITEM = 0.5
...
-        verify(popupWindow).showAtLocation(anchor, Gravity.START or Gravity.TOP, x, y)
+
+        val (targetX, targetY) = getTargetCoordinates(x, y, containerView, anchor, directionRight = false, directionUp = false)
+        verify(popupWindow).showAtLocation(anchor, Gravity.NO_GRAVITY, targetX, targetY)
     }
 
-    private fun createContainerRecyclerView(
-        itemCount: Int = 10,
+    @Test
+    fun `WHEN should completely overlap & orientation up & fits available height THEN showAtLocation with original height & bottom of the menu positioned exactly to the bottom of the anchor with leftBottomAnimation`() {
```
t-p-white, line 314:
```suggestion
    fun `WHEN should completely overlap & orientation up & fits up THEN showAtLocation with original height & bottom of the menu positioned exactly to the bottom of the anchor with leftBottomAnimation`() {
```

android-components/components/browser/menu2/src/test/java/mozilla/components/browser/menu2/ext/PopupWindowTest.kt
```diff
@@ -40,293 +41,683 @@ private const val HALF_MENU_ITEM = 0.5
...
+            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuRightBottom,
+            popupWindow.animationStyle,
+        )
+
+        val (targetX, targetY) = getTargetCoordinates(x, y, containerView, anchor, overlapStyle, directionRight = false)
+        verify(popupWindow).showAtLocation(anchor, Gravity.NO_GRAVITY, targetX, targetY)
+    }
+
+    @Test
+    fun `WHEN should completely overlap & orientation down & fits down THEN showAsDropDown with original height & positioned exactly from the top of the anchor with leftTopAnimation`() {
```
t-p-white, line 352:
```suggestion
    fun `WHEN should completely overlap & orientation down & fits down THEN showAtLocation with original height & positioned exactly from the top of the anchor with leftTopAnimation`() {
```

android-components/components/browser/menu2/src/test/java/mozilla/components/browser/menu2/ext/PopupWindowTest.kt
```diff
@@ -40,293 +41,683 @@ private const val HALF_MENU_ITEM = 0.5
...
+            R.style.Mozac_Browser_Menu2_Animation_OverflowMenuLeftTop,
+            popupWindow.animationStyle,
+        )
+
+        val (targetX, targetY) = getTargetCoordinates(x, y, containerView, anchor, overlapStyle, directionUp = false)
+        verify(popupWindow).showAtLocation(anchor, Gravity.NO_GRAVITY, targetX, targetY)
+    }
+
+    @Test
+    fun `WHEN should completely overlap & orientation down & fits down & is RTL THEN showAsDropDown with original height & positioned exactly from the top of the anchor with rightTopAnimation`() {
```
t-p-white, line 371:
```suggestion
    fun `WHEN should completely overlap & orientation down & fits down & is RTL THEN showAtLocation with original height & positioned exactly from the top of the anchor with rightTopAnimation`() {
```

t-p-white:
> @t-p-white Tests are in the `PopupWindowTest` file🙂
Nice 🙌 missed them (as it's a large change it wasn't rendered by default). Is the TODO in the PR description still relevant?

mavduevskiy:
@t-p-white 
Adjusted the naming! let me know if there is still a blocking issue or we could land the change :)

t-p-white (approved):
LGTM 👍 
NB: PopUpWindowTest changes to be made in another issue

mavduevskiy:
Had a call with @t-p-white on `PopUpWindowTest` changes:

Do a follow up to separate the test into two: `PopUpWindowTest` and a `MenuPositioningDataTest`. The first one would be a very simple one, to make sure that the `showAtLocation` is called with proper values, and the second one would contain most of the logic that is current in the `PopUpWindowTest` and checking would be done directly on the returned `MenuPositioningData` value.

Also refactor `displayPopup` to have only a `MenuPositioningData` parameter. It will make testing easier and will follow more closely [the design](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/menu/src/main/java/mozilla/components/browser/menu/BrowserMenu.kt#128) of the original method.

jonalmeida:
> Do a follow up to separate the test into two: `PopUpWindowTest` and `MenuPositioningDataTest`. The first one would be a very simple one, to make sure that the `showAtLocation` is called with proper values, and the second one would contain the most of the logic that is currently inside the `PopUpWindowTest`. Checking would be done directly on the returned `MenuPositioningData` value.

As part of this follow-up, could you we add some of [these style changes](https://github.com/mozilla-mobile/fenix/pull/28648/files#diff-bf95046416fcc223f06cc8a6bd4836ac60a3a78a1a2e714424aaefc8a51a7688R52-R56) that we have in Fenix to the Sample Browser too? That would be helpful to test this an isolated environment.

PR #344, opened 2022-12-15
Bug 1799679 - Clear search terms on url load

Alexandru2909:
For https://github.com/mozilla-mobile/fenix/issues/25812, we should clear the search terms only when loading a new URL. Currently we dismiss the search terms whenever we receive a redirected `LoadRequest`. As explained in [this comment](https://github.com/mozilla-mobile/fenix/issues/25812#issuecomment-1307157758), multiple search engines redirect the user to the search results, clearing the search terms.

This behavior was initially introduced to fix a search term [issue](https://github.com/mozilla-mobile/fenix/issues/7632). This is not reproducible after the changes in this PR.

| Focus | Fenix |
| --- | --- |
| <source>https://user-images.githubusercontent.com/35462038/207870731-6a5f9b85-d0cb-4e6c-93db-3f0372db92a1.mp4</source> | <source>https://user-images.githubusercontent.com/35462038/200655876-43db56c8-ab67-4226-a1fb-cf93445cdf81.mp4</source> |

The previosuly reproducible behavior of saving the search term when navigating through history is not reproducible:

https://user-images.githubusercontent.com/35462038/202745603-b41de905-9a0f-4df4-b0e9-814853621e84.mp4

The behavior of saving the search term for sites that use location change to navigate to different pages is still reproducible:

https://user-images.githubusercontent.com/35462038/202744904-9b757b67-5e1c-4a36-8543-7ad64ac57b6a.mp4
### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/main/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Milestone**: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
- **Breaking Changes**: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### GitHub Automation

Used by GitHub Actions.

PR #453, opened 2022-12-29
Bug 1807920 - Make newPrivateTabSearch usecase open only private new tabs

Alexandru2909:
Based on [comment](https://github.com/mozilla-mobile/firefox-android/pull/344/files#r1051204996)

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1807920


Mugurell (commented):
Seems like a good approach but I'm a bit worried that this modifies both the `newPrivateTabSearch` and `newTabSearch` invocations to remove the `isPrivate` parameter.

We will need patches to address the breaking change in all projects where we should also check to ensure that there were no mismatched calls - I do expect for a 
```
newTabSearch.invoke(
      private = true,
)
```
call to exist.
Has this happened?

Alexandru2909:
> 

This case occured [here](https://github.com/mozilla-mobile/firefox-android/pull/344/files#r1051204996), where a usecase defined as private was required to call `invoke(private = true)` to create a new private tab.

Mugurell:
> This case occured [here](https://github.com/mozilla-mobile/firefox-android/pull/344/files#r1051204996), where a usecase defined as private was required to call `invoke(private = true)` to create a new private tab.

Saw that case but I was curious if there are others.
Looked through https://searchfox.org/mozilla-mobile/search?q=newPrivateTabSearch&path=&case=false&regexp=false and https://searchfox.org/mozilla-mobile/search?q=newTabSearch&path=&case=false&regexp=false and it seems like that was the only case where the `private` parameter was used.

Mugurell (approved):

PR #488, opened 2023-01-04
Bug 1804996: Replace and remove DDG Type Tags

DreVla:
fpsa, ftas, ftsa and ffos tags have been removed and replaced with fpas.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


android-components/components/feature/search/src/main/assets/searchplugins/ddg.xml
```diff
@@ -17,7 +17,7 @@
 </Url>
 <Url type="application/x-moz-tabletsearch" method="GET" template="https://duckduckgo.com/">
   <Param name="q" value="{searchTerms}" />
-  <Param name="t" value="ftas" />
+  <Param name="t" value="fpas" />
```
DreVla, line 20: Not sure if we need this whole Url tag anymore or we should delete it completely. Open for feedback here. Same for duckduckgo.xml file

boek: I don't recall ever seeing `x-moz-tabletsearch` before. Looking through SearchFox it doesn't look like desktop uses it either. I did find [this issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1486813) in Bugzilla about potentially supporting it in WebExtensions but I couldn't find any context.

I think for the scope of this PR let's leave it as is but it might be worth opening a ticket to determine if we need to keep them around in the future.

boek (approved):

PR #505, opened 2023-01-06
Catches ForegroundServiceStartNotAllowedException and emits a fact

boek:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


boek:
Also thinking about how I might test this 🤔

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/service/MediaSessionServiceDelegate.kt
```diff
@@ -180,10 +183,29 @@ internal class MediaSessionServiceDelegate(
...
+                    Build.VERSION.SDK_INT >= Build.VERSION_CODES.S &&
+                    e is ForegroundServiceStartNotAllowedException
+                ) {
+                    // We should not encounter this exception if `android:foregroundServiceType="mediaPlayback"`
+                    // is added to the service. The crash reporter loses the stack trace for this
+                    // exception so we want to be able to track this crash independently to ensure
+                    // this case is fixed and be able to determine if there are other cases where we
+                    // might be trying to start foreground services from the background.
+                    // https://bugzilla.mozilla.org/show_bug.cgi?id=1802620
+                    emitCaughtForegroundServiceStartNotAllowed(e)
```
MatthewTighe, line 203: I don't have much context here, but just looking through I am wondering if it might make more sense to add a crash reporter as a (optional?) property to the `AbstractMediaSessionService` and inject it during instantiation like is done with the `BrowserStore`. It could then be passed down to this delegate as needed.

IMO that seems slightly more idiomatic, as AFAIK `Fact`s are intended to map more towards metrics we are recording with Glean.

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/service/MediaSessionServiceDelegate.kt
```diff
@@ -180,10 +186,32 @@ internal class MediaSessionServiceDelegate(
...
+                    Build.VERSION.SDK_INT >= Build.VERSION_CODES.S &&
+                    e is ForegroundServiceStartNotAllowedException
+                ) {
+                    // We should not encounter this exception if `android:foregroundServiceType="mediaPlayback"`
+                    // is added to the service. The crash reporter loses the stack trace for this
+                    // exception so we want to be able to track this crash independently to ensure
+                    // this case is fixed and be able to determine if there are other cases where we
+                    // might be trying to start foreground services from the background.
+                    // https://bugzilla.mozilla.org/show_bug.cgi?id=1802620
+                    crashReporter.submitCaughtException(e)
```
MatthewTighe, line 209: Will swallowing this exception cause any breakage in the user experience?

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/service/MediaSessionServiceDelegate.kt
```diff
@@ -180,10 +186,32 @@ internal class MediaSessionServiceDelegate(
     }
 
     @VisibleForTesting
-    internal fun startForeground(sessionState: SessionState) {
-        notificationScope?.launch {
+    @Suppress("TooGenericExceptionCaught")
+    internal fun startForeground(
+        sessionState: SessionState,
+        coroutineContext: CoroutineContext = EmptyCoroutineContext,
```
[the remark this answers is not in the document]
MatthewTighe: Generally I would agree with the latter route in a perfect world because I think it would increase clarity to downstream consumers, but I am not sure it's worth a breaking change in the API.

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/service/AbstractMediaSessionService.kt
```diff
@@ -35,6 +36,7 @@ internal class MediaServiceBinder(delegate: MediaSessionDelegate) : Binder() {
  */
 abstract class AbstractMediaSessionService : Service() {
     protected abstract val store: BrowserStore
+    protected abstract val crashReporter: CrashReporting
```
MatthewTighe, line 39: We could avoid a breaking change by making this property optional and using `null` as a default through the instantiation chain.

MatthewTighe: I am not necessarily suggesting that we _should_ not introduce a breaking change here but am curious whether this crash reporting is something other downstreams will want as a hard requirement vs only needing it in Fenix.

boek: This is kind of what I was thinking with my first approach using a `Fact`. I think if we're going to go the route of using `CrashReporting` the opt-out should be explicit using a no-op `CrashReporting` object versus requiring the domain knowledge of opting in.

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/service/MediaSessionServiceDelegate.kt
boek: @MatthewTighe I haven't noticed any breakage and I used the same approach PocketCasts uses here: https://github.com/Automattic/pocket-casts-android/blob/dfb5f8686d60b27adaad033977b5b772afbe826a/modules/services/repositories/src/main/java/au/com/shiftyjelly/pocketcasts/repositories/playback/PlaybackService.kt#L259. That being said I can't say for sure 🤔

MatthewTighe: Cool, I would expect keeping the app from crashing is highest priority here anyway - was just curious whether this had any side-effects like notifications disappearing unexpectedly.

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/service/AbstractMediaSessionService.kt
boek: I was thinking more generally from the point of view of controlling dependencies and exposing the API. If we're injecting something into a component the caller should be responsible for deciding what to do, not the component.

android-components/components/feature/media/src/test/java/mozilla/components/feature/media/service/MediaSessionServiceDelegateTest.kt
```diff
@@ -460,9 +469,59 @@ class MediaSessionServiceDelegateTest {
         verify(controller).pause()
     }
 
+    @Test
+    fun `WHEN on API level is 31 startForeground throws an exception THEN catch and pass the exception to the crash reporter`() = runTestOnMain {
```
MatthewTighe, line 473: nit:
```suggestion
    fun `GIVEN device is at least API level 31 WHEN startForeground throws an exception THEN catch and pass the exception to the crash reporter`() = runTestOnMain {
```

android-components/components/feature/media/src/test/java/mozilla/components/feature/media/service/MediaSessionServiceDelegateTest.kt
```diff
@@ -460,9 +469,59 @@ class MediaSessionServiceDelegateTest {
...
+        val notification: Notification = mock()
+        delegate.notificationHelper = coMock {
+            doReturn(notification).`when`(this).create(mock(), delegate.mediaSession)
+        }
+
+        val exception = ForegroundServiceStartNotAllowedException("Test thrown exception")
+        doThrow(exception).`when`(service).startForeground(anyInt(), any())
+
+        setSdkInt(31)
+        delegate.startForeground(mock())
```
MatthewTighe, line 487: nit: I think it would make sense to include setting the SDK with the setup part of the test above
```suggestion
        delegate.notificationHelper = coMock {
            doReturn(notification).`when`(this).create(mock(), delegate.mediaSession)
        }

        val exception = ForegroundServiceStartNotAllowedException("Test thrown exception")
        doThrow(exception).`when`(service).startForeground(anyInt(), any())
        setSdkInt(31)

        delegate.startForeground(mock())
```

MatthewTighe (approved):
Couple nits, otherwise this LGTM. I'm good with moving forward with the dependency strategy laid out as is, though I would prefer to have a second opinion weigh in about it in regards to [our comment chain here](https://github.com/mozilla-mobile/firefox-android/pull/505/files#r1066155873) before landing (if only for my own edification). Maybe @jonalmeida?

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/service/AbstractMediaSessionService.kt
jonalmeida: Sorry, I have a boring response: it depends.

It depends on what the thing is. My weak opinion is that having the `CrashReporting` null by default is nicer because, as an embedder of the media component, I don't want to implement crash reporting in order to use the feature (it's not part of the core functionality).

We have [examples](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/AbstractCustomTabsService.kt#42) where there is non-core functionality of a service that is null by default because of non-core functionality. 

Again, it depends. 🙂

boek: I'm good with this :)

MatthewTighe:
Thanks for polling a second opinion, I'm happy to have this land whenever 🙂

PR #540, opened 2023-01-13
Bug 1809972 - Cookie banner CFR: code should not try to replace localized text when it can pass a variable

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1809972


Mugurell:
> Would it make sense to refactor `ClickableSubstringLink()` to not search for the text, but replace the variable directly with the code you need?

The current implementation allows the app to work with just one string (including all declared substitutions) and then searching in that string and marking as a link a certain substitution.
This allows for localizers to adapt the string and change the order of the substitutions without this impacting the app which will search for a specific part of the big string, another smaller string, to mark as a link.

If I understand correctly you are proposing for us to join the strings in code instead of relying on substitutions in the `strings.xml` file? If we do that the default order for the multiple strings that would compose the displayed text might not be the one best suiting each locale.

flodolo:
> f I understand correctly you are proposing for us to join the strings in code instead of relying on substitutions in the `strings.xml` file

Not at all. 

I was suggesting to replace `%2$s` directly with the code you need (link using `cfr_cookie_banner_link`), instead or replacing it with the string from `cfr_cookie_banner_link`, then search for the text and replace it again with code. But looking at the amount of work required to add the link, it might not make sense.

focus-android/app/src/main/res/values/strings.xml
```diff
@@ -696,11 +696,11 @@
     <string name="cookie_banner_exception_panel_description_state_off_for_site">%1$s can try to automatically reject cookie requests.</string>
 
     <!-- Contextual Feature Recommendation Popups -->
-    <!-- CFR for Cookie Banner (Banner Info Message). The placeholder will be updated with the app name. -->
-    <string name="cfr_for_cookie_banner">%1$s tries to reject cookie requests to dismiss annoying cookie banners.\n\nManage cookie banner preferences in settings.</string>
+    <!-- CFR for Cookie Banner (Banner Info Message). %1$s will be replaced by the app name, %2$s will be an active link using the string cfr_cookie_banner_link as text. -->
+    <string name="cfr_cookie_banner">%1$s tries to reject cookie requests to dismiss annoying cookie banners.\n\nManage cookie banner preferences in %2$s.</string>
```
Amejia481, line 700: @iorgamgabriel Is safe to update already existing strings? In Fenix instead of updating we re-create the string with a new ID and mark it for future deletion as it could cause crash see https://github.com/mozilla-mobile/firefox-android/blob/main/focus-android/docs/Removing-strings.md

iorgamgabriel: Done

Amejia481 (approved):

PR #542, opened 2023-01-13
Bug 1810150 - don't run pull-request task graphs when we already have a push graph

jcristau:
### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/main/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Milestone**: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
- **Breaking Changes**: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1810150

PR #556, opened 2023-01-17
Revert "Bug 1810150 - don't run pull-request task graphs when we already have a push graph"

jcristau:
Reverts mozilla-mobile/firefox-android#542

The branch protection rule requires complete-pr on main, so we need to coordinate a change to that rule before we can re-land this.
https://bugzilla.mozilla.org/show_bug.cgi?id=1810150

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,9 +1,7 @@
 queue_rules:
   - name: default
     conditions:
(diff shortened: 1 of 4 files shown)


DonalMe (approved):
LGTM

PR #575, opened 2023-01-19
Bug 1811175 - Implement fallback protocol for sponsored tiles

Alexandru2909:
For sponsored tiles fallback [protocol](https://docs.google.com/document/d/1IbEgAmav_3HPXyZHRmbFP93qxwFQIQBCGvytD70SGeE/edit#), we should use the `max-age` and `stale-if-error` headers to calculate up to what time a set of sponsored tiles is valid.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1811175


android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -49,26 +57,24 @@ class ContileTopSitesProvider(
     // The last modified time of the disk cache.
     @VisibleForTesting
     @Volatile
-    internal var diskCacheLastModified: Long? = null
+    internal var diskCacheMaxAge: Long? = null
```
rocketsroger, line 60: ❓ I find I'm getting confused by `maxCacheAgeInMInutes` and `diskCacheMaxAge`.  Is it possible to have some helper functions for `maxCacheAgeInMinutes` so we don't need a separate variable `diskCacheMaxAge`?

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -24,7 +28,10 @@ import java.util.Date
 
 internal const val CONTILE_ENDPOINT_URL = "https://contile.services.mozilla.com/v1/tiles"
 internal const val CACHE_FILE_NAME = "mozilla_components_service_contile.json"
+internal const val CACHE_VALID_FOR_KEY = "valid_for"
+internal const val CACHE_TOP_SITES_KEY = "tiles"
 internal const val MINUTE_IN_MS = 60 * 1000
```
rocketsroger, line 33: 🔍 nit: We should remote these and use DateUtils' `MINUTE_IN_MILLIS` and `SECOND_IN_MILLIS`

rocketsroger (commented):
Looks good, I have some questions/suggestions.

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -160,24 +186,14 @@ class ContileTopSitesProvider(
...
+internal fun Headers.computeValidFor(): Long =
+    getAll("cache-control").sumOf {
+        val valueList = it.split("=")
+            .map { item -> item.trim() }
+
+        if (valueList.size == 2 && valueList[0] in listOf("max-age", "stale-if-error")) {
+            valueList[1].toLong()
+        } else {
+            0L
+        }
```
MatthewTighe, line 198: nit: IMO it feels a little odd to split how we're handling extension functions in this package. My preference would be to move the `JSONObject` extensions back to this file since they don't appear to be used elsewhere, and it makes sense to test the extensions in the `ContileTopSitesProviderTest` as well. They seem to still be a single logical unit to me.

I think it would also be fine to move this `Headers` extension to a separate file as well if you are concerned about file length.

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -98,12 +104,23 @@ class ContileTopSitesProvider(
...
-                    JSONObject(responseBody).getTopSites().also {
-                        if (maxCacheAgeInMinutes > 0) {
-                            writeToDiskCache(responseBody)
-                        }
-                    }
+                    writeToDiskCache(
+                        if (maxCacheAgeInMinutes == -1L) {
+                            response.headers.computeValidFor() * MILLIS_PER_SEC
+                        } else {
+                            maxCacheAgeInMinutes * MINUTE_IN_MS
```
MatthewTighe, line 118: Looking at this, I'm wondering if it would be worth changing our local cache ages to also be in seconds to better match the server and keep the number of types of conversion lower.

MatthewTighe (commented):
First pass. Thanks for taking this on! It looks a complicated domain

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
Alexandru2909: Added the `CacheState` for extra clarity, thank you for the suggestion!

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -117,51 +133,138 @@ class ContileTopSitesProvider(
...
+        /**
+         * Update local and server max age values
+         */
+        fun computeMaxAges(lastModified: Long, localMaxAge: Long, serverMaxAge: Long): CacheState =
+            this.copy(
+                isCacheValid = true,
+                localCacheMaxAge = lastModified + localMaxAge,
+                serverCacheMaxAge = lastModified + serverMaxAge,
+            )
+    }
```
MatthewTighe, line 245: I think these types and comments add a lot of semantic value, thanks again for indulging me!

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -98,12 +105,21 @@ class ContileTopSitesProvider(
...
                 return try {
-                    JSONObject(responseBody).getTopSites().also {
-                        if (maxCacheAgeInMinutes > 0) {
-                            writeToDiskCache(responseBody)
-                        }
-                    }
+                    val jsonBody = JSONObject(responseBody)
+                    writeToDiskCache(
+                        response.headers.computeValidFor() * DateUtils.SECOND_IN_MILLIS,
+                        jsonBody.getJSONArray("tiles"),
```
MatthewTighe, line 119: nit: I think `"tiles"` could be `CACHE_TOP_SITES_KEY` here as well. 

It also feels a little funky to me that we are calling `jsonBody.getJSONArray("tiles")` here and then immediately doing basically the same thing with `jsonBody.getTopSites()`. I can't immediately think of a clean way to resolve that so I think we can keep it as is unless something comes to your mind.

MatthewTighe (approved):
Couple small nits, but this otherwise LGTM! Confirmed top sites are still appearing in Fenix and AFAICT this addresses the changes defined in the new protocol as well

PR #586, opened 2023-01-20
Bug 1811448 - Expose the GeckoView release channel in EngineVersion utils

Mugurell:
This will help conditionally execute code depending on the GeckoView release channel which may have different functionalities enabled at compile time.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1811448

### DIFF
--- a/android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineTest.kt
+++ b/android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineTest.kt
@@ -27,6 +27,7 @@ import mozilla.components.concept.engine.UnsupportedSettingException
 import mozilla.components.concept.engine.content.blocking.TrackerLog
 import mozilla.components.concept.engine.mediaquery.PreferredColorScheme
 import mozilla.components.concept.engine.serviceworker.ServiceWorkerDelegate
+import mozilla.components.concept.engine.utils.EngineReleaseChannel
 import mozilla.components.concept.engine.webextension.Action
 import mozilla.components.concept.engine.webextension.WebExtension
 import mozilla.components.concept.engine.webextension.WebExtensionDelegate
@@ -1906,6 +1907,7 @@ class GeckoEngineTest {
 
         assertTrue(version.major >= 69)
         assertTrue(version.isAtLeast(69, 0, 0))
+        assertTrue(version.releaseChannel != EngineReleaseChannel.UNKNOWN)
     }
 
     @Test
--- a/android-components/components/concept/engine/src/test/java/mozilla/components/concept/engine/utils/EngineVersionTest.kt
+++ b/android-components/components/concept/engine/src/test/java/mozilla/components/concept/engine/utils/EngineVersionTest.kt
@@ -68,6 +68,100 @@ class EngineVersionTest {
         assertEquals("68.3.0esr", "68.3esr".toVersion().toString())
         assertEquals("68.0.0", "68.0".toVersion().toString())
     }
+
+    @Test
+    fun `GIVEN a nightly build of the engine WHEN parsing the version THEN add the correct release channel`() {
+        val result = EngineVersion.parse("0.0.1", "nightly")?.releaseChannel
+
+        assertEquals(EngineReleaseChannel.NIGHTLY, result)
+    }
+
+    @Test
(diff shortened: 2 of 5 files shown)


rvandermeulen:
@Mugurell Should we add a check that the channel matches Gecko.channel defined in Gecko.kt?

Mugurell:
> @Mugurell Should we add a check that the channel matches Gecko.channel defined in Gecko.kt?

That would be this line https://github.com/mozilla-mobile/firefox-android/pull/586/files#diff-721332683fd5919591f4f8e05f46a520007032faedcb30a3d9f921e3a1fb25e9R1890 which would ensure we always get passed `nightly` / `beta` / `release`.
This tests are affected by the same issue that they try to solve: they will be run with different versions of Geckoview set by the [`version` from `Gecko.kt`](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/plugins/dependencies/src/main/java/Gecko.kt#L17) but cannot know which.
Open for suggestions.

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineTest.kt
```diff
@@ -1886,6 +1887,7 @@ class GeckoEngineTest {
 
         assertTrue(version.major >= 69)
         assertTrue(version.isAtLeast(69, 0, 0))
+        assertTrue(version.releaseChannel != EngineReleaseChannel.UNKNOWN)
```
jonalmeida, line 1890: Nice!

android-components/components/concept/engine/src/test/java/mozilla/components/concept/engine/utils/EngineVersionTest.kt
```diff
@@ -68,6 +68,78 @@ class EngineVersionTest {
...
+    }
+
+    @Test
+    fun `GIVEN an nightly release type WHEN comparing to other versions THEN return the expected result`() {
+        val version = "0.0.1"
+        val nightly = EngineVersion.parse(version, "nightly")
+
+        assertEquals(0, nightly!!.compareTo(nightly))
+        assertTrue(nightly > EngineVersion.parse(version, "unknown")!!)
+        assertTrue(nightly < EngineVersion.parse(version, "beta")!!)
```
jonalmeida, line 118: Wouldn't it be the opposite way? Nightly is newer so should be greater than beta?

jonalmeida (approved):
Looks good! Just one clarification about the `compareTo` check.

This looks safe to land before the weekend if desired, since it's not code that is currently used anywhere.

android-components/components/concept/engine/src/test/java/mozilla/components/concept/engine/utils/EngineVersionTest.kt
csadilek: I think this ends up just comparing the values in the ReleaseChannel enum because of the `compareTo` changes above. In reality, Nightly will have a greater major version anyway, yielding a different result.

A Nightly of the same major version would be older than beta, so this seems correct? A Nightly of a higher major version would be newer. Maybe that's a test we should be adding :)

csadilek (approved):
Looks good to me! This discussion needs another check and ideally an additional test: https://github.com/mozilla-mobile/firefox-android/pull/586/files#r1082960102

android-components/components/concept/engine/src/test/java/mozilla/components/concept/engine/utils/EngineVersionTest.kt
jonalmeida: Right! I should have make it more clear that the reason it was giving us this result is because of the (unintended?) side-effect of the enum positioning.

Mugurell: The enum comparison was intended with the members being arranged in a preference order with the last being preferable to the ones above.
I went with thinking about two jars, with version `100.0.0`, one from Nightly, one from Release.
The `100.0.0` - `release` jar is newer (by ~2 months) and also (hopefully) more stable so for the same exact version the `release` jar is preferable and a comparison will return a positive result.

Thank you for the suggestions about a new test! In the current code the release channel comparison is done last so I didn't think a test is needed for different versions and different release channels since we already have tests for different versions but that code can be refactored and so a test would help catch regressions.

PR #640, opened 2023-01-27
Bug 1813157 - Use the new l10n import action for the monorepo

gabrielluong:
Co-authored-by: "Francesco Lodolo (:flod)" <flodolo@mozilla.com>

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1813157


.github/workflows/import-l10n.yml
```diff
@@ -0,0 +1,78 @@
...
+
+name: Import l10n translations
+
+permissions:
+  contents: write
+  pull-requests: write
+
+on:
+  schedule:
+    - cron: "10 0 * * *"
```
flodolo, line 13: I realized afterwards that this might be running 10 minutes after the current TaskCluster workflow.
https://github.com/mozilla-l10n/android-l10n-tooling/blob/master/.cron.yml#L19

Depending on how we approach this (keep them running in parallel for a few days), we might want to run this one a couple of hours later.

.github/workflows/import-l10n.yml
```diff
@@ -0,0 +1,78 @@
...
+                echo "Link: $pr_link"
+            else
+                echo "Opening new pull request."
+                gh pr create --fill
+            fi
+          else
+            echo "No changes found."
+          fi
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
flodolo, line 78: I'm not sure how Mergify determines the author of the PR
https://github.com/mozilla-mobile/firefox-android/blob/main/.mergify.yml#L27

```
author~=(mozilla-l10n-automation-bot|github-actions\[bot\])
```

The user name is set up on line 55, but it might not be enough (if it is, it's kind of a security concern, since you can impersonate others so easily…).

The right approach would be to create a specific token from the `mozilla-l10n-automation-bot` user, then store it in the repository's settings (e.g. as `L10N_RELENG_TOKEN`).

The alternative is to switch the user settings (line 55-56) to the GitHub Actions bot, and let it run with the default `GITHUB_TOKEN`. One caveat is that other GitHub workflows won't run on the pull request in this case (I don't think you have any at the moment).

Let's see what @JohanLorenzo thinks, since he's probably more used to setting up these.

gabrielluong: Another idea here would be to drop `mozilla-l10n-automation-bot` and just use the `github-actions[bot]`

flodolo: See my previous comment

> The alternative is to switch the user settings (line 55-56) to the GitHub Actions bot, and let it run with the default GITHUB_TOKEN. One caveat is that other GitHub workflows won't run on the pull request in this case (I don't think you have any at the moment).

It will work if you don't need to run other GitHub Workflows on the pull request. For example, the Bugzilla workflows won't be run on this pull request (to avoid possible loops).

JohanLorenzo: > For example, the Bugzilla workflows won't be run on this pull request (to avoid possible loops).

I'm sorry, I'm lacking context. What is a Bugzilla workflow? 

> I'm not sure how Mergify determines the author of the PR

On top of my head, Mergify uses the value given by Github. This means, the PR author is the one that created the PR no matter who made the last commit. In other words, if I click on the WebUI to create a PR, then I am the author. If something creates a PR through the Github API, then this something is the PR author. The only way the Github API knows what this something is, is thanks to the Github token.

Thus, the real question is: where does `secrets.GITHUB_TOKEN` comes from? If its a token generated on the @mozilla-l10n-automation-bot account, then the PR is owned by this account. Otherwise, it's likely going to be `github-actions[bot]`

flodolo: > > For example, the Bugzilla workflows won't be run on this pull request (to avoid possible loops).
> 
> I'm sorry, I'm lacking context. What is a Bugzilla workflow?

https://github.com/mozilla-mobile/firefox-android/tree/main/.github/workflows

There seems to be a couple of Bugzilla-related workflows, and one for Glean, that run on pull request events.

> Thus, the real question is: where does `secrets.GITHUB_TOKEN` comes from? If its a token generated on the @mozilla-l10n-automation-bot account, then the PR is owned by this account. Otherwise, it's likely going to be `github-actions[bot]`

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

`GITHUB_TOKEN` is created as part of the workflow, it doesn't belong to the user specified. For that, you need a specific token generated by the user (`mozilla-l10n-automation-bot`), and stored as a secret in the repository.

That's what we're doing here, for example
https://github.com/mozilla-l10n/android-l10n/blob/master/.github/workflows/import_ac.yaml#L62

A token was generated by the user `mozilla-pontoon` and stored as a secret named `ANDROID_GITHUB_TOKEN`.

JohanLorenzo: Thanks! In this case, if I'm not mistaken `secrets.GITHUB_TOKEN` makes the PR authored by `github-actions[bot]`. I'm fine with that. We can then remove `mozilla-l10n-automation-bot` from the mergify config, if that is indeed true.

.github/workflows/import-l10n.yml
```diff
@@ -0,0 +1,78 @@
...
+          --toml l10n/mozilla-mobile/focus-android/l10n.toml
+          --dest source/focus-android
+      - name: Commit changes and open pull request
+        run: |
+          # Only try to commit if there are pending changes
+          cd source
+          if [[ $(git diff --exit-code) || $(git ls-files --other --exclude-standard) ]]
+          then
+            branch="automation/import-l10n"
+            git config user.name "Mozilla L10n Automation Bot"
```
flodolo, line 55: Let's update these to GitHub actions then.

flodolo (approved):
I would also tweak the schedule timing, as suggested in the comment, so that it runs well before TaskCluster and Mergify as time to merge.

.github/workflows/import-l10n.yml
gabrielluong: Updated this to “0 3 * * *”

PR #657, opened 2023-01-30
Bug 1795423 - Replace deprecated Fragment.requestPermissions API in sample browser.

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1795423


android-components/components/support/utils/src/main/java/mozilla/components/support/utils/ext/Fragment.kt
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Requests permissions and handles results inside a [Fragment].
+ */
+fun Fragment.requestInPlacePermissions(
+    permissionsToRequest: Array<String>,
+    onResult: (Map<String, Boolean>) -> Unit,
+) {
+    requireActivity().activityResultRegistry.register(
+        "permissionsRequest",
```
Amejia481, line 18: I'm not that familiar with the new API, I'm a bit confused about having the [key](https://developer.android.com/reference/androidx/activity/result/ActivityResultRegistry#register(java.lang.String,androidx.lifecycle.LifecycleOwner,androidx.activity.result.contract.ActivityResultContract%3CI,O%3E,androidx.activity.result.ActivityResultCallback%3CO%3E)) parameter as a fixed string ""permissionsRequest". Could you help me to clarify why ? Thanks in advance :)

Amejia481: Before we has request code, as an identifier for each request now it seems that we has the same identifier fall of the requests?

Amejia481: Is it possible to add tests for this extensions function?

mcarare: It's not a new API, it has been used in Focus for a while.

mcarare: Thank you for your review, I will look into this more.

mcarare: As far as I have seen in tests, the key parameter in a `register` call is used to avoid mid-air collisions by requesting the same permission multiple times. This should not happen anyway, but we might send unique keys depending on the caller. maybe class name of the feature, etc

mcarare:
Also tested and the permission requests work as expected.
Would appreciate more testing besides a review, even if it is the sample browser, as I intend to use the extension in other places.

Amejia481 (approved):
✅ 
Tested on sample browser, with downloads, site permission and file uploads (Noticed bug [#1812605](https://bugzilla.mozilla.org/show_bug.cgi?id=1812605))  was working as expected, excepted but the known bug.

PR #678, opened 2023-02-01
Bug 1814338 - Improve l10n GitHub Workflows

flodolo:
CC @gabrielluong 

This pull request restores the original cronjob time for the "import l10n" workflow, and improve the commit message for clarity.

Changes to the sync workflows are mostly for clarity:
* Use of env variables to avoid repetitions, and simplify reuse for other products (e.g. Fenix)
* Use latest versions of ubuntu and actions (most use a deprecated version of node at this point)

There should be no practical changes to the workflow, I've only used `uplift` instead of `sync` in the branch name. Both actions have been tested on my fork.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1814338

### DIFF
--- a/.github/workflows/focus-sync-strings.yml
+++ b/.github/workflows/focus-sync-strings.yml
@@ -2,45 +2,49 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/
 
-name: "Focus - Sync Strings"
+name: Focus - Uplift Strings
 
 on:
   schedule:
-    - cron: '0 2 * * *'
+    - cron: "0 2 * * *"
   workflow_dispatch:
 
 jobs:
   main:
-    name: "Sync Strings"
-    runs-on: ubuntu-20.04
+    name: Uplift strings
+    runs-on: ubuntu-latest
     steps:
-      - name: "Discover Focus Beta Version"
-        id: focus-android-beta-version
+      - name: Discover Focus beta version
+        id: discover-beta-version
         uses: mozilla-mobile/fenix-beta-version@4.2.2
-      - name: "Skip non-beta versions"
+      - name: Skip non-beta versions
         uses: andymckay/cancel-action@0.2
-        if: ${{ steps.focus-android-beta-version.outputs.beta_version == '' }}
-      - name: "Checkout Main Branch"
-        uses: actions/checkout@v2
+        if: ${{ steps.discover-beta-version.outputs.beta_version == '' }}
+      - name: Store ENV variables
+        id: env-variables
+        run: |
+          echo "beta_branch=releases_v${{ steps.discover-beta-version.outputs.beta_version }}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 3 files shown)


flodolo:
> The comment message should look like:

Is this supposed to say **commit** message instead? 🤔

.github/workflows/focus-sync-strings.yml
```diff
@@ -2,45 +2,50 @@
...
-        if: ${{ steps.focus-android-beta-version.outputs.beta_version == '' }}
-      - name: "Checkout Main Branch"
-        uses: actions/checkout@v2
+        if: ${{ steps.discover-beta-version.outputs.beta_version == '' }}
+      - name: Store ENV variables
+        id: env-variables
+        run: |
+          echo "project_name=Focus Strings" >> $GITHUB_OUTPUT
+          echo "beta_branch=releases_v${{ steps.discover-beta-version.outputs.beta_version }}" >> $GITHUB_OUTPUT
+          echo "branch_name=automation/uplift-focus-strings-${{ steps.discover-beta-version.outputs.beta_version }}" >> $GITHUB_OUTPUT
```
gabrielluong, line 28: I would consider dropping `project_name` and `branch_name` as variables from these changes. I think it adds unnecessary complexity to something rather simple since these variables are only ever used once and are pretty simple strings. `beta_branch` is at least used multiple times so that is fine.

.github/workflows/focus-sync-strings.yml
```diff
@@ -2,45 +2,50 @@
...
+      - name: Checkout default branch
+        uses: actions/checkout@v3
         with:
-          path: main
-          ref: main
-      - name: "Checkout Beta Branch"
-        uses: actions/checkout@v2
+          path: default
```
gabrielluong, line 32: Could we just stick with doing what this was doing before with the exact path and ref? I think that we have more clarity when we are explicitly saying we are checking out the main branch compare to default branch because that begs the question of what is a  default branch and how does the actions/checkout parameters work. 

At the end of the day, I will defer to your judgement since I think the l10n team should ideally just own these actions. 

We will have to start moving away from GitHub actions to utilize taskcluster around March as preparation for the m-c migration so the amount of effort we should spend here should be fairly minimal. My concern here is that we are losing clarity about what we are doing here with this change.

gabrielluong (approved):

.github/workflows/focus-sync-strings.yml
flodolo: I personally don't think it adds complexity, and it puts all aspects specifics to a product in one place.
To adapt this to Fenix you would only need to update this section, instead of 2, and ending up with different versions in each place.

With that said, you're likely the one who needs to maintain these going forward, so I'll go with your preference.

flodolo:
Updated, rebased and pushed. Also tested on my fork.

I left `${{ github.ref_name }}` in the commit messages, since I think it's useful to make sure we're actually on the right branch.

PR #732, opened 2023-02-08
Bug 1813671 - Update Gradle to version 7.6 and Android Gradle Plugin to 7.4.1

mcarare:
See https://developer.android.com/studio/releases/gradle-plugin#jvm-11-bytecode for details.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


mcarare:
Due to the high number of files, I suggest reviewing one commit at a time.

android-components/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,7 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.1-bin.zip
+# this is the network timeout default value in ms.
+networkTimeout=10000
```
rahulsainani, line 5: ❓ Do we need to set it if it's already set by default to 10s?

mcarare: Not necessarily, but this is a change in the new version, the ability to configure network timeout. I added this here for surfacing the new configuration param and its default value.

fenix/app/build.gradle
```diff
@@ -250,7 +250,7 @@ android.applicationVariants.all { variant ->
 
     println("----------------------------------------------")
     println("Variant name:      " + variant.name)
-    println("Application ID:    " + [variant.mergedFlavor.applicationId, variant.buildType.applicationIdSuffix].findAll().join())
+    println("Application ID:    " + [variant.applicationId, variant.buildType.applicationIdSuffix].findAll().join())
```
rahulsainani, line 253: Nice! This deprecation used to bother me

Although I don't think this change introduced this "issue" but I see this printed in the build log
`Application ID:    org.mozilla.fenix.debug.fenix.debug`

❓ should we just print the `variant.applicationId` ?

mcarare: I noticed the same, but when upgrading Gradle I prefer not to change existing behavior. And the same value is currently printed in the build log. We could open another issue to investigate if this is intentional or if should we change it.

rahulsainani: Sounds good! 👍

rahulsainani (approved):
Really nice work! 👌 

Tested fenix and focus locally and they work well. Thanks for looking into this and fixing the deprecation warnings as well! 

Would be great to get another pair of eyes on this as well as this is a bigger change.

JohanLorenzo (approved):
Thank you so much for splitting this PR into atomic commits! ❤️ At first, I was afraid to review this +1,5k -1,3k PR. Then, I looked at each commit individually and it was became so quick to read! 

Also, kudos for explaining why we're doing things in commit messages. TIL these deprecated gradle options! 

This PR looks good to me. To me, the next step would be to remove the jdk8 from the CI docker images: 

https://github.com/mozilla-mobile/firefox-android/blob/f5663b970396dc3d39dedfadfa4c8b28389b7316/taskcluster/docker/base/Dockerfile#L33

This way, we're sure we don't build anything against openjdk8 anymore (cc @csadilek). No need to handle it in this PR, though. It can definitely be done in a followup.

android-components/build.gradle
```diff
@@ -238,7 +238,7 @@ subprojects {
 
                 tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
                     kotlinOptions {
-                        jvmTarget = "1.8"
+                        jvmTarget = "11"
```
csadilek, line 240: @mcarare I think there are a few more occurrences of `jvmTarget` to update: https://github.com/mozilla-mobile/firefox-android/search?q=jvmTarget

The nimbus-gradle-plugin we will remove from this repo this week.

csadilek: (But looks like Focus files need an update.)

mcarare: I initially did not want to touch the `jvmTarget` for glean and telemetry and I only updated the tasks lint was complaining about. I will update the target for telemetry and glean also in a separate commit (so it can be easily reverted if any side-effects occur.)

PR #793, opened 2023-02-14
Bug 1816715 - Do not optimize away UI tests task when a change happen…

JohanLorenzo:
… in either fenix or focus

In https://github.com/mozilla-mobile/firefox-android/pull/791, UI tests should have definitely run. I took the generated params file and here's what taskgraph-diff returns with these changes.

```diff
--- optimized_task_graph@d85fd9ff98fa
+++ optimized_task_graph@main
@@ -1,20 +1,22 @@
+build-apk-fenix-android-test-debug
+build-apk-fenix-debug
 build-docker-image-android-build
 build-docker-image-base
 build-docker-image-ui-tests
 complete-pr
 complete-pr-1
 complete-pr-2
 complete-pr-3
 external-gradle-dependencies-browser-domains
 external-gradle-dependencies-browser-engine-gecko
 external-gradle-dependencies-browser-engine-system
 external-gradle-dependencies-browser-errorpages
 external-gradle-dependencies-browser-icons
 external-gradle-dependencies-browser-menu
 external-gradle-dependencies-browser-menu2
 external-gradle-dependencies-browser-session-storage
 external-gradle-dependencies-browser-state
 external-gradle-dependencies-browser-storage-sync
 external-gradle-dependencies-browser-tabstray
 external-gradle-dependencies-browser-thumbnails
 external-gradle-dependencies-browser-toolbar
@@ -117,22 +119,25 @@
 external-gradle-dependencies-support-test-libstate
 external-gradle-dependencies-support-utils
 external-gradle-dependencies-support-webextensions
 external-gradle-dependencies-test-components
 external-gradle-dependencies-tooling-detekt
 external-gradle-dependencies-tooling-fetch-tests
 external-gradle-dependencies-tooling-glean-gradle
 external-gradle-dependencies-tooling-lint
 external-gradle-dependencies-tooling-nimbus-gradle
 external-gradle-dependencies-ui-autocomplete
 external-gradle-dependencies-ui-colors
 external-gradle-dependencies-ui-fonts
 external-gradle-dependencies-ui-icons
 external-gradle-dependencies-ui-tabcounter
 external-gradle-dependencies-ui-widgets
 fetch-android-sdk-3859397
 lint-compare-locales-fenix
 lint-detekt-fenix
 lint-ktlint-fenix
 lint-lint-fenix
+signing-apk-fenix-android-test-debug
+signing-apk-fenix-debug
 toolchain-linux64-android-sdk-linux-repack
+ui-test-apk-fenix-arm-debug
```
https://bugzilla.mozilla.org/show_bug.cgi?id=1816715

### DIFF
--- a/taskcluster/android_taskgraph/gradle.py
+++ b/taskcluster/android_taskgraph/gradle.py
@@ -7,6 +7,11 @@ def get_gradle_project(task):
     attributes = task.get("attributes", {})
     gradle_project = attributes.get("component")
     if not gradle_project:
+        shipping_product = attributes.get("shipping-product", "")
+        if shipping_product:
+            return shipping_product
+
+        # TODO: Use only shipping-product attribute instead
         treeherder_group = attributes.get("treeherder-group", "")
         if "focus" in treeherder_group:
             gradle_project = "focus"
--- a/taskcluster/ci/ui-test-apk/kind.yml
+++ b/taskcluster/ci/ui-test-apk/kind.yml
@@ -44,6 +44,8 @@ task-defaults:
 
 tasks:
     focus-arm-debug:
+        attributes:
+            shipping-product: focus
         description: 'Focus UI tests with firebase'
         run-on-tasks-for: [github-pull-request, github-push]
         run:
@@ -72,6 +74,7 @@ tasks:
     focus-arm-nightly:
         attributes:
             build-type: nightly-firebase
+            shipping-product: focus
         description: 'UI tests on Nightly with firebase'
         run-on-tasks-for: [github-push]
         run-on-git-branches: [main]
@@ -100,6 +103,7 @@ tasks:
     focus-arm-beta:
         attributes:
             build-type: nightly-firebase
+            shipping-product: focus
         description: 'UI tests on Beta with firebase'
         run-on-tasks-for: [github-push]
         run-on-git-branches: [main, releases_v]
@@ -126,6 +130,8 @@ tasks:
             env:
                 GOOGLE_PROJECT: 'moz-focus-android'
     fenix-arm-debug:
+        attributes:
+            shipping-product: fenix
         description: Test Fenix
         run-on-tasks-for: [github-pull-request, github-push]
         run-on-git-branches: ["^((?!releases[_/]).+)$"]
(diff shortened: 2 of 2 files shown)


taskcluster/android_taskgraph/gradle.py
```diff
@@ -7,6 +7,11 @@ def get_gradle_project(task):
     attributes = task.get("attributes", {})
     gradle_project = attributes.get("component")
     if not gradle_project:
+        shipping_product = attributes.get("shipping-product", "")
```
JohanLorenzo, line 10: `shipping-product` is a concept used in gecko. It enables us to support both `devedition` and `firefox`. We should leverage this feature here too. We technically have at least to products: `focus` and `fenix`. 

This change is incomplete. There are other places where we should use `shipping-product` instead. However, this patch is blocking PRs from landing because we don't run UI tests there anymore. That's why I just handled the UI tests.

gbrownmozilla: Thanks for the info!

gbrownmozilla (approved):

taskcluster/android_taskgraph/gradle.py
jcristau: shipping-product in gecko is used in release graphs (per https://firefox-source-docs.mozilla.org/taskcluster/attributes.html#shipping-product), it seems weird to use that same attribute for tests IMO.

JohanLorenzo:
@mergifyio backport releases_v111 releases_v110

taskcluster/android_taskgraph/gradle.py
JohanLorenzo: I agree `shipping-product` was originally defined to be used in release promotion. That said, it's also used in some tests in gecko[1]. I think it makes sense to use this attribute because we won't have to define another one to carry the same data. 

[1] https://searchfox.org/mozilla-central/rev/6444ed17e9f5e0d4e7dabc13c930d46b529fca15/taskcluster/ci/startup-test/kind.yml#54

jcristau: startup-test in gecko is special, since it blocks release tasks, it's actually part of releases, in a sense.

jcristau: So I guess my point is we're not using the same attribute as gecko here, we're reusing the attribute name to mean something different, and IMO that's worse than coming up with a new name for it.

PR #818, opened 2023-02-15
Bug 1816932 - Add Maps to app links common sub domains

rocketsroger:
Add "maps" as part of the domain that we strip when deciding if two domains are the same.  (example: maps.google.com is the same as www.google.com)

The reason here is that we do not want to trigger the prompt to the user twice when the redirect from maps.google.com to www.google.com/maps happened.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1816932

### DIFF
--- a/android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt
+++ b/android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt
@@ -18,6 +18,7 @@ import mozilla.components.support.ktx.kotlin.tryGetHostFromUrl
 private const val WWW = "www."
 private const val M = "m."
 private const val MOBILE = "mobile."
+private const val MAPS = "maps."
 
 /**
  * This feature implements use cases for detecting and handling redirects to external apps. The user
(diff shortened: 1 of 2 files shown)


android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt
```diff
@@ -156,6 +158,7 @@ class AppLinksInterceptor(
             url.startsWith(WWW) -> url.replaceFirst(WWW, "")
             url.startsWith(M) -> url.replaceFirst(M, "")
             url.startsWith(MOBILE) -> url.replaceFirst(MOBILE, "")
+            url.startsWith(MAPS) -> url.replaceFirst(MAPS, "")
```
csadilek, line 161: Reading the `doNotIntercept` logic again reminded us of https://github.com/mozilla-mobile/android-components/issues/7614.

Let's add a comment here to describe **why** we're doing this...to not handle the first redirect separately, and so we don't open an additional prompt.

csadilek (approved):
👍 Let's just document this edge case below, as it's not obvious that this is to handle the redirect.

android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt
```diff
@@ -144,18 +145,21 @@ class AppLinksInterceptor(
...
+    // Determines if the transition between the two URLs is related.  If the two URLs
+    // are from the same website then the app links interceptor will not try to find an application to open it.
+    @VisibleForTesting
+    internal fun isSameDomain(url1: String?, url2: String?): Boolean {
         return stripCommonSubDomains(url1?.tryGetHostFromUrl()) == stripCommonSubDomains(url2?.tryGetHostFromUrl())
     }
 
+    // Remove subdomains that are ignored when determining if two URLs are from the same website.
     private fun stripCommonSubDomains(url: String?): String? {
         return when {
```
Mugurell, line 157: Drive by: With this method only removing at max one subdomain it allows for others to still be present.
In a quick search I found for example
> https://www.m.999shopbd.com

which can also be accessed through

> https://999shopbd.com

but `isSameDomain` will return `false` while I would've expected to get `true`.
Should I open a new ticket for this?

rocketsroger: Yes, please open a new issue for this.  I'm not sure for the app links case we need to make that determination but it will be good to investigate.  Thanks

Mugurell: Created https://bugzilla.mozilla.org/show_bug.cgi?id=1818492.

PR #820, opened 2023-02-15
Bug 1810044 1810045 - Delay reading clipboard and allow to paste non url

Timshel:
Ported PR from https://github.com/mozilla-mobile/fenix/pull/27351

In relation with [1810045](https://bugzilla.mozilla.org/show_bug.cgi?id=1810045) dot not read the clipboard content to determine if we may have an url available to paste.

This is similar to what is done for the Clipboard suggestion of the url bar  [SearchDialogFragment.kt#L519](https://github.com/mozilla-mobile/fenix/blob/47b7336387f020607f95f43d25024d22e6e8c6f0/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt#L519) 
And [SearchDialogFragment.kt#L486]( https://github.com/mozilla-mobile/fenix/blob/47b7336387f020607f95f43d25024d22e6e8c6f0/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt#L486)

This logic was initially introduced in https://github.com/gexsi/user-agent-android/commit/ccc439e9c4119e4518e5bc63e948cd3ccb082310 but then reverted in https://github.com/mozilla-mobile/fenix/commit/5daa06f7dcd8b98fd98b4a7273f3093692c34209. To avoid NPE  switched to rely on `orEmpty()` and since the PopupWindow should almost always be short lived I believe it should be sufficient as even if the user trigger the `pasteAndGo` option with a non url it will just trigger an empty search.

Note on the use of `extractUrl()` for the `pasteAndGo` it was done to keep a similar handling to the clipboard suggestion :
https://github.com/mozilla-mobile/fenix/blob/47b7336387f020607f95f43d25024d22e6e8c6f0/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt#L366 

Additionally include the change for [1810044](https://bugzilla.mozilla.org/show_bug.cgi?id=1810044), since it concerns the same component, which is to allow to paste plain text even when the urlbar is empty, the ability to paste when there is an url was added in https://github.com/mozilla-mobile/fenix/pull/24294.
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1810044


fenix/app/src/main/java/org/mozilla/fenix/utils/ToolbarPopupWindow.kt
```diff
@@ -82,17 +85,17 @@ object ToolbarPopupWindow {
...
+                handlePaste(clipboard.text.orEmpty())
             }
         }
 
-        clipboardUrl?.let { url ->
+        if (binding.pasteAndGo.isVisible) {
             binding.pasteAndGo.setOnClickListener {
                 popupWindow.dismiss()
-                handlePasteAndGo(url)
+                handlePasteAndGo(clipboard.extractURL().orEmpty())
```
boek, line 98: Do we still want to call `handlePasteAndGo` with an empty string? We may want to check for a URL and then we can continue to do the logging `Logger("ToolbarPopupWindow").error("Clipboard contains URL but unable to read text")` if we can't find one

Timshel: My kotlin is a bit rusty, fixed it with a combo of `?.also` and `?:run`.
Could not find something cleaner to unwrap or handle the `null`, unless there is an helper ?

boek (approved):

boek:
Awesome, thanks for following up @Timshel!

Timshel:
@boek I have seen the revert, is it caused just by the deleted comment line (Sorry missed it was not expecting the PR to be merged so fast did a quick test but did not review the changes again :( ).

If if it's only this should I open another PR with the diff corrected ?
If it's something else anything I can do ?
Thank you :).

PR #825, opened 2023-02-15
Bug 1815637 - Add rtl support

mavduevskiy:
Some background on the problem: the default behavior for the `TextView` is to set the text direction based on the language direction. Let's say we have a list of `TextView`s with different languages populating each element. By default, English sentence would start from the left corner while Hebrew would start from the right. It creates a bit of inconsistency in the UI across the app for the English speaking users, and a lot of it for RtL speaking users.

Setting `textAlignment` to `viewStart` would override the default behavior and align all the text according to the layout direction, not the text direction. In that case, an RtL language would align all texts to the right, English including. It would be the desired result, but going through all the TextView in `fenix` and `a-c` and setting the parameter would be too much work to do. 

Instead, we could set `textAlignment` in styles.xml as a default for all the views, that would fix the `TextView` problem. But it creates a problem of overriding `gravity` value for text elements – elements that should have been positioned in center not align with the view start. There are much fewer elements of that sort, the problem was to find them.

I went through all the `TextView` with set gravity parameter to `center` or `center_horizontal`. Buttons have the default center gravity, so I added `textAlignment` to them as well. The last problem came from `Dialog`s, that would create buttons programmatically, with not xml files. For those, I have created `withCenterAlignedButtons` extension, and made sure that all dialogs that add buttons align them appropriately.  

FYI: The easiest way to test RtL mode is to set "Force RTL layout direction" in the dev options on your device.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1815637


android-components/components/ui/widgets/src/main/java/mozilla/components/ui/widgets/Extentions.kt
```diff
@@ -0,0 +1,28 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.ui.widgets
```
mavduevskiy, line 5: wasn't sure where to put it, is there a better place?

jonalmeida: This seems like a good place!

android-components/components/support/base/build.gradle
```diff
@@ -67,6 +67,7 @@ dependencies {
     implementation ComponentsDependencies.androidx_lifecycle_viewmodel
 
     api project(":concept-base")
+    implementation project(':ui-widgets')
```
jonalmeida, line 70: Similar comment as [this one](https://github.com/mozilla-mobile/firefox-android/pull/1132#issuecomment-1457192471). Adding this dependency here to a support library has consequences for other components that depend on support libraries too.

Litmus test: does sync-storate and browser-state need to have ui-widgets as a dependency too?

jonalmeida: From my previous comment, about where to place the change you added to ui-widgets, I didn't realize that it meant adding ui-widgets to the support library. That's my fault for not catching that.

android-components/components/support/base/src/main/java/mozilla/components/support/base/dialog/DeniedPermissionDialogFragment.kt
```diff
@@ -36,7 +37,7 @@ class DeniedPermissionDialogFragment : DialogFragment() {
             .setPositiveButton(R.string.mozac_support_base_permissions_needed_positive_button) { _, _ ->
                 openSettingsPage()
             }
-        return builder.create()
+        return builder.create().withCenterAlignedButtons()
```
jonalmeida, line 40: We could do two things here:
1. Copy over the extension functions to this component to avoid making ui-widget a dependent.
2. Move this `DeniedPermissionDialogFragment` somewhere else. It seems suspicious that it's here. cc @Amejia481 - would you know a better place to move this fragment too?

jonalmeida (commented):
Leaving as a comment which this is still a draft.

mavduevskiy:
Hi, @jonalmeida ! Thanks for the preemptive comment :)

About moving `DeniedPermissionDialogFragment`. It is just used inside downloads feature. We could move it their or, if we want to keep it come accessible for other components, move to a broader UI folder, like `ui-widgets`? I wish we would have a very general UI focused components, for stuff like `SnackbarDelegate`, `DeniedPermissionDialogFragment`, which do not exist in one particular domain of the app. 

It would help keeping the UI focused elements separately from the more logic based entities (`SnabkbarDelegate` resides in `support-utils`, which consists mostly of non ui elements), but alas we don't have it yet. 

I wonder what your thoughts on this are.

jonalmeida:
> About moving `DeniedPermissionDialogFragment`. It is just used inside downloads feature. We could move it their or, if we want to keep it come accessible for other components, move to a broader UI folder, like `ui-widgets`? I wish we would have a very general UI focused components, for stuff like `SnackbarDelegate`, `DeniedPermissionDialogFragment`, which do not exist in one particular domain of the app.

Funny enough, in the linked PR above that @t-p-white worked on, that was the direction they decided to take i.e. move `SnackbarDelegate` to `ui-widgets`.

> I wonder what your thoughts on this are.

Here, I'm wondering if it's a bit more reserved to move the `DeniedPermissionDialogFragment` to the downloads feature as you suggested.

Two reasons for this:
1. The most obvious: it's not used by other components. While it's good that the class was written for re-usability, it doesn't seem to have been used as such in the end.
2. mcarare is working on a [refactoring](https://github.com/mozilla-mobile/firefox-android/pull/849) of _notification_ permissions to include the new Android 13 permission model and that would be better place to build-off for handling _generic_ denied permissions.

android-components/components/support/base/src/main/java/mozilla/components/support/base/dialog/DeniedPermissionDialogFragment.kt
Amejia481: Yeah, it's really odd that it was part of the `support-base`,  @mavduevskiy already moved it to feature download that seems a better fit.

jonalmeida (approved):
LGTM after some local sanity testing.

There is something to consider: as we write more new code, how do we prevent regressing? How would other/new developers know to always call this extension function?

We are fixing this problem today, but there is a loophole left out for tomorrow. Do we need a base dialog that auto applies `withCenterAlignedButtons` to it?

For now, let's land this either way - it has been sitting here for too long and you've put a lot of work into it. 🙂

PR #842, opened 2023-02-16
Bug 1807324 - Add macrobenchmarks for startup metrics

rahulsainani:
### What
This PR adds introduces [jetpack macrobenchmark](https://developer.android.com/topic/performance/benchmarking/macrobenchmark-overview) library to measure startup metrics with different startup modes - [Cold](https://developer.android.com/topic/performance/vitals/launch-time#cold), [Warm](https://developer.android.com/topic/performance/vitals/launch-time#warm) & [Hot](https://developer.android.com/topic/performance/vitals/launch-time#hot).

### How
– Add dependencies and create `benchmark` module
– Fight with Glean and fix dependency conflict
– Add `startupBenchmark` to capture startup metrics.
– Run using AndroidStudio just like any other test or using gradle `./gradlew benchmark:connectedBenchmarkAndroid Test` 

### How the results look?
Screenshot 2023-02-16 at 15 47 45

### Possible Next Steps
– Run this on CI so we can measure consistently and graph the metrics
– Identify critical app flows and integrate `FrameMetrics` and `TraceMetrics` for them
– Since this is a prerequisite to add [BaselineProfiles](https://developer.android.com/topic/performance/baselineprofiles/overview), generate and add baseline profiles to improve possibly improve app startup performance, especially for the first week of the new version release (when cloud profiles have not been loaded on the device).

### Issues
– Running using gradle `./gradlew benchmark:connectedBenchmarkAndroidTest` has an issue while installing the apk for the correct ABI. Is that something we've observed before?

```
Exception thrown during onBeforeAll invocation of plugin com.google.testing.platform.plugin.android.AndroidDevicePlugin.
Failed to install APK(s): /Users/rahulsainani/StudioProjects/firefox-android/fenix/app/build/outputs/apk/fenix/benchmark/app-fenix-x86_64-benchmark.apk
INSTALL_FAILED_NO_MATCHING_ABIS: INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries, res=-113
```

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1807324


fenix/benchmark/src/main/java/org/mozilla/fenix/benchmark/utils/MacroBenchmarkRule.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+fun MacrobenchmarkRule.measureRepeatedX(
+    packageName: String = TARGET_PACKAGE,
+    metrics: List<Metric>,
+    compilationMode: CompilationMode = CompilationMode.DEFAULT,
+    startupMode: StartupMode? = null,
+    @IntRange(from = 1)
+    iterations: Int = DEFAULT_ITERATIONS,
+    setupBlock: MacrobenchmarkScope.() -> Unit = {},
+    measureBlock: MacrobenchmarkScope.() -> Unit,
```
MatthewTighe, line 22: nit: I think it could make sense to split this into two functions, `measureRepeatedDefault` and `measureRepeatedX`(with no default iterations) just to increase clarity around iterations. That, or I'd consider just dropping the `X` from the name.

fenix/app/src/main/AndroidManifest.xml
```diff
@@ -49,6 +49,10 @@
         android:usesCleartextTraffic="true"
         tools:ignore="UnusedAttribute">
 
+        <profileable
+            android:shell="true"
+            tools:targetApi="29" />
```
MatthewTighe, line 54: Without looking too deeply, I'm curious whether adding this to our normal manifest could have any impact on production performance and if so whether there is there an easy way to create a "copy" app that we could use under test.

fenix/benchmark/src/main/java/org/mozilla/fenix/benchmark/utils/Constants.kt
```diff
@@ -0,0 +1,4 @@
+package org.mozilla.fenix.benchmark.utils
+
+const val TARGET_PACKAGE = "org.mozilla.fenix"
+const val DEFAULT_ITERATIONS = 5
```
MatthewTighe, line 4: Could these files be moved into the test source tree or is there a reason they need to be in the main source tree?

MatthewTighe (commented):
Couple small drive-by comments. I am very excited by this though, thanks for bringing it in! Would love to see follow-ups around scrolling the home screen and the tabs tray. Is there a plan to add this to our pipelines as well?

MarcLeclair:
> Couple small drive-by comments. I am very excited by this though, thanks for bringing it in! Would love to see follow-ups around scrolling the home screen and the tabs tray. Is there a plan to add this to our pipelines as well?
> 
> 

So, macrobenchmark is mostly used for something that happens once in a while (I.e start up). So this test shouldn't include that in my opinion since we want a clear view of start up. However, we do have microbenchmarks that do just that! I have a patch ready to land that I never landed since there was issues with dependencies a few months ago that seems resolved now. I think if we landed that we would have a good understanding of the entire Home Screen :)

fenix/app/src/main/AndroidManifest.xml
rahulsainani: Good question, looking into the docs, the [profileable tag](﻿﻿https://developer.android.com/guide/topics/manifest/profileable-element) makes the non debuggable builds profileable on 29+, it doesn't look like it has any impact on the performance since it's designed to benchmark release builds for more accurate perf metrics.

fenix/benchmark/src/main/java/org/mozilla/fenix/benchmark/utils/MacroBenchmarkRule.kt
rahulsainani: Yeah `measureRepeatedDefault` does read better, will update that 👍  The idea was to add one extension method with default params, there is already a `measureRepeated` function on `MacroBenchmarkScope` without the app specific defaults that this function calls

fenix/benchmark/src/main/java/org/mozilla/fenix/benchmark/utils/Constants.kt
rahulsainani: This is in the `benchmarks` module, which is like `androidTests` and don't go into the compiled main apk. AFAIK benchmark installs it's own apk that controls the main apk using ui automator.

MatthewTighe: Ah I missed that. Thanks!

fenix/benchmark/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+
+    buildTypes {
+        // This benchmark buildType is used for benchmarking, and should function like your
+        // release build (for example, with minification on). It's signed with a debug key
+        // for easy local/CI testing.
+        benchmark {
+            debuggable = true
+            if (gradle.hasProperty("localProperties.autosignReleaseWithDebugKey")) {
+                signingConfig signingConfigs.debug
+            }
```
MatthewTighe, line 33: Curious about 2 things here:
1. Does this module need to be debuggable for the profiling to work correctly? Just a little surprising given the emphasis on using release builds, but I didn't actually try the setup wizard
2. I think this `signingConfig` property will be added by the `releaseTemplate` closure used in the top-level [build file](https://github.com/mozilla-mobile/firefox-android/pull/842/files#diff-3eeb974dbea8ed2ac5d859292948457f089d0211ac697f866e4cf62c55576d4bR151). Is it necessary to have it in both places?

fenix/benchmark/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+ */
+configurations.all {
+    resolutionStrategy.capabilitiesResolution.withCapability("org.mozilla.telemetry:glean-native") {
+        def toBeSelected = candidates.find { it.id instanceof ModuleComponentIdentifier && it.id.module.contains('geckoview') }
+        if (toBeSelected != null) {
+            select(toBeSelected)
+        }
+        because 'use GeckoView Glean instead of standalone Glean'
+    }
+}
```
MatthewTighe, line 55: I don't really understand what's happening here, would you mind explaining it to me?

MatthewTighe (commented):
Had just a couple more things I wanted to touch on before approving, but this is looking good and I was able to verify the workflow locally.

> Running using gradle ./gradlew benchmark:connectedBenchmarkAndroidTest has an issue while installing the apk for the correct ABI. Is that something we've observed before?

I was able to resolve this by adding `autosignReleaseWithDebugKey` to my local.properties. Let me know if that helps!

MatthewTighe:
> So, macrobenchmark is mostly used for something that happens once in a while (I.e start up). So this test shouldn't include that in my opinion since we want a clear view of start up. However, we do have microbenchmarks that do just that! I have a patch ready to land that I never landed since there was issues with dependencies a few months ago that seems resolved now.

Oh cool! I would be interested in reading through that patch as well.

@rahulsainani another thing to consider for follow-ups:

IIRC, we have historically treated time-to-interactivity as a more valuable metric than time-to-display since that captures a more realistic picture of a user experience. I believe this patch only measures the latter, so it would be great to either add that metric to these existing tests or create new ones to measure that separately. 

I would also be interested in understanding better the results between the different flows. These results show a pretty massive difference between hot and cold startup, though I don't know what kind of values to expect. My results:

```
StartupBenchmark_startupHot
timeToInitialDisplayMs   min 64.7,   median 71.6,   max 87.4
Traces: Iteration 0 1 2 3 4
StartupBenchmark_startupCold
timeToInitialDisplayMs   min 755.1,   median 812.1,   max 850.2
Traces: Iteration 0 1 2 3 4
StartupBenchmark_startupWarm
timeToInitialDisplayMs   min 161.2,   median 188.1,   max 200.6
Traces: Iteration 0 1 2 3 4
```

rahulsainani:
@MatthewTighe Good to see it working in your local env 👌 

Yes, you're right that this PR only adds Time to initial display (TTID). There's still value in measure TTID as lots happen till that time and it can give a good signal about app performance. I'd like to add tests to get time-to-interactivity aka Time to Full Display (TTFD). For that `Activity.reportFullyDrawn()` is used by the benchmark as a signal. However, for this PR, the goal is to try macrobenchmarks and explore the benefits and where it fits in our perf metrics observability setup and how this is a prerequisite to add baseline profiles. [Ref](https://developer.android.com/topic/performance/benchmarking/macrobenchmark-metrics#startup-timing)

We can also benchmark other macro flows like Login/Changing language and capture `TraceMetrics` for those, or `FrameMetrics` for scroll performance. But I'll stop myself from going into tangents here 😄 

The numbers look consistent to me comparing to [Google Play Vitals](https://play.google.com/console/u/0/developers/7083182635971239206/app/4972519468758466290/vitals/metrics/details?days=28&metric=COLD_STARTUP_TIME&device=google%2Fblueline). 

[Cold Startup](https://developer.android.com/topic/performance/vitals/launch-time#cold) is what we're interested in, when the app is not in memory. [Hot](https://developer.android.com/topic/performance/vitals/launch-time#hot) is when the app is in memory but the activity is in background, all it does it bring it to the foreground.

fenix/benchmark/build.gradle
rahulsainani: There was a build error because of glean dependency resolution conflict as it was coming from geckoview as well. The glean plugin applied on `app` module handles this case, along with other functionality. Since `benchmark` module doesn't need all the glean gradle plugin's functionalities, we added this to fix the dependency resolution conflict issue. Does that help understand better? 😅 This was an issue that took some time to figure out. Jan Erik was a big here here!

fenix/benchmark/build.gradle
rahulsainani: 1. Good question, on my Samsung A50 (android 11) the metrics can't be read/extracted when this is set to false. On Pixel 5 & 7 (android 13) it works regardless of this setting, so I set it true. The samples seem to do the [same](https://github.com/android/performance-samples/blob/main/MacrobenchmarkSample/macrobenchmark/build.gradle.kts#L40). 
2. This is the `benchmark/build.gradle`, the closure is defined in `app/build.gradle`, so that won't work right? Although i'm not sure if this condition is even required. This is something I'd like to get feedback on, based on how we sign apks on CI and how we would like to sign the benchmark apk as well.

rahulsainani:
> I was able to resolve this by adding `autosignReleaseWithDebugKey` to my local.properties. Let me know if that helps!

Somehow that's not working on my machine, the local.properties flag is needed to run from AS also, so I had it enabled. But it's good to know that it works for you 😄

MatthewTighe (approved):
This looks good from my perspective, so adding an approval in case it unblocks you. It does sound like maybe you were looking for [additional feedback](https://github.com/mozilla-mobile/firefox-android/pull/842#discussion_r1115471022) about our CI systems which I am not confident in answering, but it seems like those may be resolved in follow-ups.

rahulsainani:
> This looks good from my perspective, so adding an approval in case it unblocks you. It does sound like maybe you were looking for [additional feedback](https://github.com/mozilla-mobile/firefox-android/pull/842#discussion_r1115471022) about our CI systems which I am not confident in answering, but it seems like those may be resolved in follow-ups.

Thanks @MatthewTighe  for taking the time to review and dig into macrobenchmarks 🙌 
Yes, I'll wait for some more feedback and still need the solution to the signing issue on the build. 🚧

jonalmeida:
For this unrelated CI failure, please rebase to the latest `main` to include the fix in your branch:

```
[task 2023-03-01T17:36:01.392Z] FAILURE: Build failed with an exception.
[task 2023-03-01T17:36:01.392Z] 
[task 2023-03-01T17:36:01.392Z] * What went wrong:
[task 2023-03-01T17:36:01.392Z] Execution failed for task ':support-rusthttp:generateDebugUnitTestStubRFile'.
[task 2023-03-01T17:36:01.392Z] > Could not resolve all files for configuration ':support-rusthttp:debugUnitTestRuntimeClasspath'.
[task 2023-03-01T17:36:01.392Z]    > Failed to transform monitor-1.4.0.aar (androidx.test:monitor:1.4.0) to match attributes {artifactType=android-symbol-with-package-name, org.gradle.category=library, org.gradle.libraryelements=jar, org.gradle.status=release, org.gradle.usage=java-runtime}.
[task 2023-03-01T17:36:01.392Z]       > Could not find monitor-1.4.0.jar (androidx.test:monitor:1.4.0).
[task 2023-03-01T17:36:01.392Z]         Searched in the following locations:
[task 2023-03-01T17:36:01.393Z]             file:/builds/worker/fetches/external-gradle-dependencies/google/androidx/test/monitor/1.4.0/monitor-1.4.0.aar
[task 2023-03-01T17:36:01.393Z]             file:/builds/worker/fetches/external-gradle-dependencies/google/androidx/test/monitor/1.4.0/monitor-1.4.0.jar
[task 2023-03-01T17:36:01.394Z]    > Failed to transform monitor-1.4.0.aar (androidx.test:monitor:1.4.0) to match attributes {artifactType=android-symbol-with-package-name, org.gradle.status=release}.
[task 2023-03-01T17:36:01.394Z]       > Could not find monitor-1.4.0.aar (androidx.test:monitor:1.4.0).
[task 2023-03-01T17:36:01.394Z]         Searched in the following locations:
[task 2023-03-01T17:36:01.394Z]             file:/builds/worker/fetches/external-gradle-dependencies/google/androidx/test/monitor/1.4.0/monitor-1.4.0.aar
[task 2023-03-01T17:36:01.394Z]             file:/builds/worker/fetches/external-gradle-dependencies/google/androidx/test/monitor/1.4.0/monitor-1.4.0.jar
[task 2023-03-01T17:36:01.394Z] 
[task 2023-03-01T17:36:01.394Z] * Try:
[task 2023-03-01T17:36:01.394Z] > Run with --stacktrace option to get the stack trace.
[task 2023-03-01T17:36:01.394Z] > Run with --info or --debug option to get more log output.
[task 2023-03-01T17:36:01.394Z] > Run with --scan to get full insights.
```

rahulsainani:
Thanks for looking into it and fixing it @jonalmeida 🙌

jonalmeida (approved):
LGTM!

Left one nit about a doc, but that's not blocking being a nit. 🚀

fenix/benchmark/build.gradle
```diff
@@ -0,0 +1,71 @@
...
+    defaultConfig {
+        minSdk 23
+        targetSdk Config.targetSdkVersion
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    buildTypes {
+        // This benchmark buildType is used for benchmarking, and should function like your
+        // release build (for example, with minification on). It's signed with a debug key
+        // for easy local/CI testing.
```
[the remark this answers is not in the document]
rahulsainani: Yes the idea would be to run it on CI builds, with the next steps. 
Just mentioning, this `build.gradle` is for the `benchmark` module that generates it's own apk that performs the test on the fenix apk. Not sure if taskcluster already have a signing key for that? I'll remove it for now and we can add this again as/if we run it on CI 👍

JohanLorenzo:
Chatted with @rahulsainani. At the moment, the Mergify merge queue doesn't work anymore because of https://github.com/Mergifyio/mergify/issues/5075. I'm merging this PR manually while we figure out a way to put the merge queue back in a working state.

JohanLorenzo:
Permanent fix will be tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1822491

PR #848, opened 2023-02-16
Bug 1814116 - Add new context menu option `Open in Background Tab`

gitstart:
`Homescreen` Shortcuts need to add a one more `context menu` option -> `Open in Background Tab`.

- HomeScreen

### Issue Screenshots
[fnxoss-72-issue.webm](https://user-images.githubusercontent.com/20729878/217828334-aa23b3d4-7541-4de2-ad23-d2443d50662b.webm)

### Fix Screenshots
[fnx-72-fix.webm](https://user-images.githubusercontent.com/20729878/218725857-48da6ed5-60e0-470e-87e4-f01a1a23ab5a.webm)

Pull Request checklist
 - [x] Tests: This PR includes thorough tests or an explanation of why it does not
 - [x] Screenshots: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
 - [x] Accessibility: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.
 
**QA**
- [x] QA Needed

**GitHub Automation**
[Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1812144](https://bugzilla.mozilla.org/show_bug.cgi?id=1814116)
https://bugzilla.mozilla.org/show_bug.cgi?id=1814116
https://bugzilla.mozilla.org/show_bug.cgi?id=1814116
https://bugzilla.mozilla.org/show_bug.cgi?id=1814116
https://bugzilla.mozilla.org/show_bug.cgi?id=1814116
https://bugzilla.mozilla.org/show_bug.cgi?id=1814116
https://bugzilla.mozilla.org/show_bug.cgi?id=1814116


fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -323,6 +328,49 @@ class DefaultSessionControlController(
...
+        val tabId = addTabUseCase.invoke(
+            url = appendSearchAttributionToUrlIfNeeded(topSite.url),
+            selectTab = false,
+            startLoading = true,
+        )
+
+        if (settings.openNextTabInDesktopMode) {
+            activity.handleRequestDesktopMode(tabId)
+        }
+        activity.openToBrowser(BrowserDirection.FromTabsTray)
```
Mugurell, line 371:
```suggestion
        activity.openToBrowser(BrowserDirection.FromHome)
```

Even if the tab was not immediately selected it was still opened from the homescreen.

fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt
```diff
@@ -323,6 +328,49 @@ class DefaultSessionControlController(
         }
     }
 
+    override fun handleOpenInBackgroundTabClicked(topSite: TopSite, position: Int) {
```
Mugurell, line 331: This seems to duplicate the code from [handleSelectTopSite](https://searchfox.org/mozilla-mobile/rev/2d9dd7b862d87ac539a52b9b0d99160b61745dae/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlController.kt#391-432) with the only difference being the `selectTab = false` parameter of `addTabUseCase`.
If the goal is to keep the two methods in sync then to prevent future refactorings only modifying one behavior we should look into reusing the exesting code. 
I propose adding a new parameter to the `handleSelectTopSite` methods from the Controller, maybe `selectTab` with the two methods from the Interaction calling this with either `true` or `false`.

fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/SessionControlInteractor.kt
```diff
@@ -182,6 +182,14 @@ interface CustomizeHomeIteractor {
  * Interface for top site related actions in the [SessionControlInteractor].
  */
 interface TopSiteInteractor {
+
+    /**
+     * Opens the given top site in a new tab. Called when an user clicks on a top site.
+     *
+     * @param topSite The top site that will be open in a new tab.
```
Mugurell, line 189: The method documentation should reference both parameters.

fenix/app/src/main/java/org/mozilla/fenix/home/topsites/TopSiteItemViewHolder.kt
```diff
@@ -53,49 +53,6 @@ class TopSiteItemViewHolder(
     private val binding = TopSiteItemBinding.bind(view)
 
     init {
-        itemView.setOnLongClickListener {
```
Mugurell, line 56: There is a small advantage is cleaning up the code but this change is not needed to solve the current issue.
You could add this refactoring in a separate commit but I would favor not modifying the code to keep the git history clean and ease the blame process later.

fenix/app/src/main/java/org/mozilla/fenix/home/topsites/TopSiteItemViewHolder.kt
```diff
@@ -132,11 +89,65 @@ class TopSiteItemViewHolder(
         }
     }
 
+    /**
+     * Binds a top site to the view. This will set the title, url, and image.
+     * setOnClickListener and setOnLongClickListener are also set here.
```
Mugurell, line 94: nit: With `setOnClickListener` and `setOnLongClickListener` being methods they are not _set_ here.
This method does set a `OnClickListener`, `OnTouchListener` and `OnLongClickListener` but these details are not needed to be made public. Same as with setting the `title`, `url` and `image`.

Looking through the current state of the codebase it looks like we just avoided adding documentation for the `bind` method or went with something general like `Bind the data for displaying a particular top site and respond to user interactions`.

fenix/app/src/main/java/org/mozilla/fenix/home/topsites/TopSiteItemViewHolder.kt
```diff
@@ -132,11 +89,65 @@ class TopSiteItemViewHolder(
...
+     * @param topSite The top site to bind.
```
Mugurell, line 95: nit: We usually add a newline between the method description and the list of parameters.

fenix/app/src/main/res/values/strings.xml
```diff
@@ -1966,6 +1966,8 @@
     <string name="a11y_action_label_wallpaper_collection_learn_more">open link to learn more about this collection</string>
     <!-- Action label for links that point to an article. Talkback will append this to say "Double tap to read the article". -->
     <string name="a11y_action_label_read_article">read the article</string>
+    <!-- Action label for Context Menu in HomeScreen "Open in background tab". -->
+    <string name="open_in_backgroud_tab">Open in background tab</string>
```
Mugurell, line 1970: Making a note to check this value with UX.

Mugurell (changes requested):
Thank you for your contribution!
I would suggest not refactoring the `init` block from `TopSiteItemViewHolder` but otherwise the changes looks good!

Since the `Top Sites` also exist on other platforms which do not have this new menu option we'll have to check with UX and product the opportunity of this change in Fenix. Will keep you posted with the discussions result.

gitstart:
Thank you @Mugurell for reviewing the PR. 🙏

lime124:
Hi! Thanks so much for thinking about how to make our context menu more useful. 

The intent of the homepage is to be a central place people can use to jump off to quickly resume unfinished tasks, open frequently used sites, or discover new content. 	If someone opens a top site, they can easily return to the homepage and open a different item on that page. The intent on a search result page is different as someone peruses results and may want to open several in the background while remaining on the result page to continue finding useful webpages. 

The open tab in background is a really useful context menu item for a search results page but given the nature of the homepage has less benefit on that surface. We also want to avoid placing too many actions in the context menu which makes it harder for people to use. 

It looks like open in background is already in context menus for websites so I recommend we close this PR. Thanks!

PR #849, opened 2023-02-16
Bug 1796348,  Bug 1811795 - Target Android 13 in Fenix and AC

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1811795
https://bugzilla.mozilla.org/show_bug.cgi?id=1796348


rocketsroger:
Adding a "do not land" label so we can have a plan before landing this.  Crash reporter depends on the crash notification to report crashes.  If we update the API to 13 too early, we might lose a lot of crash reports on Android 13 devices.

There are two things we can do:
1. Make notification pre permission prompt on by default (Currently it's in v110 as an experiment)
2. Add a "Looks like Firefox crashes" prompt at start if there are unreported crashes to prompt the user to report their crashes.

I think the minimum here is to do step 1 first before landing this change.  Thanks

mcarare:
> Adding a "do not land" label so we can have a plan before landing this. Crash reporter depends on the crash notification to report crashes. If we update the API to 13 too early, we might lose a lot of crash reports on Android 13 devices.
> 

Why would we lose crash reports on Android 13 devices? This patch introduces no changes in the behavior. We actually make sure that every time we try to show a notification from anywhere in the app we ask for the notifications permission. We do not have control over that in the current state.  And we do not have callbacks for the result of our request made with `ensureMarketingChannelExists` exists calls. I do not think we should postpone this any longer for changes that can be introduced after targeting Android 13 lands. 

If anything, this allows showing any kind of prompt/ dialog/before asking the notifications permission from anywhere in the app, using the request permission method in the delegate. 

Also, for crash reports, we can have the alternative of showing a dialog, if the notification permission is denied. Because the logic can be handled within the delegate. So we can prompt for crashes we are not able to send now for Android 13 devices

rocketsroger:
> 



> > Adding a "do not land" label so we can have a plan before landing this. Crash reporter depends on the crash notification to report crashes. If we update the API to 13 too early, we might lose a lot of crash reports on Android 13 devices.
> 
> IMO this patch cannot cause losing crash reports on Android 13 devices. This patch introduces no changes in the behavior. We actually make sure that every time we try to show a notification from anywhere in the app we ask for the notifications permission. We do not have control over that in the current state. And we do not have callbacks for the result of our request made with `ensureMarketingChannelExists` exists calls. I do not think we should postpone this any longer for changes that can be introduced after targeting Android 13 lands.

Yes, but we can only ask once.  Currently there is an experiment of using a dialog showing why it's in the user's advantage to give us the notification permission and then ask for the permission.  This is a better way to ensure we have a higher opt-in for notification permission.

> If anything, this allows showing any kind of prompt/ dialog/before asking the notifications permission from anywhere in the app, using the request permission method in the delegate.
> 
> Also, for crash reports, we can have the alternative of showing a dialog, if the notification permission is denied. Because the logic can be handled within the delegate. So we can prompt for crashes we are not able to send now for Android 13 devices

This doesn't work consistently so we had to move to a notification.  Showing dialog when Fenix is not in foreground doesn't work consistently on newer OS versions.

mcarare:
> Yes, but we can only ask once. Currently there is an experiment of using a dialog showing why it's in the user's advantage to give us the notification permission and then ask for the permission. This is a better way to ensure we have a higher opt-in for notification permission.
> 

I agree. But this patch does not prevent us from using that dialog. Also, with this patch, you can ask for permission twice.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/HomeNotificationPermissionDialogFragment.kt
```diff
@@ -49,6 +53,7 @@ class HomeNotificationPermissionDialogFragment : DialogFragment() {
                         onDismiss = ::onDismiss,
                         grantNotificationPermission = {
                             ensureMarketingChannelExists(context.applicationContext)
+                            requireComponents.notificationsDelegate.requestNotificationPermission()
```
mcarare, line 56: Depending on the experiment results we could now better handle a potential rejection and show an additional message (even by just changing the text) before we ask for the permission a second time.

pwd-github:
The notification conversation is obviously scope creep. 

The thing is, most people click yes on everything. The ones that click no, need clear instructions as to how to turn notifications back on. If the reason is to submit crashes, that's a terrible reason.

But none of that has anything to do with this patch, especially given that all competitor browsers are targeting Android 13 in their release versions.

rocketsroger:
I've confirmed that we do have pre-permission notification on by default starting v112.  This is one of the requirement to unblock this PR.  

I am confirming with data-science to have the result on the pre-permission prompt experiment to make sure that the result is positive before we land this.

android-components/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -59,6 +62,8 @@ class FetchDownloadManager<T : AbstractFetchDownloadService>(
     @VisibleForTesting
     internal fun getSDKVersion() = SDK_INT
 
+    internal fun getTargetSDKVersion() = applicationContext.applicationInfo.targetSdkVersion
```
rahulsainani, line 65: Think we can remove this as we will target 33 and above and this will always be true

android-components/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
```diff
@@ -71,6 +76,13 @@ class FetchDownloadManager<T : AbstractFetchDownloadService>(
         }
         validatePermissionGranted(applicationContext)
 
+        if (SDK_INT >= Build.VERSION_CODES.TIRAMISU &&
+            getTargetSDKVersion() >= Build.VERSION_CODES.TIRAMISU &&
+            !NotificationManagerCompat.from(applicationContext).areNotificationsEnabled()
```
rahulsainani, line 81: ⚠️ This could throw a `RemoteException`.  

As it's already being checked in `notificationsDelegate.requestNotificationPermission()` maybe we can call that already without the conditions.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/HomeNotificationPermissionDialogFragment.kt
```diff
@@ -47,6 +51,7 @@ class HomeNotificationPermissionDialogFragment : DialogFragment() {
                     onDismiss = ::onDismiss,
                     grantNotificationPermission = {
                         ensureMarketingChannelExists(context.applicationContext)
```
rahulsainani, line 53: Do we need this now? cc @rocketsroger

android-components/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
mcarare: Done.

android-components/components/feature/downloads/src/main/java/mozilla/components/feature/downloads/manager/FetchDownloadManager.kt
mcarare: Removed the additional check.

rahulsainani (approved):
Works well! Thanks for persevering and working through all the different feedback cycles 🚀

rocketsroger (approved):
Removed my objection and the `do not land` label.  🚢

android-components/components/support/base/src/main/java/mozilla/components/support/base/android/NotificationsDelegate.kt
```diff
@@ -0,0 +1,174 @@
...
+ * @param notificationManagerCompat a reference to [NotificationManagerCompat].
+ * @property activity The [AppCompatActivity] this handler is bound to.
+ * @property onPermissionGranted optional callback for handling permission acceptance.
+ * @property onPermissionRejected optional callback for handling permission refusal.
+ */
+class NotificationsDelegate(
+    val notificationManagerCompat: NotificationManagerCompat,
+) {
+    private val unboundHandlerException =
+        UnboundHandlerException("You must bind the NotificationPermissionHandler to an activity")
```
jonalmeida, line 39: Why hold an instance to the exception here? 🤔

android-components/components/support/base/src/main/java/mozilla/components/support/base/android/NotificationsDelegate.kt
```diff
@@ -0,0 +1,174 @@
...
+/**
+ * Exception thrown when a [NotificationsDelegate] is not bound to any [AppCompatActivity]
+ */
+class UnboundHandlerException(message: String) : Exception(message)
+
+/**
+ * Handles showing notifications and asking permission, if needed.
+ *
+ * @param notificationManagerCompat a reference to [NotificationManagerCompat].
+ * @property activity The [AppCompatActivity] this handler is bound to.
```
jonalmeida, line 31: nit: doc for non-existent property.

android-components/components/support/base/src/main/java/mozilla/components/support/base/android/NotificationsDelegate.kt
```diff
@@ -0,0 +1,174 @@
...
+ * @property onPermissionRejected optional callback for handling permission refusal.
+ */
+class NotificationsDelegate(
+    val notificationManagerCompat: NotificationManagerCompat,
+) {
+    private val unboundHandlerException =
+        UnboundHandlerException("You must bind the NotificationPermissionHandler to an activity")
+    private var onPermissionGranted: OnPermissionGranted = { }
+    private var onPermissionRejected: OnPermissionRejected = { }
+    private val notificationPermissionHandler: MutableMap<AppCompatActivity, ActivityResultLauncher<String>> =
```
jonalmeida, line 42: Discussion only: for our apps right now, we hold a singleton reference to a NotificationDelegate so we it's a bit harder to leak here. What do you think about wrapping this in a weakref (as a follow-up)?

```suggestion
    private val notificationPermissionHandler: MutableMap<WeakReference<AppCompatActivity>, ActivityResultLauncher<String>> =
```

android-components/components/support/base/src/main/java/mozilla/components/support/base/android/NotificationsDelegate.kt
```diff
@@ -0,0 +1,174 @@
...
+            }
+
+        notificationPermissionHandler[activity] = activityResultLauncher
+    }
+
+    /**
+     * Removes activity reference from the [NotificationsDelegate]
+     */
+    @Suppress("unused")
+    fun unBindActivity(activity: AppCompatActivity) {
```
jonalmeida, line 66: Since we call this in `onDestroy`, I'm more inclined to having a weak ref wrapper since we know that `onDestroy` isn't always called.

android-components/components/tooling/lint/src/main/java/mozilla/components/tooling/lint/NotificationManagerChecks.kt
```diff
@@ -0,0 +1,84 @@
...
+    "android.app.NotificationManager"
+
+internal const val NOTIFY_ERROR_MESSAGE = "Using Android NOTIFY instead of base component"
+
+/**
+ * Custom lint that ensures [NotificationManagerCompat] and [NotificationManager]'s method [notify]
+ * is not called directly from code.
+ * Calling notify directly from code eludes the checks implemented in [NotificationsDelegate]
+ */
+class NotificationManagerChecks : Detector(), Detector.UastScanner {
```
jonalmeida, line 31: Love this! 🙌

android-components/components/tooling/lint/src/main/java/mozilla/components/tooling/lint/NotificationManagerChecks.kt
```diff
@@ -0,0 +1,84 @@
...
+            }
+
+            if (inAppPackage) {
+                context.report(
+                    ISSUE_NOTIFICATION_USAGE,
+                    node,
+                    context.getLocation(node),
+                    NOTIFY_ERROR_MESSAGE,
+                )
+            }
```
jonalmeida, line 66: Curious, why are these two checks separate? Wouldn't it be the same to do:

```kotlin
if (inComponentPackage || inAppPackage) {
    context.report(
        ISSUE_NOTIFICATION_USAGE,
        node,
        context.getLocation(node),
        NOTIFY_ERROR_MESSAGE,
    )
}
```

jonalmeida (approved):
Excellent work - this PR was much easier to read than I expected!

I've only reviewed a few core components in this change since it has already been reviewed by two others. All my comments can be addressed after this lands so it's non-blocking.

:shipit:

android-components/components/support/base/src/main/java/mozilla/components/support/base/android/NotificationsDelegate.kt
mcarare: It was left from a previous iteration, but I will change it to create an instance only when thrown, as it is a single place where we do that.

android-components/components/tooling/lint/src/main/java/mozilla/components/tooling/lint/NotificationManagerChecks.kt
mcarare: I encountered an issue when having it in the single if (or) statment would miss some valid lint issues. I am going to leave it like this and check-up on the issue as a follow-up.

PR #850, opened 2023-02-16
Bug 1792796 - Force Left to Right layout direction for Onboarding whe…

HarrisonOg:
…n running on Android API level less than 23

[Bugzilla ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1792796)

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


MozillaNoah:
CC: @mavduevskiy

jonalmeida:
Unassigning after speaking with Harrison offline - this was meant for @mavduevskiy to review.

mavduevskiy:
@jonalmeida 
Trying to sum up all the context I could get:
1. There is a bug that appeared after we enabled onboarding on fenix (and it is also present in Focus, and is also related to a similar onboarding);
2. It is present on 5.1 devices with two languages (with just 10 cases in English, which accounts for 0.03% cases);
3. Google eng said they didn't change any code and [was confused](https://issuetracker.google.com/issues/266059178#comment5) why we would get two different arrays in `placeRightOrBottom`. 

I failed to reproduce the bug as well. What's more, the onboarding screen didn't not trigger `placeRightOrBottom` for LtR languages for me, as everything is positioned to the center, except for the close button – which is positioned to the left, not right. I am trying to guess what is the element that would be positioned to the right on the onboarding screen in LtR mode, and can not. Also, the fact that the in and out array are different inside `placeRightOrBottom` also doesn't make sense to me, as they are created of the same array of `Measurable`. Something is funny.

It looks like the majority of the crashes come from <108 version, more recent versions account only for less than 1 percent. Wonder why.

Anyway, the patch looks good to me. It doesn't affect every user, changes only the position of the close icon, and overall is a good educated guess. 

tagging @Mugurell if there is something to add! thanks for a lot of work put into researching this!

Mugurell:
Thank you @HarrisonOg and @mavduevskiy!

Compose atm seems like a black box in regards to how it works and how small details may break it so this speculative fix even though not ideal because of disabling RTL seems okay to me given that relatively few users will be "affected" - still on Lollipop - by the side RTL side effect which is still preferable to a crash.
I remember this approach was confirmed by product so LGTM.

Would be nice to monitor the impact of this patch after landing (in release) to assess whether it helped or not. 
If there isn't a clear reduction in the number of crashes then maybe this can be backed out to keep the RTL support and accept the issue until future bigger refactorings of the onboarding process or compose updates that might improve things.

PR #852, opened 2023-02-16
Bug 1812144 - `Homescreen` appbar drag issue fixed

gitstart:
`Homescreen` Appbar is scrollable even if there is no content, from `customize homepage` all options disabled and then the appbar is scrolling if you tap on the bar and scroll up.

- HomeScreen

### Issue Screenshots
[fnx-60-issue.webm](https://user-images.githubusercontent.com/20729878/216605191-1fa098bd-8114-49f6-bc8b-9dac25bdf592.webm)

### Fix Screenshots
[fnx-60-fix.webm](https://user-images.githubusercontent.com/20729878/216927241-45a39792-ec1c-45df-a717-04160546db6a.webm)

Pull Request checklist
 - [ ] Tests: This PR includes thorough tests or an explanation of why it does not
 - [x] Screenshots: This PR includes screenshots or GIFs of the changes made or an explanation of why it does not
 - [x] Accessibility: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/master/android/accessibility_guide.md) or does not include any user facing features. In addition, it includes a screenshot of a successful [accessibility scan](https://play.google.com/store/apps/details?id=com.google.android.apps.accessibility.auditor&hl=en_US) to ensure no new defects are added to the product.
 
**QA**
- [x] QA Needed

**GitHub Automation**
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144
https://bugzilla.mozilla.org/show_bug.cgi?id=1812144


Mugurell (changes requested):
Overall it looks okay but it is not a complete fix.
It is still possible to have more than two items in the `RecyclerView` but with it's height still smaller than needed to overflow the screen height and so the issue will still be reproduceable even after this patch.


https://user-images.githubusercontent.com/11428869/221538168-f59d5860-c4d3-4daf-872b-6962ce5c866b.mp4




Based on the above scenario it seems to me that we'd need to disable `AppBarLayout`'s dragging based on whether it's plus the height of the `RecyclerView` is bigger than the available screen height.

*Maybe a bit more performant solution would be to set a flag only once on whether the homescreen should be draggable or not and use this in the `setDragCallback` of the `AppBarLayout`.

gitstart:
Thank you @Mugurell for reviewing the PR. 🙏
Will update the requested changes soon. 🚀

debojyoti452 (approved):

gitstart:
Hello @Mugurell,🙏
Pushed requested changes. 🚀

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -503,6 +505,21 @@ class HomeFragment : Fragment() {
...
+        if (binding.homeAppBar.layoutParams != null) {
+            val appBarLayoutParams = binding.homeAppBar.layoutParams as CoordinatorLayout.LayoutParams
+            val appBarBehavior = AppBarLayout.Behavior()
+            appBarBehavior.setDragCallback(
+                object : AppBarLayout.Behavior.DragCallback() {
+                    override fun canDrag(appBarLayout: AppBarLayout): Boolean {
+                        return false
+                    }
+                },
+            )
```
Mugurell, line 518: Interesting approach to disable dragging in all scenarios.

Looking through the initial tickets for the current functionality - https://github.com/mozilla-mobile/fenix/issues/99, https://github.com/mozilla-mobile/fenix/issues/137, https://github.com/mozilla-mobile/fenix/issues/7700 I don't see any discussions or indications about the scenarios in which we'd want the appBar to be draggable or not.

Given that currently the appBar can be dragged only when there aren't enough elements on the screen to overflow it and in this scenario the drag gesture results in the bug we need so solve disabling dragging seems like the sensible choice.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -503,6 +505,21 @@ class HomeFragment : Fragment() {
         }
     }
 
+    private fun setScrollingBehavior() {
```
Mugurell, line 508: Sorry for the being pedantic but the new approach that will disable dragging in all situations I think warrants being specific about the intent.
Can you please rename to something like `disableAppBarDragging()`. 
It's an important behavior change that I think we need to highlight.

Mugurell (commented):
Thank you for the update!
Just one naming comment to ensure this change will be clear as intended for future readers.

gitstart:
> Thank you for the update! Just one naming comment to ensure this change will be clear as intended for future readers.

Thank you @Mugurell for reviewing the PR.🙏
I have updated the function name as per the recommendation. ✅

Mugurell:
bors try

Mugurell:
bors retry

Mugurell:
Same ui test failing in both try runs:
`org.mozilla.fenix.ui.TabbedBrowsingTest#verifyContextMenuShortcuts` with the following stacktrace:

> java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:87)
	at org.junit.Assert.assertTrue(Assert.java:42)
	at org.junit.Assert.assertTrue(Assert.java:53)
	at org.mozilla.fenix.helpers.MatcherHelper.assertItemWithResIdExists(MatcherHelper.kt:48)
	at org.mozilla.fenix.helpers.MatcherHelper.assertItemWithResIdExists$default(MatcherHelper.kt:45)
	at org.mozilla.fenix.ui.robots.HomeScreenRobot.verifyHomeWordmark(HomeScreenRobot.kt:195)
	at org.mozilla.fenix.ui.TabbedBrowsingTest$verifyContextMenuShortcuts$11.invoke(TabbedBrowsingTest.kt:366)
	at org.mozilla.fenix.ui.TabbedBrowsingTest$verifyContextMenuShortcuts$11.invoke(TabbedBrowsingTest.kt:361)
	at org.mozilla.fenix.ui.robots.NavigationToolbarRobot$Transition.openTabFromShortcutsMenu(NavigationToolbarRobot.kt:242)
	at org.mozilla.fenix.ui.TabbedBrowsingTest.verifyContextMenuShortcuts(TabbedBrowsingTest.kt:361)

    
    
 Seems like a new issue which needs to be investigated more before attempting to merge.

gitstart:
> Same ui test failing in both try runs: `org.mozilla.fenix.ui.TabbedBrowsingTest#verifyContextMenuShortcuts` with the following stacktrace:
> 
> > java.lang.AssertionError
> > at org.junit.Assert.fail(Assert.java:87)
> > at org.junit.Assert.assertTrue(Assert.java:42)
> > at org.junit.Assert.assertTrue(Assert.java:53)
> > at org.mozilla.fenix.helpers.MatcherHelper.assertItemWithResIdExists(MatcherHelper.kt:48)
> > at org.mozilla.fenix.helpers.MatcherHelper.assertItemWithResIdExists$default(MatcherHelper.kt:45)
> > at org.mozilla.fenix.ui.robots.HomeScreenRobot.verifyHomeWordmark(HomeScreenRobot.kt:195)
> > at org.mozilla.fenix.ui.TabbedBrowsingTest$verifyContextMenuShortcuts$11.invoke(TabbedBrowsingTest.kt:366)
> > at org.mozilla.fenix.ui.TabbedBrowsingTest$verifyContextMenuShortcuts$11.invoke(TabbedBrowsingTest.kt:361)
> > at org.mozilla.fenix.ui.robots.NavigationToolbarRobot$Transition.openTabFromShortcutsMenu(NavigationToolbarRobot.kt:242)
> > at org.mozilla.fenix.ui.TabbedBrowsingTest.verifyContextMenuShortcuts(TabbedBrowsingTest.kt:361)
> 
> Seems like a new issue which needs to be investigated more before attempting to merge.

Hello @Mugurell,
Nothing was touched on tests. Do you need any help from our end? 🙏
Thank you!

Mugurell:
> Hello @Mugurell, Nothing was touched on tests. Do you need any help from our end? 🙏 Thank you!

Seems like the issue is real and is happening after these changes in this very specific scenario:
- have a private tab opened
- long press on the tab counter to see a contextual menu
- choose "New tab"

Expected result (as seen in the current applications):
- the Firefox header is seen at the top of the screen.

Actual result (after this patch):
- the Firefox logo is not seen at the top of the screen.

We need to have the Firefox logo shown in all cases where the users goes to the homescreen - with the least amount of changes.
Or find another way to solve the initial issue.

gitstart:
> > Hello @Mugurell, Nothing was touched on tests. Do you need any help from our end? 🙏 Thank you!
> 
> Seems like the issue is real and is happening after these changes in this very specific scenario:
> 
> * have a private tab opened
> * long press on the tab counter to see a contextual menu
> * choose "New tab"
> 
> Expected result (as seen in the current applications):
> 
> * the Firefox header is seen at the top of the screen.
> 
> Actual result (after this patch):
> 
> * the Firefox logo is not seen at the top of the screen, as seen in the below video:
> 
> We need to have the Firefox logo shown in all cases where the users goes to the homescreen - with the least amount of changes. Or find another way to solve the initial issue.

Okay @Mugurell, Let me check it.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -503,6 +505,22 @@ class HomeFragment : Fragment() {
...
+            appBarBehavior.setDragCallback(
+                object : AppBarLayout.Behavior.DragCallback() {
+                    override fun canDrag(appBarLayout: AppBarLayout): Boolean {
+                        return false
+                    }
+                },
+            )
+            appBarLayoutParams.behavior = appBarBehavior
+        }
+        binding.homeAppBar.setExpanded(true)
```
Mugurell, line 521: I see this change being added in the last commit.
Is this proposed as a fix for the issue seen in tests - described more here - https://github.com/mozilla-mobile/firefox-android/pull/852#issuecomment-1453784159 ?

`binding.homeAppBar.setExpanded(true)` would regress another behavior as described here - https://github.com/mozilla-mobile/firefox-android/pull/852#discussion_r1118488158

If the patch cannot be improved to avoid both issues we could check with UX to temporary approve one since the original issue seems a bit more important to fix.

Let me know what your thoughts are.

gitstart: Hello @Mugurell,
This fix is workaround with less code. If we need proper implementation, then for simple fix a lot more code would be added. As you mentioned on [#852 (Comment)](https://github.com/mozilla-mobile/firefox-android/pull/852#discussion_r1118488158) that make sure to add least changes.

Mugurell:
@mozilla-mobile/ux Need your help on deciding how to go forward with this. We have 3 scenarios:
- the original issue - https://bugzilla.mozilla.org/show_bug.cgi?id=1812144 seems more important since in that scenario the user could not easily scroll the homescreen to see again the header. They could navigate to a page/open settings and then go back to the homescreen as a long workaround.
Trying to fix that issue leaves us with two new ones:
- with this patch we would though break the scenario in which when navigating to a newly created collection from the snackbar the Firefox header would be hidden  - after this patch it would be shown OR
- with this patch we would break the scenario in which when we'd try to open a normal tab while being in private mode the Firefox header would be shown - after this patch it would be hidden.

Giving that we plan to refactor the homescreen to Compose is it acceptable to merge this patch to fix the issue from https://bugzilla.mozilla.org/show_bug.cgi?id=1812144 and temporarily get one of the last two issues or should we avoid merging and leave the fix for the Bugzilla issue to be included in the Compose refactoring?

lime124:
@Verdi can you provide guidance on this one please?

Verdi:
Ok @Mugurell and I went over the issues. The patch fixes the initial issue and makes the other two issues better by making the header visible. We should land this.

gitstart:
> Thank you all for the support and sticking through with this! Looks like we are good to go!
> 
> @gitstart Can you please
> 
> * remove the line about [collapsing the header](https://searchfox.org/mozilla-mobile/rev/9c5241a33119980c8a43339cf47124418a1adcca/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#627) when navigating to a new collection
> * update the commit / PR title to say that with the patch that scenario is fixed instead of the suggesting that the result will be have the homescreen scrollable even with no content
> * squash your commits locally to only remain with one that can be easily landed

Thank you @Mugurell, 
Requested all the changes are updated.

Mugurell (approved):
LGTM. Thank you!

Mugurell:
bors try

Mugurell:
Just a small remark that this will lead to another bug which will have the header expanded after navigating to a newly created collection - issue seen when there are more sections shown on the screen and the screen has to be scrolled to highlight the collection. This was discussed with @ Verdi and agreed that it's a smaller issue than the original one in which the private tabs button would not be seen and accepted until the Compose refactoring.

https://user-images.githubusercontent.com/11428869/223699514-44ff7dce-2ddd-4b58-b8fc-09488284285c.mp4

PR #898, opened 2023-02-20
Bug 1817708 - Using local contile service in MozillaOnline builds

rxumoz:
We Beijing office plan to start to use Contile service in MozillaOnline builds. However, there are some unique demands for us.

Beijing office need to do some promotion during shopping festivals like 11/11 every year, using contile service can help us make it easier.
The situation now is we have already put some sponsored top sites into the pinned default top sites. While we start to using Contile service, we want to make those duplicated pinned default topsites invisible. Also, for those pinned default top sites, if they are not sponsored anymore , we want them invisible too.

For these requirements, I wrote this draft PR, trying to add some information, about the pinned default top sites which should be hidden, into the JSON. But I'm not sure if it's appropriate because it modifies the interface TopSitesProvider .

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1817708

### DIFF
--- a/android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt
+++ b/android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt
@@ -15,6 +15,7 @@ import mozilla.components.feature.top.sites.facts.emitTopSitesCountFact
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.base.observer.Observable
 import mozilla.components.support.base.observer.ObserverRegistry
+import mozilla.components.support.ktx.kotlin.getRepresentativeSnippet
 import kotlin.coroutines.CoroutineContext
 
 /**
@@ -92,6 +93,7 @@ class DefaultTopSitesStorage(
         val topSites = ArrayList<TopSite>()
         val pinnedSites = pinnedSitesStorage.getPinnedSites().take(totalSites)
         var providerTopSites = emptyList<TopSite>()
+        var pinnedExclusions = emptyList<String>()
         var numSitesRequired = totalSites - pinnedSites.size
 
         if (topSitesProvider != null &&
@@ -105,6 +107,7 @@ class DefaultTopSitesStorage(
                     .filter { providerConfig.providerFilter?.invoke(it) ?: true }
                     .take(numSitesRequired)
                     .take(providerConfig.maxThreshold - pinnedSites.size)
+                pinnedExclusions = topSitesProvider.getPinnedExclusions()
                 topSites.addAll(providerTopSites)
                 numSitesRequired -= providerTopSites.size
             } catch (e: Exception) {
@@ -112,7 +115,10 @@ class DefaultTopSitesStorage(
             }
         }
 
-        topSites.addAll(pinnedSites)
+        val filteredPinnedTopSites = pinnedSites
+            .filter { !pinnedExclusions.contains(it.url.getRepresentativeSnippet()) }
+        topSites.addAll(filteredPinnedTopSites)
+        numSitesRequired = numSitesRequired + pinnedSites.size - filteredPinnedTopSites.size
 
         if (frecencyConfig?.frecencyTresholdOption != null && numSitesRequired > 0) {
             // Get 'totalSites' sites for duplicate entries with
--- a/android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/TopSitesProvider.kt
+++ b/android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/TopSitesProvider.kt
@@ -17,4 +17,11 @@ interface TopSitesProvider {
      * @return a list of top sites from the provider.
      */
     suspend fun getTopSites(allowCache: Boolean = true): List<TopSite>
+
+    /**
+     * Provides a list of hosts of top sites that should be excluded from default top sites.
+     *
+     * @return a list of hosts from the provider.
+     */
+    fun getPinnedExclusions(): List<String>
 }
--- a/android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
+++ b/android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
@@ -41,17 +41,20 @@ internal const val CACHE_TOP_SITES_KEY = "tiles"
  * @property maxCacheAgeInSeconds Maximum time (in seconds) the cache should remain valid
  * before a refresh is attempted. Defaults to -1, meaning the max age defined by the server
  * will be used.
+ * @param hasExclusions Whether or not the JSON response has a list of hosts to exclude to parse.
  */
 class ContileTopSitesProvider(
     context: Context,
     private val client: Client,
     private val endPointURL: String = CONTILE_ENDPOINT_URL,
     private val maxCacheAgeInSeconds: Long = -1,
+    val hasExclusions: Boolean = false,
 ) : TopSitesProvider {
 
     private val applicationContext = context.applicationContext
     private val logger = Logger("ContileTopSitesProvider")
     private val diskCacheLock = Any()
+    private var pinnedExclusions: List<String> = emptyList()
 
     // Current state of the cache.
     @VisibleForTesting
@@ -74,6 +77,11 @@ class ContileTopSitesProvider(
         } else {
             null
         }
+
+        if (allowCache && !isCacheExpired(shouldUseServerMaxAge = false) && hasExclusions) {
+            pinnedExclusions = readExclusionsFromDiskCache()!!
+        }
+
         if (!cachedTopSites.isNullOrEmpty()) {
             return cachedTopSites
         }
@@ -85,6 +93,8 @@ class ContileTopSitesProvider(
             throw e
         }
     }
+
+    override fun getPinnedExclusions(): List<String> = pinnedExclusions
 
     /**
      * Refreshes the cache with the latest top sites response from [endPointURL]
@@ -108,6 +118,10 @@ class ContileTopSitesProvider(
                     cacheState = cacheState.invalidate()
                     getCacheFile().delete()
                     return listOf()
+                }
+
+                if (hasExclusions) {
+                    pinnedExclusions = JSONObject(responseBody).getPinnedExclusions()
                 }
 
                 return try {
@@ -143,6 +157,14 @@ class ContileTopSitesProvider(
                 JSONObject(it).let { cachedObject ->
                     CachedData(cachedObject.validFor, cachedObject.getTopSites())
                 }
+            }
+        }
+    }
+
+    private fun readExclusionsFromDiskCache(): List<String>? {
+        synchronized(diskCacheLock) {
+            return getCacheFile().readAndDeserialize {
+                JSONObject(it).getPinnedExclusions()
             }
         }
     }
@@ -285,6 +307,12 @@ internal fun JSONObject.getTopSites(): List<TopSite.Provided> =
     getJSONArray(CACHE_TOP_SITES_KEY)
         .asSequence { i -> getJSONObject(i) }
         .mapNotNull { it.toTopSite() }
+        .toList()
+
+internal fun JSONObject.getPinnedExclusions(): List<String> =
+    getJSONArray("exclusions")
+        .asSequence { i -> getJSONObject(i) }
+        .mapNotNull { it.getString("host") }
         .toList()
 
 private fun JSONObject.toTopSite(): TopSite.Provided? {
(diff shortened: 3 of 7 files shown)


rxumoz:
Anyone can have a look at it and give me some advice? @gabrielluong @Alexandru2909 
Sorry I'm not authorized to tag this as `needs review`

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -44,14 +45,17 @@ internal const val CACHE_TOP_SITES_KEY = "tiles"
  */
 class ContileTopSitesProvider(
     context: Context,
+    val isMozillaOnline: Boolean = false,
```
rocketsroger, line 48: ❓ should we be passing this value around?  Since we can get this through `Config.channel.isMozillaOnline`?

rxumoz: I'm not sure if we can get `Config.channel.isMozillaOnline` in **android-components** ? I've only used that in **fenix** itself.

rocketsroger: I see, that makes sense.  Thanks

android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/TopSitesProvider.kt
```diff
@@ -17,4 +17,6 @@ interface TopSitesProvider {
      * @return a list of top sites from the provider.
      */
     suspend fun getTopSites(allowCache: Boolean = true): List<TopSite>
+
+    fun getPinnedExclusions(): List<String>
```
gabrielluong, line 21: Could you add a kDoc for this function?

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
gabrielluong, line 48: Since this is in Android Components and intended to be a generic component, I would suggest renaming this to not signal anything related to our internal Mozilla use cases. Something like `hasExclusions` with the kDoc description indicating the JSON response has a list of hosts to exclude to parse.

There's also another way we could approach this if we know this list of hosts are fairly static and could just live within Fenix. We could just add a lambda to filter the list of hosts in https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/TopSitesConfig.kt#36. This would simplify everything if this is an okay approach for your use case.

However, if the list of hosts are fairly dynamic and is preferred to be fetch through this JSON response, let me know.

gabrielluong, line 48: Could we also reorder the list of parameter here by moving this new parameter to the bottom of the list? We should order non-optional parameters before optional parameters.

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -88,6 +98,8 @@ class ContileTopSitesProvider(
         }
     }
 
+    override fun getPinnedExclusions(): List<String> = pinnedExclusions!!
```
gabrielluong, line 101: Does this need the `!!` since it's already defined as an empty list on init?

android-components/components/service/contile/src/test/java/mozilla/components/service/contile/ContileTopSitesProviderTest.kt
```diff
@@ -44,7 +44,7 @@ class ContileTopSitesProviderTest {
     fun `GIVEN a successful status response WHEN top sites are fetched THEN response should contain top sites`() =
         runTest {
             val client = prepareClient()
-            val provider = ContileTopSitesProvider(testContext, client)
+            val provider = ContileTopSitesProvider(testContext, false, client)
```
gabrielluong, line 47: We could avoid adding the `false` here once we order the list of parameters so that non-optional parameters appear before optional ones.

android-components/components/service/contile/src/test/java/mozilla/components/service/contile/ContileTopSitesProviderTest.kt
```diff
@@ -70,15 +70,15 @@ class ContileTopSitesProviderTest {
     fun `GIVEN a 500 status response WHEN top sites are fetched THEN throw an exception`() =
         runTest {
             val client = prepareClient(status = 500)
-            val provider = ContileTopSitesProvider(testContext, client)
+            val provider = ContileTopSitesProvider(testContext, false, client)
```
gabrielluong, line 73: I would like to see a test where we do fetch a JSON with this "exclusion" list and ensure we are filtering correctly.

gabrielluong (changes requested):

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
rxumoz: The default top sites we put are the ones that we cooperate relatively stable and long lasting, while the partners of the special collaboration during shopping festival might be some of them, which may cause duplicate. But the special collaboration partners might change from some to others from time to time, thus I think we need a dynamic exclusions to hide different default top sites every shopping fes. It won't be frequent though, about twice a year. @gabrielluong

gabrielluong: Okay, could you make the suggested changes and we could do another review on the changes

android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt
```diff
@@ -105,14 +107,17 @@ class DefaultTopSitesStorage(
                     .filter { providerConfig.providerFilter?.invoke(it) ?: true }
```
gabrielluong, line 107: We are only filtering the top sites provded by Contile. So, we should just filter here instead of filtering the providerTopSites + pinnedSites below.

android-components/components/service/contile/src/main/java/mozilla/components/service/contile/ContileTopSitesProvider.kt
```diff
@@ -74,6 +77,10 @@ class ContileTopSitesProvider(
         } else {
             null
         }
+        if (allowCache && !isCacheExpired(shouldUseServerMaxAge = false) && hasExclusions) {
```
gabrielluong, line 80: Let's add a new line above here.

android-components/components/service/contile/src/test/resources/contile/contile-hasExclusions.json
```diff
@@ -0,0 +1,9 @@
+{
+  "tiles": [
```
gabrielluong, line 2: I think we should add some sample tiles here since we are expecting the exclusion to filter filter the tiles from this list.

android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt
rxumoz: Yes, we would like to filter the default or pinned top sites. This is because we add some default top sites, let's say [A, B, C]; while during shopping fes, we may still need B but with a different url (for example a diffrenet reference code or something else), let's call it B' . So we put B' into the contile provider. At the same time we don't wanna users see two different Bs in their top sites, so we want to hide the original default top site B, to eventually show users [B', A, C]. 
For users who don't have contile service on, they'll just keep [A, B, C] without showing B' or hiding B. 
The reason we need this more complicated method to deal with these top sites is that we used to add our collaborations all into default top sites in MozillaOnline builds before contile service available.

android-components/components/service/contile/src/test/resources/contile/contile-hasExclusions.json
rxumoz: Like I said above, what we actually wanna filter is in default top sites. This is why I add another test in [DefaultTopSitesStorageTest.kt](https://github.com/mozilla-mobile/firefox-android/pull/898/files#diff-6fab8584751338c2c0e0e23f17cd92baaa44bd33d92dc6fe6f83486b2a4ad1ba)

android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt
gabrielluong: @rxumoz One more clarifying question from me as I continue to think more about the this. Is MozillaOnline also using the same service that the Contile team built? For example, I am wondering if the existing caching and fallback strategy that the Contile team uses are also relevant for MozillaOnline. Currently, we are piggybacking on the existing `ContileTopSitesProvider` implementation which was designed for the Contile sponsored tiles. However, if one of the services changes in the future because they weren't using the same implementation and becomes incompatible for eachother, it does raise a question whether or not we should introduce a separate `MozillaOnlineTopSitesProvider` and `MozillaOnlineTopSitesStorage` that your team could fine tune for your own use case if the services are indeed different. We would duplicate the class logic that makes sense and modify it to your own needs.

android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt
```diff
@@ -105,14 +107,17 @@ class DefaultTopSitesStorage(
...
                 numSitesRequired -= providerTopSites.size
             } catch (e: Exception) {
                 logger.error("Failed to fetch top sites from provider", e)
             }
         }
 
-        topSites.addAll(pinnedSites)
+        topSites.addAll(
+            pinnedSites.filter { !pinnedExclusions.contains(it.url.getRepresentativeSnippet()) },
+        )
```
gabrielluong, line 120: Consider the thoughts in https://github.com/mozilla-mobile/firefox-android/pull/898/files#r1203355929 first, but one of the bugs I see here is that we are potentially no longer displaying the right number of top sites because we are filtering after subtracting the `numSitesRequired` above [L97](https://github.com/mozilla-mobile/firefox-android/pull/898/files#diff-81df006dcf09133e7186c24352b59822b7b76bf49efcaf4d36ded81cc181924eR97). 

For example, if we specified that we want to display 8 top sites, we might end up only displaying 7 if one of them got filtered out and we did not adjust the count of `numSitesRequired`.

android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt
rxumoz: Our main aim is to show some sponsored top sites defined by server in MozillaOnline builds, which I believe is the same as contiles. As for caching and fallback, we do not have any special requirement. From this perspective I think current `ContileTopSitesProvider` is good for us.
For exclusions, due to historical reasons (we put sponsored sites into default top sites), we have to deal with old default top sites to avoid duplication. I chose to use the same server to do this for agile control, but it doesn't have to be this way if there is any better method. But no matter what method we use, I think we still need to handle it within `DefaultTopSitesStorage` since this is the place we used to put sponsored sites in.

gabrielluong: @rxumoz I am just checking with the contile team to confirm the server is the same as theirs. I don't really have any concerns if they come back to confirm that everything is pretty much the same as their team. My remaining ask then would be to fix the bug pointed out in https://github.com/mozilla-mobile/firefox-android/pull/898#discussion_r1203370159 so that you will get the right number of top sites after filtering.

rxumoz: I think currently our server does not have any config for this, just return a JSON, that's all. BTW I change the endpoint url to https://download-ssl.firefox.com.cn/fenix-contile/d3.json  , since the original one (https://download-ssl.firefox.com.cn/test20220920/d3.json) is for test.

android-components/components/feature/top-sites/src/main/java/mozilla/components/feature/top/sites/DefaultTopSitesStorage.kt
rxumoz: Is it acceptable that I adjust `numSitesRequired` before we get frecent top sites, so that frecent top sites can fill up?

gabrielluong:

PR #906, opened 2023-02-20
Bug 1814187 - Refractor code to toggle the save button to active when fields are properly set.

gitstart:
The PR fix the bug where the save icon on `Edit login` is always active 

### Issue Screenshots
[fnx-48-issue.webm](https://user-images.githubusercontent.com/93866435/215518790-c37a2e0a-33a0-477e-8c58-42e852393f00.webm)

### Fix Screenshots
[fnx-48-fix.webm](https://user-images.githubusercontent.com/93866435/215518511-8a2b1db2-eabf-4dcc-b4ba-f3d6c8fcedc2.webm)

Pull Request checklist
### Pull Request checklist
- [x] **Quality:** This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests:** This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog:** This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/main/docs/changelog.md) or does not need one
- [x] **Accessibility:** The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/main/android/accessibility_guide.md) or does not include any user facing features

### After merge
 - Milestone: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
 - Breaking Changes: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

**GitHub Automation**
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187

https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187
https://bugzilla.mozilla.org/show_bug.cgi?id=1814187

PR #930, opened 2023-02-21
Bug 1815421 - Clear button is active without username input on `Add login`

gitstart:
The PR removes clear button when there's no user input 

### Issue video
[fnxx-14-issue.webm](https://user-images.githubusercontent.com/93866435/219415222-c694bf28-9431-4c7d-b600-b5ee130dc01e.webm)

### Demo video after fix
[fnxx-14-fix.webm](https://user-images.githubusercontent.com/93866435/219414781-8876cd25-4dc5-4977-9a07-9e8d95d42252.webm)

### Pull Request checklist
- [x] **Quality:** This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests:** This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog:** This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/main/docs/changelog.md) or does not need one
- [x] **Accessibility:** The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/main/android/accessibility_guide.md) or does not include any user facing features

### After merge
 - Milestone: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
 - Breaking Changes: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

---
This code was written and reviewed by GitStart Community. Growing great engineers, one PR at a time.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1815421


fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
```diff
@@ -200,10 +200,20 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
...
                     // update usernameChanged to true when the text is not empty,
                     // otherwise it is not changed, as this screen starts with an empty username.
-                    usernameChanged = editable.toString().isNotEmpty()
-                    updateUsernameField()
-                    setSaveButtonState()
-                    findDuplicate()
+                    when {
+                        editable.toString().isEmpty() -> {
+                            validUsername = false
+                            binding.clearUsernameTextButton.isVisible = false
```
Mugurell, line 206: Based on the model of the password field this is where we should also call `updateUsernameField` to set an error for empty username.
(`usernameChanged` should probably be changed to `true` here also)

fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
```diff
@@ -200,10 +200,20 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
...
+                        }
+                        else -> {
+                            usernameChanged = editable.toString().isNotEmpty()
```
Mugurell, line 209: nit: editable.toString().isNotEmpty() and we know is false.
I think `usernameChanged` would always be `true` after the user interacted once with this field.

fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
```diff
@@ -200,10 +200,20 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
...
+                            binding.clearUsernameTextButton.isVisible = !editable.isNullOrBlank()
```
Mugurell, line 210: Here we are using a check on whether the text is `blank` but above only if the text is `empty`. Based on the current code it looks like we should stick with a check for just `empty`?

fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
```diff
@@ -200,10 +200,20 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
                 override fun afterTextChanged(editable: Editable?) {
                     // update usernameChanged to true when the text is not empty,
                     // otherwise it is not changed, as this screen starts with an empty username.
-                    usernameChanged = editable.toString().isNotEmpty()
-                    updateUsernameField()
-                    setSaveButtonState()
-                    findDuplicate()
+                    when {
+                        editable.toString().isEmpty() -> {
```
Mugurell, line 204: nit: no non-null assertion operator (`?`) used for the `editable` parameter.

Mugurell (changes requested):
It looks to me like in both of the scenarios (after the user changed the username fiels the username is empty or not) we should:
- update `usernameChanged` to `true`
- update the clear button's visibility based on a `isEmpty` check
- call `updateUsernameField()` (which calls `setSaveButtonState()` on it's own)
- ensure `setSaveButtonState()` is called

 `+`
Only if the username is not empty on top of all of the above call `findDuplicate(..)`

What do you think?

gitstart:
Thank you @Mugurell for reviewing🙏.
Will soon push the requested changes. 🚀

fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
Mugurell: Thanks!
But it seems that the clear button does not allows for only spaces while other functionality does - because elsewhere we only do `isEmpty` checks?

fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
```diff
@@ -200,10 +200,20 @@ class AddLoginFragment : Fragment(R.layout.fragment_add_login), MenuProvider {
                 override fun afterTextChanged(editable: Editable?) {
                     // update usernameChanged to true when the text is not empty,
                     // otherwise it is not changed, as this screen starts with an empty username.
```
Mugurell, line 202: This comment should probably be moved to after line 208, above
>  usernameChanged = editable.toString().isNotEmpty()

fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt
gitstart: @Mugurell `other functionality does` what does it mean? Do you think we need to remove blank space check at beginning of?

Mugurell: I would expect to use only one check: `isBlank()` or `isEmpty()` everywhere not a combination of them.
At line 209  `isNotEmpty` is used and then at line 210 `isNullOrBlank` is used.
Looking at the current code it seems like we are not checking for `blank`.
And checking with desktop I see that is allows for usernames or passwords composed solely from empty spaces so I recommend here also to only check whether the string is empty or not and avoid the `isBlank` check.

Mugurell:
@gitstart I see some merge commits being added to the PR.
Please note that we prefer to use `rebase` instead of `merge` to get the recent changes from `main` in order to keep the git history clean.

gitstart:
> @gitstart I see some merge commits being added to the PR. Please note that we prefer to use `rebase` instead of `merge` to get the recent changes from `main` in order to keep the git history clean.

Thank you for letting know it.🙏
Okay @Mugurell, from now we will make sure we rebase it instead merge.🚀

Mugurell (changes requested):
Thank you for updating this!
Looks like some discussions are still open (generally feel free to close discussions when addressing the feedback) but I also saw one issue remaining: the clear button doesn't appear for blank input in the username but does appear for blank input in the password.


https://user-images.githubusercontent.com/11428869/221632022-6f1e9076-eebf-47ab-b213-f8b9a9fc8b49.mp4

---

Thinking that we're trying too many changes and create the risk for other regressions I've looked at why the initial issue happens and saw that the clear button is initially active because of [this updateUsernameField() call](https://searchfox.org/mozilla-mobile/rev/9c5241a33119980c8a43339cf47124418a1adcca/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt#90) happening when the screen is opened which [would set validUsername = true here](https://searchfox.org/mozilla-mobile/rev/9c5241a33119980c8a43339cf47124418a1adcca/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/settings/logins/fragment/AddLoginFragment.kt#302) even if the username is empty - the user hasn't yet had the change to input any text for the username.

As such it seems that this is just an edgecase we need to consider - _only update the username field (errors / clear button) after the user entered some text_ - and I tried doing that with the smallest amount of changes [in this patch](https://gist.github.com/Mugurell/b3bc24384c293ead954c10f7debe85e7). Please check that and let me know if you think that could be a solution. Feel free to also base your patch on that.

zmckenney:
@iorgamgabriel Adding you as assignee because of your work on the bugzilla ticket associated with this PR.

iorgamgabriel:
The bug was fixed in this [pr](https://github.com/mozilla-mobile/firefox-android/pull/906)

PR #936, opened 2023-02-21
Bug 1815185 - Remove fixed height from `RecentSyncedTab`

MozillaNoah:
This content originally had a fixed height of 180, but it does not actually need this to be set as a `Modifier` property, since the card's contents will naturally make this larger than 180.

The text cut-off manifested after [replacing the fixed height of 36 from the top-level `Button` Composable](https://bugzilla.mozilla.org/show_bug.cgi?id=1813905) with the up-to-date padding values. The padding increase (in order to match the figma) caused the button to be bigger than what was allowed by the card's fixed height, causing the text to be vertically truncated. The [figma](https://www.figma.com/file/hJfV1KGZuOK7OYr29pooL5/Synced-tabs?node-id=334%3A14730&t=aKEEiisG4lp6r04a-0) for this homescreen component does not have the accurate padding represented, making the value of 180 misleading and unrepresentative of what the height of this default state should be.

The size discrepancy noted below is due to the button growing to its proper size but also the `Button` Composable adding some invisible tap area above and below the button. It obtains this via `Button` (ours) -> `Button` (Material) -> `Surface` -> `Modifier.minimumTouchTargetSize()`.

Removing the `Modifier.height(180)` resolves this issue, but this could also open the discussion for what we need to consider when translating figmas to Compose while incorporating `Modifier.minimumTouchTargetSize()`.

image

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1815185

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/view/RecentSyncedTab.kt
@@ -86,7 +86,6 @@ fun RecentSyncedTab(
     Card(
         modifier = Modifier
             .fillMaxWidth()
-            .height(180.dp)
             .combinedClickable(
                 onClick = { tab?.let { onRecentSyncedTabClick(tab) } },
                 onLongClick = { isDropdownExpanded = true },


gabrielluong (approved):

MatthewTighe (approved):
Thanks for fixing this! I'm curious whether the 192 measurement is dynamic or static, and if so where the value can be found. I am mostly just curious how to avoid issues with Figma in the future

MozillaNoah:
> Thanks for fixing this! I'm curious whether the 192 measurement is dynamic or static, and if so where the value can be found. I am mostly just curious how to avoid issues with Figma in the future

My assumption would be this:
| Static sizing | Dynamic sizing |
| -- | -- |
| Tab icon  | Text |  
| Synced tab drawable | Button |
| White space from `Spacer` | |
| Pre-programmed padding | |

The dynamic stuff would be mostly driven by their use of `sp` vs `dp` (e.g. text scaling from the device using a smaller or larger font size than default), but also the framework inserting any shadow padding via `Modifier.minimumTouchTargetSize()`. I think the shadow padding is the biggest offender of this, but it's not something easily accounted for (unless we can somehow create a smart modifier to do this for us), as it may only be needed with certain font sizes or other odd edge cases. The shadow padding is not even added when the button has an icon.

In the screenshot, this is

MozillaNoah:
@Mergifyio backport releases_v111

PR #952, opened 2023-02-22
Bug 1646242 - Avoid crashes when requireContext is called and fragment is no longer attached.

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1646242

### DIFF
--- a/focus-android/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/focus-android/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -926,7 +926,7 @@ class BrowserFragment :
             },
             showConnectionInfo = ::showConnectionInfo,
             showCookieBannerExceptionsDetailsPanel = ::showCookieBannerExceptionDetailsPanel,
-        ).also { currentEtp -> currentEtp.show() }
+        ).also { currentEtp -> context?.let { currentEtp.show() } }
     }
 
     private fun reloadCurrentTab() {
--- a/focus-android/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
+++ b/focus-android/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
@@ -654,9 +654,9 @@ class BrowserFragment :
             ),
         )
 
-        snackbar.setAction(getString(R.string.download_snackbar_open)) {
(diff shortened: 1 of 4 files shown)


focus-android/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
```diff
@@ -616,7 +616,9 @@ class BrowserFragment :
 
                 TelemetryWrapper.downloadDialogDownloadEvent(sentToDownload = true)
 
-                showDownloadCompletedSnackbar(state, extension)
+                context?.let {
```
jonalmeida, line 619: Which crash signature is this fix meant for? I'm having a tough time sifting through the reports from the linked bug since it has multiple attached.

jonalmeida (approved):

focus-android/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt
mcarare: [@ java.lang.IllegalStateException: at androidx.fragment.app.Fragment.requireContext(Unknown Source:17) ]

Some of the reports contain this path in the crash signature(https://crash-stats.mozilla.org/report/index/9820ec02-e358-438a-a16d-088430230226, one of them).

jonalmeida: Thanks!

It looks like the crash is happening at [this](https://searchfox.org/mozilla-mobile/rev/3d5542b4f51f83da68ea168039f08da4951f6e4c/firefox-android/focus-android/app/src/main/java/org/mozilla/focus/fragment/BrowserFragment.kt#659) `requireContext`.


I think what will happen is that the early context check will still pass and we run into the same crash because the onClick lambda happens at a later point (when the context is no longer available).

What we could do is either:
* Add an early check in the lambda itself.
* Use the `View` in the onClick lambda to find the context:
  ```kotlin
  snackbar.setAction(getString(R.string.download_snackbar_open)) { view ->
    val opened = AbstractFetchDownloadService.openFile(
      applicationContext = view.context.applicationContext,
      download = state,
    )
  }
  ```

mcarare: updated.

jonalmeida (approved):

PR #958, opened 2023-02-22
Bug 1817523 – fix application search engines background

mavduevskiy:
### Screenshots
 Old | New
 --- | --- 
 |   
 |   

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1817523


fenix/app/src/main/res/drawable/ic_bookmarks_search.xml
```diff
@@ -7,15 +7,8 @@
...
-        android:fillColor="@color/photonBlack"/>
+  <path
+      android:pathData="M4.365,8.949L7.225,11.324L7.398,11.922L6.49,15.488C6.367,15.972 6.549,16.474 6.953,16.767C7.354,17.059 7.888,17.078 8.314,16.812L11.44,14.835H12.061L15.188,16.812C15.387,16.938 15.612,17 15.835,17C16.085,17 16.334,16.922 16.547,16.766C16.951,16.473 17.133,15.972 17.01,15.487L16.1,11.912L16.28,11.32L19.135,8.949C19.519,8.63 19.668,8.116 19.513,7.64C19.359,7.165 18.937,6.837 18.439,6.805L14.742,6.566L14.241,6.191L12.877,2.764V2.763C12.691,2.3 12.249,2 11.75,2C11.251,2 10.809,2.3 10.623,2.764L9.259,6.191L8.758,6.566L5.061,6.805C4.563,6.837 4.141,7.165 3.987,7.64C3.832,8.116 3.981,8.63 4.365,8.949ZM4.982,19.7L5.782,20.5H17.717L18.517,19.7V17.75C18.517,17.336 18.853,17 19.267,17C19.681,17 20.017,17.336 20.017,17.75V19.5C20.017,20.881 18.898,22 17.517,22H5.982C4.601,22 3.482,20.881 3.482,19.5V17.75C3.482,17.336 3.818,17 4.232,17C4.646,17 4.982,17.336 4.982,17.75V19.7Z"
+      android:fillColor="@color/mozac_ui_icons_fill"
```
mavduevskiy, line 12: Overall in the app `android:fillColor="?attr/textPrimary" ` is used for the automatic support for dark/light colours, but it doesn't work in this case because drawables are [converted](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/components/Core.kt#217) to `Bitmap` to be stored in `SearchEngine`, and then converted back to drawables to be used in `TextMenuCandidate`.

I am not sure why this double conversion was created, maybe should remove as a follow up.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -655,6 +656,11 @@ class HomeFragment : Fragment() {
                     text = it.name,
                     start = DrawableMenuIcon(
                         drawable = it.icon.toDrawable(resources),
+                        tint = if (it.type == SearchEngine.Type.APPLICATION) {
+                            requireContext().getColorFromAttr(R.attr.textPrimary)
```
gabrielluong, line 660:
```suggestion
                            requireContext().getColorFromAttr(R.attr.iconPrimary)
```
We should use the correct variable here. textPrimary is for text whereas we want the icon variable here.

mavduevskiy: let's merge it as it is, I will file a bug to add the attribute. currently there is no `iconPrimary` attribute, and the majority is the icons in the xml are using `android:fillColor="?attr/textPrimary"` as a filler.

gabrielluong: Yes, this makes sense to me. Historically, we have been using `textPrimary` for both text and icons before the introduction of the `iconPrimary` with the design system. So, you will see in almost everywhere that we're using `textPrimary` for the icon color.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -655,6 +656,11 @@ class HomeFragment : Fragment() {
                     text = it.name,
                     start = DrawableMenuIcon(
                         drawable = it.icon.toDrawable(resources),
+                        tint = if (it.type == SearchEngine.Type.APPLICATION) {
```
gabrielluong, line 659: You will also need to do this in https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogFragment.kt#772. Right now, this will only affect the Home screen, but when you navigate to a website and click on the search bar. You will see a mismatch of colors because the search dialog search selector menu isn't updated with the appropriate tint.

As a result, I can see the following after doing a simple search after navigating to a webpage:
Screen Shot 2023-02-23 at 1 26 27 PM

gabrielluong (changes requested):
Just gonna flag this as changes requested in case it wasn't clear because of https://github.com/mozilla-mobile/firefox-android/pull/958/files#r1116085252.

STR:
1. Start with dark mode
2. Enter an address and navigate to the site. Observe that search selector menu looks right.
3. Change to light theme
4. Go back to a site and tap on the address bar
5. Observe that the icons in the search selector menu of the search dialog fragment are using the dark mode colours.

mavduevskiy:
@gabrielluong 
oh, that's a bit emberassing :) but thank you for noticing that!!
fixed

fenix/app/src/main/res/drawable/ic_bookmarks_search.xml
```diff
@@ -7,15 +7,8 @@
...
-        android:strokeWidth="1" />
-    <path
-        android:pathData="m7.225,11.324 l-2.86,-2.375a1.207,1.207 0,0 1,-0.378 -1.309,1.205 1.205,0 0,1 1.074,-0.835l3.697,-0.239 0.501,-0.375 1.364,-3.427A1.209,1.209 0,0 1,11.75 2c0.499,0 0.941,0.3 1.127,0.763v0.001l1.364,3.427 0.501,0.375 3.697,0.239c0.498,0.032 0.92,0.36 1.074,0.835 0.155,0.476 0.006,0.99 -0.378,1.309L16.28,11.32l-0.18,0.592 0.91,3.575c0.123,0.485 -0.059,0.986 -0.463,1.279a1.202,1.202 0,0 1,-1.359 0.046l-3.127,-1.977h-0.621l-3.126,1.977c-0.426,0.266 -0.96,0.247 -1.361,-0.045a1.205,1.205 0,0 1,-0.463 -1.279l0.908,-3.566 -0.173,-0.598z"
-        android:fillColor="@color/photonBlack"/>
-    <path
-        android:pathData="m5.782,20.5 l-0.8,-0.8v-1.95a0.75,0.75 0,0 0,-1.5 0v1.75a2.5,2.5 0,0 0,2.5 2.5h11.535a2.5,2.5 0,0 0,2.5 -2.5v-1.75a0.75,0.75 0,0 0,-1.5 0v1.95l-0.8,0.8H5.782z"
-        android:fillColor="@color/photonBlack"/>
+  <path
```
gabrielluong, line 10: Before you land this, could you add a reference to where these new icons are sourced from? This is mainly to provide some context for anybody in the future that is looking back at these icon changes and being able to determine the requirements and the source of these icons at the time for whatever reason they needed to do this. 

I am asking for this because there were no clear directions in the bug description about where to grab these icons from. Ideally, we would have this in the bug description, but I also see some benefits to linking a figma in the PR description as well.

gabrielluong (approved):

fenix/app/src/main/res/drawable/ic_bookmarks_search.xml
gabrielluong, line 12: We don't typically use `@color/mozac_ui_icons_fill` for the icons in Fenix. You might be better off just using `?attr/textPrimary` here. I don't believe there's anything wrong with it, but `mozac_ui_icons_fill` is only defined within ui-icons https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/ui/icons/src/main/res/values/colors.xml#3 and defaults to only white.

Using `textPrimary` gives you the benefit of being able to see what the icon will look like under light and dark theme in Design mode of Android Studio.

gabrielluong: I am probably missing the most obvious benefit of using `textPrimary` here. If this icon was to be used elsewhere in Fenix, then `textPrimary` would be the appropriate color without having to change the tint programmatic.

fenix/app/src/main/res/drawable/ic_bookmarks_search.xml
mavduevskiy: design (and icons) could be found [here](https://www.figma.com/file/epu0Sj6mCPk7K282sNpHgI/Site?node-id=874%3A14793&t=KJyX7Whx81P5rg5Z-0)

fenix/app/src/main/res/drawable/ic_bookmarks_search.xml
mavduevskiy: It feels like you skipped over this [comment](https://github.com/mozilla-mobile/firefox-android/pull/958#discussion_r1114969826)

mavduevskiy:
@gabrielluong 
thanks for a review, gl! especially noticing that it was broken in the edit mode, I should have noticed that myself.
Will land on Monday

fenix/app/src/main/res/drawable/ic_bookmarks_search.xml
gabrielluong: Should the fillColor really affect the fact that we are programmatic changing the tint though?

mavduevskiy: In this case using an alias as fillColor results in blank icons. We are converting drawables to bitmaps, to convert those to drawables again, and my hunch is the first conversion results in a blank bitmap

gabrielluong: Gotcha gotcha. I had understood the original comment, but it never fully landed.

PR #975, opened 2023-02-23
Bug 1818056 - Delay TCP CFR until dismissing the cookie banner re-engagement dialog

Amejia481:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1818056


Mugurell:
Looking at the [failing ui tests](https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5943330685776557177):
- cookieProtectionHintTest fails because it was actually checking for the TCP message [here](https://searchfox.org/mozilla-mobile/rev/1f15b04e5c88c592f54ab8e749995c8cc01b8ef8/firefox-android/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt#708-715). I suggest disabling the test for now.
-  jumpBackInCFRTest fails because of the same check [here](https://searchfox.org/mozilla-mobile/rev/1f15b04e5c88c592f54ab8e749995c8cc01b8ef8/firefox-android/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ContextualHintsTest.kt#53). I suggest removing this line for now.

 cc @AndiAJ

fenix/app/src/androidTest/java/org/mozilla/fenix/helpers/FeatureSettingsHelperDelegate.kt
```diff
@@ -30,7 +30,7 @@ class FeatureSettingsHelperDelegate : FeatureSettingsHelper {
         isRecentTabsFeatureEnabled = settings.showRecentTabsFeature,
         isRecentlyVisitedFeatureEnabled = settings.historyMetadataUIFeature,
         isPWAsPromptEnabled = !settings.userKnowsAboutPwas,
-        isTCPCFREnabled = settings.shouldShowTotalCookieProtectionCFR,
+        isTCPCFREnabled = settings.shouldShowTotalCookieProtectionCFR && settings.shouldShowCookieBannerReEngagementDialog(),
```
Mugurell, line 33: I think this would normally be
```suggestion
        isTCPCFREnabled = settings.shouldShowTotalCookieProtectionCFR && !settings.shouldShowCookieBannerReEngagementDialog(),
```
the same as in the `BrowserToolbarCFRPresenter.kt`
but right now there is no way to show / considered as already shown the CBH and that's why these were combined this way?

I'm just thinking that if so then these lines don't need any change?
cc @AndiAJ also.

Mugurell (approved):
LGTM.
Just some questions about the UI tests.

Amejia481:
I tried to address the test by disabling the cookie banner dialog on it.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarCFRPresenter.kt
```diff
@@ -68,7 +65,7 @@ class BrowserToolbarCFRPresenter(
      */
     @Suppress("MagicNumber")
     fun start() {
-        if (settings.shouldShowTotalCookieProtectionCFR) {
+        if (settings.shouldShowTotalCookieProtectionCFR && !settings.shouldShowCookieBannerReEngagementDialog()) {
```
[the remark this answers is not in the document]
Mugurell: After talking with Andi a new interesting scenario was seen: What if the cookie protection banner never shows? 
We could potentially wait quite a long time to show the TCP CFR.
Not seeing a way to mitigate this other than just deciding to add a new condition to show the TCP CFR after a number of tabs being opened which would still allow for it to be shown at the same time with CBH but hopefully with reduced chances?

AndiAJ:
Hi guys,

Yes, as @Mugurell mentioned, in both cases the tests fail when trying to verify the TCP CFR.

Based on the fact that the TCP CFR should appear after the Cookie banner reduction dialog is dismissed, the easiest way to make te tests work would be to disable the cookie banner before the tests starts, rather than doing it in the tests.

```
@get:Rule
    val activityTestRule = HomeActivityTestRule(
        isJumpBackInCFREnabled = true,
        isTCPCFREnabled = true,
        isPocketEnabled = false,
        isRecentlyVisitedFeatureEnabled = false,
        isCookieBannerReductionDialogEnabled = false, ⬅️
    )
```

Also, as @Mugurell mentioned I think the changes made to `isTCPCFREnabled` in `FeatureSettingsHelperDelegate.kt` as well as in `HomeActivityTestRule.kt` are not needed, they should remain separate for better control.

Thanks !  😊

Amejia481:
I'll update the PR for adding an additional fallback, in case the user don't see a cookie banner after certain X amount of pages visited :) .

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarCFRPresenter.kt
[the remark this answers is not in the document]
Amejia481: That sounds fair!

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarCFRPresenter.kt
```diff
@@ -86,6 +88,13 @@ class BrowserToolbarCFRPresenter(
         }
     }
 
+    private fun shouldShowCFR(): Boolean {
+        return settings.shouldShowTotalCookieProtectionCFR && (
+            !settings.shouldShowCookieBannerReEngagementDialog() ||
+                settings.openTabsCount >= CRF_MINIMUM_NUMBER_OPENED_TABS
+            )
+    }
```
Mugurell, line 96: Thank you! 👍

Amejia481: Thanks for proposing an alternative solution :)

Amejia481:
@mergify backport releases_v111

PR #985, opened 2023-02-24
Bug 1813252 - Refactor MessagingMiddlewareTest to use a real Store in…

sarah541:
…stance

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1813252


fenix/app/src/test/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddlewareTest.kt
```diff
@@ -178,27 +195,36 @@ class MessagingMiddlewareTest {
...
-    fun `WHEN onMessageDismissed THEN updateMetadata,removeMessage , UpdateMessages and removeMessageToShowIfNeeded`() = runTestOnMain {
+    fun `WHEN onMessageDismissed THEN updateMetadata,removeMessage , UpdateMessages and consumeMessageToShowIfNeeded`() = runTestOnMain {
         val message = Message(
             "control-id",
             mockk(relaxed = true),
             action = "action",
             mockk(relaxed = true),
             listOf("trigger"),
-            Message.Metadata("same-id"),
+            Message.Metadata("same-id", displayCount = 0),
```
jonalmeida, line 205: For `onMessageDismissed` to happen, wouldn't we need to have displayed the message at least once?

```suggestion
            Message.Metadata("same-id", displayCount = 1),
```

fenix/app/src/test/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddlewareTest.kt
```diff
@@ -178,27 +195,36 @@ class MessagingMiddlewareTest {
...
+        store.waitUntilIdle()
 
-        coVerify { messagingController.onMessageDismissed(message.metadata) }
-        coVerify { messagingStorage.updateMetadata(message.metadata.copy(dismissed = true)) }
-        verify { middlewareContext.dispatch(UpdateMessages(emptyList())) }
-        verify { spiedMiddleware.removeMessage(middlewareContext, message) }
+        // removeMessages causes messages size to be 0
+        assertEquals(0, store.state.messaging.messages.size)
+        // consumeMessageToShowIfNeeded causes messageToShow map size to be 0
+        assertEquals(0, store.state.messaging.messageToShow.size)
```
jonalmeida, line 227: The assertion here is right, but if you look up at your state, you start off with `messageToShow` as empty. So we're asserting here that nothing has changed? Shouldn't we have had a `messageToShow` first, then after the dismissal, the message is removed from `messages` and `messageToShow`?

jonalmeida: Let's add an assertion before the `dispatch(MessageDismissed(message))` that the `messageToShow` has a message in there first.

sarah541: Yes, totally agree here

fenix/app/src/test/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddlewareTest.kt
```diff
@@ -48,240 +38,226 @@ class MessagingMiddlewareTest {
...
 
-        coEvery { messagingStorage.getMessages() } returns messages
+        val message = Message(
+            "message1",
+            mockk(relaxed = true),
+            "",
+            StyleData(),
+            emptyList(),
+            Message.Metadata("same-id", displayCount = 0),
+        )
```
jonalmeida, line 73: So here is the other bit of refactoring we could do if you think it's worthwhile.

We create virtually identical messages for each test with only one/two changes. A nice way to make these tests easier to parse or easier to write more in the future is with simple helper functions for our tests:

```kotlin
private fun createMessage(
    messageId: String = "control-id",
    data: MessageData = mockk(relaxed = true),
    action: String = "action",
    styleData: StyleData = StyleData(),
    triggers: List<String> = listOf("triggers"),
    metadata: Message.Metadata = createMetadata(),
) = Message(messageId, data, action, styleData, triggers, metadata)

private fun createMetadata(
    id: String = "same-id",
    displayCount: Int = 0,
    pressed: Boolean = false,
    dismissed: Boolean = false,
    lastTimeShown: Long = 0L,
    latestBootIdentifier: String? = null,
) = Message.Metadata(
    id,
    displayCount,
    pressed,
    dismissed,
    lastTimeShown,
    latestBootIdentifier,
)
```

This means that this line or the ones below turn into one/two liners like this:

```suggestion
        val message = createMessage()
```

jonalmeida (approved):
Great! This looks good to me to land.

I left some nits which aren't blocking but worth considering.

cc: @t-p-white who has worked on messaging and might want to know that we've updated these tests to properly utilizee the messaging middleware code in them.

fenix/app/src/test/java/org/mozilla/fenix/gleanplumb/state/MessagingMiddlewareTest.kt
sarah541: Yes makes sense

PR #1020, opened 2023-02-28
Bug 1812149 - Add ContextWrapper before updating locale

jonalmeida:
From ohall's learnings, we know that we need to use the original Android Activity Context with the system PrintManager. The most straight-forward way to do that is to use a ContextWrapper which internally holds a reference to the original (base) context.

Added the `EngineView.setActivityContext` to sample-browser, fenix and focus as well.

I didn't see any valuable tests to write for this; open to suggestions.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1812149


jonalmeida:
Firebase UI testing has [a temporary outage](https://status.firebase.google.com/incidents/CaJr3TT3ayqXeADqd5jf) causing UI test failures. Will retry when those are resolved.

jonalmeida:
This might not be ready to review - looks like the ui test failures are legitimate and this change has somehow broken the tabs tray and toolbar's ability to change locale RTL.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/activity/GeckoViewActivityContextDelegate.kt
```diff
@@ -35,7 +35,7 @@ class GeckoViewActivityContextDelegate(
      */
     override fun getActivityContext(): Context? {
         val context = contextRef.get()
-        if ((context == null) || (context as Activity).isDestroyed) {
+        if ((context == null)) {
```
jonalmeida, line 38: Had to remove this casting and check because the context can no longer be an original activity one. Maybe I'm doing something wrong here, but this casting fails even though we're able to use the context (which means its an activity context).

android-components/components/support/locale/src/main/java/mozilla/components/support/locale/ActivityContextWrapper.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.app.Activity
+import android.content.Context
+import android.content.ContextWrapper
+
+/**
+ * A [ContextWrapper] that holds the original [Activity] Context.
+ */
+class ActivityContextWrapper(
+    baseContext: Context,
+    val originalContext: Context = baseContext,
```
jonalmeida, line 16: This default might come off as confusing, so happy to remove it.

jonalmeida (commented):
After some investigation, it looks like there is no way around holding a separate reference to the original context. We can still store this in a `ContextWrapper` and recursively retrieve it from multiple wrappers.

ohall-m (approved):
Thanks! Tested against a geckoview with `window.print` enabled and Fenix and Focus responded as expected. (Sample browser just needs a slight change as noted and it responds as expected too.) 🖨

android-components/samples/browser/src/main/java/org/mozilla/samples/browser/BrowserActivity.kt
```diff
@@ -25,7 +25,7 @@ import org.mozilla.samples.browser.ext.components
 /**
  * Activity that holds the [BrowserFragment].
  */
-open class BrowserActivity : AppCompatActivity(), ComponentCallbacks2 {
+open class BrowserActivity : LocaleAwareAppCompatActivity(), ComponentCallbacks2 {
```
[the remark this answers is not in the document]
jonalmeida: Oh yes, good catch!

jonalmeida:
Offline, mcarare suggested using the `ActivityLifecycleCallback` to see if the activity retrieved from there would work. Unfortunately, the activity the callbacks provide there have the same problem. Where we fail at [this line](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/print/PrintManager.java;l=517-519;drc=b7b880315a81073a724db2dbba1ce9b7e94dfdcd) in the Android PrintManager.

Holding this patch back to land in 113.

PR #1060, opened 2023-03-01
Bug 1806741 - Expose hasRuleForBrowsingContext API

owlishDeveloper:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1806741

### DIFF
--- a/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -441,6 +441,37 @@ class GeckoEngineSession(
     }
 
     /**
+     * See [EngineSession.hasCookieBannerRuleForSession]
+     */
+    override fun hasCookieBannerRuleForSession(
+        onResult: (Boolean) -> Unit,
+        onException: (Throwable) -> Unit,
+    ) {
+        geckoSession.hasCookieBannerRuleForBrowsingContextTree().then(
+            { response ->
+                if (response == null) {
+                    logger.error(
+                        "Invalid value: unable to get response from hasCookieBannerRuleForBrowsingContextTree.",
+                    )
+                    onException(
+                        java.lang.IllegalStateException(
+                            "Invalid value: unable to get response from hasCookieBannerRuleForBrowsingContextTree.",
+                        ),
+                    )
+                    return@then GeckoResult()
+                }
+                onResult(response)
+                GeckoResult<Boolean>()
+            },
+            { throwable ->
+                logger.error("Checking for cookie banner rule failed.", throwable)
+                onException(throwable)
+                GeckoResult()
+            },
+        )
+    }
+
+    /**
      * Checks and returns a non-mobile version of the url.
      */
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
--- a/android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
+++ b/android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
@@ -97,6 +97,7 @@ import org.mozilla.geckoview.WebRequestError.ERROR_MALFORMED_URI
 import org.mozilla.geckoview.WebRequestError.ERROR_UNKNOWN
 import org.mozilla.geckoview.WebResponse
 import org.robolectric.Shadows.shadowOf
+import java.io.IOException
 import java.security.Principal
 import java.security.cert.X509Certificate
 
@@ -2291,6 +2292,80 @@ class GeckoEngineSessionTest {
 
         engineSession.toggleDesktopMode(false, reload = true)
         verify(engineSession, atLeastOnce()).reload()
+    }
+
+    @Test
+    fun `hasCookieBannerRuleForSession should call onSuccess callback for a valid GV response`() {
+        val engineSession = GeckoEngineSession(
+            mock(),
+            geckoSessionProvider = geckoSessionProvider,
+        )
+        var onResultCalled = false
+        var onExceptionCalled = false
+
+        val ruleResult = GeckoResult<Boolean>()
+        whenever(geckoSession.hasCookieBannerRuleForBrowsingContextTree()).thenReturn(ruleResult)
+
+        engineSession.hasCookieBannerRuleForSession(
+            onResult = { onResultCalled = true },
+            onException = { onExceptionCalled = true },
+        )
+
+        ruleResult.complete(true)
+        shadowOf(getMainLooper()).idle()
+
+        assertTrue(onResultCalled)
+        assertFalse(onExceptionCalled)
+    }
+
+    @Test
+    fun `hasCookieBannerRuleForSession should call onError callback in case GV returns an exception`() {
+        val engineSession = GeckoEngineSession(
+            mock(),
+            geckoSessionProvider = geckoSessionProvider,
+        )
+
+        var onResultCalled = false
+        var onExceptionCalled = false
+
+        val ruleResult = GeckoResult<Boolean>()
+        whenever(geckoSession.hasCookieBannerRuleForBrowsingContextTree()).thenReturn(ruleResult)
+
+        engineSession.hasCookieBannerRuleForSession(
+            onResult = { onResultCalled = true },
+            onException = { onExceptionCalled = true },
+        )
+
+        ruleResult.completeExceptionally(IOException())
+        shadowOf(getMainLooper()).idle()
+
+        assertTrue(onExceptionCalled)
+        assertFalse(onResultCalled)
+    }
+
+    @Test
+    fun `hasCookieBannerRuleForSession should call onError callback in case GV returns a null`() {
+        val engineSession = GeckoEngineSession(
+            mock(),
+            geckoSessionProvider = geckoSessionProvider,
+        )
+
+        var onResultCalled = false
+        var onExceptionCalled = false
+
+        val ruleResult = GeckoResult<Boolean>()
+        whenever(geckoSession.hasCookieBannerRuleForBrowsingContextTree()).thenReturn(ruleResult)
+
+        engineSession.hasCookieBannerRuleForSession(
+            onResult = { onResultCalled = true },
+            onException = { onExceptionCalled = true },
+        )
+
+        ruleResult.complete(null)
+        shadowOf(getMainLooper()).idle()
+
+        assertTrue(onExceptionCalled)
+        assertFalse(onResultCalled)
     }
 
     @Test
--- a/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
+++ b/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
@@ -781,6 +781,16 @@ abstract class EngineSession(
     abstract fun toggleDesktopMode(enable: Boolean, reload: Boolean = false)
 
     /**
+     * Checks if there is a rule for handling a cookie banner for the current website in the session.
+     *
+     * @param onSuccess callback invoked if the engine API returned a valid response. Please note
+     * that the response can be null - which can indicate a bug, a miscommunication
+     * or other unexpected failure.
+     * @param onError callback invoked if there was an error getting the response.
+     */
+    abstract fun hasCookieBannerRuleForSession(onResult: (Boolean) -> Unit, onException: (Throwable) -> Unit)
+
+    /**
      * Finds and highlights all occurrences of the provided String and highlights them asynchronously.
      *
      * @param text the String to search for
--- a/android-components/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
+++ b/android-components/components/browser/engine-system/src/main/java/mozilla/components/browser/engine/system/SystemEngineSession.kt
@@ -405,6 +405,13 @@ class SystemEngineSession(
         }
     }
 
+    override fun hasCookieBannerRuleForSession(
+        onResult: (Boolean) -> Unit,
+        onException: (Throwable) -> Unit,
+    ) {
+        throw UnsupportedOperationException("Cookie Banner handling is not available in this engine")
+    }
+
     /**
      * See [EngineSession.exitFullScreenMode]
      */
--- a/android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/EngineObserverTest.kt
+++ b/android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/EngineObserverTest.kt
@@ -68,6 +68,10 @@ class EngineObserverTest {
             override fun toggleDesktopMode(enable: Boolean, reload: Boolean) {
                 notifyObservers { onDesktopModeChange(enable) }
             }
+            override fun hasCookieBannerRuleForSession(
+                onResult: (Boolean) -> Unit,
+                onException: (Throwable) -> Unit,
+            ) {}
             override fun findAll(text: String) {}
             override fun findNext(forward: Boolean) {}
             override fun clearFindMatches() {}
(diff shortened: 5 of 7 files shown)


android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -440,6 +440,13 @@ class GeckoEngineSession(
         }
     }
 
+    /**
+     * See [EngineSession.hasCookieBannerRuleForSession]
+     */
+    override fun hasCookieBannerRuleForSession(): Boolean {
+        return geckoSession.hasCookieBannerRuleForBrowsingContextTree().poll()!!
```
Amejia481, line 447: As `hasCookieBannerRuleForBrowsingContextTree` is async call we will need to expose  a way for consumers to get the value without blocking, we could use callbacks you can find an similar example in [GeckoEngine](https://github.com/mozilla-mobile/firefox-android/blob/68f6309018768326683fa16a45ccc0c046ac5226/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#L226-L263).

Amejia481:
I couldn't take a deep look today, but in general this is looking good, I think we have to check the CI failures and try to address them. I'll  promote the PR to "Ready for review" to all the checks to run.

https://github.com/mozilla-mobile/firefox-android/pull/1060/checks?check_run_id=11729009989
https://github.com/mozilla-mobile/firefox-android/pull/1060/checks?check_run_id=11729007101
https://github.com/mozilla-mobile/firefox-android/pull/1060/checks?check_run_id=11729010639

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -440,6 +440,30 @@ class GeckoEngineSession(
...
+     */
+    override fun hasCookieBannerRuleForSession(
+        onSuccess: (Boolean) -> Unit,
+        onError: (Throwable) -> Unit,
+    ) {
+        geckoSession.hasCookieBannerRuleForBrowsingContextTree().then(
+            { response ->
+                if (response == null) {
+                    logger.error("No result from GeckoView for cookie banner rule.")
+                    return@then GeckoResult<Boolean>()
```
Amejia481, line 454: Should we call `onError()` for this case? As it looks like if hit this section neither `onError()` or `onSuccess()` will be called?

Amejia481: Could we cover this flow with test? :)

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
```diff
@@ -2293,6 +2294,43 @@ class GeckoEngineSessionTest {
...
+        engineSession.hasCookieBannerRuleForSession(
+            onSuccess = { onSuccessCalled = true },
+            onError = { onErrorCalled = true },
+        )
+
+        ruleResult.completeExceptionally(IOException())
+        shadowOf(getMainLooper()).idle()
+
+        assertTrue(onErrorCalled)
+    }
```
Amejia481, line 2332: Nit: Maybe we could break down the test in multiple smaller tests that only cover one test case. This could improve readability for future maintenance. Using [backticks](https://stackoverflow.com/questions/67273691/does-kotlin-allow-the-use-of-backtick-names-for-test-classes) (``), could help to clarify the test's aim, in other [tests files](https://github.com/mozilla-mobile/firefox-android/blob/4a86be5594cfa7bb0ff9a132f6c762b31c5fcc3c/android-components/components/compose/awesomebar/src/test/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionFetcherTest.kt#L29) we try to adopt the pattern [Given When Then](https://martinfowler.com/bliki/GivenWhenThen.html).

Amejia481 (commented):
This looks good ✅ !
I tested locally and it was working as expected!
I added a couple of comments not blocking, only one clarification before approving https://github.com/mozilla-mobile/firefox-android/pull/1060/files#r1126929730

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
owlishDeveloper: > Should we call `onError()` for this case? As it looks like if hit this section neither `onError()` nor `onSuccess()` will be called?

I don't think so. First of, when we implemented `checkForFormData()` [here](https://github.com/mozilla-mobile/firefox-android/pull/264/files) we didn't. Second, `onError` is for handling the error that comes from the API. In this case we are handling a legit response (which is not that likely, I added this null check mostly to get rid of the `Boolean?` to improve ergonomics of the API)

owlishDeveloper: > Could we cover this flow with test? :)

What exactly do you hope to test there? The behaviour of the logger? In the patch I mentioned above similar code path is also not tested as far as I can tell.

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
owlishDeveloper: We already have similar tests in the code base, [for example](https://github.com/mozilla-mobile/firefox-android/blob/055b6b67c690d7f8f1741ad4494d61ac73d91ccd/android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineTest.kt#L1914-L1980) - that one is even longer.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
Amejia481: > What exactly do you hope to test there? The behaviour of the logger? In the patch I mentioned above similar code path is also not tested as far as I can tell.

I was looking to exercise how we are notifying the caller when this case happens see https://github.com/mozilla-mobile/firefox-android/pull/1060#discussion_r1127028167.

I'm bit worry to confirm if we are in a state, *where we don't notify the caller, this function will block the door hanger from showing until we get the value* see https://github.com/mozilla-mobile/firefox-android/blob/22563c8ae26b6de2b0eabd9b1c5f7ddbd1012038/fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt#L379

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
Amejia481: That's true, the test is really old, ideally we would like to improve the code base whenever is possible, to make it more maintainable 🧹 ✨  :)

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
owlishDeveloper: How is the caller notified in the case of `checkForFormData()`? I think the way in both cases the caller gets notified is `onSuccess` is not getting called (+ in the case of `checkForFormData()` we don't notify observers either). Why would the function block? It returns in the case of `null`: `return@then GeckoResult<Boolean>()`, but maybe you have a better understanding?

owlishDeveloper: I mean, I certainly can add the test case to assert we don't call `onSuccess` in the case of `null` if you think that's valuable

Amejia481: For the `checkForFormData()`I think we missed that and we should file a bug (**I'll do it in a bit**), that functionality is not a user facing one, it's only use to [report telemetry](https://github.com/mozilla-mobile/fenix/pull/28697/files), it will fail silently.  On the other hand, `hasCookieBannerRuleForSession` will be use directly on the UI if we don't report all cases, we may miss some part of the UI.

`hasCookieBannerRuleForSession` will be use to construct this UI.
image

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -780,6 +780,13 @@ abstract class EngineSession(
      */
     abstract fun toggleDesktopMode(enable: Boolean, reload: Boolean = false)
 
+    /**
+     * Checks if there is a rule for handling a cookie banner for the current website in the session.
+     *
+     * @return true if there is a cookie banner handling rule, false otherwise.
```
Amejia481, line 786: nit: I think we should update this as we are not returning any more :)

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -780,6 +780,13 @@ abstract class EngineSession(
...
+     */
+    abstract fun hasCookieBannerRuleForSession(onSuccess: (Boolean?) -> Unit, onError: (Throwable) -> Unit)
```
Amejia481, line 788: Nit: We should document both parameters, specially what `onSuccess(null)` indicates.

Amejia481: Just a head up, as this is a generic API we don't propagate specifics implementation details about internal behaviours of the engines (`GeckoEngine` or `SystemEngine`), if the `GeckoEngine` provide `null` values we don't expose it as part of public APIs unless the same behaviour could apply for both engines.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -780,6 +780,16 @@ abstract class EngineSession(
      */
     abstract fun toggleDesktopMode(enable: Boolean, reload: Boolean = false)
 
+    /**
+     * Checks if there is a rule for handling a cookie banner for the current website in the session.
+     *
+     * @param onSuccess callback invoked if the engine API returned a valid response. Please note
+     * that the response can be null - which can indicate a bug, a miscommunication
+     * or other unexpected failure.
```
Amejia481, line 788: >      * @param onSuccess callback invoked if the engine API returned a valid response. Please note
>      * that the response can be null - which can indicate a bug, a miscommunication
>      * or other unexpected failure.
If `null` indicates `"indicates a bug, a miscommunication or other unexpected failure"` should we report it as an error instead (e.g `onError`)?  This will make the API concise and consistent for consumers.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -780,6 +780,16 @@ abstract class EngineSession(
...
+     * @param onError callback invoked if there was an error getting the response.
+     */
+    abstract fun hasCookieBannerRuleForSession(onSuccess: (Boolean?) -> Unit, onError: (Throwable) -> Unit)
```
jonalmeida, line 791: Drive-by 🚗 : we use `onSuccess` and `onError` as an existing pattern, however here, since the result is a boolean, this API looks a bit confusing.

It reads a bit like, "the result of hasCookieBannerRuleForSession if true will return in `onSuccess` and false in `onError`". Maybe I'm reading too much into it, and in which case, ignore. 🙂 

On the other hand, possibly consider two options for naming:
1. Rename `onError` to `onException` so it's clear the result from the second callback is an exceptional one.
2. Rename `onSuccess` to `onResult` so it's then clear where to expect the result of the boolean to come from (preferable).

Amejia481: Thanks ❤️ , completely agree!
`onResult` reads much nicer!

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
owlishDeveloper: I don't think so because `onError` is called for the API error. I am not sure how likely it is that we return null, tbh

owlishDeveloper: We can file a bug for improving the error handling in AC, and improve all the APIs (`checkForFormData()`) etc. I am not sure if the pattern that is already being used in the codebase should be holding up this patch

Amejia481: > Please note that the **response** can be null - which can indicate a bug, a miscommunication or other unexpected failure.

I think this looks like an implementation detail that shouldn't be pass to consumers, as they are not aware of a **response** is, this is a low detail on the `GeckoEngineSession` layer. Instead of us handling the GeckoView `null` response on our end, we are passing the handling to Fenix, Focus and all the apps that use AC.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
Amejia481: Bug for [checkForFormData()](https://bugzilla.mozilla.org/show_bug.cgi?id=1820803)

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
owlishDeveloper: It's not an implementation detail :) It's one possible return value. Which I'm not sure is likely at all.

owlishDeveloper: Currently, we have `(Boolean?) -> Unit` - which makes it obvious that the parameter can be null, and so null needs to be handled. The caller can handle that in any way they want - they can display an error, they can have some default value, they can call the API again etc.

The variant `(Throwable?) -> Unit` implies we can have a nullable error there. Which raises a question, if there is no error, why are we calling `onError`? And please note that the null still needs to be handled, but in this case we have less options - because we are in an error handler, so the only way to handle this now is as if it is an error.

Can you explain to me why you think it's more convenient?

owlishDeveloper: Even if we call `onError` with a certain kind of exception like invalid data exception - that is still less options to handle the null

Amejia481: For example, I'm seeing it from the perspective of the developer that will be consuming the API.

**Using the actual approach:** 
```kotlin
val onSuccess: (Boolean?) -> Unit = { 
    if (it == null){
        showErrorMessage() 
     } else {
        doSomethingWithTheValue(it)
     }
}

val onError: (Throwable)  -> Unit = { 
   logger(it)
   showErrorMessage() 
}

session.hasCookieBannerRuleForSession(onSuccess, onError)
```

**Use handling the `null` `response` and exposing a  `onSuccess: (Boolean) -> Unit`**

```kotlin
val onSuccess: (Boolean) -> Unit = { 
   doSomethingWithTheValue(it)
}

val onError: (Throwable)  -> Unit = { 
   logger(it)
   showErrorMessage() 
}

session.hasCookieBannerRuleForSession(onSuccess, onError)
```

We are not indicating enough information with  `onSuccess: (Boolean?) -> Unit` for consumers to handle the `null` diffidently than an error, it looks like we are making the API less consistent without providing any additional value to the consumer.

owlishDeveloper: Your suggestion is this though, isn't it?

```
val onError: (Throwable?)  -> Unit = { 
   logger.error(it)
   showErrorMessage() 
}
```

Or are you suggesting we do this in `then`: 

```
geckoResult.then(
  {value -> 
      if (value == null) onError(IllegalArgumentException)
     else doSomethingWith(value) 
   },
)
```

BTW, which exception are you suggesting?

Amejia481: Addressed via zoom!

Amejia481 (approved):

PR #1061, opened 2023-03-01
Bug 1819776 - Remove unused test dependencies from support-rusthttp

jonalmeida:
A ~test~ patch to see if removing the failing dependencies from the support-rusthttp has any effect. 

Forcing the build as well for added safety, but this shouldn't change with the cherry-picked GV and Glean bumps.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1819776


rvandermeulen (approved):
Ship it! Thanks for finding a solution, Jon!

PR #1082, opened 2023-03-02
Bug 1820036 - Add Matrix labeling on Firebase Test Lab

AaronMT:
To help one identify and locate one's test matrices in the Firebase console, you can use the optional --client-details matrixLabel="Example matrix label" flag to label a test matrix.

We can start by adding a PR number through the existing transform from `fenix` repo.

https://bugzilla.mozilla.org/show_bug.cgi?id=1820036

Screenshot 2023-03-07 at 11 48 55 AM

@JohanLorenzo transform import works well; review to double check the transform exits gracefully if push/direct push without PR number available is OK
https://bugzilla.mozilla.org/show_bug.cgi?id=1820036
https://bugzilla.mozilla.org/show_bug.cgi?id=1820036


JohanLorenzo (approved):
Looks good to me!

PR #1087, opened 2023-03-03
Bug 1813418 - Normalize search widget previews and microphone icon

Alexandru2909:
Normalized search widget previews and used the newer microphone icon from A-C. [Link](https://www.figma.com/file/ORQ1e04zJZFOITsULYlsF2/Search-engine-widget-preview-%5BAndroid%5D?node-id=7%3A3401&.t=LOtm7gE2GkyFg7k7-0) to Figma designs.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1813418


fenix/app/src/main/res/layout/search_widget_large.xml
```diff
@@ -38,6 +38,6 @@
         android:layout_alignParentEnd="true"
         android:layout_centerVertical="true"
         android:layout_marginEnd="1dp"
-        android:background="@drawable/ic_microphone_widget"
+        android:background="@android:color/transparent"
```
mavduevskiy, line 41: I wonder if there is a way to add the A-C icon through xml as well?

fenix/app/src/main/java/org/mozilla/gecko/search/SearchWidgetProvider.kt
```diff
@@ -163,19 +168,19 @@ class SearchWidgetProvider : AppWidgetProvider() {
         }
     }
 
-    private fun RemoteViews.setIcon(context: Context) {
+    private fun RemoteViews.setIcon(context: Context, viewId: Int, resourceId: Int) {
```
mavduevskiy, line 171: This change has some cons to it. Line 188 and below always sets the same content description (giving `R.id.button_search_widget_voice` the wrong description). Also, a nit, microphone icon doesn't have gradient so there is no value to run it through the same version check.

If we can't sen the mic icon through xml, maybe changing the function to `setIcons` and just adding the icon and content here would make more sense? 

It looks like `setIcon` method was supposed to encapsulate icon logic. Making a `setIcons` method or two separate methods would help to keep the icon related code in one place, and make the `createRemoteViews` more straightforward.

mavduevskiy (changes requested):

mavduevskiy:
Moving into the right direction! Let's not break the description and it's good to go!

Are there plans to align the widget itself with the new previews? They are a bit different – paddings, text size, and – the biggest – widget icon in dev build is different from the one in the widget preview :) The preview has the new blue one, and the widget  itself has the android with wings and a beak.

fenix/app/src/main/res/layout/search_widget_large.xml
Alexandru2909: While testing this I noticed that when the device is set to light theme, the icon remains white. The way Focus manages this is by overriding the `mozac_ui_icons_fill` color. Currently we are not doing this in Fenix, however the color is used in many drawables throughout both A-C and Fenix. The alternative to this would be to keep `ic_microphone_widget` as a duplicate of `mozac_ic_microphone` but using `fx_mobile_icon_color_primary` as `fill_color`. What do you think?

fenix/app/src/main/java/org/mozilla/gecko/search/SearchWidgetProvider.kt
Alexandru2909: I believe this doesn't change the content description for `button_search_widget_voice`, only for `button_search_widget_new_tab_icon`, however I agree we should try to use the xml to change the icon.

Alexandru2909 (commented):
I'm not sure about any plans to further improve the search widget - preview consistency, maybe [`previewLayout`](https://developer.android.com/develop/ui/views/appwidgets/enhance#add-scalable-widget-previews) might come in handy for that. Should we file a follow-up ticket?

The icon for the debug variant was updated based on [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1813418#c6) discussion.

fenix/app/src/main/res/layout/search_widget_large.xml
mavduevskiy: I like the idea of overriding the `mozac_ui_icons_fill` value! Is there any downside to this? Overriding a color value we can not theme sounds like a good idea, and a better idea than having a duplicate icon just with a different color.

Btw, we could also use tint to override the color. It would require suppressing the warning, but it looks like using `android:tint` could affect negatively only <21 versions, so we should be safe.
```
android:tint="@color/fx_mobile_text_color_primary"
```

mavduevskiy (commented):
I believe we should open up follow up tickets, to make the widgets consistent with the new previews :)

I read through the conversation about the icon, but you were talking about the icon in the preview, right? We used to have a Firefox Preview logo in the widget preview. But when you actually add the widget with the Firefox Developer Edition logo on preview to the home screen, the logo changes – to the android with a beak and wings. Which renders preview to be misleading.

fenix/app/src/main/res/layout/search_widget_large.xml
Alexandru2909: Thanks for the `android:tint` suggestion! I tried using `app:tint` in a previous iteration but that didn't work since it is part of the `AppCompatImageView`.

Because `mozac_ui_icons_fill` is mostly [used](https://searchfox.org/mozilla-mobile/search?q=mozac_ui_icons_fill&path=&case=false&regexp=false) in the `ui-icons` module, by overriding it, the only risk would be to use `fx_mobile_icon_color_primary` for icons which we may otherwise want to display as white but I'm not sure if there are any such cases in Fenix.

Whether we decide on using `android:tint` or overriding `mozac_ui_icons_fill`, we won't need to update `SearchWidgetProvider`, so I'll dismiss it's changes.

mavduevskiy: I vote for `android:tint` in this case. It is safe – we are targeting 21+. It is localized – it only affects tinting the widget icons, while overriding colors for all icons is riskier at this point. 

If we decide to do that, we should test the change first to see if it doesn't affect the existing coloring. Given the number of icons and existing theming, it looks a bit complex.

PR #948, opened 2023-02-22
Bug 1809305 - Allow user to copy an image to the clipboard

t-p-white:
[copy image snackbar.webm](https://user-images.githubusercontent.com/1833156/221835892-6e2192f6-d483-4e42-acf6-5b57f8d68ab0.webm)

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/android-components/blob/main/docs/changelog.md) or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/shared-docs/blob/main/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Milestone**: Make sure issues closed by this pull request are added to the [milestone](https://github.com/mozilla-mobile/android-components/milestones) of the version currently in development.
- **Breaking Changes**: If this is a breaking change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1809305
https://bugzilla.mozilla.org/show_bug.cgi?id=1819746

PR #1090, opened 2023-03-03
Bug 1820125 - Let `.buildconfig.yml` contain transitive dependencies

JohanLorenzo:
Bug found in https://github.com/mozilla-mobile/firefox-android/pull/948

The parser in `taskcluster/scripts/lint/update_buildconfig_from_gradle.py` originally comes from https://github.com/mozilla-mobile/android-components/pull/7617 and https://github.com/mozilla-mobile/android-components/pull/8594. 

@csadilek could you take a look at the generated `.buildconfig.yml` and let me know if you see anything suspicious? 
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125
https://bugzilla.mozilla.org/show_bug.cgi?id=1820125

### DIFF
--- a/android-components/.buildconfig.yml
+++ b/android-components/.buildconfig.yml
@@ -4,62 +4,104 @@ projects:
     path: components/browser/domains
     publish: true
     upstream_dependencies:
+    - concept-base
     - concept-toolbar
+    - lib-publicsuffixlist
+    - support-base
+    - support-ktx
     - support-test
+    - support-utils
+    - tooling-lint
   browser-engine-gecko:
     description: Engine implementation based on GeckoView.
     path: components/browser/engine-gecko
     publish: true
     upstream_dependencies:
+    - browser-errorpages
+    - concept-base
     - concept-engine
     - concept-fetch
+    - concept-storage
+    - lib-publicsuffixlist
     - service-glean
+    - support-base
     - support-ktx
     - support-test
     - support-utils
     - tooling-fetch-tests
+    - tooling-lint
+    - ui-icons
   browser-engine-system:
     description: Engine implementation based on the system WebView.
     path: components/browser/engine-system
     publish: true
     upstream_dependencies:
+    - browser-errorpages
+    - concept-base
     - concept-engine
+    - concept-fetch
+    - concept-storage
+    - lib-publicsuffixlist
+    - support-base
     - support-ktx
     - support-test
     - support-utils
+    - tooling-lint
+    - ui-icons
   browser-errorpages:
     description: Responsive browser error pages for Android apps.
     path: components/browser/errorpages
     publish: true
     upstream_dependencies:
+    - concept-base
+    - lib-publicsuffixlist
+    - support-base
     - support-ktx
     - support-test
+    - support-utils
+    - tooling-lint
     - ui-icons
   browser-icons:
     description: A component for loading and storing website icons.
     path: components/browser/icons
     publish: true
     upstream_dependencies:
+    - browser-errorpages
     - browser-state
     - concept-base
     - concept-engine
     - concept-fetch
+    - concept-storage
     - lib-fetch-httpurlconnection
     - lib-fetch-okhttp
+    - lib-publicsuffixlist
+    - lib-state
+    - support-base
     - support-images
     - support-ktx
     - support-test
+    - support-utils
+    - tooling-lint
+    - ui-icons
   browser-menu:
     description: A customizable menu for browsers.
     path: components/browser/menu
     publish: true
     upstream_dependencies:
+    - browser-errorpages
     - browser-state
+    - concept-base
     - concept-engine
+    - concept-fetch
     - concept-menu
+    - concept-storage
+    - lib-publicsuffixlist
+    - lib-state
     - support-base
     - support-ktx
     - support-test
+    - support-utils
+    - tooling-lint
     - ui-colors
     - ui-icons
   browser-menu2:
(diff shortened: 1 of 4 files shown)


taskcluster/scripts/lint/update_buildconfig_from_gradle.py
```diff
@@ -18,51 +19,45 @@
 
 logger = logging.getLogger(__name__)
 
-CONFIGURATIONS_WITH_DEPENDENCIES = (
-    "api",
-    "compileOnly",
-    "implementation",
-    "testImplementation",
-)
```
JohanLorenzo, line 26: I removed the configurations introduced in https://github.com/mozilla-mobile/android-components/pull/8594 because it didn't show transitive dependencies.

For instance: `./gradlew --parallel dependencies :compose-engine:dependencies` shows `support-ktx` but none of these commands do:

 * `./gradlew --parallel dependencies :compose-engine:dependencies --configuration api`
 * `./gradlew --parallel dependencies :compose-engine:dependencies --configuration compileOnly`
 * `./gradlew --parallel dependencies :compose-engine:dependencies --configuration implementation`
 * `./gradlew --parallel dependencies :compose-engine:dependencies --configuration testImplementation`

We actually benefit from this simplification: this script runs now much faster since it only shells out once instead of 4 times.

taskcluster/scripts/lint/update_buildconfig_from_gradle.py
```diff
@@ -18,51 +19,45 @@
...
+    for line in subprocess.check_output(
+        cmd, universal_newlines=True, cwd=gradle_root
+    ).splitlines():
+        # If we find the start of a new component section, update our tracking
+        # variable
+        if line.startswith("Project"):
+            current_project_name = line.split(":")[1].strip("'")
+
+        # If we find a new local dependency, add it.
+        local_dep_match = _LOCAL_DEPENDENCY_PATTERN.search(line)
```
JohanLorenzo, line 56: `seach()` makes sure we search the full string and not just the beginning. Transitive dependencies are indented in the gradle output. E.g.:

```
releaseUnitTestRuntimeClasspath - Runtime classpath of compilation 'releaseUnitTest' (target  (androidJvm)).
[...]
+--- project :browser-state
|    +--- project :lib-state
|    |    +--- org.jetbrains.kotlin:kotlin-stdlib:1.7.20 (*)
[...]
|    |    +--- project :support-base (*)
|    |    \--- project :support-ktx
```

taskcluster/scripts/lint/update_buildconfig_from_gradle.py
```diff
@@ -18,51 +19,45 @@
...
+        if local_dep_match:
+            local_dependency_name = local_dep_match.group("local_dependency_name")
+            if local_dependency_name != current_project_name:
```
JohanLorenzo, line 59: The current project is part the local dependencies output. For instance:

```
------------------------------------------------------------
Project ':compose-engine'
------------------------------------------------------------
[...]

releaseUnitTestRuntimeClasspath - Runtime classpath of compilation 'releaseUnitTest' (target  (androidJvm)).
+--- project :compose-engine (*)
```

fenix/.buildconfig.yml
```diff
@@ -62,6 +64,7 @@ projects:
     - lib-publicsuffixlist
     - lib-push-firebase
     - lib-state
+    - mozilla-lint-rules
```
JohanLorenzo, line 67: @csadilek I see the decision task is busted because of this line. I'm not sure why gradle reports it on Fenix not anywhere else:

```
./gradlew --parallel dependencies :app:dependencies
[...]
lintChecks - Configuration to apply external lint check jar
\--- project :mozilla-lint-rules
```

I guess it's related to this line but I'm unsure what to do: 

https://github.com/mozilla-mobile/firefox-android/blob/aa3adc0e486c55efa18f26e4a29615f2ba4aac5a/fenix/settings.gradle#L29

csadilek: @JohanLorenzo This is a real project in Fenix from https://github.com/mozilla-mobile/firefox-android/tree/main/fenix/mozilla-lint-rules. We will have to handle this in our script and ignore it.

csadilek: I don't think we can do this using Gradle, so it would have to be filtered out in our parser, or alternatively, we can add the project explicitly so it's known to avoid the `KeyError`?

android-components/.buildconfig.yml
```diff
@@ -4,150 +4,257 @@ projects:
     path: components/browser/domains
     publish: true
     upstream_dependencies:
```
bhearsum, line 6: This now appears to be full dependencies? If so, `upstream_dependencies` ought to be renamed. (I know that technically they're all still upstream dependencies...but we usually mean "direct upstream" when we say such things.)

bhearsum (approved):

android-components/.buildconfig.yml
JohanLorenzo: Oh TIL! I'd always assumed "upstream dependencies" included both direct and transitive ones

bhearsum: Now I'm doubting myself. I will double check this assumption later...

fenix/.buildconfig.yml
JohanLorenzo: Yeah, let's filter out this gradle project for now. It's not ideal but at least we'll make progress on the rest transitive dependencies.

android-components/.buildconfig.yml
bhearsum: OK, this is correct on `main` at the moment. For example:
- `browser-engine-gecko` depends on `concept-engine`
- `concept-engine` currently depends on `support-base`, which `browser-engine-gecko` does not explicitly depend on

In any case -- feel free to keep `upstream_dependencies` if you want. I was mainly pointing out that in a lot of our usage in other places (and currently here), that it usually means direct upstream only (regardless of whether or not that's technically correct).

taskcluster/scripts/lint/update_buildconfig_from_gradle.py
```diff
@@ -18,51 +19,49 @@
...
-                current_project_name = line.split(":")[1].strip("'")
-
-            # If we find a new local dependency, add it.
-            if line.startswith("+--- project") or line.startswith(r"\--- project"):
-                project_dependencies[current_project_name].add(line.split(" ")[2])
+    logger.info(f"Looking for dependencies in {gradle_root}")
+
+    # This is eventually going to fail if there's ever enough projects to make the
+    # command line too long. If that happens, we'll need to split this list up and
+    # run gradle more than once.
```
csadilek, line 36: nit: Should we maybe log the command? I wonder how we close we are to the limit (4k?), given we have over a hundred projects already.

csadilek (approved):
👍 Thank you!

PR #1097, opened 2023-03-03
Bug 1805683 - Upstream Nimbus messaging code to Android Components

titooan:
This PR aims to upstream Nimbus Messaging from Fenix to Android Components, so that it can be used from Fenix, Focus or any other browser using Android Components. 

"Gleanplumb" references in Fenix and Android Components have been renamed into Nimbus Messaging. 

Thanks a lot @jhugman and @jeddai for your help on this task!

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1805683


android-components/components/support/utils/src/main/java/mozilla/components/support/utils/BootUtils.kt
```diff
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-package org.mozilla.fenix.utils
+package mozilla.components.support.utils
```
jhugman, line 5: 😻

fenix/app/src/main/AndroidManifest.xml
```diff
@@ -333,7 +333,7 @@
                 android:exported="false" />
 
         <service
-            android:name=".gleanplumb.NotificationDismissedService"
+            android:name=".messaging.NotificationDismissedService"
```
jhugman, line 336: Does this `AndroidManifest.xml` fragment need to go to `service-nimbus`?
Answer: No. 

It might do if MessageNotificationWorker needs to go to an Android Component (though likely not the `service-nimbus` one).

jhugman (commented):
Really nice, good job.

There are a few comments here (mostly likely issues to file and then followup on).

Since I paired on quite a lot of this, I don't think I can give you a ✅ so would kindly ask @jonalmeida be the final arbiter of quality.

fenix/app/src/test/java/org/mozilla/fenix/nimbus/NimbusSystemTest.kt
```diff
@@ -10,6 +10,7 @@ import io.mockk.just
 import io.mockk.mockk
 import io.mockk.runs
 import io.mockk.slot
+import mozilla.components.service.nimbus.messaging.NimbusSystem
```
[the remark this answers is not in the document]
titooan: Actually this test depends on Fenix-specific Settings, thus I think it needs to live inside Fenix. Does it seem right to you?

fenix/app/src/test/java/org/mozilla/fenix/messaging/FenixNimbusMessagingControllerTest.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.fenix.BuildConfig
+import org.robolectric.RobolectricTestRunner
+import java.util.*
+
+@RunWith(RobolectricTestRunner::class)
+@ExperimentalCoroutinesApi
+class FenixNimbusMessagingControllerTest {
```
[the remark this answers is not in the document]
titooan: Oh right, I'll put those tests back to NimbusMessagingControllerTest inside AC. Thanks!

fenix/app/nimbus.fml.yaml
```diff
@@ -9,8 +9,126 @@ channels:
...
+              USER_FR_SPEAKER: "'fr' in locale"
+              DEVICE_ANDROID: os == 'Android'
+              DEVICE_IOS: os == 'iOS'
+              ALWAYS: "true"
+              NEVER: "false"
+              DAY_1_AFTER_INSTALL: days_since_install == 1
+              DAY_2_AFTER_INSTALL: days_since_install == 2
+              DAY_3_AFTER_INSTALL: days_since_install == 3
+              DAY_4_AFTER_INSTALL: days_since_install == 4
+              DAY_5_AFTER_INSTALL: days_since_install == 5
```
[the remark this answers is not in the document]
titooan: I created a bug and will handle it: https://bugzilla.mozilla.org/show_bug.cgi?id=1823691

Thanks!

android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt
```diff
@@ -208,8 +206,7 @@ class NimbusMessagingStorage(
         }
     }
 
-    @VisibleForTesting
-    internal fun isMessageEligible(
+    fun isMessageEligible(
```
Amejia481, line 209: Are we going to use this function outside of the `NimbusMessagingStorage`? If so we should have some comments documenting the function and its parameters :)

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
```diff
@@ -11,11 +11,11 @@ import mozilla.components.service.pocket.ext.recordNewImpression
 import org.mozilla.fenix.components.AppStore
 import org.mozilla.fenix.ext.filterOutTab
 import org.mozilla.fenix.ext.getFilteredStories
-import org.mozilla.fenix.gleanplumb.state.MessagingReducer
 import org.mozilla.fenix.home.pocket.PocketRecommendedStoriesSelectedCategory
 import org.mozilla.fenix.home.recentsyncedtabs.RecentSyncedTabState
 import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem
 import org.mozilla.fenix.home.recentvisits.RecentlyVisitedItem.RecentHistoryGroup
+import org.mozilla.fenix.messaging.state.MessagingReducer
```
Amejia481, line 18: I was wondering if we should also uplift, the state bits to AC, like `MessagingMiddleware`and `MessagingState` as this will need to be duplicated in Focus when we need to used it.

Amejia481: No suggesting that it should be added as part of this PR as it's already big, but maybe something that we could consider for re-usability 😄 .

fenix/app/src/main/java/org/mozilla/fenix/messaging/FenixNimbusMessagingController.kt
```diff
@@ -0,0 +1,19 @@
...
+ * used by Fenix
+ */
+class FenixNimbusMessagingController(
+    messagingStorage: NimbusMessagingStorage,
+    now: () -> Long = { System.currentTimeMillis() },
+) : NimbusMessagingController(
+    messagingStorage = messagingStorage,
+    deepLinkScheme = BuildConfig.DEEP_LINK_SCHEME,
+    now = now,
+)
```
Amejia481, line 19: Maybe, we may not need a new specific type of `NimbusMessagingController`? As we are not adding new behaviour just passing values to the super class. Maybe there is an use case that I'm missing could you help me to clarify :)

android-components/components/service/nimbus/messaging.fml.yaml
```diff
@@ -151,13 +151,4 @@ enums:
...
-  MessageSurfaceId:
-    description: The identity of a message surface
-    variants:
-      homescreen:
-        description: A banner in the homescreen.
-      notification:
-        description: A local notification in the background, like a push notification.
-      fullscreen:
-        description: A full screen dialog that is intended to be disruptive.
+        description: The surface should show no message.
```
jhugman, line 154: Nit: needs an empty last line.

android-components/components/service/nimbus/src/test/java/mozilla/components/service/nimbus/messaging/OnDiskMessageMetadataStorageTest.kt
```diff
@@ -20,9 +20,10 @@ import org.junit.Assert.assertTrue
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
-import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
+import org.robolectric.RobolectricTestRunner
 
-@RunWith(FenixRobolectricTestRunner::class)
+@RunWith(RobolectricTestRunner::class)
```
jonalmeida, line 25: AndroidJUnit4 should be enough, I think? It's faster than Robolectric.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/BootUtilsTest.kt
```diff
@@ -1,20 +1,20 @@
...
 import org.junit.runner.RunWith
-import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
-import org.mozilla.fenix.utils.BootUtils.Companion.getBootIdentifier
+import org.robolectric.RobolectricTestRunner
 import org.robolectric.annotation.Config
 
 private const val NO_BOOT_IDENTIFIER = "no boot identifier available"
 
-@RunWith(FenixRobolectricTestRunner::class)
+@RunWith(RobolectricTestRunner::class)
```
jonalmeida, line 17: We don't need any special test runner here. The Junit default one would work and is always best to use when possible.

jonalmeida (commented):
Nice work! I did a first pass and I can go through it again. I see there are other reviewers as well so apologies for any other overlap of review comments.

fenix/app/src/main/AndroidManifest.xml
titooan: I filed a bug to track that [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1823736)

android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt
```diff
@@ -12,10 +12,7 @@ import org.mozilla.experiments.nimbus.GleanPlumbInterface
 import org.mozilla.experiments.nimbus.GleanPlumbMessageHelper
 import org.mozilla.experiments.nimbus.internal.FeatureHolder
 import org.mozilla.experiments.nimbus.internal.NimbusException
-import org.mozilla.fenix.nimbus.ControlMessageBehavior
-import org.mozilla.fenix.nimbus.MessageSurfaceId
-import org.mozilla.fenix.nimbus.Messaging
-import org.mozilla.fenix.nimbus.StyleData
+import mozilla.components.service.nimbus.GleanMetrics.Messaging as GleanMessaging
```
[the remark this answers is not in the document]
titooan: I fixed that in a new commit, thanks for your insights!

android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt
titooan: Actually I only turned this method visible outside of tests because the test that uses it is now in another module. 

I [filed a bug to improve that](https://bugzilla.mozilla.org/show_bug.cgi?id=1823472) but for now I couldn't think of a cleaner way to solve the problem. 

In this case, would the best solution for now  be to document the function and its parameters?

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
titooan: Do you think that those suggestions can fall under the scope of [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1823736) ? 
If so, I can add it in its description

fenix/app/src/main/java/org/mozilla/fenix/messaging/FenixNimbusMessagingController.kt
titooan: The only reason I created this specific type was to unsure that Fenix uses NimbusMessagingController with the same configuration everywhere, from the app code and from the test code.  
For example, I wanted to avoid the further possibility of specifying a specific `httpActionToDeepLinkUriConvertor` somewhere and forgetting to do it all across the app. Does it make sense?

android-components/components/service/nimbus/.experimenter.yaml
```diff
@@ -0,0 +1,124 @@
+---
```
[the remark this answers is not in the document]
jhugman: Filed [EXP-3305](https://mozilla-hub.atlassian.net/browse/EXP-3305).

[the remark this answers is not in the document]
jhugman: Closing EXP-3305 now this is deleted.

android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt
Amejia481: Let's leave it as it is, this a good enough workaround until we move the tests to AC :)

android-components/components/service/nimbus/src/test/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorageTest.kt
```diff
@@ -2,56 +2,69 @@
...
+import org.robolectric.RobolectricTestRunner
+
+@RunWith(RobolectricTestRunner::class)
+@kotlinx.coroutines.ExperimentalCoroutinesApi
 class NimbusMessagingStorageTest {
     private lateinit var storage: NimbusMessagingStorage
-    private lateinit var metadataStorage: MessageMetadataStorage
-    private lateinit var gleanPlumb: GleanPlumbInterface
+    private val metadataStorage: MessageMetadataStorage = mock()
+    private val gleanPlumb: GleanPlumbInterface = mock()
```
Amejia481, line 41: Having these initialled as member class could cause flakyness on tests, as we will be sharing those instances between tests. Ideally we can initialize them  on [setup](https://junit.org/junit4/javadoc/4.12/org/junit/Before.html) function or using the [@Mock](https://javadoc.io/doc/org.mockito/mockito-core/latest/org/mockito/Mock.html) annotation to re-initialize each each instance per tests.

Amejia481:
@titooan thanks for all the effort! 
Maybe to reduce the size of the PR and risk, we could break down the work to first add the minimum necessary for messaging to be in AC, then we could cover the remaining tasks.

What about this?

1. Bring the core functionality.
2. Migrate tests. 
3. Cover any missing bits

What do you think?

android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt
titooan: I updated the description of the bug to take this in account

fenix/app/nimbus.fml.yaml
```diff
@@ -9,8 +9,133 @@ channels:
   - beta
   - nightly
   - developer
-includes:
-  - messaging.fml.yaml
+import:
+  - path: ../../android-components/components/service/nimbus/messaging.fml.yaml
```
Amejia481, line 13: Nice ✨  I didn't know you could use import :)

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
Amejia481: Sorry, for the delay I think we may need another bug, as I was referring to uplift the state part of the messages, like `MessagingMiddleware`, and `MessagingState` which are app agnostic.

fenix/app/src/main/java/org/mozilla/fenix/messaging/MessageNotificationWorker.kt
```diff
@@ -147,28 +148,15 @@ class MessageNotificationWorker(
          * Initialize the [Worker] to begin polling Nimbus.
          */
         fun setMessageNotificationWorker(context: Context) {
-            val messaging = FxNimbus.features.messaging
-            val featureConfig = messaging.value()
-            val notificationConfig = featureConfig.notificationConfig
-            val pollingInterval = notificationConfig.refreshInterval.toLong()
+//            val messaging = FxNimbusMessaging.features.messaging
```
Amejia481, line 151: nit: I think this was pushed by mistake :)

fenix/app/src/main/java/org/mozilla/fenix/messaging/FenixNimbusMessagingController.kt
Amejia481: Got it!
We could also create a single instance on `Analytics`, and reduce it everywhere else, similar to what we do with [messagingStorage](https://github.com/mozilla-mobile/firefox-android/blob/5bfbab24d37489c001da8f63471c3e22f8400a63/fenix/app/src/main/java/org/mozilla/fenix/components/Analytics.kt#L149-L160).

Amejia481:
@titooan and me tested it locally and it was working like a charm 🎉 .
I think we still need to cover [this](https://github.com/mozilla-mobile/firefox-android/pull/1097/files#r1143548852) before landing after that It ✅ !

Amejia481 (approved):
Great job 🎖️ !

PR #1132, opened 2023-03-06
Bug 1820467 - Removed dependency to google_material from support-ktx gradle.

t-p-white:
Removed dependency to google_material from support-ktx gradle. Added SNACKBAR_LENGTH_LONG to SnackbarDelegate.kt

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1820467


jonalmeida:
Thanks for working on this follow-up so quickly!

This patch does rightly remove the Material dependency from `support-ktx` but leaves the `support-util` dependency in place which doesn't quite solve the problem of our support libraries depending on this UI library.

What we currently have in-tree now has a dependency graph that sort of looks like this:

```mermaid
flowchart TD
    A[browser-engine] --> B[feature-contextmenu]
    C[support-ktx] --> B
    D[support-utils] --> B
    E[google_material] --> C
    E[google_material] --> D
```

What this change is requesting will make it a bit better, reducing some of the duplication, but leaves us with the support component depending on the UI material library:

```mermaid
  flowchart TD
      A[browser-engine] --> B[feature-contextmenu]
      C[support-ktx] --> B
      D[support-utils] --> B
      E[google_material] --> D
```

Do you think it's possible instead to "lift" this dependency up higher to a component that touching the UI layer (i.e. our feature component)? The thinking behind this suggestion is around our [architecture documentation](https://mozac.org/contributing/architecture) that makes feature components the highest level which depend on specific use cases that need to be implemented.

```mermaid
flowchart TD
    A[browser-engine] --> B[feature-contextmenu]
    C[support-ktx] --> B
    D[support-utils] --> B
    E[google_material] --> B
```

Our support libraries are used by many non-UI components (e.g. browser-state, browser-session-storage, concept-engine), and now we have made them all indirectly depend on this third-party library. See the [full list generated here](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/.buildconfig.yml) by our build system. 
In general, I'd say, we try to avoid `support` libraries from building a long list of dependencies because they are shared around and are often used like "utils" component, that are also traditionally a dumping ground and become ever increasingly large.

t-p-white:
@jonalmeida thank you very much for the detailed feedback 🙂 As discussed, I've implemented it as:

```mermaid
flowchart TD
    A[browser-engine] --> B[feature-contextmenu]
    C[support-ktx] --> B
    D[ui-widgets] --> B
    E[google_material] --> D

```

android-components/.buildconfig.yml
```diff
@@ -190,7 +190,6 @@ projects:
     upstream_dependencies:
     - browser-state
     - concept-engine
-    - support-ktx
```
t-p-white, line 193: @jonalmeida I'm assuming this safe to remove now? This was added due to the transitive dependency build issue we were seeing https://github.com/mozilla-mobile/firefox-android/pull/948/files#diff-07bb9c7b5be613874316fa2adac2ac55725f5fd9dcd4f17c958a73bc1c8c44ef

android-components/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
```diff
@@ -195,11 +190,7 @@ fun Context.copyImage(
     // Only show a Snackbar for Android 12 and lower.
     // See developer.android.com/develop/ui/views/touch-and-input/copy-paste .
     if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.S_V2) {
```
jonalmeida, line 192: If we're keeping this version check and comment here, then we should document that in the `param` for `onCopyConfirmation` and update the comment too.

The alternative is to also lift this check to the caller (which is probably preferable?).

android-components/.buildconfig.yml
jonalmeida: I think so too.

Maybe after rebasing on top of https://github.com/mozilla-mobile/firefox-android/pull/1090 the build errors might go away.

jonalmeida (commented):
This looks great, thanks for taking the time to get this right!

I left one comment for consideration, but otherwise, this should be good to land after the 113 cut.

android-components/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt
t-p-white: Good suggestion on lifting this check 👍

android-components/components/support/ktx/build.gradle
```diff
@@ -41,7 +41,6 @@ dependencies {
     implementation ComponentsDependencies.androidx_core_ktx
 
     implementation ComponentsDependencies.kotlin_coroutines
-    implementation ComponentsDependencies.google_material
```
jonalmeida, line 44: 🙌

android-components/components/support/ktx/src/main/res/values/strings.xml
```diff
@@ -5,13 +5,13 @@
...
+<resources xmlns:tools="http://schemas.android.com/tools" xmlns:moz="http://schemas.android.com/apk/res-auto">
     <!-- Text displayed when choosing which app to call with after selecting a phone number-->
     <string name="mozac_support_ktx_menu_call_with">Call with…</string>
     <!-- Text displayed when choosing which app to email with after selecting an email address-->
     <string name="mozac_support_ktx_menu_email_with">Email with…</string>
     <string name="mozac_support_ktx_menu_share_with">Share with…</string>
     <string name="mozac_support_ktx_share_dialog_title">Share via</string>
     <!-- Text for confirmation "snackbar" shown after copying an image to the clipboard. -->
-    <string name="mozac_support_ktx_snackbar_delegate_image_copied">Image copied to clipboard</string>
+    <string name="mozac_support_ktx_snackbar_delegate_image_copied" moz:removedIn="114" tools:ignore="UnusedResources">Image copied to clipboard</string>
```
jonalmeida, line 16: Why did we need to add this?

android-components/components/support/utils/build.gradle
```diff
@@ -31,7 +31,6 @@ dependencies {
     implementation ComponentsDependencies.androidx_annotation
     implementation ComponentsDependencies.androidx_core
     implementation ComponentsDependencies.androidx_core_ktx
-    implementation ComponentsDependencies.google_material
```
jonalmeida, line 34: 🙌

fenix/app/src/main/res/values/strings.xml
```diff
@@ -1160,6 +1160,8 @@
     <string name="share_copy_link_to_clipboard">Copy to clipboard</string>
     <!-- Toast shown after copying link to clipboard -->
     <string name="toast_copy_link_to_clipboard">Copied to clipboard</string>
+    <!-- Text for confirmation "snackbar" shown after copying an image to the clipboard. -->
+    <string name="snackbar_copy_image_to_clipboard_confirmation">Image copied to clipboard</string>
```
jonalmeida, line 1164: Was this meant to be part of different change?

jonalmeida (approved):
The changes look great!

Lets remove the string changes though before landing. r+ since everything apart from that looks right.

android-components/components/support/ktx/src/main/res/values/strings.xml
jonalmeida: To clarify a bit better, why not move this string to the ui-widget component as well? That way "Image copied to clipboard" can be used shared with both apps without it needing to be re-translated for each.

```xml
    <string name="mozac_ui_widget_image_copied">Image copied to clipboard</string>
```

t-p-white: Agreed, moving it to ui-widget component makes sense 👍

fenix/app/src/main/res/values/strings.xml
t-p-white: Was part of the above change but no longer required with your suggestion 🙂

jonalmeida:
I've tried to restart the build - feel free to land this whenever is convenient as it's already been approved. 🙂

PR #1207, opened 2023-03-13
Bug 1815428 - able to read the pbm shortcut label in german 

sharmaachintya08:
bug 1815428 
https://bugzilla.mozilla.org/show_bug.cgi?id=1815428

Earlier the text in the button in pbm shortcut dialog was overflowing in german language, but now the text is completely readable and not overflowing from the button after changing the layout_height to wrap_content for the button.

### DIFF
--- a/taskcluster/ci/startup-test/kind.yml
+++ b/taskcluster/ci/startup-test/kind.yml
@@ -35,7 +35,7 @@ task-defaults:
               json: true
     run-on-tasks-for: []
     attributes:
-        build-type: nightly
+        build-type: nightly-firebase
         shipping-product: fenix
         # cannot safely run this in Firebase
         # turning off til we can safely run test in taskcluster


Mugurell (approved):
Looks like longer text were not considered at the time when this was first implemented.
Changes looks good, thank you!

fenix/app/src/main/res/layout/pbm_shortcut_popup.xml
```diff
@@ -42,7 +42,7 @@
             android:id="@+id/cfr_pos_button"
             style="@style/NeutralButton"
             android:layout_width="0dp"
-            android:layout_height="36dp"
+            android:layout_height="wrap_content"
```
Mugurell, line 45: This would lead to a significant change in UI:

| before | after |
| --- | --- |
| image | image |

Easiest way to keep the current UI would be to add 6dp vertical insets to the background similar to [what we're doing now to increase tap area](https://github.com/mozilla-mobile/firefox-android/blob/795f3950b671b102de99b47684a7089ffef10b96/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#L910) (line of code which after this changes should be removed).

Could also check with UX for an approval for visually increasing the button height.
@gitstart Let me know what you prefer.

Mugurell (changes requested):
Looked a bit more into the proposed solution and looks like there is still a bit more work to do.

fenix/app/src/main/res/layout/pbm_shortcut_popup.xml
sharmaachintya08: @Mugurell can i look into it and correct the bug without changing much into the ui ?

sharmaachintya08: okay, I'll look into it:).

PR #1208, opened 2023-03-13
Bug 1822091 - Execute tailored Fenix/Focus UI test task on release promotion

AaronMT:
MTE-440 & https://bugzilla.mozilla.org/show_bug.cgi?id=1822091

https://bugzilla.mozilla.org/show_bug.cgi?id=1822091
https://bugzilla.mozilla.org/show_bug.cgi?id=1822091
https://bugzilla.mozilla.org/show_bug.cgi?id=1822091
https://bugzilla.mozilla.org/show_bug.cgi?id=1822091

### DIFF
--- a/taskcluster/ci/ui-test-apk/kind.yml
+++ b/taskcluster/ci/ui-test-apk/kind.yml
@@ -326,7 +326,6 @@ tasks:
         worker:
             env:
                 GOOGLE_PROJECT: moz-fenix
-
     fenix-robo-arm-debug:
         attributes:
             shipping-product: fenix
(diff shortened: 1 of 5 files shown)


taskcluster/ci/ui-test-apk/kind.yml
```diff
@@ -351,3 +350,79 @@ tasks:
         worker:
             env:
                 GOOGLE_PROJECT: moz-fenix
+    fenix-arm-release:
+        attributes:
+            build-type: beta-firebase
+            shipping-phase: push
```
AaronMT, line 356: @JohanLorenzo Just want to verify again if this is all that's needed to schedule this `task` to be part of a `Push` phase Graph (e.g, [here](https://firefox-ci-tc.services.mozilla.com/tasks/groups/SCnpCb59S0u2UaYbRkUZ4A)). 

Looking at `release-notify-push` `task` ([here](https://github.com/mozilla-mobile/firefox-android/blob/main/taskcluster/ci/release-notify-push/kind.yml)) for example, are there any other attributes needed?

For context, @sv-ohorvath looks for the `release-notify-push` email to signal QA to start testing and running automation on candidates (so this is automating some of that)

JohanLorenzo: Thanks for asking! You can test this out locally 😃 That's one of the perks of taskgraph.

1. First, create a python virtual environment.
2. Activate it
3. Install taskgraph: https://github.com/taskcluster/taskgraph/#installation
4. Run `taskgraph target-graph -p taskcluster/test/params/main-repo-promote-beta.yml` ➡️ make sure `ui-test-apk-fenix-arm-release` isn't there
5. Run `taskgraph target-graph -p taskcluster/test/params/main-repo-push-beta.yml` ➡️ this time, it should be here.
6. You can do the same with `main-repo-promote-release.yml` for the release graph (instead of the beta one)

You can even see the impact of your patch with `taskgraph target-graph -p taskcluster/test/params/main-repo-push-beta.yml --diff origin/main`, `origin` being the name of the remote repository that points to `mozilla-mobile/firefox-android`. You can figure it out with `git remote -v`

AaronMT:
I think this is ready to try out and iterate from there on future adjustments and further automation.

What we're looking for here is to run smoke-tests (part of `ui-test-apk`) on the `releases` branches when the associated `shipping-product` (Focus/Fenix) enters the `push` phase (`shipping-phase`) as I understand it.

This is part of adding automation for QA when they receive the `release-notify-push` email, it's a signal for them to start testing.

@sv-ohorvath to look over the Flank configs
@JohanLorenzo to look over the `ui-test-apk` tasks

sv-ohorvath (approved):
Flank configs look ok to me. Thanks

JohanLorenzo (approved):
Taskgraph changes look good to me.

I see `taskcluster/ci/ui-test-apk/kind.yml` now exceeds 400 lines and most of them are very similar, as in: they follow the same rules. I'd recommend to refactor this file by extracting the common logic in a transforms. To do so: 
 1. Copy this [existing transforms](https://github.com/mozilla-mobile/firefox-android/blob/e83ae923d15c9f4fb5ed34255e1d991b995d6756/taskcluster/android_taskgraph/transforms/test.py)
 2. Then you can use python functions (like the one called `add_pr_number()` in the previous file) to change the task config as if it were a Python dictionary. 

This way, we could inject the same type of config no matter if we're on focus, fenix, beta, release, etc. The rules will then exist as Python code. This will make `taskcluster/ci/ui-test-apk/kind.yml` easier to read and less error prone next time we add/modify a job.

AaronMT:
Coming back to this – @JohanLorenzo I'm happy to look at refactoring on a follow-up if you don't mind as we'd like to continue to try this out. For each of the tasks defined they are structured similarily but there's varying differences (e.g, the APKs used, task triggers, build-type etc as you know) but I'm not sure there's (perhaps you have more insight than me) on how to consolidate these (I understand what you're suggesting as I agree the yaml config file is growing).

JohanLorenzo:
Moving the configuration from YAML (in the `kind.yml`) to Python (in a transforms file) is the way to go, in my opinion. I'm happy to chat on a call to help clearing this up 🙂

gabrielluong:
@AaronMT Should this be landed or closed if this is something we don't want to land?

PR #1216, opened 2023-03-13
Bug 1822148 - Android Components: exception in release builds when se…

afranchuk:
…rializing crash pings

The kotlinx.serialization proguard rules aren't transitively provided, so they are copied to the proguard rules of the lib-crash library.

There's some updated generated code, I figure it doesn't hurt to include it.

Unfortunately, I tried a number of things but couldn't find a way to create a unit test which would catch this issue (short of manually writing some gradle tasks to run minification on the test suite).

Note that at one point I was just using `@Keep` from androidx, and I *thought* I had that working but I came in this morning and it wasn't, so I changed to proguard rules.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1822148


android-components/components/lib/crash/proguard-rules.pro
```diff
@@ -19,3 +19,40 @@
 # If you keep the line number information, uncomment this to
 # hide the original source file name.
 #-renamesourcefileattribute SourceFile
+
+# From https://github.com/Kotlin/kotlinx.serialization/blob/6333daf8bb4ee7e7c8cd66619c7c66cc8c69dddf/rules/common.pro
```
csadilek, line 23: > The kotlinx.serialization proguard rules aren't transitively provided, so they are copied to the proguard rules of the lib-crash library.

@afranchuk ah, it looks like the proguard rules have only recently been bundled with the library in version 1.5.0, and we're still on [1.4.1](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt#L19). See release notes for [1.5.0](https://github.com/Kotlin/kotlinx.serialization/releases) (need to scroll down to the first RC which mentions it).

Let's try and upgrade to 1.5.0 instead? Then we don't need to ship our own rules. Since this dependency was only recently introduced for lib-crash anyway, there's not really a reason to wait.

Wdyt? If there's a reason not to, happy to approve and land otherwise.

csadilek: Copying relevant section from 1.5.0 release notes:

**Bundled Proguard rules**

The kotlinx-serialization-core-jvm JAR file now includes consumer Proguard rules,
so manual Proguard configuration is no longer necessary for most of the setups.

🤞

afranchuk: Oh! That wasn't clear to me at all, I wish the docs mentioned that. I will try that.

afranchuk: Updating to 1.5.0 works!

csadilek (approved):
🚢

csadilek:
We're dealing with a mergify problem so I will merge this manually when green.

PR #1217, opened 2023-03-13
Bug 1814994 - Create Tabs Tray list and grid layouts

MozillaNoah:
Note: the disabling of the Private theme will be examined at-large in [Bug 1820897](https://bugzilla.mozilla.org/show_bug.cgi?id=1820897)https://bugzilla.mozilla.org/show_bug.cgi?id=1820897 

https://user-images.githubusercontent.com/87384386/224806154-a60466ef-54f1-4b47-b6fc-b5fb4bf38b63.mov

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1814994

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
@@ -12,29 +12,39 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.input.nestedscroll.nestedScroll
+import androidx.compose.ui.platform.rememberNestedScrollInteropConnection
 import androidx.compose.ui.unit.dp
 import com.google.accompanist.pager.ExperimentalPagerApi
 import com.google.accompanist.pager.HorizontalPager
 import com.google.accompanist.pager.rememberPagerState
 import kotlinx.coroutines.launch
+import mozilla.components.browser.state.state.ContentState
+import mozilla.components.browser.state.state.TabSessionState
 import mozilla.components.lib.state.ext.observeAsComposableState
 import org.mozilla.fenix.compose.Divider
 import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
 
 /**
  * Top-level UI for displaying the Tabs Tray feature.
  *
  * @param tabsTrayStore [TabsTrayStore] used to listen for changes to [TabsTrayState].
+ * @param displayTabsInGrid Whether the normal and private tabs should be displayed in a grid.
  */
-@OptIn(ExperimentalPagerApi::class)
+@OptIn(ExperimentalPagerApi::class, ExperimentalComposeUiApi::class)
 @Composable
 fun TabsTray(
     tabsTrayStore: TabsTrayStore,
+    displayTabsInGrid: Boolean,
 ) {
     val multiselectMode = tabsTrayStore
         .observeAsComposableState { state -> state.mode }.value ?: TabsTrayState.Mode.Normal
+    val normalTabs = tabsTrayStore
+        .observeAsComposableState { state -> state.normalTabs }.value ?: emptyList()
     val pagerState = rememberPagerState(initialPage = 0)
     val scope = rememberCoroutineScope()
     val animateScrollToPage: ((Page) -> Unit) = { page ->
@@ -48,10 +58,12 @@ fun TabsTray(
             .fillMaxSize()
             .background(FirefoxTheme.colors.layer1),
     ) {
-        TabsTrayBanner(
-            isInMultiSelectMode = multiselectMode is TabsTrayState.Mode.Select,
-            onTabPageIndicatorClicked = animateScrollToPage,
-        )
+        Box(modifier = Modifier.nestedScroll(rememberNestedScrollInteropConnection())) {
+            TabsTrayBanner(
+                isInMultiSelectMode = multiselectMode is TabsTrayState.Mode.Select,
+                onTabPageIndicatorClicked = animateScrollToPage,
+            )
+        }
 
         Divider()
 
@@ -64,12 +76,17 @@ fun TabsTray(
             ) { position ->
                 when (Page.positionToPage(position)) {
                     Page.NormalTabs -> {
-                        Text(
-                            text = "Normal tabs",
-                            modifier = Modifier.padding(all = 16.dp),
-                            color = FirefoxTheme.colors.textPrimary,
-                            style = FirefoxTheme.typography.body1,
-                        )
+                        FirefoxTheme(theme = Theme.getTheme(allowPrivateTheme = false)) {
+                            if (displayTabsInGrid) {
+                                TabGrid(
+                                    tabs = normalTabs,
+                                )
+                            } else {
+                                TabList(
+                                    tabs = normalTabs,
+                                )
+                            }
+                        }
                     }
                     Page.PrivateTabs -> {
                         Text(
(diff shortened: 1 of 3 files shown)


MatthewTighe (approved):
LGTM considering the private theming issue will be resolved separately, as will the layouts for synced tabs and private tabs.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -48,10 +58,12 @@ fun TabsTray(
             .fillMaxSize()
             .background(FirefoxTheme.colors.layer1),
     ) {
-        TabsTrayBanner(
-            isInMultiSelectMode = multiselectMode is TabsTrayState.Mode.Select,
-            onTabPageIndicatorClicked = animateScrollToPage,
-        )
+        Box(modifier = Modifier.nestedScroll(rememberNestedScrollInteropConnection())) {
```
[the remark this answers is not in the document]
MozillaNoah: It's to communicate with the `BottomSheetBehavior` that wraps the `ComposeView` to give the UI the feel of a bottom sheet without explicitly implementing a `BottomSheet` View or Composable. (see: [`TabSheetBehaviorManager`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabSheetBehaviorManager.kt) and [`TabsTrayFragment`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt#249-268)) This hook allows *just* the banner to be the thing that can be scrolled to expand/collapse the Tabs Tray.

Otherwise, we have something like this, where the tabs tray will collapse when swiping down on the tab list:

https://user-images.githubusercontent.com/87384386/224833665-990d26b4-2a73-4ddf-b04a-99cc50470d5a.mov

It took a couple hours of trial and error to figure out where that `Modifier.nestedscroll` needed to live.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -12,29 +12,39 @@ import androidx.compose.foundation.layout.padding
...
+import org.mozilla.fenix.theme.Theme
 
 /**
  * Top-level UI for displaying the Tabs Tray feature.
  *
  * @param tabsTrayStore [TabsTrayStore] used to listen for changes to [TabsTrayState].
+ * @param displayTabsInGrid Whether the normal and private tabs should be displayed in a grid.
  */
-@OptIn(ExperimentalPagerApi::class)
+@OptIn(ExperimentalPagerApi::class, ExperimentalComposeUiApi::class)
```
rahulsainani, line 38: We'll have a bunch of these soon and the callers would need to opt in as well, I think it might be good to add them in the build.gradle file as an opt in.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -64,12 +76,17 @@ fun TabsTray(
             ) { position ->
                 when (Page.positionToPage(position)) {
                     Page.NormalTabs -> {
-                        Text(
-                            text = "Normal tabs",
-                            modifier = Modifier.padding(all = 16.dp),
-                            color = FirefoxTheme.colors.textPrimary,
-                            style = FirefoxTheme.typography.body1,
-                        )
+                        FirefoxTheme(theme = Theme.getTheme(allowPrivateTheme = false)) {
```
rahulsainani, line 79: ❓ Do we need this? The theme is already a top level composable in `TabsTrayFragment` ?

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
```diff
@@ -0,0 +1,139 @@
...
+    tabs: List<TabSessionState>,
+) {
+    val tabListBottomPadding = dimensionResource(id = R.dimen.tab_tray_list_bottom_padding)
+    val state = rememberLazyListState()
+
+    LazyColumn(
+        modifier = Modifier.fillMaxSize(),
+        state = state,
+    ) {
+        items(tabs) { tab ->
```
rahulsainani, line 49: We can use `tab.id` as key for the item. This will take care of reorderings and avoid unnecessary recomposition, and then animations can be supported easily as well --> https://developer.android.com/jetpack/compose/lists#item-keys

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: This is to manually prevent the private theme from being used on the normal tabs page. I'm currently discussing with UX for another ticket, Bug 1820897, what parts of the Tabs Tray, if any, will actually utilize the private theme.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: I'm not 100% certain that the callers need to opt-in (my understanding is it's just the Composable functions that use them), but how does opting-in via a `build.gradle` file work? Is it essentially a global opt-in?

rahulsainani: You can add the opt in compiler flag [here](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/build.gradle#L139). It has to be fully qualified name. 

Something like:
```
"-opt-in=com.google.accompanist.pager.ExperimentalPagerApi",
"-opt-in=androidx.compose.foundation.ExperimentalFoundationApi",
```

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
rahulsainani: okay! thanks for the explanation!

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: Gotcha gotcha. Should we file a follow-up ticket then? I can add it to the big meta ticket and the kanban board so anyone can pick it up and work on it asap.

rahulsainani: I'd recommend updating in this PR itself.

MozillaNoah: My thoughts are adding the compiler flags + doing any of the refactoring work to remove all of the `@OptIn` flags is kind of out of scope for this particular ticket/PR and detract from the current changelog.

MatthewTighe: IMO adding a compiler flag to globally opt-in to an experimental API would do a good job of reducing explicit information in the code itself. Without personally knowing the full list of `ExperimentalComposeUIApi`s for example, it could lead to a situation where code is added without the gut check of whether we _should_ be added API calls that aren't finalized. 

That said, if we are organizationally prepared to accept that we will be ingesting experimental APIs throughout the codebase, I still agree that it should be done as a follow-up ticket so that it can be more easily tracked. I also think a separate follow-up ticket should be filed suggesting periodic investigation as to whether it's still needed and a comment linking to the second follow-up should be included in the patch for the first. 

I do see that we have at least one global opt-in for experimental coroutines API, but that to me is a better usecase than Compose opt-ins because there seems to be more churn with Compose spread across many different packages.

MatthewTighe: I also think that if the app were fully modularized and we could opt-in in a build script that was package specific there would be less risk, like what is done in A-C. Turning on an opt-in at the build level for the entire app just feels like a larger decision, to me.

rahulsainani: Fair point! as discussed on slack, there are other references than this file, let's do it as a separate ticket 👍

rahulsainani: @MatthewTighe i hear you, let's discuss on slack to not overtake this PR with this discussion. Maybe even a good topic of conversation for android engg meeting?

PR #1237, opened 2023-03-14
Bug 1821718 - Add feature and nimbus flag for juno onboarding

rahulsainani:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1821718

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -71,4 +71,9 @@ object FeatureFlags {
      * Enables the notification pre permission prompt.
      */
     const val notificationPrePermissionPromptEnabled = true
+
+    /**
+     * Enables the redesigned onboarding.
+     */
+    const val junoOnboardingEnabled = false
 }
--- a/fenix/app/.experimenter.yaml
+++ b/fenix/app/.experimenter.yaml
@@ -31,6 +31,14 @@ homescreen:
     sections-enabled:
       type: json
       description: "This property provides a lookup table of whether or not the given section should be enabled. If the section is enabled, it should be toggleable in the settings screen, and on by default."
+juno-onboarding:
+  description: A feature that shows juno onboarding flow.
+  hasExposure: true
+  exposureDescription: ""
+  variables:
+    enabled:
+      type: boolean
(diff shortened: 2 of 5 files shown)


fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -71,4 +71,9 @@ object FeatureFlags {
      * Enables the notification pre permission prompt.
      */
     const val notificationPrePermissionPromptEnabled = true
+
+    /**
+     * Enables the redesigned onboarding.
+     */
+    const val junoOnboardingEnabled = false
```
rocketsroger, line 78: ❓ Should this be Config.channel.isNightlyOrDebug?

rocketsroger (approved):

fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
rahulsainani: Yeah it can be, my thinking was to keep it disabled till the last patch is in.

rocketsroger: I think both are fine.  I would probably enable this in Nightly/Debug once there are for onboarding dialogs so we can make sure it's disabled during UI testing.

rahulsainani: yeah, will update it to `nightlyOrDebug` when the last patch is in 👍 
Good point about UI testing, will keep that in mind as this goes into beta/release.

rahulsainani: That's good to know, thanks! 
And for the usage, could we then just call `FxNimbus.features.junoOnboarding.value().enabled` instead of
using  `lazyfeatureFlagPreference()`?

jonalmeida: Yes, exactly. We don't need to check two values then - just the one.

rocketsroger: But we still want to experiment on it though.  Don't I need the feature flag to be true and then use the Nimbus Flag to experiment?

PR #1246, opened 2023-03-14
Bug 1809809 - debounce click events on wallpaper settings

HarrisonOg:
Added debouncing to the button click event for `WallpaperThumbnailItem` so that a rapid second click doesn't trigger the snackbar again before the new `currentWallpaper` is set from the initial click.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1809809

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
@@ -4,7 +4,14 @@
 
 package org.mozilla.fenix.compose.ext
 
+import android.os.SystemClock
+import androidx.compose.foundation.clickable
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.composed
 import androidx.compose.ui.draw.drawBehind
 import androidx.compose.ui.geometry.CornerRadius
 import androidx.compose.ui.graphics.Color
@@ -49,3 +56,29 @@ fun Modifier.dashedBorder(
         )
     },
 )
+
+/**
+ * Used when clickable needs to be debounced to prevent rapid successive clicks
+ * from calling the onClick function.
+ *
+ * @param debounceInterval The length of time to wait between click events in milliseconds
+ * @param onClick Callback for when item this modifier effects is clicked
+ */
+fun Modifier.debouncedClickable(
+    debounceInterval: Long = 1000L,
+    onClick: () -> Unit,
+) = composed {
+    var lastClickTime: Long by remember { mutableStateOf(0) }
+
+    this.then(
+        Modifier.clickable(
+            onClick = {
+                val currentSystemTime = SystemClock.elapsedRealtime()
+                if (currentSystemTime - lastClickTime > debounceInterval) {
+                    onClick()
+                    lastClickTime = currentSystemTime
+                }
+            },
+        ),
+    )
+}
--- a/fenix/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
@@ -10,7 +10,6 @@ import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.background
 import androidx.compose.foundation.border
-import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
@@ -51,6 +50,7 @@ import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import org.mozilla.fenix.R
 import org.mozilla.fenix.compose.ClickableSubstringLink
+import org.mozilla.fenix.compose.ext.debouncedClickable
 import org.mozilla.fenix.theme.FirefoxTheme
 import org.mozilla.fenix.wallpapers.Wallpaper
 
@@ -300,7 +300,7 @@ private fun WallpaperThumbnailItem(
             .fillMaxWidth()
             .aspectRatio(aspectRatio)
             .then(border)
-            .clickable { onSelect(wallpaper) }
+            .debouncedClickable { onSelect(wallpaper) }
             .then(contentDescriptionModifier),
     ) {
         bitmap?.let {


MatthewTighe (commented):
This feels like a useful enough pattern that I'm wondering if it might be worth abstracting as a new `Modifier` in `Modifier.kt`. What do you think?

HarrisonOg:
> This feels like a useful enough pattern that I'm wondering if it might be worth abstracting as a new `Modifier` in `Modifier.kt`. What do you think?

this sounds like a good idea, I'll see if I can rewrite this into a `Modifier`

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
```diff
@@ -49,3 +56,30 @@ fun Modifier.dashedBorder(
...
+    factory = {
+        var lastClickTime: Long by remember { mutableStateOf(0) }
+
+        this.then(
+            Modifier.clickable(
+                onClick = {
+                    if (SystemClock.elapsedRealtime() - lastClickTime > debounceInterval) {
+                        onClick()
+                    }
+                    lastClickTime = SystemClock.elapsedRealtime()
```
MozillaNoah, line 80: should this line be inside the `if`?

HarrisonOg: I think the lastClickTime should be updated every click right?

MozillaNoah: it should be every click that's successfully debounced right? otherwise, if the click was spammed, the `if` would never be true?

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
```diff
@@ -49,3 +56,30 @@ fun Modifier.dashedBorder(
...
+
+/**
+ * Used when clickable needs to be debounced to prevent rapid successive clicks
+ * from calling the onClick function.
+ *
+ * @param debounceInterval The length of time to wait between click events
+ * @param onClick Callback for when item this modifier effects is clicked
+ */
+fun Modifier.debouncedClickable(
+    debounceInterval: Long,
```
MatthewTighe, line 68: nit: I think we should specify units for this parameter and we can also have a sane default like 300 or 500 millis.

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
MatthewTighe: That could potentially end up debouncing clicks forever, which we may not want in every case even if we theoretically want for the wallpapers case. I don't think we need to accommodate both cases though, and can just err towards the general case.

fenix/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -274,6 +274,10 @@ private fun WallpaperThumbnailItem(
         Modifier
     }
 
+    // Used to debounce the click events for setting wallpaper
+    @Suppress("MagicNumber")
+    val debounceInterval = 1000L
```
MatthewTighe, line 279: If we use a default parameter as suggested above then we can potentially remove this magic number as well

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
HarrisonOg: ah I see what you mean, the debounce should be based on last time the click actually calls `onClick()` without being debounced.
I'll make the change to move it inside the loop.

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
HarrisonOg: sounds good, I'll add a default.
unfortunately setting the default minimum to anything less than 900 milliseconds seems to not prevent the immediate second click from being registered and sent a `onClick` event.

fenix/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
HarrisonOg: removed this magic number

fenix/app/src/main/java/org/mozilla/fenix/settings/wallpaper/WallpaperSettings.kt
```diff
@@ -300,7 +300,7 @@ private fun WallpaperThumbnailItem(
             .fillMaxWidth()
             .aspectRatio(aspectRatio)
             .then(border)
-            .clickable { onSelect(wallpaper) }
+            .debouncedClickable(onClick = { onSelect(wallpaper) })
```
MozillaNoah, line 303:
```suggestion
            .debouncedClickable {
                onSelect(wallpaper) 
            }
```

I think this can be simplified slightly.

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
```diff
@@ -49,3 +56,30 @@ fun Modifier.dashedBorder(
...
+ * from calling the onClick function.
+ *
+ * @param debounceInterval The length of time to wait between click events in milliseconds
+ * @param onClick Callback for when item this modifier effects is clicked
+ */
+fun Modifier.debouncedClickable(
+    debounceInterval: Long = 1000L,
+    onClick: () -> Unit,
+) = composed(
+    factory = {
```
MatthewTighe, line 71: nit: Since `composed` is a [SAM](https://kotlinlang.org/docs/fun-interfaces.html#sam-conversions) if we remove the `factory` named parameter here we can also remove the parens for the function call. Looks a little cleaner IMO but that's not a super strong opinion

MatthewTighe (approved):
one nit but otherwise LGTM! Thanks for cycling on this 🙂

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
```diff
@@ -49,3 +56,30 @@ fun Modifier.dashedBorder(
...
+        var lastClickTime: Long by remember { mutableStateOf(0) }
+
+        this.then(
+            Modifier.clickable(
+                onClick = {
+                    if (SystemClock.elapsedRealtime() - lastClickTime > debounceInterval) {
```
MozillaNoah, line 77: maybe extract the system clock time to a local value since it could change between this `if` and the `lastClickTime` reassignment?

MozillaNoah (commented):
Nits aside, this looks good to me so ship this whenever Matt gives his final sign off.

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
HarrisonOg: ah yes, good call, removed the `factory` parameter

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
HarrisonOg: good point, I moved system clock time to a local value

MozillaNoah (approved):

PR #1258, opened 2023-03-15
Bug 1821479 - Add support for fullscreen custom tabs

Alexandru2909:
Added support for fullscreen custom tabs.

https://user-images.githubusercontent.com/35462038/226352974-ed1f2042-5eff-4423-b5f3-4de1af3808e1.mp4

https://user-images.githubusercontent.com/35462038/225335562-aa709caf-0e33-43a6-91c0-1fd8e44dc42d.mp4

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1821479

### DIFF
--- a/android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
+++ b/android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
@@ -6,8 +6,8 @@ package mozilla.components.feature.session
 
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.cancel
-import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.map
+import mozilla.components.browser.state.selector.findCustomTab
 import mozilla.components.browser.state.selector.findTabOrCustomTabOrSelectedTab
 import mozilla.components.browser.state.state.SessionState
 import mozilla.components.browser.state.store.BrowserStore
@@ -27,17 +27,23 @@ open class FullScreenFeature(
     private val fullScreenChanged: (Boolean) -> Unit,
 ) : LifecycleAwareFeature, UserInteractionHandler {
     private var scope: CoroutineScope? = null
+    private var isFullscreenCustomTab: Boolean = false
     private var observation: Observation = createDefaultObservation()
 
     /**
      * Starts the feature and a observer to listen for fullscreen changes.
      */
     override fun start() {
-        scope = store.flowScoped { flow ->
-            flow.map { state -> state.findTabOrCustomTabOrSelectedTab(tabId) }
-                .map { tab -> tab.toObservation() }
-                .ifChanged()
-                .collect { observation -> onChange(observation) }
+        if (tabId != null && store.state.findCustomTab(tabId)?.content?.fullScreen == true) {
+            isFullscreenCustomTab = true
+            fullScreenChanged(true)
+        } else {
+            scope = store.flowScoped { flow ->
+                flow.map { state -> state.findTabOrCustomTabOrSelectedTab(tabId) }
+                    .map { tab -> tab.toObservation() }
+                    .ifChanged()
+                    .collect { observation -> onChange(observation) }
+            }
         }
     }
 
@@ -65,12 +71,21 @@ open class FullScreenFeature(
     override fun onBackPressed(): Boolean {
         val observation = observation
 
-        if (observation.inFullScreen && observation.tabId != null) {
+        if (!isFullscreenCustomTab && observation.inFullScreen && observation.tabId != null) {
             sessionUseCases.exitFullscreen(observation.tabId)
             return true
         }
 
         return false
+    }
+
+    /**
+     * Method used to exit fullscreen mode, allowing fullscreen custom tabs to remain fullscreen.
+     */
+    fun exitFullscreenMode() {
+        if (!isFullscreenCustomTab) {
+            fullScreenChanged(false)
+        }
     }
 }
 
--- a/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
+++ b/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/CustomTabSessionState.kt
@@ -73,12 +73,14 @@ fun createCustomTab(
     private: Boolean = false,
     webAppManifest: WebAppManifest? = null,
     initialLoadFlags: EngineSession.LoadUrlFlags = EngineSession.LoadUrlFlags.none(),
+    fullscreen: Boolean = false,
 ): CustomTabSessionState {
     return CustomTabSessionState(
         id = id,
         source = source,
         content = ContentState(
             url = url,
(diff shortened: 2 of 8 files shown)


android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -65,13 +71,22 @@ open class FullScreenFeature(
...
         }
 
         return false
     }
+
+    /**
+     * Method used to exit fullscreen mode, allowing fullscreen custom tabs to remain fullscreen.
+     */
+    fun exitFullscreenMode() {
+        if (!isFullscreenCustomTab) {
```
rocketsroger, line 86: ❓ Is there a way to exit fullscreen custom tab then?

rocketsroger (approved):
Looks good, one question but don't have to block from landing.  We can always update this when integrating with fullscreen research surface.

android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
Alexandru2909: For full-screen custom tabs, we only exit full-screen mode when the browser fragment is [destroyed](https://github.com/mozilla-mobile/firefox-android/pull/1258/files#diff-4385d548241ec2239de72ceb324eeb29f03cd502b71a29a1285be2f0f350c71dR1491). I added the call in `onDestroyView` as not to disrupt current `FullScreenFeature.onBackPressed()` [interactions](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#1207). Should we allow the user to exit full-screen mode on back pressed for example?

jonalmeida: Huh, so if I open a youtube video in a custom tab, I can never leave it? 🙃

Alexandru2909: I don't know how we could currently open a youtube video as directly fullscreen in a custom tab. I know we can use `youtube.com/watch_popup?v=` to open a video in fullscreen on dekstop but it doesn't work on mobile. 

In the case when we open a youtube link in a custom tab and then watch the video fullscreen, pressing back would call the `FullscreenFeature.onBackPressed()` with `isFullscreenCustomTab` set to `false` (the tab would be created as non-fullscreen). 

For the case when we open a fullscreen video in a fullscreen custom tab, I updated the first video from the PR description.

Should we maybe update the naming of the `isFullscreenCustomTab` flag or move the logic to the `BaseBrowserFragment` ?

android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -27,17 +27,23 @@ open class FullScreenFeature(
...
     override fun start() {
-        scope = store.flowScoped { flow ->
-            flow.map { state -> state.findTabOrCustomTabOrSelectedTab(tabId) }
-                .map { tab -> tab.toObservation() }
-                .ifChanged()
-                .collect { observation -> onChange(observation) }
+        if (tabId != null && store.state.findCustomTab(tabId)?.content?.fullScreen == true) {
+            isFullscreenCustomTab = true
+            fullScreenChanged(true)
+        } else {
```
jonalmeida, line 40: Why do we need this? If we are sending a CustomTabState with fullscreen enabled, (what is currently in the else) should be enough to handle this case.

android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -27,17 +27,23 @@ open class FullScreenFeature(
     private val fullScreenChanged: (Boolean) -> Unit,
 ) : LifecycleAwareFeature, UserInteractionHandler {
     private var scope: CoroutineScope? = null
+    private var isFullscreenCustomTab: Boolean = false
```
jonalmeida, line 30: We have a concept of state holding here in this feature one line below - the Observation. We can add extra state into that instead of creating a new concept of state holding. See below.

android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
```diff
@@ -65,13 +71,22 @@ open class FullScreenFeature(
     override fun onBackPressed(): Boolean {
         val observation = observation
 
-        if (observation.inFullScreen && observation.tabId != null) {
+        if (!isFullscreenCustomTab && observation.inFullScreen && observation.tabId != null) {
```
jonalmeida, line 74: Related to the above comment, if we fold the state holding into the Observation, we just need to check if we're in a custom tab. So we could add a `Observation.isCustomTab` property then.

jonalmeida (changes requested):
Left some review comments below that we should address if we're going to land this. There is a lot of code that we don't need here.

> Add support for fullscreen custom tabs

Nit: this commit message doesn't make a lot of sense because we can fullscreen in a custom tab already.

android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
jonalmeida: If we remove the if-check, everything still works as expected:

```diff
diff --git a/android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt b/android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
index 536a31043a..34f85aa4bf 100644
--- a/android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
+++ b/android-components/components/feature/session/src/main/java/mozilla/components/feature/session/FullScreenFeature.kt
@@ -34,17 +34,17 @@ open class FullScreenFeature(
      * Starts the feature and a observer to listen for fullscreen changes.
      */
     override fun start() {
-        if (tabId != null && store.state.findCustomTab(tabId)?.content?.fullScreen == true) {
-            isFullscreenCustomTab = true
-            fullScreenChanged(true)
-        } else {
+//        if (tabId != null && store.state.findCustomTab(tabId)?.content?.fullScreen == true) {
+//            isFullscreenCustomTab = true
+//            fullScreenChanged(true)
+//        } else {
             scope = store.flowScoped { flow ->
                 flow.map { state -> state.findTabOrCustomTabOrSelectedTab(tabId) }
                     .map { tab -> tab.toObservation() }
                     .ifChanged()
                     .collect { observation -> onChange(observation) }
             }
-        }
+//        }
     }

     override fun stop() {

```

Alexandru2909:
Based on discussions with product, support for fullscreen custom tabs is no longer required. Thank you @jonalmeida for reviewing this!

PR #1276, opened 2023-03-16
Bug 1822750 - Add option to open default browser help page in custom tab

rahulsainani:
### What
As the Juno Onboarding Cards include a CTA to make firefox the default browser, it needs to use Activity.openSetDefaultBrowserOption() extension function. This currently defaults to opening a new tab. As the user would be in the Juno Onboarding Flow, a new tab would not work. So this PR adds option to open default browser help page in a custom tab.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1822750


t-p-white (approved):
lgtm 👍

Mugurell:
What happens if from a custom tab users then tap on the menu to choose "Open in browser"?

rahulsainani:
> What happens if from a custom tab users then tap on the menu to choose "Open in browser"?



> What happens if from a custom tab users then tap on the menu to choose "Open in browser"?

Good question! It opens the new tab, that's not what we want, is there a way to disable that from the caller?

Mugurell:
> > What happens if from a custom tab users then tap on the menu to choose "Open in browser"?
> 
> Good question! It opens the new tab, that's not what we want, is there a way to disable that from the caller?

Don't think so.
We can add options but I don't think we can remove the default ones.
https://developer.chrome.com/docs/android/custom-tabs/integration-guide/
Given that a custom tab opens in the default browser which may not be Fenix it doesn't seem like we can safely avoid this.

Not sure about the overall UX but maybe we could introduce in the app another functionality for showing one webpage at a time, similar to a custom tab but implemented really through just a fragment/dialog showing an `EngineView` in private mode (so that history etc would not be kept)
This scenario would be useful in other similar scenarios also.
Personally I would really like something like a modal dialog with a bit of empty space around it to suggest that the page the user is seeing is somewhat temporary, in-app content and not part of the browsing UX.

image

Piratesarereal:
Yup, I understand, trying to be mindful of unintentionally leaking users out of onboarding if they discover this by themselves, they would be able to skip onboarding entirely. I am not sure about the complexity of petru's suggested implementation, but I would say it can be tabled for now. But I wouldn't close this issue, because it can still possibly cause user leaks from onboarding. We can come back to this post MVP implementation of Android Onboarding for Juno.

rahulsainani:
Thanks @Piratesarereal We've created this [ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1823892) to solve the issue of leaking out the users out of onboarding, and that's been worked on. 
So now, the caller can use this to show the SUMO page in a custom tab and when the mentioned ticket is done, the caller can also pass in the flag to hide the "Open in Firefox" menu item. @Mugurell wdyt?

Mugurell:
> Thanks @Piratesarereal We've created this [ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1823892) to solve the issue of leaking out the users out of onboarding, and that's been worked on. So now, the caller can use this to show the SUMO page in a custom tab and when the mentioned ticket is done, the caller can also pass in the flag to hide the "Open in Firefox" menu item. @Mugurell wdyt?

LGTM.
We've discussed "offline" and this seems to be a great middle ground approach for cases in which we need to have a modal for displaying webpages.
Not sure if this behavior of hiding that we're using a custom tab should be used for all the other similar scenarios (being able to switch to a full tab which would allow for functionalities like history, bookmarks, collections, sharing has advantages) but in this case where we need to prevent the user exiting the onboarding early (which could be possible through a custom tab's "open in .." menu option) it seems like a nice solution.

PR #1327, opened 2023-03-20
Bug 1823470 - Create pager indicator component

rahulsainani:
This adds the Pager indicator composable that takes in `pagerState` to draw the indicator dots. By default this shows the current page as the active indicator. This is same as the [accompanist pager indicator](https://github.com/google/accompanist/blob/main/pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt#L78), with the only difference being that it takes in  a param `leaveTrail` that shows the previous pages as "active" as well, showing a trail of progress.

### Video
https://user-images.githubusercontent.com/38040960/226432528-bde10da3-e5f9-4f6c-8f28-b643fd7cfb48.mp4

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/compose/HorizontalPagerIndicator.kt
```diff
@@ -0,0 +1,136 @@
...
+ * @param pageCount the size of indicators should be displayed, defaults to [PagerState.pageCount].
+ * If you are implementing a looping pager with a much larger [PagerState.pageCount]
+ * than indicators should displayed, e.g. [Int.MAX_VALUE], specify you real size in this param.
+ * @param pageIndexMapping describe how to get the position of active indicator by the giving page
+ * from [PagerState.currentPage], if [pageCount] is not equals to [PagerState.pageCount].
+ * @param activeColor the color of the active Page indicator
+ * @param inactiveColor the color of page indicators that are inactive. This defaults to
+ * [activeColor] with the alpha component set to the [ContentAlpha.disabled].
+ * @param indicatorWidth the width of each indicator in [Dp].
+ * @param indicatorHeight the height of each indicator in [Dp]. Defaults to [indicatorWidth].
```
HarrisonOg, line 50: small nit: `This defaults to [indicatorWidth]."

rahulsainani: Yes, it's correct, it makes it a square :)

fenix/app/src/main/java/org/mozilla/fenix/compose/HorizontalPagerIndicator.kt
```diff
@@ -0,0 +1,136 @@
...
+import androidx.compose.ui.unit.IntOffset
+import androidx.compose.ui.unit.dp
+import com.google.accompanist.pager.HorizontalPager
+import com.google.accompanist.pager.PagerState
+import kotlin.math.absoluteValue
+import kotlin.math.sign
+
+/**
+ * An horizontally laid out indicator for a [HorizontalPager], that's almost the same as the
+ * accompanist HorizontalPagerIndicator, but with the ability to leave the trail of indicators
```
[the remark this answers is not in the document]
gabrielluong: I can't speak to how we could resolve the license issue. It could escalate to legal due to the nature of things. Taking a step back, it could be easier to just write a simplified more opinionated pager indicator since we don't need all the functionality and customization such as leaving a trail provided by this component. Ideally, the generic components that we do add are aligned with the design system - we would be more opinionated here because we don't need the extra customization parameters to indicate the colors, size or shape since we know all the pager indicator would just have a very specific look according to the design system.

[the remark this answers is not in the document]
rahulsainani: As discussed "offline" updated with the changes to simplify the component, Let me know your thoughts :)

fenix/app/build.gradle
```diff
@@ -460,7 +460,10 @@ nimbus {
 
 tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
     kotlinOptions {
-        freeCompilerArgs += "-opt-in=kotlinx.coroutines.ExperimentalCoroutinesApi"
+        freeCompilerArgs += [
+                "-opt-in=kotlinx.coroutines.ExperimentalCoroutinesApi",
+                "-opt-in=com.google.accompanist.pager.ExperimentalPagerApi",
```
gabrielluong, line 465: Let's opt for the manual `@OptIn` instead of adding a compiler args here. The reason being is that this accompanist library will soon be migrated into the Compose library and it would see easier to me to remove an one liner `@OptIn` than to revert the args here again.

fenix/app/src/main/java/org/mozilla/fenix/compose/PagerIndicator.kt
```diff
@@ -0,0 +1,132 @@
...
+                { it == pagerState.currentPage }
+            }
+
+        repeat(pageCount) {
+            if (conditionForActiveModifier(it)) {
+                Box(activeModifier)
+            } else {
+                Box(inactiveModifier)
+            }
+        }
```
gabrielluong, line 82:
```suggestion
        val showActiveModifier: (pageIndex: Int) -> Boolean =
            if (leaveTrail) {
                { it <= pagerState.currentPage }
            } else {
                { it == pagerState.currentPage }
            }

        repeat(pageCount) {
            Box(modifier = Modifier
                .size(6.dp)
                .background(
                    color = if (showActiveModifier(it) {
                        activeColor                  
                    } else {
                        inactiveColor
                   },
                    shape = CircleShape
                )
            )
        }
```
I think we could simplify this much further since the only thing that changes between these modifier variables is the color so we could avoid the extra variables and if statement checks by inlining. I also think we can just inline the size, spacing and shape here and remove the need of the constants. Part of my thinking here is to really move away from the prior accompanist code.

I think you improve upon my sample suggestion by changing the function to return the indicator color instead of doing the if statement check here.

fenix/app/src/main/java/org/mozilla/fenix/compose/PagerIndicator.kt
```diff
@@ -0,0 +1,132 @@
...
+ * @param pageCount The size of indicators should be displayed, defaults to [PagerState.pageCount].
+ * If you are implementing a looping pager with a much larger [PagerState.pageCount]
+ * than indicators should displayed, e.g. [Int.MAX_VALUE], specify you real size in this param.
+ * @param activeColor The color of the active page indicator, and the color of previous page
+ * indicators in case [leaveTrail] is set to true.
+ * @param inactiveColor The color of page indicators that are inactive.
+ * @param leaveTrail Whether to leave the trail of indicators to show progress.
+ * This defaults to false and just shows the current one as active.
+ */
+
```
gabrielluong, line 48: nit: new line

fenix/app/src/main/java/org/mozilla/fenix/compose/PagerIndicator.kt
rahulsainani: Fair point, updated and that removed the need for constants 👍  Left the if condition as creating a function required passing in the colors as params anyway.

fenix/app/build.gradle
rahulsainani: The [compose foundation pager](https://developer.android.com/jetpack/compose/layouts/pager) is also experimental, I think it would be easier to just update the flag here when we update to compose 1.4, as we will use the pager api.

fenix/app/src/main/java/org/mozilla/fenix/compose/PagerIndicator.kt
```diff
@@ -0,0 +1,128 @@
...
+    activeColor: Color,
+    inactiveColor: Color,
+    leaveTrail: Boolean = false,
+) {
+    Row(
+        modifier = modifier,
+        horizontalArrangement = Arrangement.spacedBy(8.dp),
+        verticalAlignment = Alignment.CenterVertically,
+    ) {
+        val showActiveModifier: (pageIndex: Int) -> Boolean =
```
gabrielluong, line 58: I am curious if we were able to change this function to return the indicator color so we could avoid the additional if statement when we set the color.

gabrielluong (approved):

rahulsainani:
Thanks for the review @gabrielluong !

PR #1329, opened 2023-03-20
Bug 1823482 - collect Ad ID opt outs

rosahbruno:
If the user opts of out an Ad ID, then their Ad ID will be returned as a string of 0's. In this case, we want to treat the ID as if we couldn't fetch an ID and collect no ID.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsUtils.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsUtils.kt
@@ -28,6 +28,10 @@ object MetricsUtils {
     private const val PBKDF2_ITERATIONS = 1000
     private const val PBKDF2_KEY_LEN_BITS = 256
 
+    // The default string returned by the device if the user has opted out of
+    // having an advertising ID.
+    private const val OPT_OUT_DEFAULT_AD_ID = "00000000-0000-0000-0000-000000000000"
+
     /**
      * Possible sources for a performed search.
      */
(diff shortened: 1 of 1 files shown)


gabrielluong:
@rocketsroger I think this is fairly straight change for you to review or help shepherd to completion. Since getAdvertisingID() is testable, I would suggest adding a test to ensure we do get null for this `OPT_OUT_DEFAULT_AD_ID`

rocketsroger:
@rosahbruno how do I reproduce?  I opted out in Settings -> Privacy-> Ads -> Opt out of Ads Personalization in Android 12 and 13 but `AdvertisingIdClient.getAdvertisingIdInfo(context).id` still returns the advertising ID.  Though `AdvertisingIdClient.getAdvertisingIdInfo(context).isLimitAdTrackingEnabled` is returning `true`. 

Should we check `isLimitAdTrackingEnabled` and if it's true then return null?

rosahbruno:
> @rosahbruno how do I reproduce? I opted out in Settings -> Privacy-> Ads -> Opt out of Ads Personalization in Android 12 and 13 but `AdvertisingIdClient.getAdvertisingIdInfo(context).id` still returns the advertising ID. Though `AdvertisingIdClient.getAdvertisingIdInfo(context).isLimitAdTrackingEnabled` is returning `true`.

I was not able to reproduce this on an emulator, I was only able to reproduce running on a real device where I opted out.

> Should we check `isLimitAdTrackingEnabled` and if it's true then return null?

I think having that as an additional check would be a good idea. I can update this to include that check.

rocketsroger:
> I think having that as an additional check would be a good idea. I can update this to include that check.

Thanks!

PR #1345, opened 2023-03-21
[Focus] Bug 1822619: fix crash due to NPE

Alexandra-Virvara:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/focus-android/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
+++ b/focus-android/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
@@ -19,7 +19,6 @@ import android.widget.TextView
 import androidx.core.content.ContextCompat
 import androidx.core.view.isVisible
 import androidx.recyclerview.widget.ItemTouchHelper
-import androidx.recyclerview.widget.ItemTouchHelper.SimpleCallback
 import androidx.recyclerview.widget.LinearLayoutManager
 import androidx.recyclerview.widget.RecyclerView
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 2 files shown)


focus-android/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
```diff
@@ -308,7 +256,7 @@ open class AutocompleteListFragment : BaseSettingsLikeFragment(), CoroutineScope
             handleView.isVisible = isSelectionMode
             handleView.setOnTouchListener { _, event ->
                 if (event.actionMasked == MotionEvent.ACTION_DOWN) {
-                    itemTouchHelper.startDrag(this)
+                    itemTouchHelper?.startDrag(this)
```
Mugurell, line 259: This seems to be the proposed fix for the crash?
In which case we could avoid the refactoring by just checking `isSelectionMode` before setting the `onTouchListener` or before trying to drag the item, similar to [what we do for the click listener](https://searchfox.org/mozilla-mobile/rev/9649a03c3b0044af635ad9d06eb2d3b98866ff1d/firefox-android/focus-android/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt#316-320).

This got me to think about why we don't have the same behavior for the two listeners and seems like we leveraged the `handleView` not  being visible while in selection mode and as such users could not interact with it and so the `OnTouchListener` would not be called.
This behavior of setting the proper visibility for the `handleView` seems to be a regression after https://github.com/mozilla-mobile/firefox-android/commit/08a4accd95245aefee80c037142d8e73ca59f993#diff-50d6207d57d99bc04260668b6c87256cb402dafb8a2c7e67d701fe8b3adb1cd5R308. So if we set the proper visibility back then we'd solve both a UI issue and the crash.

Mugurell (changes requested):
We usually prefer to keep the changes to a minimal to avoid the risk of other regressions and keep the git history clean to easily allow tracking changes.
This seems like quite a big refactoring which could be avoided for the proposed solution but looking further it seems like the fix is even smaller if we concentrate on just the regression from 109, left a comment about that.

Regarding the form of the commits and PRs I would suggest for the future to try to add commit comments explaining the fix and when possible screenshots / videos of the before and after behavior which would help the reviewer, QA and any people after to easily understand what changed.

focus-android/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
```diff
@@ -136,9 +136,7 @@ open class AutocompleteListFragment : BaseSettingsLikeFragment(), CoroutineScope
             adapter = DomainListAdapter()
             setHasFixedSize(true)
 
-            if (!isSelectionMode()) {
-                itemTouchHelper.attachToRecyclerView(this)
-            }
+            itemTouchHelper.attachToRecyclerView(this)
```
Mugurell, line 139: This would fix the crash but regress the original specs and implementation from https://github.com/mozilla-mobile/focus-android/issues/1781 - https://github.com/mozilla-mobile/focus-android/commit/d1803f0ae7a7e8a18851ae52f5ddfbad66fc3d10. Dragging items should not be possible while in selection mode.

Mugurell (changes requested):
If we follow the regression - https://github.com/mozilla-mobile/firefox-android/commit/08a4accd95245aefee80c037142d8e73ca59f993#diff-50d6207d57d99bc04260668b6c87256cb402dafb8a2c7e67d701fe8b3adb1cd5R308 it seems like the only change we need to do is reverse the visibility set at in [AutocompleteListFragment.kt#L308](https://github.com/mozilla-mobile/firefox-android/blob/main/focus-android/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt#L308):
> handleView.isVisible = isSelectionMode

should be
> handleView.isVisible = !isSelectionMode

as we do want the handle view visible and allowing to drag items if not in selection mode.

focus-android/app/src/main/java/org/mozilla/focus/autocomplete/AutocompleteListFragment.kt
Alexandra-Virvara: This was actually a fix for the effect of what had caused the crash. After a longer investigation, I discovered the cause of the problem: the itemTouchHelper was being attached to the recycler view only when selectionMode was false. The handlerView becomes visible only when selectionMode becomes true, so the itemTouchHelper wasn't configured when displaying the list of removable elements (hence the crash when trying to drag the element of the list). 
I have also tested the code before https://github.com/mozilla-mobile/firefox-android/commit/08a4accd95245aefee80c037142d8e73ca59f993#diff-50d6207d57d99bc04260668b6c87256cb402dafb8a2c7e67d701fe8b3adb1cd5R308 and the problem was still there.

Mugurell:
Seeing the stack of 10 commits here I also want to reiterate what https://github.com/mozilla-mobile/firefox-android/pull/1344#pullrequestreview-1353193543 said regarding standardizing the commit/PR naming scheme - the "[Focus]" part should be dropped and I'd also recommend squashing commits and rebasing them on top of `main` locally to keep the git history and changes clean and easy to follow.
If you want still integrate the new changes but use the quicker Github options you can also choose to rebase the PR commits instead of by default merging the new changes through a merge commit.
image

Mugurell (approved):
LGTM.
Thank you!
Don't forget to squash your commits or use the `needs landing (squashed)` label to land.

PR #1374, opened 2023-03-23
Bug 1815013 - Port over the Tabs Tray FAB 

Alexandru2909:
Added FAB to composed tabs tray, business logic remains to be added in [1819196](https://bugzilla.mozilla.org/show_bug.cgi?id=1819196).

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1815013

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
@@ -1,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.tabstray
+
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.padding
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.unit.dp
+import mozilla.components.lib.state.ext.observeAsComposableState
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.compose.button.FloatingActionButton
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Floating action button for tabs tray.
+ *
+ * @param tabsTrayStore [TabsTrayStore] used to listen for changes to [TabsTrayState].
+ */
+@Composable
+fun TabsTrayFab(
+    tabsTrayStore: TabsTrayStore,
+) {
+    val currentPage: Page = tabsTrayStore.observeAsComposableState { state ->
+        state.selectedPage
+    }.value ?: Page.NormalTabs
+    val isSyncing: Boolean = tabsTrayStore.observeAsComposableState { state ->
+        state.syncing
+    }.value ?: false
+    val isInNormalMode: Boolean = tabsTrayStore.observeAsComposableState { state ->
+        state.mode == TabsTrayState.Mode.Normal
+    }.value ?: false
+
+    val icon: Painter
+    val contentDescription: String
+    val label: String?
+
+    when (currentPage) {
+        Page.NormalTabs -> {
+            icon = painterResource(id = R.drawable.ic_new)
+            contentDescription = stringResource(id = R.string.add_tab)
+            label = null
+        }
+
+        Page.SyncedTabs -> {
+            icon = painterResource(id = R.drawable.ic_fab_sync)
+            contentDescription = stringResource(id = R.string.tab_drawer_fab_sync)
+            label = if (isSyncing) {
+                stringResource(id = R.string.sync_syncing_in_progress)
+            } else {
+                stringResource(id = R.string.resync_button_content_description)
+            }.uppercase()
+        }
+
+        Page.PrivateTabs -> {
+            icon = painterResource(id = R.drawable.ic_new)
+            contentDescription = stringResource(id = R.string.add_private_tab)
+            label = stringResource(id = R.string.tab_drawer_fab_content).uppercase()
+        }
+    }
+
+    Box(Modifier.fillMaxSize()) {
+        if (isInNormalMode) {
+            FloatingActionButton(
+                icon = icon,
+                modifier = Modifier
+                    .align(Alignment.BottomEnd)
+                    .padding(16.dp),
+                contentDescription = contentDescription,
+                label = label,
+            ) {}
+        }
+    }
+}
+
+@LightDarkPreview
+@Composable
+private fun TabsTraySyncFabPreview() {
+    val store = TabsTrayStore(
+        initialState = TabsTrayState(
+            selectedPage = Page.SyncedTabs,
+            syncing = true,
+        ),
+    )
+
+    FirefoxTheme {
+        TabsTrayFab(store)
+    }
+}
+
+@LightDarkPreview
+@Composable
+private fun TabsTrayPrivateFabPreview() {
+    val store = TabsTrayStore(
+        initialState = TabsTrayState(
+            selectedPage = Page.PrivateTabs,
+        ),
+    )
+    FirefoxTheme {
+        TabsTrayFab(store)
+    }
+}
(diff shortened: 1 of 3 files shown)


MozillaNoah (commented):
I'm questioning whether it makes more sense to create a one-off `TabsTrayFab` + implementing it via XML vs adding a `FloatingActionButton` directly to the `Box` in `TabsTray` Composable and forgoing any more XML. 

This way, we'd be able to have the FAB in all of the top-level Tabs Tray previews, as well as avoid having to mess with more XML boilerplate.

Alexandru2909:
I believe by adding the FAB directly to the `TabsTray Box` we would also have the FAB move with the `TabsTray` composable, because the `ComposeView` where `TabsTray` lives [uses](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/layout/component_tabstray3.xml#11) the `BottomSheetBehavior`. I'm not sure if we have a method to specify the absolute position for the FAB while having it inside `TabsTray`. To achieve the same result, I think we would need to use `BottomSheetScaffold` for the `TabsTray` composable while still having the FAB outside it. @MozillaNoah do you know any method to position the FAB in the current `TabsTray`, or should we update `TabsTray` to live in a composable `BottomSheet` in this ticket?

MozillaNoah:
> I believe by adding the FAB directly to the `TabsTray Box` we would also have the FAB move with the `TabsTray` composable, because the `ComposeView` where `TabsTray` lives [uses](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/layout/component_tabstray3.xml#11) `BottomSheetBehavior`. I'm not sure if we have a method to specify the absolute position for the FAB while having it inside `TabsTray`. To achieve the same result, I think we would need to use `BottomSheetScaffold` for the `TabsTray` composable while still having the FAB outside it. @MozillaNoah do you know any method to position the FAB in the current `TabsTray`, or should we update `TabsTray` to live in a composable `BottomSheet` in this ticket?

Ugh, I completely forgot about `BottomSheetBehavior` 🤦‍♀️. Yeah, this will be tricky then. Unfortunately, the Composable we'd want to use, `ModalBottomSheet`, is still in [beta](https://developer.android.com/jetpack/androidx/releases/compose-material3#1.1.0-alpha06). I also tried used [`ModalBottomSheetLayout`](https://developer.android.com/reference/kotlin/androidx/compose/material/package-summary#ModalBottomSheetLayout(kotlin.Function1,androidx.compose.ui.Modifier,androidx.compose.material.ModalBottomSheetState,androidx.compose.ui.graphics.Shape,androidx.compose.ui.unit.Dp,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,androidx.compose.ui.graphics.Color,kotlin.Function0)), but I couldn't get it to work properly.

My wish list is to have everything under a singular Composable, but the FAB is not a deal breaker for this. It would (hopefully) be fairly trivial to refactor in May/June once the library we need is officially released. 

There could probably be some weird hack to insert the `BottomSheetBehavior`  into the Compose code, but that's probably not the best path. 

All this is to say, I'm cool with reviewing this PR as-is then. What are your thoughts?

Alexandru2909:
I didn't look too much into inserting the `BottomSheetBehavior` in the Compose code, but I agree having everything in a singular Composable when `ModalBottomSheet` will be stable sounds much better. Until then we could go with the two XMLs from the current implementation.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
```diff
@@ -0,0 +1,109 @@
...
+    tabsTrayStore: TabsTrayStore,
+) {
+    val trayState = tabsTrayStore
+        .observeAsComposableState { state ->
+            TrayState(
+                state.selectedPage,
+                state.syncing,
+                state.mode == TabsTrayState.Mode.Normal,
+            )
+        }.value ?: TrayState()
```
MozillaNoah, line 38: do we benefit from having a wrapper for the selected page, syncing, and browsing mode vs observing these variables individually?

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
```diff
@@ -0,0 +1,109 @@
...
+            R.drawable.ic_fab_sync
+        } else {
+            R.drawable.ic_new
+        },
+    )
+    val contentDescription = stringResource(
+        id = when (trayState.currentPage) {
+            Page.SyncedTabs -> R.string.tab_drawer_fab_sync
+            Page.PrivateTabs -> R.string.add_private_tab
+            else -> R.string.add_tab
```
MozillaNoah, line 50: nit: since this is a `when` over an enum, we can enforce the exhaustive check part of that by having the last condition as `Page.NormalTabs` instead of `else`. (not that I foresee us adding a 4th tab page any time soon).

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
```diff
@@ -0,0 +1,109 @@
...
+    )
+    val label = when (trayState.currentPage) {
+        Page.SyncedTabs -> if (trayState.isSyncing) {
+            stringResource(id = R.string.sync_syncing_in_progress)
+        } else {
+            stringResource(id = R.string.resync_button_content_description)
+        }
+        Page.PrivateTabs -> stringResource(id = R.string.tab_drawer_fab_content)
+        else -> null
+    }?.uppercase()
```
MozillaNoah, line 61: alternatively, this could be a singular `when` based on the current page, and have each val initialized inside. something like:

```suggestion
    val icon: PainterResource
    val contentDescription: String
    val label: String
    
    when (trayState.currentPage) {
            Page.SyncedTabs -> {
                icon = ...
                contentDescription = ...
                label = ...
            }
            Page.PrivateTabs -> {
                icon = ...
                contentDescription = ...
                label = ...
            }
            Page.NormalTabs -> {
                 icon = ...
                contentDescription = ...
                label = ...
             }
    }
```

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
```diff
@@ -0,0 +1,109 @@
...
+@Composable
+fun TabsTrayFab(
+    tabsTrayStore: TabsTrayStore,
+) {
+    val trayState = tabsTrayStore
+        .observeAsComposableState { state ->
+            TrayState(
+                state.selectedPage,
+                state.syncing,
+                state.mode == TabsTrayState.Mode.Normal,
```
MozillaNoah, line 36: nit: add the named parameters

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
```diff
@@ -0,0 +1,109 @@
...
+            stringResource(id = R.string.sync_syncing_in_progress)
+        } else {
+            stringResource(id = R.string.resync_button_content_description)
+        }
+        Page.PrivateTabs -> stringResource(id = R.string.tab_drawer_fab_content)
+        else -> null
+    }?.uppercase()
+
+    Box(Modifier.fillMaxSize()) {
+        if (trayState.isInNormalMode) {
```
MozillaNoah, line 64: does this show/hide behavior line up with the XML version (that it's a simple show/hide when the user exits/enters tab multi-selection)?

MozillaNoah (commented):
First real pass. Have a question about using a state data class wrapper and a couple organizational nits, but this is looking good!

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
Alexandru2909: It aligns, see attached video: 
| XML version | Compose version |
| --- | --- |
| <video src="https://user-images.githubusercontent.com/35462038/227981797-c237cb4d-3c9f-4bf7-90fd-7fc05e4c4b17.mp4" /> | <video src="https://user-images.githubusercontent.com/35462038/227981881-f9584ddf-d515-471c-8dfa-989b11646e2e.mp4" /> |

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFab.kt
Alexandru2909: We don't really, I removed it and added individual variables.

MozillaNoah (approved):
Tested locally and this looks great!!

The FAB text doesn't update yet, but that'll be functional after the tabs tray page selectors are worked on in [Bug 1814991](https://bugzilla.mozilla.org/show_bug.cgi?id=1814991)

:shipit: :shipit: :shipit:

PR #1378, opened 2023-03-23
Bug 1824134 - Add support for clickable linkified text in OnboardingPage

rahulsainani:
### What
– Makes part of description text clickable.
– Also makes `onDismiss` nullable and doesn't show the close button when this is null.

### How
– Add `LinkifiedText` property in `OnboardingPageState`.
– If that is present in the state, `OnboardingPage` uses the information to make the part of the text clickable and linkifies it. An annotated strings is constructed with a tag and the supplied link. 
– `ClickableText` is used and the passed in lambda is only called when the tagged part of the text is clicked.

### Video
https://user-images.githubusercontent.com/38040960/227269770-4ca601f6-0246-4de1-983c-d57765afad48.mp4

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1824134


fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -65,15 +75,19 @@ fun OnboardingPage(
...
+                ) {
+                    Icon(
+                        painter = painterResource(id = R.drawable.mozac_ic_close),
+                        contentDescription = stringResource(R.string.onboarding_home_content_description_close_button),
+                        tint = FirefoxTheme.colors.iconPrimary,
+                    )
+                }
+            } else {
+                Spacer(Modifier)
             }
```
t-p-white, line 91: ❓What does the Spacer do here, is it required?

rahulsainani: It's a dummy composable. Since the parent is using `verticalArrangement = Arrangement.SpaceBetween` and the design is such that the buttons must be at the bottom of the available screen space. This spacer acts as a first element in the parent in the absence of the close button.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -131,6 +143,70 @@ fun OnboardingPage(
...
+    }
+}
+
+/**
+ * A composable for displaying text that has a clickable linkified part.
+ * @param text The complete text.
+ * @param linkifiedText The linkified part of the text.
+ */
+@Composable
+private fun LinkifiedText(
```
t-p-white, line 172: I wonder if having the Model and Func of the same name could be a little ambiguous?

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -131,6 +143,70 @@ fun OnboardingPage(
...
+    text: String,
+    linkifiedText: LinkifiedText,
+) {
+    val annotatedString = buildAnnotatedString {
+        val startIndex = text.indexOf(linkifiedText.text)
+        val endIndex = startIndex + linkifiedText.text.length
```
t-p-white, line 178: Optional:
```suggestion
        val endIndex = text.lastIndexOf(linkifiedText.text)
```

t-p-white (approved):
lgtm 👍 I've approved, but you may want someone else with more Compose experience to also take a look

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
t-p-white: Gotcha, thanks for explaining

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -65,15 +75,19 @@ fun OnboardingPage(
...
+                    modifier = Modifier.align(Alignment.End),
+                ) {
+                    Icon(
+                        painter = painterResource(id = R.drawable.mozac_ic_close),
+                        contentDescription = stringResource(R.string.onboarding_home_content_description_close_button),
+                        tint = FirefoxTheme.colors.iconPrimary,
+                    )
+                }
+            } else {
+                Spacer(Modifier)
```
gabrielluong, line 90: Do we need to actually pass in the `Modifier` here?

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -38,18 +42,24 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
+ * The tag used for links in the text for annotated strings.
+ */
+private const val URL_TAG = "URL_TAG"
+
 /**
  * A composable for displaying onboarding page content.
  *
  * @param pageState [OnboardingPageState] The page content that's displayed.
- * @param onDismiss Invoked when the user clicks the close button.
+ * @param onDismiss Invoked when the user clicks the close button. This defaults to null. When null,
```
gabrielluong, line 54: Could we order this since the parameter orders have swapped?

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -131,6 +143,70 @@ fun OnboardingPage(
...
+            color = FirefoxTheme.colors.textSecondary,
+            textAlign = TextAlign.Center,
+            style = FirefoxTheme.typography.body2,
+        )
+    }
+}
+
+/**
+ * A composable for displaying text that has a clickable linkified part.
+ * @param text The complete text.
```
gabrielluong, line 168: nit: new line between the description and @param

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
gabrielluong, line 172: LinkifiedText sounds a bit strange to me. Could we just call this `LinkText` or `LinkLabel`?

rahulsainani: Yeah, thanks for the link. These are better, updated to `LinkText`.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
rahulsainani: Hmm the edge case with this change would be that `endIndex` could be the `lastIndex` of the subsequent or last occurrence of the string. e.g in "A firefox logo is not really a fox", the part "fox logo is not really a fox" would then be covered by start and end indices, instead of just the first occurrence of "fox". The component doesn't support multiple links but won't error and default to the first occurrence with the existing implementation.

PR #1429, opened 2023-03-27
Bug 1824818 - Update manifest version strings to respect the simple versioning format

willdurand:
~~Not sure whether I got it right, and sorry for repeating the same code
in 3 projects but assuming the first function is correct, the other ones
should also be correct.~~
=> Fixed by #1486 

Examples:

- old/current version string: `113.0.0324141525`
- new version string: `113.0.20230327.185052`

I verified locally that the versions were correctly generated (for
Fenix). That being said, it looks like the previous versions are
"higher" than the new ones, which is potentially a problem until we
update the major or minor version in `version.txt`. Until then, I
_think_ extensions won't be updated.

~~Not sure what we could do, though? Maybe we could bump the minor?~~
=> We do nothing, we'll merge this at the very beginning of a new cycle.

There is a "big" assumption here: `version.txt` will always contain a
version string that has at least 2 parts separated with a dot, e.g.
`110.1`, `110.0`, `110.0a1`, etc.

---

### Pull Request checklist

- ~~[ ] **Tests**: This PR includes thorough tests or an explanation of why it does not~~

Not sure we need a changelog entry, this change impacts built-in (hidden) extensions.

There is no test for the gradle scripts I think? That being said, an error would result in a build failure.

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1824818


JohanLorenzo (approved):
Changes look good to me. I agree it's better to include the year in the version number injected of the builtin extensions we ship in Focus/Fenix. 

I chatted with @willdurand on Element. The current implementation will make nightly extensions have a lower version number than what we already shipped. To circumvent that, we should land this PR as close as possible to the next cycle (114.0a1). Therefore, let's merge this PR on Friday April 7th, 2023. 114.0a1 will start on Monday April 10th. This way, we live with this issue just for a single weekend. The problem will go away once the first 114 nightly goes out.

Amejia481 (approved):
LGTM 👍🏽 
Let's lands as as Johan suggested to avoid issue on Friday April 7th, 2023.

csadilek (approved):
Thank you!

> That being said, it looks like the previous versions are "higher" than the new ones, which is potentially a problem until we update the major or minor version in version.txt

Just a drive-by to point out that afaik we only check for the version number to be different for built-in extensions, we don't seem to check for greater/less than, so it should be OK either way:
https://searchfox.org/mozilla-central/source/mobile/android/modules/geckoview/GeckoViewWebExtension.jsm#669

I agree with the plan to wait for the next cycle though.

csadilek:
Adding a note that https://github.com/mozilla-mobile/firefox-android/pull/1486 will likely land before this now (early next week). I've moved the `updateExtensionVersion` function declaration to the new `shared-settings.gradle` so we now have single declaration and can get rid of the the duplication. So, the patch here can then be applied once, instead of three times.

shared-settings.gradle
```diff
@@ -55,6 +55,22 @@ def setupProject(name, path, description) {
...
+    // each part, and we make sure that there is no letter, e.g. `"0a2"` would
+    // become `"0"`.
+    String[] parts = componentsVersion.split("\\.").collect {
+        part -> part.split("a|b")[0]
+    };
+    // Note the single `H` to avoid leading zeros, which are not allowed.
+    String dateAndTime = new Date().format("YYYYMMdd.Hmmss");
+
+    return "${parts[0]}.${parts[1]}.${dateAndTime}";
+}
```
Amejia481, line 72: ✨

JohanLorenzo:
Today's the 7th. Let's get this landed 🙂

PR #1440, opened 2023-03-28
Bug 1825028 - Delay init of MozillaOnline build until privacy notice is accepted

csadilek:
See ticket for motivation.

We're essentially introducing a new activity for the `MozillaOnline` build so we can show our privacy notice prompt before initializing `FenixApplication` and `HomeActivity`, which in turn would otherwise initialize all our submodules e.g., Nimbus, Glean, Gecko. This change prevents unnecessary initialization.

Verify before landing:
- [x] MozillaOnline build uses new `MozillaOnlineHome` activity, all other builds do NOT.
- [x] Only MozillaOnline build shows privacy notice.
- [x] App initialization starts and completes only after user taps on "Agree and Continue", subsequent launches should not show the prompt. 
- [x] If the user taps on "Disagree and exit", Fenix exits and subsequent launches will continue to display the prompt.
- [x] No sub modules are initialized when notice is shown.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1825028

### DIFF
--- a/fenix/app/build.gradle
+++ b/fenix/app/build.gradle
@@ -31,6 +31,11 @@ android {
         testBuildType project.property("testBuildType")
     }
 
+    // This allows overriding the target activity for MozillaOnline builds, which happens
+    // as part of the defaultConfig below, and applies to all other configurations (Nightly,
+    // Beta, and Release.)
+    def targetActivity = "HomeActivity"
+
     defaultConfig {
         applicationId "org.mozilla"
         minSdkVersion Config.minSdkVersion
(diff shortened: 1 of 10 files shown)


rxumoz (approved):
LGTM

csadilek (commented):
I went through all the test case above and everything worked. Looks ready to land from my side.

jonalmeida (approved):
Likewise, tested this separate one more time for a sanity check. Let's land it then! 🙂

csadilek:
@mergify backport releases_v112

PR #1442, opened 2023-03-28
Bug 1821720 - Add juno onboarding pager ui

rahulsainani:
### How
• Add `JunoOnboardingScreen` composable.
• Add `JunoOnboardingPageType` enum .
•  Add mapper that converts `JunoOnboardingPageType` to `OnboardingPageState` that is expected by `OnboardingPage`. 

### Video
https://user-images.githubusercontent.com/38040960/228324660-20624914-8364-43b1-9500-9d4ac018ba05.mp4

### Next steps 
• Add navigation logic and host fragment
• Add telemetry 

### Next phase steps
• Update mapper to convert Nimbus object to `OnboardingPageState` to use text coming from nimbus. Currently, we use resource strings for phase 1. 

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/JunoOnboardingPageType.kt
```diff
@@ -0,0 +1,14 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.onboarding.view
+
+/**
+ * Model for different types of Onboarding Pages.
+ */
+enum class JunoOnboardingPageType {
```
gabrielluong, line 10: This is just a thought. I think we should have a plan on whether or not we should drop the 'Juno' prefix here. It doesn't have to happen right now, but if this goes beyond just experimentation and becomes the default first run onboarding, then we should ideally drop the these types of project prefixes because they won't provide much context a year or two from when the project phase ended. For example, we have had MR1/MR2 2021 and 2022 but once the feature releases for awhile, the meaning of those major release and the contents within them simply becomes a historical event with no context for anyone to really grasp what went into each release.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/JunoOnboardingScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+    onMakeFirefoxDefaultClick: () -> Unit,
+    onPrivacyPolicyClick: (String) -> Unit,
+    onSignInButtonClick: () -> Unit,
+    grantNotificationPermissionButtonClick: () -> Unit,
+    onFinish: () -> Unit,
+) {
+    val context = LocalContext.current
+    val coroutineScope = rememberCoroutineScope()
+    val pagerState = rememberPagerState()
+    val isSignedIn: State<Boolean?> = context.components.backgroundServices.syncStore
```
gabrielluong, line 64:
```suggestion
    val isSignedIn: State<Boolean?> = context.components.backgroundServices.syncStore
```
There should be an existing `components` getter in a @Composable. https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/samples/compose-browser/src/main/java/org/mozilla/samples/compose/browser/Components.kt#59

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/JunoOnboardingScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+
+@Composable
+@Suppress("LongParameterList")
+private fun JunoOnboardingContent(
+    pagerState: PagerState,
+    scrollToNextPageOrDismiss: () -> Unit,
+    onMakeFirefoxDefaultClick: () -> Unit,
+    onPrivacyPolicyClick: (String) -> Unit,
+    onSignInButtonClick: () -> Unit,
+    grantNotificationPermissionButtonClick: () -> Unit,
```
gabrielluong, line 107:
```suggestion
    onGrantNotificationPermissionButtonClick: () -> Unit,
```
We should ideally just follow our callback "on<X>" convention.

gabrielluong:
Will defer final reviews to another reviewer.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/JunoOnboardingScreen.kt
rahulsainani: Good to know, thanks! updated!

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/JunoOnboardingScreen.kt
```diff
@@ -0,0 +1,179 @@
...
+private fun JunoOnboardingContent(
+    pagerState: PagerState,
+    scrollToNextPageOrDismiss: () -> Unit,
+    onMakeFirefoxDefaultClick: () -> Unit,
+    onPrivacyPolicyClick: (String) -> Unit,
+    onSignInButtonClick: () -> Unit,
+    onNotificationPermissionButtonClick: () -> Unit,
+) {
+    val nestedScrollConnection = remember {
+        object : NestedScrollConnection {
```
t-p-white, line 108: nit: could be extracted to a standalone func making `JunoOnboardingContent` easier to navigate?

```
fun createNestedScrollConnection(pagerState : PagerState) = object : NestedScrollConnection {
    override fun onPreScroll(available: Offset, source: NestedScrollSource): Offset {
    ...
    }
}
       
```

t-p-white (approved):
Really great work Rahul! 🚀 just the single nit from me

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/JunoOnboardingPageType.kt
t-p-white: This is a good point @gabrielluong. I've come across a few legacy project names in the code that are quite confusing without the context, definitely something we should be mindful of. 
@rahulsainani agree that dropping it once it becomes the default is a good way forward - for now I think the prefix is useful for identifying the 'new' onboarding elements

PR #1444, opened 2023-03-28
Bug 1814993 - Add inactive tabs to the composified `TabsTray`

MozillaNoah:
Changes in this PR:
- Refactored away `TabGrid` +`TabList` in favor of just a `TabList` with a boolean to display the tabs as a grid.
- Added an optional `header` Composable to `TabList` so the `InactiveTabList` Composable can be assimilated into the Lazy grid/list.

The preview for `TabsTray` with the `InactiveTabList` is completely functional! Its basic business logic functionality is plumbed-in using stores + remember variables.

https://user-images.githubusercontent.com/87384386/228334596-98005421-07d7-4a4b-abf7-74f64d059503.mov

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1814993


fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -664,6 +680,17 @@ class TabsTrayFragment : AppCompatDialogFragment() {
...
 
+    private fun showInactiveTabsAutoCloseConfirmationSnackbar() {
+        val text = getString(R.string.inactive_tabs_auto_close_message_snackbar)
+        val snackbar = FenixSnackbar.make(
+            view = tabsTrayComposeBinding.root,
+            duration = FenixSnackbar.LENGTH_SHORT,
+            isDisplayedWithBrowserToolbar = true,
+        ).setText(text)
+        snackbar.view.elevation = ELEVATION
+        snackbar.show()
```
MozillaNoah, line 691: I'm going to open a ticket to port this to Compose if this PR is approved

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -232,6 +235,19 @@ class TabsTrayFragment : AppCompatDialogFragment() {
                             tabsTrayInteractor.onMultiSelectClicked(tab, TABS_TRAY_FEATURE_NAME)
                         },
                         onTabLongClick = tabsTrayInteractor::onTabLongClicked,
+                        onInactiveTabsHeaderClick = tabsTrayInteractor::onInactiveTabsHeaderClicked,
+                        onDeleteAllInactiveTabsClick = tabsTrayInteractor::onDeleteAllInactiveTabsClicked,
+                        onInactiveTabsAutoCloseDialogShown = {
+                            TabsTray.autoCloseSeen.record(NoExtras())
+                        },
```
MozillaNoah, line 242: we may want to refactor this later to the interactor/controller

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
gabrielluong: We used to have a Snackbar implementation that @MatthewTighe did for wallpapers https://github.com/mozilla-mobile/fenix/pull/23196, but I believe it got removed since it wasn't used after further changes to wallpapers as part of MR2. If we were to port this again, I think we should also consider https://bugzilla.mozilla.org/show_bug.cgi?id=1814198. Perhaps utilize the same ticket if it makes sense to, you could reset the assignee since I am not actively working on it, but just assigned myself due to my general interest in it and to help me easily find it. 

My understanding of snackbars is that it is shown at the app level so if the tab trays fragment got dismissed, the snackbar should still be displayed for the set duration. That said I think introducing a composable snackbar does require some more thoughts. It might make sense to keep using `FenixSnackbar` in the meantime until we have a comprehensive understanding of things

MozillaNoah: Appreciate your thoughts here. I'll hold off for now then.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -97,42 +128,57 @@ fun TabsTray(
...
+                            }
                         } else {
-                            TabList(
-                                tabs = normalTabs,
-                                onTabClose = onTabClose,
-                                onTabMediaClick = onTabMediaClick,
-                                onTabClick = handleTabClick,
-                                onTabLongClick = onTabLongClick,
-                            )
+                            optionalInactiveTabsHeader = null
```
MatthewTighe, line 160: nit: Switching these cases so that the short null case is first might be it a little easier to read

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -97,42 +128,57 @@ fun TabsTray(
...
-                        if (displayTabsInGrid) {
-                            TabGrid(
-                                tabs = normalTabs,
-                                onTabClose = onTabClose,
-                                onTabMediaClick = onTabMediaClick,
-                                onTabClick = handleTabClick,
-                                onTabLongClick = onTabLongClick,
-                            )
+                        val showInactiveTabsDialog = showInactiveTabsAutoCloseDialog(inactiveTabs.size)
+                        var showAutoCloseDialog by remember { mutableStateOf(showInactiveTabsDialog) }
```
MatthewTighe, line 132: This function is starting to feel a little crowded, and I'm wondering if it would make sense/be possible to move this dialog logic into `InactiveTabsList`. It might allow us to shrink some of these long parameter lists too, but I don't know if this state needs to memoized at this level for this to work

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
MatthewTighe: Or into a middleware 😉

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
```diff
@@ -34,85 +34,86 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
-        ) { tab ->
-            TabListItem(
-                tab = tab,
-                onCloseClick = onTabClose,
-                onMediaClick = onTabMediaClick,
-                onClick = onTabClick,
-                onLongClick = onTabLongClick,
-            )
-        }
+    if (displayTabsInGrid) {
```
MatthewTighe, line 57: It still kinda feels to me like these separate cases should be separate file-private Composables. If it's a PITA though I think we can just leave as is

MatthewTighe (commented):
Throwing a few comments in here for now. What's a good way to try out the inactive tabs dialog workflows?

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -232,6 +235,19 @@ class TabsTrayFragment : AppCompatDialogFragment() {
...
+                        onInactiveTabAutoCloseDialogCloseButtonClick =
+                        tabsTrayInteractor::onAutoCloseDialogCloseButtonClicked,
+                        onEnableInactiveTabAutoCloseClick = {
+                            tabsTrayInteractor::onEnableAutoCloseClicked
```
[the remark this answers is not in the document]
MozillaNoah: 🤦‍♀️ 🤦‍♀️ 🤦‍♀️  Nope lol. I'll fix this; thanks for catching it.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -38,25 +46,48 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
     onTabClick: (TabSessionState) -> Unit,
     onTabMultiSelectClick: (TabSessionState) -> Unit,
     onTabLongClick: (TabSessionState) -> Unit,
+    onInactiveTabsHeaderClick: (Boolean) -> Unit,
+    onDeleteAllInactiveTabsClick: () -> Unit,
+    onInactiveTabsAutoCloseDialogShown: () -> Unit,
+    onInactiveTabAutoCloseDialogCloseButtonClick: () -> Unit,
+    onEnableInactiveTabAutoCloseClick: () -> Unit,
+    onInactiveTabClick: (TabSessionState) -> Unit,
+    onInactiveTabClose: (TabSessionState) -> Unit,
```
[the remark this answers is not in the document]
MozillaNoah: Yeah, and we haven't even done the lambdas for the banners yet 😓. Moving away from the lambdas + interactors could certainly help.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: I'll look into reorganizing it, now that I've figured out a good pattern to use stores + Compose.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
MozillaNoah: It's not biggie to separate them back out.

Are you thinking like:

```
 fun TabLayout() {
    if (displayTabsInGrid) {
       TabGrid()
    } else {
        TabList()
    }
}

private fun TabList() 

private fun TabGrid() 

```

Still keeping the root conditional node with private leafs for the actual layouts?

MozillaNoah:
> What's a good way to try out the inactive tabs dialog workflows?

The inactive tabs dialog shows when you have 20+ inactive tabs, you don't already have auto-close on, and you haven't dismissed the dialog. So force a bunch of tabs inactive (DM if you haven't done this before).

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: so the main trade off I can foresee is that by ditching the parameters from `InactiveTabList` and moving the logic out of `InactiveTabsViewHolder` and `TabsTray`, is that two stores need to be passed-down. I think this will be inevitable if we ditch interactors, but it also requires the previews to instantiate these large objects.

Not a deal breaker by any means, but just some quick thoughts. I'll have to spin off a different ticket to handle this refactor though.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
MatthewTighe: Yeah, that's what I'd been thinking!

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MatthewTighe: Would it make sense to pass the entire Composable in? something like:

```
fun TabsTray(
    ...
    inactiveTabsContent: @Composable () -> Unit
)
```

There's lots of examples in Compose of this kind of "slot API" like in `Button`, `Scaffold`, etc. I think the question is whether this would be an appropriate use case for that kind of thing

MatthewTighe: We could potentially do this for the banner as well, and help alleviate the concerns you shared above about the parameter list getting even larger.

MozillaNoah: I think it'd be strange to include a Composable as a lambda parameter in this situation, when it'd only be serving the purpose of abstracting out some interactor lambdas, since `TabsTray` won't be used anywhere else in Fenix.

I think it's overall a larger architectural problem to solve of interactors + stores + data flow.

What do we pass for responding to user actions?

Do we only have one store to pass in? Do we sacrifice one degree of testing, if the actions are sent directly from the Composable?

How much data do we pass down to each node of the Composable? Just the raw data it needs? An entire store?

MozillaNoah:
Just pushed changes to:

- Split out the grid/list composables into their own functions again.
- Overhauled the preview for `TabsTray`, so it's one parameterized Composable that the actual previews call into.
- Fixed the lambda call for `onEnableAutoCloseClicked`
- Rebase with main

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: Actually, I'd be greatly interested in revisiting this if the [interactor RFC](https://github.com/mozilla-mobile/firefox-android/pull/1466) becomes a thing. It would greatly simplify the lambdas that have to be passed-in, as well as reduce the business logic duplication inside of the previews.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -219,8 +219,11 @@ class TabsTrayFragment : AppCompatDialogFragment() {
             tabsTrayComposeBinding.root.setContent {
                 FirefoxTheme(theme = Theme.getTheme(allowPrivateTheme = false)) {
                     TabsTray(
+                        appStore = requireComponents.appStore,
                         tabsTrayStore = tabsTrayStore,
                         displayTabsInGrid = requireContext().settings().gridTabView,
+                        shouldShowInactiveTabsAutoCloseDialog =
+                        requireContext().settings()::shouldShowInactiveTabsAutoCloseDialog,
```
MatthewTighe, line 226: nit: It looks like there's a tab missing here

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -232,6 +235,19 @@ class TabsTrayFragment : AppCompatDialogFragment() {
                             tabsTrayInteractor.onMultiSelectClicked(tab, TABS_TRAY_FEATURE_NAME)
                         },
                         onTabLongClick = tabsTrayInteractor::onTabLongClicked,
+                        onInactiveTabsHeaderClick = tabsTrayInteractor::onInactiveTabsHeaderClicked,
+                        onDeleteAllInactiveTabsClick = tabsTrayInteractor::onDeleteAllInactiveTabsClicked,
+                        onInactiveTabsAutoCloseDialogShown = {
+                            TabsTray.autoCloseSeen.record(NoExtras())
+                        },
+                        onInactiveTabAutoCloseDialogCloseButtonClick =
+                        tabsTrayInteractor::onAutoCloseDialogCloseButtonClicked,
```
MatthewTighe, line 244: nit: tab here too

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -97,42 +129,56 @@ fun TabsTray(
...
                         }
+
+                        TabLayout(
+                            tabs = normalTabs,
+                            displayTabsInGrid = displayTabsInGrid,
+                            onTabClose = onTabClose,
+                            onTabMediaClick = onTabMediaClick,
+                            onTabClick = handleTabClick,
+                            onTabLongClick = onTabLongClick,
+                            header = optionalInactiveTabsHeader,
```
MatthewTighe, line 170: Ah I hadn't really put it together, but this is `header` parameter is basically what I was suggesting we pass down to the `TabsTray` Composable. `optionalInactiveTabsHeader` would become a parameter instead being put together in this function. Just pointing it out though, I'm still cool with doing it as an optional follow-up later. I suppose it ultimately doesn't matter, but it does feel like that would match up with the pattern of keeping Composables dumb and passing in state/actions.

MatthewTighe (approved):
couple nits but LGTM otherwise

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
MozillaNoah: funnily enough, detekt complained about the tab when I originally had it indented here (and with the subsequent comment) 🤷‍♀️

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: yeah, there's definitely some Goldilocks to be found here, where Composables that are coupled from a feature perspective only get the callbacks/data flow they need and aren't necessarily all coupled to the implementation of a top-level composable.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
MozillaNoah: Opened Bug 1826152

PR #1466, opened 2023-03-29
Bug 1825425 - Add RFC 0009 to remove interactors and controllers

MatthewTighe:
This PR introduces a proposal through RFC to streamline the Fenix architecture by re-focusing entirely on the `lib-state` library and removing interactors and controllers.

https://bugzilla.mozilla.org/show_bug.cgi?id=1825425
https://bugzilla.mozilla.org/show_bug.cgi?id=1825425
https://bugzilla.mozilla.org/show_bug.cgi?id=1825425
https://bugzilla.mozilla.org/show_bug.cgi?id=1825425


Mugurell:
In the presented examples a View can dispatch an action that would trigger the MVI loop and update the app.
But in many other cases we need logic to interact with other parts of the app, trigger functionalities outside of MVI and `State`s. How will these scenarios be treated?
- will `View`s also be responsible with and have all the business logic (path started by Views deciding which `Action`s to dispatch)
- will `View`s only dispatch actions about the user interaction/system events and then reducers will update the state and `Middleware`s will trigger side effects and other functionalities? (in which case middlewares get to replace part of the current `Controller`s functionalties)

MatthewTighe:
> In the presented examples a View can dispatch an action that would trigger the MVI loop and update the app. But in many other cases we need logic to interact with other parts of the app, trigger functionalities outside of MVI and `State`s. How will these scenarios be treated?
> 
>     * will `View`s also be responsible with and have all the business logic (path started by Views deciding which `Action`s to dispatch)
> 
>     * will `View`s only dispatch actions about the user interaction/system events and then reducers will update the state and `Middleware`s will trigger side effects and other functionalities? (in which case middlewares get to replace part of the current `Controller`s functionalties)
> 
> 
> Asking this because the reducer is currently a pure function that can't interact with other app dependencies. Passing maybe an app context from which it could get access to `components` or allowing all `components` to be asked from anywhere such that all reducers would know about anything in the app seems like a dangeorus route.

Yes, reducers should be kept pure. I am imagining that all side-effects would be handled by middleware. I had tried to communicate that with this sentence in the guide-level explanation: `All state changes would be handled by reducers, and side-effects like telemetry or disk writes would be handled in middlewares.` I can update that to make it clear that reducers should be pure functions, and I am also happy to expand on the idea of using middlewares to handle side-effects if we think that this RFC would benefit from greater detail in this area. Would more text explanations suffice, or would examples be useful?

> Having all of the app controller through a `State` seems great but I'm thinking also about scenarios in which users might expect immediate updates, for example when clicking on the menu button to open the menu or tapping to navigate to another screen. Would this user interactions dispatch an action to update the current state with "isMenuOpened = true" and "destination = next_screen"? 

I think these need to be approached on a case-by-case basis. Generally speaking, I don't think we necessarily need to treat every UI-level change (like opening a menu) as part of the state. I think there is nuance to debate here - but 'this menu is currently open' does not need to be considered business logic or part of the app state in every case. This does get a little more complex if we consider adding telemetry to an action like that, however.

We could also try to establish some rules-of-thumb for when to include UI state in the state tree. Off the top of my head, a few suggestions could be:
- we want this state to be cacheable so that it can be recreated on app startup
- we want this state to be exactly reproducible by following a list of dispatched actions
- we want the action that triggers this state change to flow through middlewares (like local logging or remote telemetry)

> Having a big state with multiple middlewares might add some noticeable delay.

I agree that we want to be very careful about performance, and I actually plan to continue discussions about Stores, their performance, and the possibility of combining some Stores. These discussions will follow this RFC and hopefully be a topic during the upcoming work week. 

That said, I think we should avoid pre-optimizing here, and instead track performance metrics closely as we begin refactors towards this goal. 

It should also be noted that this is *not* a proposal to adopt an app-wide single Store as is generally suggested by Redux documentation. I think the likelihood of introducing performance issues when using feature-local Stores is relatively low. We will need to do deeper perf investigations and further design work if we begin to think about using a single Store pattern.

MozillaNoah:
I think this could be a huge boon for the UI code!

Working on the Tabs Tray to Compose rewrite, I've had to explode the API for the root [`TabsTray` Composable](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt#65-82) with lambdas to keep the interactor outside of the UI. It would greatly simplify the Composables if we instead solely relied on dispatching Actions to the appropriate Stores that `TabsTray` depends on. 

image

Moving to 100% Redux in Compose would also greatly benefit the previews. Since the Store/State/Action code is pure Kotlin, the entire data flow pattern can work within previews and allow them to function with all of the business logic necessary to do UI updates. This means that we wouldn't have to do any maintenance of `remember` variables in the previews. Instead, we'd just initialize our stores, ignore adding any middlewares which aren't necessary for UI code (e.g. not adding the middleware for telemetry) and the previews would function the same as when running the full app!

The primary nuance with doing 100% Redux, is as Petru pointed out above. Do we maintain _every_ piece of state inside a Store, stuff that needs to live throughout the lifetime of the app (`AppState.inactiveTabsExpanded` comes to mind), or just mission critical state data? This will likely be a case by case basis for a couple screens/feature flows, but we should come to an agreement on the 90-95% use case.

We'll also need to find a solid way to organize the couple top-level stores we have (e.g. `AppStore` and `BrowserStore`) and the feature-level scope ones (e.g. `TabsTrayStore` and `LocaleSettingsStore`) while consolidating the ways at which the feature level Stores are initialized. TabsTrayStore's initialization has been a minor pain point for the rewrite.

MatthewTighe:
> The primary nuance with doing 100% Redux, is as Petru pointed out above. Do we maintain every piece of state inside a Store, stuff that needs to live throughout the lifetime of the app (AppState.inactiveTabsExpanded comes to mind), or just mission critical state data? This will likely be a case by case basis for a couple screens/feature flows, but we should come to an agreement on the 90-95% use case.
> We'll also need to find a solid way to organize the couple top-level stores we have (e.g. AppStore and BrowserStore) and the feature-level scope ones (e.g. TabsTrayStore and LocaleSettingsStore) while consolidating the ways at which the feature level Stores are initialized. TabsTrayStore's initialization has been a minor pain point for the rewrite.

I've started thinking of the distinction between top-level stores and feature-level stores as 'global' vs 'non-global'. The phrasing ties in to whether the store is accessible through `Components`, in which case it is always in memory. Generally, I think we should aim to have any non-global store should be easily (re)construct-able without having a noticeable performance impact from things like loading a bunch of data from disk or the network. 

`AppStore` is in kind of a weird place. Is it global app state? Is it home screen state? Is it tabs tray state? For your specific example of `inactiveTabsExpanded`, I would think that would be better suited to the `TabsTrayStore`, and that we should consider combining our global stores into a single global store. I did a bit of prototyping around that in the lead up to this RFC and think it should be possible. Non-global stores should be ephemeral to keep memory bloat to a minimum.

If we can achieve goals like that, it would make sense to me to keep as much UI state in Stores as possible - it opens the door for lots of benefits like reproducibility, local logging, and telemetry.

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,193 @@
...
+            .observeAsComposableState { state -> state.pocketStoriesCategories }.value
+        val categoriesSelections = components.appStore
+            .observeAsComposableState { state -> state.pocketStoriesCategoriesSelections }.value
+
+        PocketTopics(
+            categoryColors = categoryColors,
+            categories = categories ?: emptyList(),
+            categoriesSelections = categoriesSelections ?: emptyList(),
+            onCategoryClick = { name -> 
+                components.appStore.dispatch(AppAction.TogglePocketStoriesCategory(name))
```
jonalmeida, line 180: Here, you have made a change to replace the interactor with the dispatch call. Can you add an example of what the other end of the store looks like that connects to this change?

I'm trying to visualize how would features like bookmarks connect to the storage and the snackbar UI layer, or navigate.

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,193 @@
...
+            onCategoryClick = { name -> 
+                components.appStore.dispatch(AppAction.TogglePocketStoriesCategory(name)) 
+            },
+        )
+    }
+}
+```
+
+This should simplify the search for underlying logic by:
+- moving business logic into an expected and consistent component (the reducer)
```
jonalmeida, line 188: Related to the point above, it would help to add examples of what this looks like with a reducer that needs to connect to multiple things that the controllers take care of today.

jonalmeida (commented):
I like this direction! Left some open questions below to maybe reduce some ambiguity. Nice work overall!

jonalmeida:
> For your specific example of `inactiveTabsExpanded`, I would think that would be better suited to the TabsTrayStore.

The UX requirements at the time was to honour the expanded state across every open/close of the tabstray which is why it was there (and was the first state item used to find a reason to create the AppStore! :).

If we move it to the TabsTrayStore, we do not have the ability to know that state when the store is re-created on re-open.

jonalmeida:
One last note (sorry!): I noticed there isn't any concrete "next steps" planned. What happens after this RFC is accepted? Is there a timeline or some place where we can move this forward? 

This isn't part of the RFC process to have this, but it's a nice-to-have something sketched out towards the end. When accepting a change like this, other priorities can sometimes get in the way, so after accepting, we might end up with no changes or partial changes before the effort is lost.

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,193 @@
...
+Generally, the interactor/controller pattern is also intended to move business logic out of classes tied to the Android framework and into classes that are more testable, re-usable, composable, digestible, and which handle a single responsibility. More reading on architecture goals is available [in our Architecture Decisions](../../fenix/docs/Architecture-Decisions.md#goals).
+
+These goals are all met reasonably well by the existing pattern, but it also contributes to unnecessary class explosion, general code complexity, and confusion about responsibility for each architectural type. This becomes especially true when wondering how responsibility should be split between interactors and controllers. It seems that interactors are often included as a matter of precedent and not to necessarily facilitate all the goals mentioned above, and this has lead to a large amount passthrough interactors.
+
+### Proposal goals
+
+1. Increase code comprehensibility
+2. Have clear delineation of responsibility between architectural components
+3. Retain ability to manage state outside of fragments/activities/etc
+4. Ability for proposal to be adopted incrementally
```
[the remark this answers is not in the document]
MatthewTighe: My intent here was to list some of the expected outcomes if we adopted this RFC. I agree that 'the ability to manage state outside of fragments/activities/etc' is not a new problem that this RFC will be solving, which is actually why I didn't list it as a motivation. That said, I thought it was worth calling out that the goal is to not lose anything in terms of a testable, decoupled (from the Android framework) architecture

docs/rfcs/0009-remove-interactors-and-controllers.md
MatthewTighe: I had originally intended for this to not necessarily be a deep dive into Redux patterns since there is so much prior art, but I will add some additional examples for clarity.

MatthewTighe:
> The UX requirements at the time was to honour the expanded state across every open/close of the tabstray which is why it was there (and was the first state item used to find a reason to create the AppStore! :).

I wasn't aware of this requirement, but that makes sense then! I see now that the kdoc for that property explains that as well. 

I will try to add some additional clarity to the distinction between global, session-persistent stores and feature-based, ephemeral ones and how each of those might tie in to disk storage differently.

gabrielluong:
This makes a lot of sense to me and has my blessing. I would like to see some more Middleware examples fleshed out. There's 2 instances that I would like to examine: (1) Middleware for AppStore and (2) Middleware for a fragment Store.

Concretely, I am looking to see what the drawbacks are for our middlewares in both instances. Handling of navigation use case seems fairly straight forward in either case where we would just pass the NavController. I am wondering for other use cases where we might be initializing classes within the fragment or have lambdas that are causing a side effect to the View Binding that lives in the fragment class that are passed into existing controllers. I believe these parameters would be easily passed into (2), but I can imagine this might be harder for where we need these parameters in the (1) global Middleware.

MatthewTighe:
> Concretely, I am looking to see what the drawbacks are for our middlewares in both instances. Handling of navigation use case seems fairly straight forward in either case where we would just pass the NavController. I am wondering for other use cases where we might be initializing classes within the fragment or have lambdas that are causing a side effect to the View Binding that lives in the fragment class that are passed into existing controllers. I believe these parameters would be easily passed into (2), but I can imagine this might be harder for where we need these parameters in the (1) global Middleware.

@gabrielluong  I am happy to work up some more examples. It seems like you've potentially got some concrete instances that you're thinking about here. Do you have anything in particular you'd like to see worked through?

MatthewTighe:
>  I am wondering for other use cases where we might be initializing classes within the fragment or have lambdas that are causing a side effect to the View Binding that lives in the fragment class that are passed into existing controllers. I believe these parameters would be easily passed into (2), but I can imagine this might be harder for where we need these parameters in the (1) global Middleware.

Reading through this more closely, I think I can address the underlying point more directly:

As things are now, any middleware that needed to affect global state or respond to global actions would need to be attached to the global Store during instantiation. We could investigate attaching additional middleware during runtime, but I would assume there are concerns there about action processing order. That is, we would lose any strong guarantees about middlewares being attached by the time actions they are interested in responding to are dispatched. 

Ultimately, I think that there is a strong likelihood that this means that we would need to attach more middlewares to the global store than we would necessarily like. I would be hesitant to make assumptions about any performance impacts without measuring them, however, especially given the (nearly) single-screen nature of our app. 

Longer-term, I would like us to investigate consolidating all of our stores into a single store, essentially moving from a "Flux" model to a "Redux" model. I believe we should be able to create some mechanisms for scoping Feature Stores from the Global Store, such that global and feature-local properties were combined, and the equivalent for actions. This would allow us to simplify the model a bit further but retain the ability to keep our memory footprint low by having the Feature portion of the Store fall out of scope appropriately. It would also mean that we could more easily attach "temporary" middlewares to the single store which would fall out of scope when the feature did.

MatthewTighe:
> but I can imagine this might be harder for where we need these parameters in the (1) global Middleware.

The more I think about this, the less concerned I am that it should represent a blocker in accepting this RFC. To restate what I believe your concern to be: we should be wary of bloating a global store with feature-specific middlewares or state. Some more thoughts on the matter:

- Generally, middleware or state that need to be applied to the global state would represent a situation where that state was required by multiple components.  That seems like an appropriate case to me.
- Global stores and feature stores already co-mingle, so this should be a pattern that is relatively familiar. 
- Feature middlewares could potentially map intercepted actions to global actions and re-dispatch them. 
- Frankly, our memory footprint is already fairly biased towards global components. Though we might lose something some benefit in lazy instantiations, those could still be performed within the middlewares themselves. Additionally, I believe we're likely to find performance wins as we de-dupe code and solidify into a single architectural pattern.

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -188,6 +188,150 @@ This should simplify the search for underlying logic by:
...
+    3. Return early
+3. If the number of categories selected would be over max
+    1. Dispatch an action deselecting the oldest
+4. Dispatch an action with the newly selected category
+5. Record a metric that a category was selected
+
+In order to transform this algorithm into an appropriate usage of the Redux pattern, we can separate this list into two parts: steps that have side-effects and steps which can be represented by pure state updates. In this case, only the steps that record metrics represent side-effects. This will mean that those steps should be moved into a middleware, and the rest can be moved into a reducer.
+
+Here's how that might look:
+```kotlin
```
jonalmeida, line 259: From [my original comment](https://github.com/mozilla-mobile/firefox-android/pull/1466#discussion_r1170798740):
> Here, you have made a change to replace the interactor with the dispatch call. Can you add an example of what the other end of the store looks like that connects to this change?
> 
> I'm trying to visualize how would features like bookmarks connect to the storage and the snackbar UI layer, or navigate.

I might have not been clear, sorry. It wasn't the redux/store part that is hard to visualize. It's the other end of the store where we now have a state change and we need to tell the UI (for example) to update. How will that work?

In the example of Pocket you've used, it falls under the Compose extensions that observe state changes. What about Android View code? We have plenty of that, so how will we observe and react to state changes there? We should be adding an example of that connecting piece more than the internals of how to use `lib-state` which we have plenty of examples in our code base.

If you're looking for ideas: I've tried tackling this problem by using individual [`AbstractBinding`](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/lib/state/src/main/java/mozilla/components/lib/state/helpers/AbstractBinding.kt) classes which observe, collect, and update which ever system needs to know about it. [`TabCounterBinding`](https://github.com/mozilla-mobile/firefox-android/blob/46f3836ae308bd5f050a226c1857363810e62000/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabCounterBinding.kt) is one of many other "binding" classes in the tabs tray package. What I like about these types of classes are that they are small and easy to grok, testable, and not particularly tied to be part of the UI; use them for anything.

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,375 @@
...
+Overall, this should convey the following improvements
+- All state updates are guaranteed to be pure and in a single component, allowing for easy testing and reproduction.
+- All side-effects are logically grouped into middlewares, allowing for testing strategies specific to the type of side-effect.
+- Several indirect abstraction layers are removed, minimizing mental model of code.
+
+---
+
+### Single- vs multi-store patterns and mobile constraints
+
+One big difference between Redux for web and for mobile is that mobile is generally under tighter resource constraints. These constraints are the primary driver for our current usage of `lib-state`, which more closely mirrors the "Flux" pattern instead of the "Redux" pattern. One of the primary differences between the two patterns is simply the number of stores they suggest using. Redux will usually have a single store, while Flux would have multiple.
```
jonalmeida, line 348: > One big difference between Redux for web and for mobile is that mobile is generally under tighter resource constraints. These constraints are the primary driver for our current usage of `lib-state` ...

Huh? Where did this come from? 🙂

jonalmeida: I don't really know what this paragraph adds. I would remove it.

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,375 @@
...
+- All side-effects are logically grouped into middlewares, allowing for testing strategies specific to the type of side-effect.
+- Several indirect abstraction layers are removed, minimizing mental model of code.
+
+---
+
+### Single- vs multi-store patterns and mobile constraints
+
+One big difference between Redux for web and for mobile is that mobile is generally under tighter resource constraints. These constraints are the primary driver for our current usage of `lib-state`, which more closely mirrors the "Flux" pattern instead of the "Redux" pattern. One of the primary differences between the two patterns is simply the number of stores they suggest using. Redux will usually have a single store, while Flux would have multiple. 
+
+Multiple stores allows us to manage memory better. For example, state for Settings will be be automatically collected when a `SettingsStore` falls out of scope, instead of being kept permanently in a global store.
```
jonalmeida, line 350:
```suggestion
Multiple stores allows us to manage memory better. For example, state for a Settings UI screen can be held in a `SettingsStore` which can be cleared from memory (garbage collected) when no longer in that UI context.
```

I think what you're trying to say here is that there is no need to hold screen-specific state in a global store.

jonalmeida (commented):
I am unsure about some of the new sections added in the latest draft, but this is moving forward in a nice direction. 🙂

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,375 @@
...
+        )
+    }
+}
+```
+
+---
+
+### Interacting with the storage layer through middlewares
+
+What if a feature requires some async data for its initial state, or needs to write changes to disk? There are no restrictions on running impure methods in middleware, so we can respond to an actions that would changes to the storage layer from middlewares. A good example is [ContainerMiddleware](../../android-components/components/feature/containers/src/main/java/mozilla/components/feature/containers/ContainerMiddleware.kt)
```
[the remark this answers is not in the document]
MatthewTighe: Ah, good point. I had thought to use a relative link for the exact opposite reason - so that it would reference the up-to-date version. I suppose that doesn't account for the path breaking either though

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,375 @@
...
+
+## Drawbacks
+
+Redux-like patterns have a bit of a learning curve, and some problems can be more difficult to solve than others. For example, handling side-effects like navigation or loading state from disk. The benefits of streamlining our architecture should outweigh this, especially once demonstrative examples of solving these problems are common in the codebase.
+
+---
+
+## Proposal acceptance and future plans
+
+To demonstrate some of the concepts discussed in this proposal in more detail, an example patch will be produced that shows a refactor of a prominent feature of the app - history. Following this, a series of meetings will be hosted to discuss in detail any questions or concerns.
```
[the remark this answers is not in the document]
MatthewTighe: I don't necessarily mean for this to prolong the process. I'll set a deadline for comments once the example becomes available and host those during the time leading up to the deadline. It does seem like that level of detail should be left out of this though, thanks

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,375 @@
...
+
+This proposal is _not_ suggesting that we move to a pure Redux model yet. There are huge wins we can achieve in terms of architectural consistency by only removing interactors/controllers and switching them to a store model. 
+
+This will mean that we will continue to have "global" stores like the `AppStore` and the `BrowserStore` co-existing with "local" stores like the `HistoryFragmentStore`. This can introduce complexity in terms of determining the best place to add state, middlewares, and actions. Generally speaking, prefer using local stores over global stores whenever possible.
+
+Moving into the future, we will investigate options for consolidating our stores. It should be relatively straightforward to at least combine our global stores. Investigation is also ongoing into whether we can create "scoped" stores that combine a local store with the global store such that the local portion of the store can still automatically fall out of scope and respect the constraints of developing on mobile.
+
+---
+
+## Drawbacks
```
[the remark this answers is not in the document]
MatthewTighe: I do think these are worth noting, if only for awareness. Thanks!

docs/rfcs/0009-remove-interactors-and-controllers.md
MatthewTighe: Ah gotcha, thanks for clarifying. On the one hand, it almost seems like if we are investing in the code health of a feature by refactoring it for architectural consistency it would also make sense to me to update that area to Compose - but that probably won't match reality in every case. I'll work up a View example as well

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,374 @@
...
+Following this proposal, the above would change to something like:
+
+```kotlin
+// 1. in LibrarySiteItemView
+setOnClickListener {
+    when {
+        selected.isEmpty() -> store.dispatch(HistoryFragmentAction.AddItemForRemoval(item))
+        ...
+    }
+}
```
Amejia481, line 115: 🙌🏽

Amejia481 (approved):
Awesome job, Really well-detailed 📝 !  
I really like this proposal, many times while writing interactors and controllers had the same feeling.

docs/rfcs/0009-remove-interactors-and-controllers.md
jonalmeida: > ..to update that area to Compose - but that probably won't match reality in every case.

Not all state is tied to UI, it's just one example of something that needs to react to a state change. We can have other systems that need to observe state changes and do other work. Having all of those in middlewares do not always make sense too.

I would use the tabs tray as [another](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt#L61) [example](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/app/src/main/java/org/mozilla/fenix/tabstray/browser/AbstractBrowserTrayList.kt#L35-L37), where we need to observe and react to state but it doesn't always mean a middleware or constructor dependency makes sense.

MatthewTighe: > Not all state is tied to UI, it's just one example of something that needs to react to a state change.

While true, I think the context of thinking about this in terms of interactors/controllers tends to frame it as more of a discussion of UI state.

> We can have other systems that need to observe state changes and do other work. Having all of those in middlewares do not always make sense too.

If we need to react to state changes, that means observing them from somewhere. We could have free-floating global components that observe for the lifetime of the app, or we could have middleware react to the changes. Otherwise, even if they aren't necessarily tied to the UI these components seem like they would at least be tied to the lifecycle of UI components.

> I would use the tabs tray as [another](https://github.com/mozilla-mobile/firefox-android/blob/910afa889ebc5222a1b9a877837de5c55244d441/fenix/app/src/main/java/org/mozilla/fenix/tabstray/browser/BrowserTabsAdapter.kt#L61) [example](https://github.com/mozilla-mobile/firefox-android/blob/910afa889ebc5222a1b9a877837de5c55244d441/fenix/app/src/main/java/org/mozilla/fenix/tabstray/browser/AbstractBrowserTrayList.kt#L35-L37), where we need to observe and react to state but it doesn't always mean a middleware or constructor dependency makes sense.

These are good examples, thank you! But I assume this statement was unrelated to the rest of the comment, since this is UI code? Were you thinking these should just be included as examples in the doc?

jonalmeida: Yeah, fair points. Maybe lets assume they are all UI components: you have Compose covered, so maybe add an example how an observation to update and Android View would work, and that should be enough. 🙂

jonalmeida: I see you went with using the `AbstractBinding` direction for connecting Android Views with store observations - that's fair. FWIW, just stating it in this doc like the Compose example would have sufficed. 🙂

jonalmeida (approved):
LGTM, thanks for writing this up. 🙂

MozillaNoah (approved):

docs/rfcs/0009-remove-interactors-and-controllers.md
```diff
@@ -0,0 +1,401 @@
...
+
+There are already some existing mechanisms for doing this included in the lib state library:
+
+1. `Fragment.consumeFrom` will setup a Store observation that is triggered on _every_ state update in the Store. The [HistoryFragment updating its view hierarchy with all state updates](https://github.com/mozilla-mobile/firefox-android/blob/910afa889ebc5222a1b9a877837de5c55244d441/fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt#L196) is one example.
+2. `Flow.ifChanged` can be combined with things like `Fragment.consumeFlow` or `Store.flowScoped` for finer-grained state observations. This will allow reactions to specific state properties updates, like when the [HistorySearchDialogFragment changes the visibility of the AwesomeBar](https://github.com/mozilla-mobile/firefox-android/blob/910afa889ebc5222a1b9a877837de5c55244d441/fenix/app/src/main/java/org/mozilla/fenix/library/history/HistorySearchDialogFragment.kt#L189).
```
gabrielluong, line 151: I think this might need to be updated when https://github.com/mozilla-mobile/firefox-android/pull/2479 lands

PR #1470, opened 2023-03-30
Bug 1825453 - Use the same Gradle jvmargs for AC as Fenix

rvandermeulen:
https://bugzilla.mozilla.org/show_bug.cgi?id=1825453
https://bugzilla.mozilla.org/show_bug.cgi?id=1825453
https://bugzilla.mozilla.org/show_bug.cgi?id=1825453

### DIFF
--- a/android-components/gradle.properties
+++ b/android-components/gradle.properties
@@ -9,7 +9,7 @@
 
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx4096m -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC
+org.gradle.jvmargs=-Xmx7g -Xms2g -XX:MaxMetaspaceSize=6g -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC
 org.gradle.daemon=true
 
 # When configured, Gradle will run in incubating parallel mode.


android-components/gradle.properties
```diff
@@ -9,7 +9,7 @@
 
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
-org.gradle.jvmargs=-Xmx4096m -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC
```
jonalmeida, line 12: I think we should be keeping the `UseParallelGC`. Seems useful to have it in CI from what I've read so far:

> The parallel collector (also referred to here as the throughput collector) is a generational collector similar to the serial collector; the primary difference is that multiple threads are used to speed up garbage collection.

https://docs.oracle.com/javase/8/docs/technotes/guides/vm/gctuning/parallel.html

> The parallel scavenge collector (Enabled using -XX:UseParallelGC). This is like the previous parallel copying collector, but the algorithm is tuned for gigabyte heaps (over 10GB) on multi-CPU machines. This collection algorithm is designed to maximize throughput while minimizing pauses.

https://www.javaperformancetuning.com/news/qotm026.shtml

jonalmeida: Ah the `HeapDumpOnOutOfMemoryError` is also what helps us know when CI runs out of memory (like in the situations that we're currently seeing).

> The -XX:+HeapDumpOnOutOfMemoryError command-line option tells the HotSpot VM to generate a heap dump when an allocation from the Java heap or the permanent generation cannot be satisfied. There is no overhead in running with this option, and so it can be useful for production systems where OutOfMemoryError takes a long time to surface.

https://www.oracle.com/java/technologies/javase/clopts.html#gbzrr

jonalmeida: So probably?

```
org.gradle.jvmargs=-Xmx7g -Xms2g -XX:MaxMetaspaceSize=6g -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC
```

rvandermeulen: I feel like we dropped those for Fenix because they were the defaults already anyway.

jonalmeida (approved):

android-components/gradle.properties
jonalmeida: I hope so! If it's the default, then there is no harm in keeping them for now.

PR #1482, opened 2023-03-30
Bug 1821005 - Upgrade Gradle to 8.3 version and Android Gradle Plugin to 8.0.2

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1821005
https://bugzilla.mozilla.org/show_bug.cgi?id=1849833


android-components/components/support/base/build.gradle
```diff
@@ -5,7 +5,7 @@
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
-def generatedSrcDir = new File(buildDir, "generated/components/src/main/java")
+def generatedSrcDir = new File(layout.buildDirectory.get().asFile, "generated/components/src/main/java")
```
gbrownmozilla, line 8: Just out of curiosity: Wouldn't `layout.buildDirectory` work just as well here?

fenix/app/src/main/AndroidManifest.xml
```diff
@@ -43,7 +47,6 @@
     <application
         android:name=".FenixApplication"
         android:allowBackup="false"
-        android:extractNativeLibs="true"
```
gbrownmozilla, line 46: I see this is now useLegacyPackaging, https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs - lgtm.

gbrownmozilla (approved):
I don't know much about Gradle, but this works for me and looks just fine. Thanks!

android-components/components/support/base/build.gradle
mcarare: That would throw a runtime exception as` buildDirectory` is a `DirectoryProperty` and there is no `File` constructor that takes that combination of arguments.

android-components/samples/browser/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,10 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
+    <uses-feature
```
jonalmeida, line 8: Leaving a comment here, but it applies to all the other apps in this commit:

So I think this is not enough for us.

We use the camera permission for the front/back/external cameras.

What I'm reading on [the docs here](https://developer.android.com/guide/topics/manifest/uses-feature-element#hw-features), is that this permission is only for the back camera.

We either need to use `android.hardware.camera.any` instead, or we need to add three separate `uses-feature` for `android.hardware.camera`, `android.hardware.camera.front`, and `android.hardware.camera.external`. It's not clear to me from this doc at least which one we need.

jonalmeida, line 8: Should this be in the library module that users these hardware features instead?

fenix/gradle.properties
```diff
@@ -27,3 +27,5 @@ libUrl=https://github.com/mozilla-mobile/firefox-android/tree/main/fenix
 libVcsUrl=https://github.com/mozilla-mobile/firefox-android.git
 libLicense=MPL-2.0
 libLicenseUrl=https://www.mozilla.org/en-US/MPL/2.0/
+android.nonTransitiveRClass=false
+android.nonFinalResIds=false
```
jonalmeida, line 31: > From Gradle 8 the default value has changed from false to true.

Does this mean that we are failing to use non-transitive R classes and non-final res Ids? Can we file a follow-up to fix this if that's true?

jonalmeida (commented):
Thanks for separate commits and comments, it made this review very easy and less intimidating!

Left some open questions, but once addressed, we could land this.

fenix/build.gradle
```diff
@@ -175,6 +175,19 @@ allprojects {
     }
 }
 
+subprojects {
+    afterEvaluate {
+        if (it.hasProperty('android')) {
```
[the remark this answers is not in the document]
mcarare: Because otherwise, building projects that do not have the `android` property would throw a `MissingMethodException`, as the extension does not exist in those types of Gradle projects.

fenix/gradle.properties
mcarare: Yes, we don't use them in Fenix or Focus. Moving to non-transitive R classes and non-final res Ids implies changing hundreds of files, so I thought it would be best not to do it here. And we still should decide if we want to.

android-components/samples/browser/src/main/AndroidManifest.xml
mcarare: This was added because the manifest specifies a permission that [implies](https://developer.android.com/topic/arc/manifest#implied-features) a feature requirment. Meaning that without specifying that the feature is not required, it would imply that it is required, and the app would not be able to be found on devices without that specific hardware (in our case Chromebook).
We could also suppress the lint warning but this seems more clean.

mcarare: I originally thought so myself, but I preferred to add it to the app manifests as the permission was declared directly in the manifest, not in the merged one; so for visibility sake.

fenix/gradle.properties
jonalmeida: Sounds good, let's file a bug then so we can track if we want to do it or not. I'm not sure what either direction implies right now.

android-components/samples/browser/src/main/AndroidManifest.xml
jonalmeida: > Meaning that without specifying that the feature is not required, it would imply that it is required, and the app would not be able to be found on devices without that specific hardware (in our case Chromebook).

Even then, we would need to say that we need (non-required) hardware to all cameras in web content, right?

fenix/gradle.properties
mcarare: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1863329 and https://bugzilla.mozilla.org/show_bug.cgi?id=1863330

android-components/samples/browser/src/main/AndroidManifest.xml
mcarare: Added a comment and commit message to make this more clear for future references.

jonalmeida (approved):

PR #1486, opened 2023-03-30
Bug 1825649 - Introduce shared-settings.gradle to allow reusing build logic.

csadilek:
This allows code re-use in our Gradle files, which currently have quite a bit of code duplication. For now, I just unified our settings.gradle files, but as a next step we can define shared functions in this new `shared-setting.gradle` e.g., for the `getManifestVersionString` function introduced in https://github.com/mozilla-mobile/firefox-android/pull/1429/.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1825649

### DIFF
--- a/shared-settings.gradle
+++ b/shared-settings.gradle
@@ -1,0 +1,104 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import org.yaml.snakeyaml.Yaml
+
+buildscript {
+    repositories {
+        mavenCentral()
+    }
+
+    dependencies {
+        classpath 'org.yaml:snakeyaml:1.33'
+    }
+}
+
+// Synchronized library configuration for all modules
+// This "componentsVersion" number is defined in "version.txt" and should follow
+// semantic versioning (MAJOR.MINOR.PATCH). See https://semver.org/
+class Config {
+
+    public final String componentsVersion
+    public final String componentsGroupId
+    public final Integer compileSdkVersion
+    public final Integer minSdkVersion
+    public final Integer targetSdkVersion
+
+    Config(
+            String componentsVersion,
+            String componentsGroupId,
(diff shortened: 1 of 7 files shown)


shared-settings.gradle
```diff
@@ -0,0 +1,57 @@
+import org.yaml.snakeyaml.Yaml
```
gabrielluong, line 1: Let's add the license here.

csadilek: OK, Just saw our build scripts actually don't have one either, but I added it as the other settings.gradle files have one.

gabrielluong (approved):

shared-settings.gradle
```diff
@@ -0,0 +1,103 @@
...
+            project.ext.description = description
+        }
+    }
+}
+
+def yaml = new Yaml()
+def buildconfig = yaml.load(new File(rootDir, '../android-components/.buildconfig.yml').newInputStream())
+
+buildconfig.projects.each { project ->
+    // If we're building A-C let's setup all project, otherwise exclude samples e.g., if we're building Fenix or Focus.
```
gabrielluong, line 62:
```suggestion
    // If we're building A-C, setup all projects including samples. Otherwise, exclude samples e.g., if we're building Fenix or Focus.
```
I think I needed this extra clarity to fully spell it out for me.

shared-settings.gradle
```diff
@@ -0,0 +1,103 @@
...
+                rename { 'manifest.json' }
+                into extDir
+
+                def values = ['version': rootProject.ext.config.componentsVersion.split("a|b")[0] + "." + new Date().format('MMddHHmmss')]
+                inputs.properties(values)
+                expand(values)
+            }
+        }
+    }
+}
```
gabrielluong, line 103: nit: add a new line to EOF

gabrielluong (approved):

shared-settings.gradle
gabrielluong: Ya, I know we added this change, but for whatever reason when I was just comparing it all, I needed it the extra wording for it to fully click.

titooan (approved):

PR #1493, opened 2023-03-31
Bug 1822268 – Juno onboarding fragment and navigation logic

rahulsainani:
### What
– Add `JunoOnboardingFragment` that passes in `List<JunoOnboardingPageType>` to the `JunoOnboardingScreen` composable. The notification permission page is only part of the list for android 13 devices where the notification permission is not yet granted.
– Adds navigation logic in `HomeActivity`. Navigate to `JunoOnboardingFragment` if the user has not yet seen the onboarding and is in the experiment. 
– Tested on beta with a draft nimbus experiment.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1822268


fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -363,15 +364,28 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         StartupTimeline.onActivityCreateEndHome(this) // DO NOT MOVE ANYTHING BELOW HERE.
     }
 
+    private fun navigateToHomeOrJunoOnboarding() {
+        if (!settings().isJunoOnboardingShown) {
+            FxNimbus.features.junoOnboarding.recordExposure()
+            if (settings().junoOnboardingEnabled) {
```
t-p-white, line 370: Should we have the`junoOnboardingEnabled` check first, else we record the exposure when a user is not enrolled in Juno Onboarding?

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -363,15 +364,28 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+                logger.debug("navigating to juno onboarding.")
+                navHost.navController.navigate(NavGraphDirections.actionGlobalHomeJunoOnboarding())
+            } else {
+                logger.debug("navigating to home, existing users or not in experiment.")
```
t-p-white, line 374: nit: 
```suggestion
                logger.debug("navigating to home, existing user or not in experiment.")
```

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -387,6 +401,12 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         }
     }
 
+    private fun shouldExposeToNotificationPrePromptExperiment() =
```
t-p-white, line 404: 👍

t-p-white (approved):
Looks great! My main comment here is around the `navigateToHomeOrJunoOnboarding` if-else logic

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
t-p-white: Gotcha, thanks for the info 🙏

rahulsainani: Also, updated the top condition to only show juno onboarding when the user clicks on the app icon. In the super rare edge case that the new user makes firefox their default browser and opens the app through a url, then the usual behavior will occur, as opposed to showing juno onboarding and then forgetting the url user clicked. cc @rocketsroger

rocketsroger (commented):
Got a few questions.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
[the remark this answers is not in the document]
rahulsainani: Yeah, that's the condition we added, extracted to a function as `detekt` detected that it was a [complex condition](https://detekt.dev/docs/rules/complexity/#complexcondition) (4 or more combined checks). 

Hmm do you mean just keeping `!settings().junoOnboardingEnabled && !NotificationManagerCompat.from(applicationContext).areNotificationsEnabledSafe()`? Wouldn't that be true when user is not in the experiment and is on android 12 or below? 

Saying that, would prefer to keep it like this for now and update it to simplify it, if required in a separate PR.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -269,7 +270,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         // Unless the activity is recreated, navigate to home first (without rendering it)
         // to add it to the back stack.
         if (savedInstanceState == null) {
-            navigateToHome()
+            navigateToHomeOrJunoOnboarding()
```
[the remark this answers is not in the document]
rahulsainani: Good point! updated it! This simplifies the nav calls. It also allowed to move the function to `Settings.kt` and add tests for it.

rocketsroger (approved):
LGTM 🚢

fenix/app/src/main/res/navigation/nav_graph.xml
```diff
@@ -29,6 +29,18 @@
...
+        android:id="@+id/action_global_home_juno_onboarding"
+        app:destination="@id/junoOnboardingFragment"
+        app:popUpTo="@id/startupFragment"
+        app:popUpToInclusive="true" />
+
+    <action
+        android:id="@+id/action_onboarding_home"
+        app:destination="@id/homeFragment"
+        app:popUpTo="@id/junoOnboardingFragment"
+        app:popUpToInclusive="true" />
```
jonalmeida, line 42: @rahulsainani @rocketsroger folks, there are some oddities here that I ran into when trying to debug a completely different issue which crossed into this Juno work.

* `action_global_home_juno_onboarding` this naming is confusing - if you look at the examples around this file, we name the action to where it is going to avoid misinterpretting the direction as "home juno onboarding" which could be HomeActivity or HomeFragment. So a better name would be `action_global_juno_onboarding`.
* `action_onboarding_home` is also confusing for the same reasons above. It reads (via the JunoOnboardingFragment code) that we are at the onboarding fragment and now going to the onboarding home direction. From what I understand, it looks like we want to go to the HomeFragment and remove the onboarding fragment from our backstack. Further, it is missing the "global" namespace which is what helps read the generated code better in the kotlin side.
* For both these actions, why do they need to be _global_ actions? 
  * `action_global_home_juno_onboarding` can be a action under the `junoOnboardingFragment` fragment declaration. No other fragment needs to be able to navigate to it, if I understand correctly (please correct me if I'm wrong).
  * `action_onboarding_home` might be unnecessary, I think? We have a global direction to go home right above called `action_global_home`, so maybe that is enough. I'm not certain though if that will do the same popUpTo action as intended by this action. If not, please do the same as above and remove it from the global actions to be a fragment direction.

Let me know if you'd like me to file a bug for the above mentioned as well. There are examples in this file for some of the recommendations.

Please take time to discuss and understand the implications of a change before merging it. Pushing back deadlines is less work that tech debt that can lead to bugs.

rahulsainani: I've created this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1830083) here to track and noted down my initial findings.

PR #1535, opened 2023-04-04
Bug 1794077/1780350/1820404 Persist Readerview position

jackyzy823:
1.  Keep Readerview's position between active and inactive
2.  Persiste Readerview's position between restart fenix

Note:  Value of position (scrollY) is accurate, however after fenix restart , the Readerview may not remain in same place ( a bit above). This may be caused by re-generating DOM. There's no such problem when toggling between active and inactive.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1820404

### DIFF
--- a/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/ReaderState.kt
+++ b/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/ReaderState.kt
@@ -16,6 +16,8 @@ package mozilla.components.browser.state.state
  * content script is required.
  * @property baseUrl the base URL of the reader view extension page.
  * @property activeUrl the URL of the page currently displayed in reader view.
+ * @property scrollY the vertical scroll position of the page currently
+ * displayed in reader view.
  */
 data class ReaderState(
     val readerable: Boolean = false,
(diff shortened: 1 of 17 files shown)


Amejia481:
@jackyzy823 👋🏽 ,
Thanks for the patch!

I think we can't have multiple bug numbers in the commit message e.g `Bug 1794077/1780350`, I think we should target one, you could mention in the main bug the other bugs will be address when the main one lands.

jackyzy823:
@Amejia481  updated.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/ReaderState.kt
```diff
@@ -24,4 +24,5 @@ data class ReaderState(
     val connectRequired: Boolean = false,
     val baseUrl: String? = null,
     val activeUrl: String? = null,
+    val scrollY: Int? = null,
```
Amejia481, line 27: Nit: Lets add docs for this new property :)

Amejia481 (approved):
This looks good!
Just a small nit :)

PR #1547, opened 2023-04-04
Bug 1819194 - Open `TabsTray` to the initial page in `TabsTrayState`

MozillaNoah:
The changes to `TabCounterView` are because the navigation arguments were not being set to `Private` when opening the Tabs Tray via the awesomebar. The XML version is only opening the Private page because the intermediary object, `TabLayoutMediator`, was taking care of this via `TabsTrayAction.PageSelected` instead. 

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1819194


fenix/app/src/main/java/org/mozilla/fenix/home/TabCounterView.kt
```diff
@@ -63,7 +64,12 @@ class TabCounterView(
 
             navController.nav(
                 R.id.homeFragment,
-                HomeFragmentDirections.actionGlobalTabsTrayFragment(),
+                HomeFragmentDirections.actionGlobalTabsTrayFragment(
+                    page = when (browsingModeManager.mode) {
+                        BrowsingMode.Normal -> Page.NormalTabs
+                        BrowsingMode.Private -> Page.PrivateTabs
+                    },
```
MatthewTighe, line 71: This made me do a bit of thinking, so I'm finishing what I wrote for posterity and conversation. I don't think we need to make any changes to this for now.

Just so I understand - this is required so that the `TabsTrayStore` has the correct mode in the `selectedPage` property? That seems like it represents a slightly larger issue, that the `selectedPage` property and the `AppState.mode` and the various other places (😮‍💨)  mode is managed can get out of sync. 

Part of me is wondering if there's any real benefit to tracking the selected page state inside the `TabsTrayStore`. If we only care about current mode during initialization of the tabs tray, it seems like we could just handle that state directly in the Composable. I think this begs a couple questions:

1. Should every piece of UI State be tracked in a Store? Idk if this is a fully settled question, but the answer does kinda seem like yes.
2. What is the lifetime of the `TabsTrayStore`. If it were re-initialized every time the tabs tray was opened, then we could rely on it properly reading the `AppState.mode` and the two states could not functionally get of of sync.

Ultimately, I think we should be trying to build the habits of using Stores where possible. This had led me to wonder whether it's better to extract the current mode from existing state during creation of the Tabs Tray. Something like:

```
val initialPage = when (args.page) {
    // sync -> sync
    // private -> private
    // normal -> if (mode == private) private else normal
}
```

but that is honestly more hacky. In the future I imagine we would want navigation to be handled by a Middleware, in which case it should be packaging nav args similar to what you're doing here. It makes me wonder how we could test the initialization flows of Stores a little better in the future, as the (nav middleware) -> (tabs tray initialization) flow seems like it would likely continue to live in a fragment.

MatthewTighe (approved):
One stream-of-consciousness rant aside, I think this looks good for now!

fenix/app/src/main/java/org/mozilla/fenix/home/TabCounterView.kt
MozillaNoah: This is required _just_ for the scenario where the tab counter button is pressed to open the Tabs Tray. There's a couple paths to opening `TabsTrayFragment`, and 95% of them just use the default values to open the Tabs Tray on the normal tabs page. There's one exception: recent synced tabs opens the Tabs Tray to the Synced Tabs. 

So in the primary user flow of tab counter button -> Tabs Tray, the private page was not being set and instead relied on a Store action for this to be the initial page.

TabsTrayStore is only created and managed while the Tabs Tray is alive, so I think we could definitely take another look whether or not `selectedPage` needs to be maintained in a formal store/state (outside of maybe weird lifecycle gotchas). 

1. I think we'll need to re-orient ourselves around how much statefulness lives inside stores and how much can just be relegated to the composables. It likely also depends on the store itself, as ones like `AppStore` and `BrowserStore` will probably have different implications than feature dependent ones like `TabsTrayStore` (which ultimately creates more degrees of nuance).
2. `TabsTrayStore` is only initialized/maintained during the scope/lifetime of `TabsTrayFragment`

A navigation middleware sounds pretty interesting though!

PR #1566, opened 2023-04-05
Bug 1821475 - Show research surface when survey message is available

rocketsroger:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1821475

### DIFF
--- a/fenix/app/nimbus.fml.yaml
+++ b/fenix/app/nimbus.fml.yaml
@@ -33,6 +33,7 @@ import:
               DAY_3_AFTER_INSTALL: days_since_install == 3
               DAY_4_AFTER_INSTALL: days_since_install == 4
               DAY_5_AFTER_INSTALL: days_since_install == 5
+              MORE_THAN_24H_SINCE_INSTALLED_OR_UPDATED: days_since_update >= 1
 
               # Using custom attributes for the browser
               I_AM_DEFAULT_BROWSER: "is_default_browser"
@@ -86,7 +87,7 @@ import:
                 max-display-count: 5
               SURVEY:
                 priority: 55
-                max-display-count: 10
+                max-display-count: 1
               PERSISTENT:
                 priority: 50
                 max-display-count: 20
--- a/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -41,6 +41,7 @@ import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.NavigationUI
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
@@ -72,6 +73,7 @@ import mozilla.components.support.ktx.android.content.share
 import mozilla.components.support.ktx.kotlin.isUrl
 import mozilla.components.support.ktx.kotlin.toNormalizedUrl
 import mozilla.components.support.locale.LocaleAwareAppCompatActivity
+import mozilla.components.support.utils.BootUtils
 import mozilla.components.support.utils.ManufacturerCodes
 import mozilla.components.support.utils.SafeIntent
 import mozilla.components.support.utils.toSafeIntent
@@ -85,9 +87,11 @@ import org.mozilla.fenix.addons.AddonPermissionsDetailsFragmentDirections
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
 import org.mozilla.fenix.browser.browsingmode.BrowsingModeManager
 import org.mozilla.fenix.browser.browsingmode.DefaultBrowsingModeManager
+import org.mozilla.fenix.components.appstate.AppAction
 import org.mozilla.fenix.components.metrics.BreadcrumbsRecorder
 import org.mozilla.fenix.databinding.ActivityHomeBinding
 import org.mozilla.fenix.exceptions.trackingprotection.TrackingProtectionExceptionsFragmentDirections
+import org.mozilla.fenix.experiments.ResearchSurfaceDialogFragment
 import org.mozilla.fenix.ext.alreadyOnDestination
 import org.mozilla.fenix.ext.areNotificationsEnabledSafe
 import org.mozilla.fenix.ext.breadcrumb
@@ -110,6 +114,8 @@ import org.mozilla.fenix.library.bookmarks.DesktopFolders
 import org.mozilla.fenix.library.history.HistoryFragmentDirections
 import org.mozilla.fenix.library.historymetadata.HistoryMetadataGroupFragmentDirections
 import org.mozilla.fenix.library.recentlyclosed.RecentlyClosedFragmentDirections
+import org.mozilla.fenix.messaging.FenixMessageSurfaceId
+import org.mozilla.fenix.messaging.FenixNimbusMessagingController
 import org.mozilla.fenix.messaging.MessageNotificationWorker
 import org.mozilla.fenix.nimbus.FxNimbus
 import org.mozilla.fenix.onboarding.FenixOnboarding
@@ -271,6 +277,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
                 navHost.navController.navigate(NavGraphDirections.actionGlobalHomeJunoOnboarding())
             }
         } else {
+            lifecycleScope.launch(IO) {
+                showFullscreenMessageIfNeeded(applicationContext)
+            }
+
             // Unless the activity is recreated, navigate to home first (without rendering it)
             // to add it to the back stack.
             if (savedInstanceState == null) {
@@ -1167,6 +1177,49 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     private fun shouldNavigateToBrowserOnColdStart(savedInstanceState: Bundle?): Boolean {
         return isActivityColdStarted(intent, savedInstanceState) &&
             !processIntent(intent)
+    }
+
+    private suspend fun showFullscreenMessageIfNeeded(context: Context) {
+        val messagingStorage = context.components.analytics.messagingStorage
+        val messages = messagingStorage.getMessages()
+        val nextMessage =
+            messagingStorage.getNextMessage(FenixMessageSurfaceId.SURVEY, messages)
+                ?: return
+
+        val fenixNimbusMessagingController = FenixNimbusMessagingController(messagingStorage)
+        val researchSurfaceDialogFragment = ResearchSurfaceDialogFragment.newInstance(
+            keyMessageText = nextMessage.data.text,
+            keyAcceptButtonText = nextMessage.data.buttonLabel,
+            keyDismissButtonText = null,
+        )
+
+        researchSurfaceDialogFragment.onAccept = {
+            processIntent(fenixNimbusMessagingController.getIntentForMessage(nextMessage))
+            components.appStore.dispatch(AppAction.MessagingAction.MessageClicked(nextMessage))
+        }
+
+        researchSurfaceDialogFragment.onDismiss = {
+            components.appStore.dispatch(AppAction.MessagingAction.MessageDismissed(nextMessage))
+        }
+
+        lifecycleScope.launch(Main) {
+            researchSurfaceDialogFragment.showNow(
+                supportFragmentManager,
+                ResearchSurfaceDialogFragment.FRAGMENT_TAG,
+            )
+        }
+
+        // Update message as displayed.
+        val currentBootUniqueIdentifier = BootUtils.getBootIdentifier(context)
+        val updatedMessage =
+            fenixNimbusMessagingController.updateMessageAsDisplayed(
+                nextMessage,
+                currentBootUniqueIdentifier,
+            )
+
+        fenixNimbusMessagingController.onMessageDisplayed(updatedMessage)
+
+        return
     }
 
     companion object {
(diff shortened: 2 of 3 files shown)


fenix/app/src/main/java/org/mozilla/fenix/experiments/view/ResearchSurfaceSurvey.kt
```diff
@@ -61,50 +99,52 @@ fun ResearchSurfaceSurvey(
     onDismiss: () -> Unit,
     onAccept: () -> Unit,
```
jhugman, line 100: Nit: The optional message `title` is not represented here.

jhugman (commented):
Well this is exciting!

I've left some comments about things that need to be fixed, and I'd like to see a UI test to match the [`NimbusMessagingNotificationTest`](https://github.com/mozilla-mobile/firefox-android/blob/daecfa9199d2d33331e8718b3f6338d7db578084/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/NimbusMessagingNotificationTest.kt) and [`NimbusMessagingHomescreenTest`](https://github.com/mozilla-mobile/firefox-android/blob/5b74938c67f5da8a57bf924eedb40736ba8d0ea9/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/NimbusMessagingHomescreenTest.kt).

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -1175,6 +1185,54 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+
+            lifecycleScope.launch(Main) {
+                researchSurfaceDialogFragment.showNow(
+                    supportFragmentManager,
+                    ResearchSurfaceDialogFragment.FRAGMENT_TAG,
+                )
+            }
+
+            // Update message as displayed.
+            val updatedMessage =
```
sarah541, line 1224: I am little confused here. Why are we calling `updateMessageAsDisplayed` here when it is automatically called in `onMessageDisplayed` below? This will cause the metadata `displayCount` to be incremented by 2. Not that it matters since the `maxDisplayCount` = 1

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -1175,6 +1185,54 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
     }
 
+    private fun showFullscreenMessageSurveyIfNeeded(context: Context) {
+        val messagingStorage = context.components.analytics.messagingStorage
+        val messages = runBlockingIncrement { messagingStorage.getMessages() }
+        val nextMessage =
+            messagingStorage.getNextMessage(FenixMessageSurfaceId.SURVEY, messages)
+                ?: return
+
+        if (nextMessage.id == "fullscreen-viewpoint-invitation-message") {
```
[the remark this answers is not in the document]
rocketsroger: We do in this case.  Since `FenixMessageSurfaceId.SURVEY` can be used for either the homescreen survey or the fullscreen survey.  The id is the one to determine which one we're showing.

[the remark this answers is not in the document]
rocketsroger: I'll double check.  I was worried that we might use the homescreen survey but you're right this won't happen.

rocketsroger:
> 



> Well this is exciting!
> 
> I've left some comments about things that need to be fixed, and I'd like to see a UI test to match the [`NimbusMessagingNotificationTest`](https://github.com/mozilla-mobile/firefox-android/blob/daecfa9199d2d33331e8718b3f6338d7db578084/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/NimbusMessagingNotificationTest.kt) and [`NimbusMessagingHomescreenTest`](https://github.com/mozilla-mobile/firefox-android/blob/5b74938c67f5da8a57bf924eedb40736ba8d0ea9/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/NimbusMessagingHomescreenTest.kt).
> 
> Thanks for jumping on this so quickly!

Thanks, I'll create an issue to track the work of adding new UI tests for the functionalities we introduced in this feature.

fenix/app/src/main/java/org/mozilla/fenix/experiments/view/ResearchSurfaceSurvey.kt
rocketsroger: Looking at the [experiment](https://experimenter.services.mozilla.com/nimbus/viewpoint-mobile-messaging-experiment/summary) I don't see title is being used in the experiment.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -1175,6 +1185,54 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+            val currentBootUniqueIdentifier = BootUtils.getBootIdentifier(context)
```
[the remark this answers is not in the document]
rocketsroger: I see what you mean.  Since we're not expected to show it again.  I'll still set it just in case something changed and we didn't notice that it's not set.  That said, I'll move it back since this is not needed until we mark the message as displayed.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rocketsroger: Let me double check.  Thanks for point this out.

rocketsroger: Confirmed that this is not an issue.  Thanks for the help @sarah541

rocketsroger:
> Thanks, I'll create an issue to track the work of adding new UI tests for the functionalities we introduced in this feature.

[Issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1826631) created

jhugman (approved):
I've left a couple of things that need to be done before this is landed, but I don't want to slow you down with another review cycle:

- write a UI test to demonstrate the surface working (or link to an issue with it on).
- add a title to the card. Since we have multiple surfaces that all work the same, this should work the same too.

rocketsroger:
>     * write a UI test to demonstrate the surface working (or link to an issue with it on).

Thanks, [Issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1826631) created.

fenix/app/src/main/java/org/mozilla/fenix/experiments/view/ResearchSurfaceSurvey.kt
rocketsroger: I see.  currently we've build this to display something that doesn't have a title to it.  I'll create an issue (not blocking the feature) that tracks the work of supporting title in case it's set.  Thanks

rocketsroger: As agreed.  I've created an [issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1826637) to track this as well.  Thanks

fenix/app/nimbus.fml.yaml
```diff
@@ -86,7 +87,7 @@ import:
                 max-display-count: 5
               SURVEY:
                 priority: 55
-                max-display-count: 10
+                max-display-count: 1
```
Amejia481, line 90: Should we create a new style, for this surface?  I guess not all surface will want to show only once, I think this will be a specific behaviour for fullscreen message.

rocketsroger: This was a decision made as a requirement to use this style.  And since we don't to show it more than once, I think 1 here is the correct value even for the homescreen version of the survey.  Thanks

sarah541:
Question : What happens if the `action` (url in this case) comes out to be empty?

rocketsroger:
> Question : What happens if the `action` (url in this case) comes out to be empty?

Good catch.  I didn't think about this case.  If `action` is empty, then I don't think we should create a `actionUri` and process the intent from that.  I'll fix.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -1175,6 +1185,51 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+        val fenixNimbusMessagingController = FenixNimbusMessagingController(messagingStorage)
+        val researchSurfaceDialogFragment = ResearchSurfaceDialogFragment.newInstance(
+            keyMessageText = nextMessage.data.text,
+            keyAcceptButtonText = nextMessage.data.buttonLabel,
+            keyDismissButtonText = null,
+        )
+
+        researchSurfaceDialogFragment.onAccept = {
+            val actionUri = fenixNimbusMessagingController.processMessageActionToUri(nextMessage)
+            processIntent(Intent(Intent.ACTION_VIEW, actionUri))
```
sarah541, line 1204:
```suggestion
            processIntent(getIntentForMessage(nextMessage))
```

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -1175,6 +1185,51 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             !processIntent(intent)
     }
 
+    private fun showFullscreenMessageSurveyIfNeeded(context: Context) {
```
Amejia481, line 1188: nit: Maybe just`showFullscreenMessageIfNeeded` just to be generic :)

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -272,6 +278,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
                 navHost.navController.navigate(NavGraphDirections.actionGlobalHomeJunoOnboarding())
             }
         } else {
+            lifecycleScope.launch(IO) {
+                showFullscreenMessageSurveyIfNeeded(applicationContext)
```
Amejia481, line 282: No sure if this is the expected behaviour, we will check once, if we haven't loaded the messages before this point we my not show the message.

I'm not sure if should reuse what we have for the home screen message, by dispatching a new action for this new surface.

https://github.com/mozilla-mobile/firefox-android/blob/865b1d75f1376d043538882b6001863f7a22f980/fenix/app/src/main/java/org/mozilla/fenix/messaging/MessagingFeature.kt#L17

and observe [messageToShow](https://github.com/mozilla-mobile/firefox-android/blob/865b1d75f1376d043538882b6001863f7a22f980/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt#L70) changes?

That will warranty that the message will show whenever we re enter the Home Screen, but I'm not sure if this is what we want?

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -1175,6 +1185,51 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
+        // Update message as displayed.
+        val currentBootUniqueIdentifier = BootUtils.getBootIdentifier(context)
+        val updatedMessage =
+            fenixNimbusMessagingController.updateMessageAsDisplayed(
+                nextMessage,
+                currentBootUniqueIdentifier,
+            )
+
+        runBlockingIncrement { fenixNimbusMessagingController.onMessageDisplayed(updatedMessage) }
+
```
titooan, line 1229: I think that defining `showFullscreenMessageSurveyIfNeeded` as `suspend` would avoid the need of using `runBlockingIncrement` here, but maybe I'm missing something.
Does it make sense to you?

fenix/app/nimbus.fml.yaml
Amejia481: Thanks for the clarification 👍🏽

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rocketsroger: Nice, I'll update.  Thanks

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rocketsroger: That is a good idea but I don't think it's for this case.  Homescreen messages are handled differently because it's part of the session adapter.  In this case we want to navigate to the full screen message directly.

rocketsroger:
> > Question : What happens if the `action` (url in this case) comes out to be empty?
> 
> Good catch. I didn't think about this case. If `action` is empty, then I don't think we should create a `actionUri` and process the intent from that. I'll fix.

From testing looks like Nimbus does protect us from these issues.  I couldn't get a empty `action` pass in and `action` can't be null either.  I won't do any extra checks then.

sarah541:
> > > Question : What happens if the `action` (url in this case) comes out to be empty?
> > 
> > 
> > Good catch. I didn't think about this case. If `action` is empty, then I don't think we should create a `actionUri` and process the intent from that. I'll fix.
> 
> From testing looks like Nimbus does protect us from these issues. I couldn't get a empty `action` pass in and `action` can't be null either. I won't do any extra checks then.

Even if Nimbus doesn't, I found a check in our code as a fallback. Linking : https://github.com/mozilla-mobile/firefox-android/blob/865b1d75f1376d043538882b6001863f7a22f980/android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt#L165

sarah541 (approved):

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rocketsroger: Good catch.  Thanks I'll update.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
Amejia481: 👍🏽

Amejia481 (approved):

PR #1578, opened 2023-04-06
Bug 1807350 - Generate baseline profiles for nightly

rahulsainani:
### How
– Create `StartupOnlyBaselineProfileGenerator` that generates baseline profile for the path of app startup.
– Define gradle managed virtual device in `benchmark/build.gradle`
– Generate baseline profiles by running `generateBaselineProfile` which execute the following gradle task with params.
```
 ./gradlew :benchmark:pixel3Api31BenchmarkAndroidTest -P android.testInstrumentationRunnerArguments.class=org.mozilla.fenix.benchmark.baselineprofile.StartupOnlyBaselineProfileGenerator -P baselineProfileGeneration
 ````
Note: when the flag `baselineProfileGeneration` is used, only `arm64-v8a` abi is created as that's what the gradle managed virtual device needs. Baseline profiles are generated for Kotlin/Java code, so any abi can be picked based on the device available.
– Add Copy gradle task `copyBaselineProfile` to copy generated baseline profiles to `app/src/main`. 
– Run these two gradle tasks in the fenix-nightly pipeline, before the build step as the apk needs to be build with the generated `baseline-prof.txt`.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1807350

### DIFF
--- a/taskcluster/ci/generate-baseline-profiles/kind.yml
+++ b/taskcluster/ci/generate-baseline-profiles/kind.yml
@@ -1,0 +1,54 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - android_taskgraph.transforms.gradle_optimization:transforms
+    - taskgraph.transforms.job:transforms
+    - taskgraph.transforms.task:transforms
+
+kind-dependencies:
+    - toolchain
+    - external-gradle-dependencies
+
+task-defaults:
+    attributes:
+        retrigger: true
+    fetches:
+        toolchain:
+            - android-sdk-linux
+        external-gradle-dependencies:
+            - external-gradle-dependencies.tar.xz
+    optimization:
+        skip-unless-changed: [] # Paths are dynamically added by transforms
+    run-on-tasks-for: []
+    run:
+        using: gradlew
+        use-caches: false
+    treeherder:
+        kind: test
+        tier: 1
+    worker-type: b-android-large
+    worker:
+        docker-image: {in-tree: base}
+        max-run-time: 7200
+
+tasks:
+    fenix:
+        description: 'Generate baseline profile for fenix.'
+        dependencies:
+            external-gradle-dependencies: external-gradle-dependencies-fenix
+        attributes:
+            build-type: nightly
+            shipping-product: fenix
+        run:
+            pre-gradlew:
+                - ["cd", "fenix"]
+                - ['java', '-version']
+            gradlew: ['setupEmulator', 'generateBaselineProfile', 'copyBaselineProfile']
+        run-on-tasks-for: []
+        treeherder:
+            symbol: fenix-baseline-profile
+            platform: fenix-android-all/opt
--- a/fenix/.buildconfig.yml
+++ b/fenix/.buildconfig.yml
@@ -137,13 +137,13 @@ variants:
   name: fenixBeta
 - apks:
   - abi: arm64-v8a
-    fileName: app-fenix-arm64-v8a-benchmark-unsigned.apk
+    fileName: app-fenix-arm64-v8a-benchmark.apk
   - abi: armeabi-v7a
-    fileName: app-fenix-armeabi-v7a-benchmark-unsigned.apk
+    fileName: app-fenix-armeabi-v7a-benchmark.apk
   - abi: x86
-    fileName: app-fenix-x86-benchmark-unsigned.apk
+    fileName: app-fenix-x86-benchmark.apk
   - abi: x86_64
-    fileName: app-fenix-x86_64-benchmark-unsigned.apk
+    fileName: app-fenix-x86_64-benchmark.apk
   build_type: benchmark
   name: fenixBenchmark
 - apks:
--- a/fenix/app/build.gradle
+++ b/fenix/app/build.gradle
@@ -160,6 +160,7 @@ android {
         benchmark releaseTemplate >> {
             initWith buildTypes.nightly
             applicationIdSuffix ".fenix"
+            signingConfig signingConfigs.debug
             debuggable false
         }
     }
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: taskcluster/ci/generate-baseline-profiles/kind.yml [hunk 1]
```diff
@@ -0,0 +1,56 @@
...
+            shipping-product: fenix
+        run:
+            pre-gradlew:
+                - ["cd", "fenix"]
+                - ['java', '-version']
+            gradlew: ['generateBaselineProfile', 'copyBaselineProfile']
+        run-on-tasks-for: []
+        include-shippable-secrets: true
+        include-nightly-version: true
+        source-project-name: "fenix"
```
ahal, line 53: So the [decision task error](https://firefox-ci-tc.services.mozilla.com/tasks/T_t3NstJQv2pm9wwx7pt-g/runs/0/logs/public/logs/live.log) you're getting is caused by the presence of these keys.

These tasks run through each of the transform files defined above in order. Some transforms also define schemas that the task most conform to at the start of the transformation process. In this case, the schema error is coming out of the `taskgraph.transforms.job` transforms.

If we look over at [this schema in the Taskgraph repo](https://github.com/taskcluster/taskgraph/blob/526f4d82906eef73ee69505c406c03f5f35de973/src/taskgraph/transforms/job/__init__.py#L31), we can see that none of these three keys are allowed there.

I suspect you copy/pasted these from the `build-apk` kind. Those tasks work because they apply the `build_apk` transforms. If we inspect those transforms, we'll see they actually [pop these keys off of the task](https://github.com/mozilla-mobile/firefox-android/blob/f68469b548ede5dce657e3e1be337596597610e9/taskcluster/android_taskgraph/transforms/build_apk.py#L49), so that by the time those tasks need to satisfy the `job_description_schema`, the offending keys are no longer present.

So if this task doesn't need these keys, you can just delete them and it should work. Otherwise you'll either need to also run them through the `build_apk` transforms (which likely doesn't make sense here), or you'll have to write your own transform file that processes these values and removes them from the task definition.

Let me know if you have any questions!

ahal (changes requested):

taskcluster/ci/generate-baseline-profiles/kind.yml
rahulsainani: Thank you for the explanation and for sharing all the docs, offline as well! 🙌 Will open this PR for review, would really appreciate your review.

rahulsainani:
Requesting RelEng folks to also review the changes/additions in the taskgraph tasks :)

taskcluster/ci/generate-baseline-profiles/kind.yml
```diff
@@ -0,0 +1,54 @@
...
+            build-type: nightly
+            shipping-product: fenix
+        run:
+            pre-gradlew:
+                - ["cd", "fenix"]
+                - ['java', '-version']
+            gradlew: ['generateBaselineProfile', 'copyBaselineProfile']
+        run-on-tasks-for: []
+        treeherder:
+            symbol: fenix-baseline-profile
```
ahal, line 53: This is fine for now especially since there is only one task. But typically Treeherder symbols are short so they don't take up too much real estate in the UI. Also if you plan to add other similar tasks in the future, you may consider grouping them together, e.g:
```
symbol: BP(fenix)
```
(This would necessitate registering the symbol in `taskcluster/ci/config.yml` as well)

ahal (approved):
The taskgraph changes look reasonable to me. I'm not qualified to speak for the build/gradle changes so please wait for a separate review on that.

gbrownmozilla (approved):
Task configuration looks good to me.

rahulsainani:
Thanks for reviewing @ahal @gbrownmozilla . Could we test that this is working as intended before merging? I'd have to inspect the resulting apk to make sure it's working as intended and we can also make sure it doesn't break the pipeline, in case this task fails.

gbrownmozilla:
One thing I did was run ```taskgraph target-graph -p taskcluster/test/params/main-repo-cron-nightly.yml --diff``` which showed that the ```generate-baseline-profiles-fenix``` task was added; I feel pretty confident the task won't break the decision task or do bad things to the task graph. 

There is a staging repo for firefox-android, but it is very out of date and might take some effort to get it working again ... I'm not sure it's worth the effort.

ahal:
I think it's the actual build tasks he wants to test (not the taskgraph setup).

@rahulsainani I pushed your PR to the `main` branch of [staging-firefox-android](https://github.com/mozilla-releng/staging-firefox-android) and triggered a nightly cron task there. You can find the graph here:
https://firefox-ci-tc.services.mozilla.com/tasks/RF9ur2gQQqOQvSBxLAdQwA

Does that contain the task you need to test?

rahulsainani:
Thanks both! Yeah I'd like to test the gradle tasks are working correctly. The gradle tasks will 
1. Create a gradle managed device (android emulator) so we can run the test without a real android device. 
2. Then it will create the apk, install it on the gradle managed device and run the test to generate baseline profiles.
3. Copy the generated baseline profiles to the app directory so they can be part of the actual fenix apk that will be shipped. 

`build-apk` and the rest of the steps can proceed further.

It seems to be failing on [the first step](https://firefox-ci-tc.services.mozilla.com/tasks/F9l0cIJyTjGumFHrZFGB3A/runs/0/logs/public/logs/live.log), but not sure what's the cause of the issue. 🤔

jonalmeida:
> It seems to be failing on [the first step](https://firefox-ci-tc.services.mozilla.com/tasks/F9l0cIJyTjGumFHrZFGB3A/runs/0/logs/public/logs/live.log), but not sure what's the cause of the issue. 🤔

> chattr: Inappropriate ioctl for device while reading flags on

This reads that the SDK is trying to set permissions on the AVD image but it can't do that because of.. permissions or strange filesystem limitations? @ahal who would be a good person to ping to understand how we could fix something like this?

ahal:
I was chatting with Rahul, and he mentioned that we might need to enable kvm in the docker-worker config. This will require creating new pools, but is definitely something we can try. I'm unsure if this requires privileged access to the host though.. if so then this will be a definite no-go for any tasks in the release pipeline.

ahal:
Discussion on this is happening in:
https://bugzilla.mozilla.org/show_bug.cgi?id=1830928

Apparently we hit the same issue when building generate-profile tasks for geckoview, and worked around it by disabling hardware acceleration.

gabrielluong:

PR #1581, opened 2023-04-06
Bug 1821425 - Add the logic to show the selected tab UI and the multiselection tab UI

MozillaNoah:
I discovered that Private tabs do not support multi selection, so rather than adding this part of business logic into the Compose code, I opted for adding it to the Controller with some unit test coverage. 

Note that the close tab UI does not properly hide itself with the grid items (list items function correctly). This will be addressed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1826670).

https://user-images.githubusercontent.com/87384386/230489340-2fa25c13-bc16-49be-9ccf-5d6e40665e39.mov

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1821425

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
@@ -27,8 +27,10 @@ import androidx.compose.ui.unit.dp
 import com.google.accompanist.pager.ExperimentalPagerApi
 import com.google.accompanist.pager.HorizontalPager
 import com.google.accompanist.pager.rememberPagerState
+import mozilla.components.browser.state.state.BrowserState
 import mozilla.components.browser.state.state.ContentState
 import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.lib.state.ext.observeAsComposableState
 import org.mozilla.fenix.components.AppStore
 import org.mozilla.fenix.components.appstate.AppState
@@ -41,6 +43,8 @@ import org.mozilla.fenix.theme.FirefoxTheme
 /**
  * Top-level UI for displaying the Tabs Tray feature.
  *
+ * @param appStore [AppStore] used to listen for changes to [AppState].
+ * @param browserStore [BrowserStore] used to listen for changes to [BrowserState].
  * @param tabsTrayStore [TabsTrayStore] used to listen for changes to [TabsTrayState].
  * @param displayTabsInGrid Whether the normal and private tabs should be displayed in a grid.
  * @param onTabClose Invoked when the user clicks to close a tab.
@@ -64,6 +68,7 @@ import org.mozilla.fenix.theme.FirefoxTheme
 @Composable
 fun TabsTray(
     appStore: AppStore,
+    browserStore: BrowserStore,
     tabsTrayStore: TabsTrayStore,
     displayTabsInGrid: Boolean,
     shouldShowInactiveTabsAutoCloseDialog: (Int) -> Boolean,
@@ -81,6 +86,8 @@ fun TabsTray(
     onInactiveTabClick: (TabSessionState) -> Unit,
     onInactiveTabClose: (TabSessionState) -> Unit,
 ) {
+    val selectedTabId = browserStore
+        .observeAsComposableState { state -> state.selectedTabId }.value
     val multiselectMode = tabsTrayStore
         .observeAsComposableState { state -> state.mode }.value ?: TabsTrayState.Mode.Normal
     val selectedPage = tabsTrayStore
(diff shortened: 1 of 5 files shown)


fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -300,8 +324,16 @@ private fun TabsTrayPreviewRoot(
...
             onTabMediaClick = {},
             onTabClick = {},
-            onTabMultiSelectClick = {},
-            onTabLongClick = {},
+            onTabMultiSelectClick = { tab ->
+                if (tabsTrayStore.state.mode.selectedTabs.contains(tab)) {
+                    tabsTrayStore.dispatch(TabsTrayAction.RemoveSelectTab(tab))
+                } else {
+                    tabsTrayStore.dispatch(TabsTrayAction.AddSelectTab(tab))
+                }
```
MozillaNoah, line 332: this was me playing around with previews and relying on the the store + actions rather than a `remember` variable to maintain the selected tabs.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
```diff
@@ -48,6 +48,8 @@ import org.mozilla.fenix.theme.FirefoxTheme
 fun TabLayout(
     tabs: List<TabSessionState>,
     displayTabsInGrid: Boolean,
+    selectedTabId: String?,
+    selectionMode: TabsTrayState.Mode,
```
MatthewTighe, line 52: nit: kdocs

MatthewTighe (commented):
Before I get too deep into a review, I wanted to mention a couple discrepancies between this and the current UI.

Current:
- entering multiselect mode makes the selected tab become non-highlighted
- each multiselected item has a dark gradient overlaying the hero image

Neither of this seem to be true of this patch. Is that intentional, will it be addressed in a follow-up, or should we handle it here?

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: > By retool, do you mean instead of passing in the `browserStore` to `TabsTray`? That seems fine to me, and you've already established precedent by passing in the `appStore` and `tabsTrayStore`.
>
> When I was talking about using actions directly in the `Composable`s, I had meant that the composables could directly call things like `tabsTrayStore.dispatch(TabsTrayAction.AddSelectTab(tab))` instead of having them passed in as lambdas.
>
> Now it feels like we're kind of in between both - the top level of composable is extracting state from the store but relying on the component above it to define actions to change that state.

I was more meaning that here, locally in the preview, retooling the usage of `Actions` in favor of keeping the pattern of using `remember` variables before landing the PR, if that is desired based on your and whoever's preference/recommendation. These changes were more of a small proof of concept for the interactor RFC, of like "can this function _just_ with actions?". I was in no way trying to set a new pattern of lambdas + Store actions or try to have this Preview be a model for this. 

I hope that makes sense.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
```diff
@@ -37,6 +37,9 @@ import org.mozilla.fenix.theme.FirefoxTheme
  *
  * @param tabs The list of [TabSessionState] to display.
  * @param displayTabsInGrid Whether the tabs should be displayed in a grid.
+ * @param selectedTabId The ID of the currently selected tab.
+ * @param selectionMode [TabsTrayState.Mode] indicating whether the Tabs Tray is in single selection
+ * or multi-selection and contains the set of selected tabs.
```
MatthewTighe, line 42: I don't think this is the place the solve this problem, but I feel like "selected" is annoyingly overloaded between the current active tab and the (multi-) selected tabs. "Selected" actually makes more sense here IMO, so I think this would mean updating the BrowserState to use a different name for the (last) active tab.

MatthewTighe (approved):
one nit but LGTM otherwise!

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
MozillaNoah: I think we're in agreement here. This is also evident with `TabListItem` and `TabGridItem` currently having:
```
isSelected: Boolean = false,
multiSelectionEnabled: Boolean = false,
multiSelectionSelected: Boolean = false,
```

I can file a ticket for a post-release or "or when time allows" follow-up.

MozillaNoah: https://bugzilla.mozilla.org/show_bug.cgi?id=1828527

PR #1624, opened 2023-04-12
Bug 1827030 - Enable juno onboarding in nightly for new users

rahulsainani:
### How
– Use `FenixOnboarding` preferences to show juno onboarding to users who haven't completed the existing onboarding yet. This means, it will show to new users or existing users who haven't completed the onboarding yet.
– Delete `isJunoOnboardingShown` flag as that's not used anymore.
– Updated tests to use `hasUserBeenOnboarded` param that's passed in the function.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -271,7 +271,11 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             it.start()
         }
 
-        if (settings().shouldShowJunoOnboarding(intent.toSafeIntent().isLauncherIntent)) {
+        if (settings().shouldShowJunoOnboarding(
+                hasUserBeenOnboarded = onboarding.userHasBeenOnboarded(),
```
rahulsainani, line 275: This seems to be causing the `verifyRunBlockingAndStrictModeSuppresionCount` to fail as it's reading the pref. We need to read the pref here to decide whether to show onboarding or not. Any suggestions @jonalmeida @rocketsroger ?

jonalmeida: I thought about this for some time last week and I think we can increment the suppression count here.

Reasoning: it's unfortunate, but this check is already made earlier in our start process in FenixApplication and even much later in the HomeFragment and BaseBrowserFragment. That said, we can't seem to avoid it here because we are creating the new `FenixOnboarding` object (which uses a separate pref file for historical reasons) and reading prefs from IO during that startup.

We can't move this check to be cached somewhere else quite as easily either. We could try to hide it within the `Settings.kt` class so that the first accessor would be only read of this pref and every follow-up one would be from in-memory caches. _However_, I would not do that in this change if this is something of interest.

We could accept this suppression count here and try the above in a follow-up. What do you think? Maybe there are other ways around it?

jonalmeida (approved):
Thanks again for the discussion!

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rahulsainani: Sounds good! 👍 Thanks for the review and discussion!

jonalmeida: Could you link the follow-up bug here for completeness? It'll help keep track of where the source of the issue came from and how to follow the trail.

rahulsainani: Sure, [here's the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1829301).

PR #1651, opened 2023-04-14
Bug 1823678 - Integrate Server Knobs into Fenix

rosahbruno:
Add the Server Knobs configuration to Fenix. This includes adding the required Nimbus feature to pass remote configurations.

**Testing**
You can replace the `setMetricsEnabledConfig()` call with the line below.
`Glean.setMetricsEnabledConfig("{\"search.default_engine.name\": false}")`

If you use the Debug Ping Viewer you can verify normal baseline pings are sent with that metric. Once you disable it, baseline pings will no longer include that metric.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1823678


travis79 (changes requested):
A couple of naming changes requested, other than that this looks good to me for the basic integration.  One thing I think you might also like to do is wiring up the "// Pass the JSON configuration for Server Knobs from Nimbus." to the onUpdatesApplied similar to what is done here: https://github.com/mozilla-mobile/firefox-android/blob/71d72cad3b76bf5110b195e53524e2d77930b566/focus-android/app/src/main/java/org/mozilla/focus/settings/privacy/studies/StudiesViewModel.kt#L53

travis79 (approved):
Looks good to me! Let's get a Fenix engineer to take a look also. I believe there is a GitHub label ready-for-review or similar...

fenix/app/nimbus.fml.yaml
```diff
@@ -326,6 +326,14 @@ features:
         type: List<OnboardingPanel>
         default: ["themes", "toolbar-placement", "sync", "tcp", "privacy-notice"]
 
+  glean:
+    description: "A feature that provides server-side configurations for Glean metrics (aka Server Knobs)."
+    variables:
+      glean-metric-configuration:
+        description: "A map of metric base-identifiers to booleans representing the state of the 'enabled' flag for that metric."
+        type: String
+        default: "{}"
```
jhugman, line 335: Could we use `Map<String, Boolean>` instead here? I'm thinking about the experimenter interface for experiment owners here: having an opaque string they type in, with not even JSON syntax to validate.

fenix/app/nimbus.fml.yaml
```diff
@@ -326,6 +326,14 @@ features:
         type: List<OnboardingPanel>
         default: ["themes", "toolbar-placement", "sync", "tcp", "privacy-notice"]
 
+  glean:
+    description: "A feature that provides server-side configurations for Glean metrics (aka Server Knobs)."
+    variables:
+      glean-metric-configuration:
```
jhugman, line 332: Nit: `glean-` is redundant, and `configuration` is somewhat over-general.

Suggest `metrics-enabled`.

android-components/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -109,6 +109,16 @@ object Glean {
         GleanCore.setExperimentInactive(experimentId = experimentId)
     }
 
+    /**
+     * Set configuration to override metrics' enabled/disabled state, typically from a remote_settings
+     * experiment or rollout.
+     *
+     * @param json Stringified JSON map of metrics and their associated `disabled` property.
```
jhugman, line 116: Not sure if this is meant to conflict here:

> A map of metric base-identifiers to booleans representing the state of the 'enabled' flag for that metric

> JSON map of metrics and their associated `disabled` property.

jhugman (commented):
Drive-by 🚌 🚚 🛻 🦼  Review.

fenix/app/nimbus.fml.yaml
rosahbruno: @jhugman 
I talked with Travis a bit earlier today. The purpose of using the JSON string rather than something more structured is intentional to avoid too tight of coupling between Nimbus and Glean. This is a bit different since desktop has a Nimbus JSON type, but we are parsing and validating the JSON structure on the Glean side to ensure nothing unexpected happens.

That being said, if this is seen as a must-have I can make updates and then just convert the Nimbus value to a JSON string before we pass it to Glean.

rosahbruno:
@jhugman @travis79 
I updated this now to use the `Map<String, Boolean>` type instead.
- Nimbus config - https://github.com/mozilla-mobile/firefox-android/pull/1651/files#diff-843e3c4b7233c5042bb66c11951c32a19f5ad30bf9369e945079cab4034f4f0eR334
- Converting map to JSON string - https://github.com/mozilla-mobile/firefox-android/pull/1651/files#diff-c07dbf42136256cfd1eea0fe95204636f9327f15255ec1fab0843e18bd2aaec7R990

android-components/components/service/glean/src/main/java/mozilla/components/service/glean/Glean.kt
```diff
@@ -109,6 +109,16 @@ object Glean {
...
 
+    /**
+     * Set configuration to override metrics' enabled/disabled state, typically from a remote_settings
+     * experiment or rollout.
+     *
+     * @param json Stringified JSON map of metrics and their associated `enabled` property.
+     */
+    fun setMetricsEnabledConfig(json: String) {
+        GleanCore.setMetricsEnabledConfig(json)
+    }
```
jhugman, line 120: Personally, I think I'd call the `toString()` in this method:

```suggestion
    fun setMetricsEnabledConfig(enabled: Map<String, Boolean>) {
        GleanCore.setMetricsEnabledConfig(JSONObject(enabled).toString())
    }
```

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -197,6 +201,18 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+        setGleanMetricsConfiguration()
+
+        // Register an observer to update the metric configuration whenever Nimbus is updated.
+        applicationContext.components.analytics.experiments.register(
+            object : NimbusInterface.Observer {
+                override fun onUpdatesApplied(updated: List<EnrolledExperiment>) {
+                    setGleanMetricsConfiguration()
+                }
+            },
+        )
```
jhugman, line 214: Nimbus configuration is not changed during runtime, so this is unnecessary.

(This `Observer` API is there for historical purposes, and not really needed now).

jhugman (approved):
There are a couple of comments, but I won't block you on another review cycle. ✅ Good job!

PR #1562, opened 2023-04-05
Bug 1819196 - Plug-in tabs tray fab business logic

Alexandru2909:
Added logic for compose tabs tray click interactions.

https://user-images.githubusercontent.com/35462038/230094210-74ecb963-5fbb-4c1e-aa1a-195ba38c3b3f.mp4

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1819196

PR #1687, opened 2023-04-18
Bug 1814992 - Add `SyncedTabsList` to `TabsTray`

MozillaNoah:
Note: the business logic for the FAB has yet to land via https://github.com/mozilla-mobile/firefox-android/pull/1562, so any syncs during testing will need to be manual syncs via Settings.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1814992

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
@@ -8,9 +8,7 @@ import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.shape.RoundedCornerShape
-import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
@@ -31,6 +29,7 @@ import mozilla.components.browser.state.state.BrowserState
 import mozilla.components.browser.state.state.ContentState
 import mozilla.components.browser.state.state.TabSessionState
 import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.browser.storage.sync.TabEntry
 import mozilla.components.lib.state.ext.observeAsComposableState
 import org.mozilla.fenix.components.AppStore
 import org.mozilla.fenix.components.appstate.AppState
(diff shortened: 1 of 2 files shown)


fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -71,6 +75,7 @@ fun TabsTray(
     browserStore: BrowserStore,
     tabsTrayStore: TabsTrayStore,
     displayTabsInGrid: Boolean,
+    taskContinuityEnabled: Boolean,
```
MatthewTighe, line 78: Tangential, but do we know the status of this flag? I would maybe assume it's safe to remove by now. Is there a ticket tracking it or anything?

MatthewTighe (approved):
LGTM!

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: As far as I know, this is a zombie secret setting. It was enabled by default in last September, but we don't have a ticket to clean it up.

jonalmeida: Let's not make more work here by passing around this flag. You could just hardcode true in the `SyncedTabsList` and file a cleanup issue for the backlog.

Is there a reason to propagate this all the way up to the `TabsTray` level?

MozillaNoah: Passing in true down to `SyncedTabsList` could make sense.

If being derived from Settings, the flags have to propagate from the fragment + `TabsTray` in order to not break the previews. (calls to `Components` break the preview compilation)

jonalmeida: I'm not sure I follow, sorry. I was suggesting: removing this flag from this line and making it `true` on line 211 below.

Does this have an effect on previews?

MozillaNoah: that made sense and I'm pushing that now - and no that doesn't break previews.

Sorry for the confusion. I was more trying to explain that having this flag as a parameter in `TabsTray` was just to abstract `Settings` (and therefore `Components`) away from the Compose code.

jonalmeida: Sounds good! 

> we don't have a ticket to clean it up.

Did you already file a bug for this? It would be a good-first-bug for an external contributor but also something we can pick up during some downtime.

jonalmeida: File a bug and add the `good-first-bug` keyword.

MozillaNoah: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1829018

PR #1699, opened 2023-04-19
Bug 1826560 - consistently sign android-test APKs with the dep key

jcristau:
These APKs need to be signed with the same key as the ones used in firebase for ui-test and startup-test, which already use dep signing.  Stop discriminating between different tasks_for and just keep a list of production build types for release signing.

Also for simplicity, remove the conditional for adding to the index since I can't think of a reason this is necessary.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1826560


taskcluster/android_taskgraph/transforms/signing_apk.py
```diff
@@ -90,13 +75,7 @@ def set_signing_type(config, tasks):
 @transforms.add
 def set_index(config, tasks):
     for task in tasks:
-        index = {}
-        if (
-            config.params["tasks_for"] in ("cron", "github-release", "action")
```
JohanLorenzo, line 95: Question: IIUC, removing this if statement will add signing index to all tasks which includes `debug` on `git-push` for instance. I'm fine with that. I wonder if there's any other side effects. What does `taskgraph target-graph --diff` return?

gbrownmozilla (approved):
Glad we can go this way. lgtm.

taskcluster/android_taskgraph/transforms/signing_apk.py
JohanLorenzo: Thanks! This diff looks good to me. I don't see anything problematic with exposing these signed builds on the Taskcluster index.

JohanLorenzo (approved):

PR #1702, opened 2023-04-19
Bug 1822193 - Scroll to the active tab when opening the Tabs Tray

MozillaNoah:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1822193


fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -145,60 +146,77 @@ fun TabsTray(
...
+                                onTabClick = handleTabClick,
+                                onTabLongClick = onTabLongClick,
+                                header = optionalInactiveTabsHeader,
+                            )
+                        } else {
+                            Text(
+                                text = "Empty state",
+                                color = FirefoxTheme.colors.textPrimary,
+                                style = FirefoxTheme.typography.body1,
+                            )
```
MozillaNoah, line 198: I had to add a placeholder empty state UI before the official ticket because of the `TabLayout` getting composed when there's no tabs, breaking the initial scroll because of how the tab lists are populated in `TabsTrayStore`

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
```diff
@@ -59,10 +59,20 @@ fun TabLayout(
...
     header: (@Composable () -> Unit)? = null,
 ) {
+    var selectedTabIndex = 0
+    selectedTabId?.let {
+        tabs.forEachIndexed { index, tab ->
+            if (tab.id == selectedTabId) {
+                selectedTabIndex = index
+            }
+        }
+    }
```
MatthewTighe, line 69: nit: We could avoid a `var` here with something like:
```
val selectedTabIndex = tabs.indexOf(tabs.find { tab ->
    tab.id == selectedTabId
})
```

Feels a little hacky with all the nullability implied in there, but might be preferable.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MatthewTighe: This could be done in the empty state ticket, but I think we should split these conditional cases into their own functions for readability. Something like:

```
if (normalTabs.isNotEmpty() || inactiveTabs.isNotEmpty()) {
    TabsUI()
} else {
    EmptyUI()
}
```

That, or the empty UI case should come first. My complaint is mostly that the nesting in the `TabsLayout` case makes it a little hard to parse

MatthewTighe: I do also see the "Empty state" text flash briefly when loading the tray, but I see that that is not technically a functionality degradation vs the XML version. Still, I can't help wondering if there's a way we could distinguish a "loading" state from an "empty" state to avoid that flash of text.

MatthewTighe (approved):
Couple nits, but this otherwise LGTM.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
MozillaNoah: For sure. We almost kinda need a tri-state here of:
1. Empty
2. Loading
3. Populated

I'm not immediately sure how to engineer that with the store observations.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTabLayouts.kt
MozillaNoah: I wanted to avoid this, since doing `indexOf` + `tabs.find` causes two linear iterations over the list of tabs, but also `find` can return `-1` when the tab isn't in the list (thinking of the situation it's searching for a Private tab in the normal list) which would cause an exception.

PR #1704, opened 2023-04-19
Bug 1826457 - Update Compose version to 1.4.2 and Lifecycle version to 2.6.1

MozillaNoah:
Bumping this library to fix a crash when rotating the Composified Tabs Tray while the tabs are in a grid.

I looked a bit into some of the non-obvious side effect changes (`HomeActivity` being one of them), and the only hypothesis I have for those is that there may have been some underlying dependency from `Foundation` library that updated with it, causing some weird compiler errors. The change to `LayoutCoordinates.isVisible` inside `PocketStoriesComposables` was an inline replacement from the deprecation docs + IDE. 

Working through the Compose version bump, the Lifecycle version had to be bumped as well.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1826457


rvandermeulen:
Should we also bump Focus and AC?

MozillaNoah:
> Should we also bump Focus and AC?

Sure, I'll look into bumping those, too. (Maybe that will help with the CI error)

ankit-gautam23 (approved):
LGTM

rvandermeulen (changes requested):
Fenix has some bustage still, but it looks like things are getting close!
https://firefox-ci-tc.services.mozilla.com/tasks/JR298elrTlaOPPM6kEHWVg

MozillaNoah:
> Fenix has some bustage still, but it looks like things are getting close! https://firefox-ci-tc.services.mozilla.com/tasks/JR298elrTlaOPPM6kEHWVg

Thanks for that link! I wasn't immediately sure what the gray square icon was indicating for `external-gradle-dependencies-fenix`.

Just pushed again so 🤞

rvandermeulen (approved):
LGTM, thanks for getting all 3 products and slogging through all the mechanical changes and CI failures! Deferring to Matthew for the primary code review, though.

android-components/components/feature/tabs/src/test/java/mozilla/components/feature/tabs/toolbar/TabCounterToolbarButtonTest.kt
```diff
@@ -54,7 +54,8 @@ class TabCounterToolbarButtonTest {
             currentState = initialState
         }
 
-        override fun getLifecycle(): Lifecycle = lifecycleRegistry
+        override val lifecycle: Lifecycle
+            get() = lifecycleRegistry
```
MatthewTighe, line 58: We should be able to avoid the explicit `get() =` and do something like `override val lifecycle: Lifecycle = lifecycleRegistry` or even `override val lifecycle = lifecycleRegistry` if we want to save space. That would be my preference but I don't think either are required changes

MatthewTighe (approved):
Optional nit but LGTM otherwise

PR #1728, opened 2023-04-21
Bug 1829361 - Prevent Talkback focus behind find in page bar

DreVla:
The best solution I could find for this after many tries and help was to hide the toolbar while the FIP bar is present. By hiding the browser toolbar while the find in page bar is visible, we make sure that talkback won't be able to focus on any elements present underneath, preventing any confusing experiences for the user. It still won't focus correctly on the edit text, but it at least focuses on the button right next to it.

https://user-images.githubusercontent.com/32488956/233658516-ef76ea2b-d7a8-4e7b-be68-e2f560b7f6b9.mp4

The issue was reported [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1802584) also, but the fix there was made for the Fenix app. The issue here is similar but not exactly the same, trying the patch on Fenix did not really work here.

### Pull Request checklist

- [X] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1829361


focus-android/app/src/main/java/org/mozilla/focus/browser/integration/FindInPageIntegration.kt
```diff
@@ -39,10 +41,12 @@ class FindInPageIntegration(
 
     fun show(sessionState: SessionState) {
         findInPageView.isVisible = true
+        browserToolbar.isVisible = false
```
gabrielluong, line 44: Could we add a test for this? https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/test/java/org/mozilla/fenix/components/FindInPageIntegrationTest.kt

MatthewTighe, line 44: I think a comment here would be helpful.

MatthewTighe (approved):
Other than my nit, this LGTM

PR #1750, opened 2023-04-24
Bug 1796434  - PWA supports display-mode media query via display value

makotokato:
Fennec and GeckoView support this value, But Fenix doesn't use GeckoView API, so this PR implements https://developer.mozilla.org/en-US/docs/Web/Manifest/display and https://developer.mozilla.org/en-US/docs/Web/CSS/@media/display-mode via PWA.

To support this, I add WebAppContentFeature. If I shouldn't add something feature for it, let me know. (Or access EngineSession from Fenix directly?)

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1796434


gabrielluong:
@jonalmeida Flagged you for review on this, but if you know someone else who is better suited to review this, please redirect. Ty

android-components/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/feature/WebAppContentFeature.kt
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Feature used to handle web content settings from manifest file.
+ */
+class WebAppContentFeature(
+    private val store: BrowserStore,
+    private val tabId: String? = null,
+    private val manifest: WebAppManifest,
+) : DefaultLifecycleObserver {
+
+    override fun onCreate(owner: LifecycleOwner) {
```
jonalmeida, line 22: Is `onCreate` the right place for this? My guess is probably yes since we want to keep the that display mode from creation - the value doesn't change until we re-fetch a new manifest.

This would not affect the entry point from the main application in Fenix because we add this to the ExternalAppBrowserFragment.

android-components/components/feature/pwa/src/test/java/mozilla/components/feature/pwa/feature/WebAppContentFeatureTest.kt
```diff
@@ -0,0 +1,61 @@
...
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.engine.manifest.WebAppManifest
+import mozilla.components.support.test.mock
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mockito.doReturn
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.verify
+
+@RunWith(AndroidJUnit4::class)
```
jonalmeida, line 23: I don't think we need this (slow) test runner, right?

android-components/components/feature/pwa/src/test/java/mozilla/components/feature/pwa/feature/WebAppContentFeatureTest.kt
```diff
@@ -0,0 +1,61 @@
...
+class WebAppContentFeatureTest {
+    private val customTabId = "custom-id"
+
+    @Test
+    fun `display mode is fullscreen based on PWA manifest`() {
+        val engineState: EngineState = mock()
+        val engineSession = mock<EngineSession>()
```
jonalmeida, line 30:
```suggestion
        val engineState: EngineState = mock()
        val engineSession: EngineSession = mock()
```

Only a nit: since you started creating a mock using type inference extension above, you could do the same with this one too.

jonalmeida (approved):
Thanks for pinging me, gl.

This change looks good to me - just two nits with the tests and we can land it. I'm happy to make the changes for you and land it if it makes it easier for you. 🙂

android-components/components/feature/pwa/src/test/java/mozilla/components/feature/pwa/feature/WebAppContentFeatureTest.kt
```diff
@@ -0,0 +1,61 @@
...
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.verify
+
+@RunWith(AndroidJUnit4::class)
+class WebAppContentFeatureTest {
+    private val customTabId = "custom-id"
+
+    @Test
+    fun `display mode is fullscreen based on PWA manifest`() {
+        val engineState: EngineState = mock()
```
jonalmeida, line 29: Just realized we don't need to mock this; it's just a data class. We can add the `EngineSession` mock inside it too.

android-components/components/feature/pwa/src/test/java/mozilla/components/feature/pwa/feature/WebAppContentFeatureTest.kt
```diff
@@ -0,0 +1,61 @@
...
+        val engineSession = mock<EngineSession>()
+
+        val tab = spy(
```
jonalmeida, line 32: `spy` (and the below `doReturn`) aren't needed when we have a real `EngineState`.

jonalmeida (commented):
I'm going to clean this up and land it along with a few more nits I noticed during a second pass.

PR #1754, opened 2023-04-24
Bug 1829619 - Fix external ActivityNotFoundException for default browser settings

rahulsainani:
Details in the bug.

### How
– Create `startExternalActivitySafe` extension function that takes in lamda `onActivityNotPresent`.
– Use the above extension function in existing `navigateToDefaultBrowserAppsSettings`
– Extract `openDefaultBrowserSumoPage` as a private extension function for ease of use and re-usability. 

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1829619


Amejia481 (commented):
I noticed with have a similar function on AC, should we have to update it too?
https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/support/utils/src/main/java/mozilla/components/support/utils/ext/Context.kt#L27

rahulsainani:
> I noticed with have a similar function on AC, should we have to update it too? https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/support/utils/src/main/java/mozilla/components/support/utils/ext/Context.kt#L27

Good to know! We should try to commonalize this and try to use the A-C one going forward.

rahulsainani:
@Amejia481 Thanks for the review! 

As discussed, removing the `HomeDeepLinkIntentProcessorTest.make_default_browser deep_link for above API 23` test as it is tricky to mock `ResolveInfo` for `PackageManager` and this PR is fixing a crash that's important. 

Adding my thoughts here: the test was testing the implementation, verifying the calling of `startActivity`. If the implementation changes, which it did, but the output didn't, the test failed. Which brings attention to whether we were testing the right thing? Should we test the implementation or test the output result?

Amejia481:
> @Amejia481 Thanks for the review!
> 
> As discussed, removing the `HomeDeepLinkIntentProcessorTest.make_default_browser deep_link for above API 23` test as it is tricky to mock `ResolveInfo` for `PackageManager` and this PR is fixing a crash that's important.
> 
> Adding my thoughts here: the test was testing the implementation, verifying the calling of `startActivity`. If the implementation changes, which it did, but the output didn't, the test failed. Which brings attention to whether we were testing the right thing? Should we test the implementation or test the output result?
> 
> Added the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1829793) for tracking.

Yeah, I think the main goal with our tests is to have a safety net that alert us when when something is broken, ideally we should test the expected behaviour, but if it's near to impossible or super hard to do so, I prefer that we have tests that verify the implementation (Implicitly check the expected behaviour) than no having no tests no safety net.

rahulsainani:
@Mergifyio backport releases_v113

PR #1758, opened 2023-04-24
1826766 -  Add onboarding feature manifest

t-p-white:
Requirements for 114:

We build 3 cards, in the order:

- Set to Default
- Sign into Sync
- Opt-in to Push Notifications

Each card has the ability for Nimbus to configure:

- title copy
- body copy
- button copy

UX Requirements:

- user can swipe to a previous card
- user cannot swipe to a future card -> user must take a button action
- user understands how many steps are involved in the onboarding process

**Note**: Limitations for this PR - the privacy policy link is still handled internally by the app code. 

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/.experimenter.yaml
```diff
@@ -36,6 +36,9 @@ juno-onboarding:
   hasExposure: true
   exposureDescription: ""
   variables:
+    cards:
+      type: json
+      description: "Collection of user facing onboarding 'views'."
```
rahulsainani, line 41: 🔍 Should we call it cards, since it's called cards above? Views can be confused to the individual view, wdyt?

fenix/app/nimbus.fml.yaml
```diff
@@ -9,6 +9,8 @@ channels:
   - beta
   - nightly
   - developer
+includes:
+  - onboarding.fml.yaml
```
rahulsainani, line 13: 👏 Nice!

fenix/app/onboarding.fml.yaml
```diff
@@ -0,0 +1,119 @@
...
+        default: false
+      cards:
+        description: Collection of user facing onboarding 'views'.
+        type: Map<String, OnboardingCardData>
+        default:
+            default-browser:
+              card-type: default-browser
+              telemetry:
+                id: "default"
+                sequence-position: "1"
```
rahulsainani, line 20: 💭 This works, but I'm a afraid if this is changed by mistake by the experiment creators. For instance, if the ordering is updated without updating the sequence-position, it's a "bug" that results in wrong telemetry.

An alternative would be to get the `sequence-position` from position of the card in the list, advantage would be that it can be tested and that functionality won't need to change regardless of the number of cards added/removed. 

It does make sense to keep the `id` here as that's based on the card itself and whenever a new card is added, it makes people think of figuring out the telemetry id, although there might be risk of someone adding a really long id, IIRC, there's a glean extras character limit 🤔

fenix/app/src/main/res/values/strings.xml
```diff
@@ -285,6 +285,9 @@
     <!-- Title for set firefox as default browser screen.
         The first parameter is the name of the app defined in app_name (for example: Fenix) -->
     <string name="juno_onboarding_default_browser_title">Make %s your go-to browser</string>
+    <!-- Title for set firefox as default browser screen.
+        Note: The word 'Firefox' should not be translated -->
```
rahulsainani, line 289: 💡 Maybe we should add a note that this is used with Nimbus and that doesn't allow usage of arguments

fenix/app/onboarding.fml.yaml
t-p-white: > IIRC, there's a glean extras character limit

Interesting, I did not know this!

rahulsainani (approved):
Looks good! ✅

fenix/app/onboarding.fml.yaml
```diff
@@ -0,0 +1,100 @@
...
+      primary-button-label:
+        type: Text
+        description: The text to display on the primary button.
+        # This should never be defaulted.
+        default: ""
+      secondary-button-label:
+        type: Text
+        description: The text to display on the secondary button.
+        # This should never be defaulted.
+        default: ""
```
jhugman, line 88: For the sake of forwards-compatibility, I'd suggest a `primary-button` and `secondary-button` as button _objects_.

fenix/app/onboarding.fml.yaml
```diff
@@ -0,0 +1,100 @@
+---
+features:
+
+  juno-onboarding:
```
jhugman, line 4: This is a feature, but juno is release/marketing name. Suggest a simpler `onboarding` or `onboarding-cards`.

(I haven't thought deeply on whether it's okay to just add to the existing `onboarding` feature, or not; though I expect it's okay).

jhugman (approved):
This is looking really nice!

fenix/app/onboarding.fml.yaml
t-p-white: That was the initial plan, but following discussions with @rahulsainani we think it's best to keep it basic for now until more data fields are required 👍

fenix/app/onboarding.fml.yaml
t-p-white: Agreed, having juno in the name is something that there were previous discussions around. We have an [existing issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1826883) for this (thanks @rahulsainani) which will cover this 👍

PR #1773, opened 2023-04-25
Bug 1829862 - Add ApplicationServices.kt to the Mergify config and clean up some stale entries

rvandermeulen:


jcristau (approved):
Thanks!

PR #1811, opened 2023-04-27
Bug 1822349 - monorepo-migration: add convert-to-hg script

jcristau:
This script rewrites the git repo's history and converts it to mercurial, but stops short of merging it with mozilla-central for now.

The rewrite steps are:
- replacing octopus merges with a series of 2-parent merges, since mercurial doesn't have octopus merges, appending " (part N)" to the first line of the commit message for each
- moving all files from the repository root to `mobile/android/`
- replacing references to pull request numbers in commit messages with their full URL

Git-cinnabar is used for the conversion to mercurial, with a custom change to better detect file copies/moves.

cc @glandium @JohanLorenzo @gbrownmozilla 

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


monorepo-migration/convert-to-hg.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/sh
+
+set -e
+set -x
+
+CINNABAR=https://github.com/jcristau/git-cinnabar
+CINNABAR_BRANCH=detect-more-renames
```
JohanLorenzo, line 7: Nit: Let's add a code comment to remind future readers these value will eventually be changed.

monorepo-migration/convert-to-hg.sh
```diff
@@ -0,0 +1,88 @@
...
+SOURCE_REPO=https://github.com/mozilla-mobile/firefox-android
+SOURCE_REPO_NAME=${SOURCE_REPO##*/}
+SOURCE_BRANCH=main
+DEST_REPO="$(readlink -f "$1")"
+DEST_SUBDIR=mobile/android
+EXPRESSIONS_FILE_PATH="$(dirname $(readlink -f "$0"))/data/message-expressions.txt"
+
+TEMPDIR=$(mktemp -d)
+#trap 'rm -rf ${TEMPDIR} EXIT'
```
JohanLorenzo, line 16: Nit: Commented code. Let's remove it.

monorepo-migration/convert-to-hg.sh
```diff
@@ -0,0 +1,88 @@
...
+CINNABAR_PATH="${TEMPDIR}/git-cinnabar"
+SOURCE_REPO_PATH="${TEMPDIR}/${SOURCE_REPO_NAME}"
+
+setup_temporary_repo() {
+  git clone "${SOURCE_REPO}" "${SOURCE_REPO_PATH}"
+  git -C "${SOURCE_REPO_PATH}" config user.email=release@mozilla.com --config user.name="Mozilla Release Engineering"
```
JohanLorenzo, line 22: Nit: missing `--` to `config user.email`?

monorepo-migration/convert-to-hg.sh
```diff
@@ -0,0 +1,88 @@
...
+      'if test "$#" -lt 7; then
+         git commit-tree "$@"
+       else
+         commit_message=$(mktemp)
+         cat > $commit_message
+             
+         final_tree="$1"
+         shift
+         shift
+         p1="$1"
```
JohanLorenzo, line 45: Question: does `p1` stands for `parent_revision_1`?

monorepo-migration/convert-to-hg.sh
```diff
@@ -0,0 +1,88 @@
...
+         shift
+         part=1
```
JohanLorenzo, line 47: Nit: Future reader may believe `$` missing from `part=$1` (I just did 😅). Let's move the shift-statement below `part` to make the intent more explicit.

monorepo-migration/convert-to-hg.sh
```diff
@@ -0,0 +1,88 @@
...
+  make -C "${CINNABAR_PATH}"
+}
+
+rewrite_git_history() {
+  # replace octopus merges with a series of two-parent merges
+  FILTER_BRANCH_SQUELCH_WARNING=1 \
+  git -C "${SOURCE_REPO_PATH}" filter-branch \
+    --force \
+    --commit-filter \
+      'if test "$#" -lt 7; then
```
JohanLorenzo, line 36: Nit: I'd recommend documenting what the 7+ parameters are since we shift quite a lot below.

monorepo-migration/convert-to-hg.sh
```diff
@@ -0,0 +1,88 @@
...
+         git diff-tree --exit-code $tree $final_tree
+         echo "$p1"
+     fi' \
+        -- "$SOURCE_BRANCH"
+  # to-subdirectory-filter: move everything to $DEST_SUBDIR
+  # replace-message: replace references to pull requests with the full URL
+  # preserve-commit-hashes: because the filter-branch step also changes
+  #                         hashes, as does the conversion to hg, rewriting
+  #                         hashes here would only point at temporary commit
+  #                         objects which wouldn't help
```
JohanLorenzo, line 66: Question: Is `preserve-commit-hashes` faster than the default behavior? If not, I'm not sure I follow this comment 😅

JohanLorenzo (commented):
Looks good to me overall. I left some questions just to make sure I understand the PR correctly. If so, I'll r+ quickly 🙂

monorepo-migration/convert-to-hg.sh
jcristau: I'll probably uncomment it instead.

monorepo-migration/convert-to-hg.sh
jcristau: Well, I'm not sure they will...

monorepo-migration/convert-to-hg.sh
jcristau: Hmm, no, but this does look wrong, will fix.

monorepo-migration/convert-to-hg.sh
jcristau: It stands for first parent, yes.

monorepo-migration/convert-to-hg.sh
jcristau: How is that more explicit?  To me the shift goes along with `p1="$1"`, so should stay next to it.

monorepo-migration/convert-to-hg.sh
jcristau: Can do.

monorepo-migration/convert-to-hg.sh
jcristau: preserve-commit-hashes means we don't rewrite hashes in commit messages to refer to temporary commits that won't be living outside the mid-rewrite git repository, and they'll keep referring to pre-migration commit hashes in the firefox-android repo; that seems preferable to me.

monorepo-migration/convert-to-hg.sh
glandium: I'm open to adding a config to tweak rename/copy detection in git-cinnabar.

jcristau: Oh cool.  I'll look into that thenm thanks @glandium

monorepo-migration/convert-to-hg.sh
JohanLorenzo: Understood. In this case, we should fix it this way, if I'm not mistaken.

```suggestion
         shift
         part=$1
```

monorepo-migration/convert-to-hg.sh
JohanLorenzo: Okay, let's rename `p1` into `parent1` to be more explicit, then.

monorepo-migration/convert-to-hg.sh
jcristau: Nope, that's very different...

jcristau: (`part` is the loop iteration counter)

monorepo-migration/convert-to-hg.sh
JohanLorenzo: Okay. Will commit hashes still be present in commit messages? Would you have an example of the final result?

jcristau: Yes, for example https://github.com/mozilla-mobile/firefox-android/commit/5620f848cfc8e82ffa7fb22a402eeeeb945cccb1 becomes something like https://hg.mozilla.org/users/jcristau_mozilla.com/firefox-android-hg/rev/d3adbd8f0c94 and the reference to 3c660fe68b872a9f20618846a96528ab20bfefcf is left unchanged.

monorepo-migration/convert-to-hg.sh
jcristau: mercurial very commonly calls them p1/p2, including in user-facing revsets and templates, so I'd prefer to keep it that way.

monorepo-migration/convert-to-hg.sh
JohanLorenzo: Understood! I agree it's better to keep the original hash here rather than referring to one that will only exist on the machine handling the migration. However, I think it's going to confuse future readers. For instance, future readers may believe the referenced commit above is an HG rev, meaning https://hg.mozilla.org/users/jcristau_mozilla.com/firefox-android-hg/rev/3c660fe68b872a9f20618846a96528ab20bfefcf should be valid.

I think we can fix that by pre-prepending `https://github.com/mozilla-mobile/firefox-android/commit/` to the referenced commit. This seems to be feasible with `git filter-repo --commit-callback`[1]. Let me know what you think! 

[1] https://github.com/newren/git-filter-repo/blob/4bc9022afce5e2e138596bbecf4df310212ae164/Documentation/git-filter-repo.txt#L1178

jcristau: I'll take a stab at that, maybe in a followup PR?

monorepo-migration/convert-to-hg.sh
jcristau: Added a comment for now.

monorepo-migration/convert-to-hg.sh
JohanLorenzo: All right, sounds good to me.

JohanLorenzo (approved):
LGTM 👍 I agree, let's handle the original hash in a follow-up PR 🙂

PR #1814, opened 2023-04-27
Bug 1810197 - Improve DownloadUtils.changeExtension.

Amejia481:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1810197


gabrielluong:
@Amejia481 Do you have another reviewer you could ping?

Amejia481:
No one specific,  but happy to provide context to anyone that it's interested.

MatthewTighe:
I can take a look next week if you'd like to sync up on the context Arturo

Amejia481:
Awesome, thanks @MatthewTighe !

Epale10 (approved):

Amejia481:
For more context: 

This patch have two main goals:

1) Refactor the actual `changeExtension` function to make it easy to understand.
2) Address [Bug #1810197](https://bugzilla.mozilla.org/show_bug.cgi?id=1810197).

With the previous implementation, we were updating some file extensions when was not needed, causing [Bug #1810197](https://bugzilla.mozilla.org/show_bug.cgi?id=1810197).

The goal of `changeExtension` function is to update the file extension, if we have a clear indication that the actual file extension is wrong based on the `providedMimeType`, for that reason we verify if the file has the right extension based on the `providedMimeType`, if so me don't do anything, otherwise we take the extension from the `providedMimeType` and update the file extension.

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
```diff
@@ -335,26 +335,30 @@ object DownloadUtils {
...
-        return if (extension != null) {
-            filename.substring(0, dotIndex) + extension
+    @Suppress("ReturnCount")
+    private fun changeExtension(filename: String, providedMimeType: String?): String {
+        if (providedMimeType == null) return filename
+
+        val file = File(filename)
+        val mimeTypeMap = MimeTypeMap.getSingleton()
+        val extensionFromMimeType =
+            mimeTypeMap.getExtensionFromMimeType(providedMimeType) ?: return filename
```
MatthewTighe, line 345: nit: If we moved this line above the first return, we could combine them and potentially get rid of the lint suppression:

```
        val extensionFromMimeType =
            mimeTypeMap.getExtensionFromMimeType(providedMimeType) 
       if (providedMimeType == nul || extensionFromMimeType == null) return filename
```

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
```diff
@@ -335,26 +335,30 @@ object DownloadUtils {
...
+        val mimetypeFromFileName = mimeTypeMap.getMimeTypeFromExtension(file.extension) ?: ""
```
MatthewTighe, line 346: nit:

```suggestion
        val mimeTypeFromFilename = mimeTypeMap.getMimeTypeFromExtension(file.extension) ?: ""
```

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
```diff
@@ -335,26 +335,30 @@ object DownloadUtils {
...
+
+        val fileHasPossibleExtension = filename.contains(extensionFromMimeType, ignoreCase = true)
+        val isFileMimetypeDifferentFromProvidedMimeType =
```
MatthewTighe, line 349: nit:

```suggestion
        val isFileMimeTypeDifferentFromProvidedMimeType =
```

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
```diff
@@ -335,26 +335,30 @@ object DownloadUtils {
...
+                ignoreCase = true,
+            )
+
+        // Mimetypes could have multiple possible file extensions, for example: text/html could have
+        // either .htm or .html extensions, we try to rename only when there is a clear indication
+        // that the file's extension is wrong, as [getExtensionFromMimeType] returns the first possible,
+        // extension we could adding the wrong extension.
+        // More examples can be found https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+        return if (isFileMimetypeDifferentFromProvidedMimeType && !fileHasPossibleExtension) {
+            return "${file.nameWithoutExtension}.$extensionFromMimeType"
```
MatthewTighe, line 361: I'm not sure I'm understanding what this patch does differently in order to fix the underlying bug. It looks like they both:
1. convert the filename extension and provided mime type to an "official" mime type
2. check if the mime type matches 
    1. if they do return the original filename
    2. if they do not return the corrected filename using the provided mimetype

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/DownloadUtilsTest.kt
```diff
@@ -128,6 +140,11 @@ class DownloadUtilsTest {
         assertEquals("file.data", DownloadUtils.guessFileName(null, null, "http://example.com/file.data", "application/octet-stream"))
         assertEquals("file.data", DownloadUtils.guessFileName(null, null, "http://example.com/file.data", "binary/octet-stream"))
         assertEquals("file.data", DownloadUtils.guessFileName(null, null, "http://example.com/file.data", "application/unknown"))
+
+        assertEquals("file.jpg", DownloadUtils.guessFileName(null, null, "http://example.com/file.zip", "image/jpeg"))
+        // Should not change to file.dll
+        assertEquals("file.exe", DownloadUtils.guessFileName(null, null, "http://example.com/file.exe", "application/x-msdos-program"))
```
MatthewTighe, line 146: Should we also have a test that demonstrates that `.exe` will be correctly converted to? Something like:
`assertEquals("file.exe", DownloadUtils.guessFileName(null, null, "http://example.com/file.bin", "application/x-msdos-program"))`

MatthewTighe (commented):
The refactor is much more understandable, thanks for the efforts there! I have a few nits but also want to just check my understanding of how this patch is solving the underlying bug

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
Amejia481: If I remember correctly (When I tested 😅 ) , The main difference is that now we are checking ahead we that we already have the right extension applied, before we failed to do so, causing the [bug 1810197](https://bugzilla.mozilla.org/show_bug.cgi?id=1810197).

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/DownloadUtilsTest.kt
Amejia481: Sure, added!

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
MatthewTighe: Okay, I puzzled through it because I was still a little confused. As suggested in the bug, I think this line was wrong in the original:
`(typeFromExt?.equals(mimeType, ignoreCase = true) != false)`. 

That could also be written as
`(typeFromExt?.equals(mimeType, ignoreCase = true) == true)`, which means we would only look up the extension for the provided mime type when the extension already matched it. This would cause us to never replace the extension with a different type as specified by the provided mime type.

I think in your version the logic is actually handled correctly, but I am not sure why the contributor patch was reverted now. I am happy to accept that your version is doing the thing we want, but I want to just double check again that I'm not missing anything. Thanks!

MatthewTighe (approved):
I'm going to approve this now in case my review is blocking you, but I do want to call attention to the [comment I left in an earlier thread](https://github.com/mozilla-mobile/firefox-android/pull/1814#discussion_r1198356515) in case you hadn't seen it.

android-components/components/support/utils/src/main/java/mozilla/components/support/utils/DownloadUtils.kt
Amejia481: I added a comment on the contributor's [PR](https://github.com/mozilla-mobile/firefox-android/pull/1572/files#r1179593362) :)

PR #1843, opened 2023-04-29
Bug 1830602 - Update AndroidX Core to version 1.10.0

rvandermeulen:
https://bugzilla.mozilla.org/show_bug.cgi?id=1830602
https://bugzilla.mozilla.org/show_bug.cgi?id=1830602


mcarare (approved):
LGTM!

mcarare:
I have also opened https://bugzilla.mozilla.org/show_bug.cgi?id=1830838 to look into the `ParcelCompat ` methods introduced in this update, and if we can use them instead of our AC compat methods.

PR #1862, opened 2023-05-01
Bug 1830780 - Add support for adding distribution based default search engines list

rocketsroger:
This is to allow us to add a distribution search engine list in list.json.  This allow us to support different engines when a distribution is detected.  (Example: through Adjust campaign ID).

The new list would be added to the list.json like so:

        "visibleDefaultEngines": [
          "google-b-m", "bing", "amazondotcom", "ddg", "ebay", "wikipedia"
        ],
        "testing": [
          "google-b-m", "ebay", "wikipedia"
        ]

And on the Fenix side if we specifies "testing" as the distribution, we would retrieve the newly added search engines list.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1830780


android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/BundledSearchEnginesStorage.kt
```diff
@@ -37,9 +37,10 @@ internal class BundledSearchEnginesStorage(
     override suspend fun load(
         region: RegionState,
         locale: Locale,
+        distribution: String?,
         coroutineContext: CoroutineContext,
     ): SearchMiddleware.BundleStorage.Bundle = withContext(coroutineContext) {
-        val localizedConfiguration = loadAndFilterConfiguration(context, region, locale)
+        val localizedConfiguration = loadAndFilterConfiguration(context, region, locale, distribution)
         val searchEngineIdentifiers = localizedConfiguration.visibleDefaultEngines
```
sarah541, line 44: `val searchEngineIdentifiers = localizedConfiguration.visibleDefaultEngines`

Everything looks good apart from this naming convention now. Since it might not be `visibleDefaultEngines` everytime, we should probably think of a more generic one.. thoughts?

sarah541 (approved):

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/BundledSearchEnginesStorage.kt
rocketsroger: That's a good point, I'll try out some names to see if that makes the code more clear.  Thanks

rocketsroger: I've renamed it to `visibleSearchEngines`  Thanks

PR #1867, opened 2023-05-02
Bug 1830112 - Display full text in compose button when large font

DreVla:
When using the compose button and the size of the font is set to largest for accessibility purposes, the font should not cut if it doesn't fit. Instead, the button accommodates the entirety of the text.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1830112

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
@@ -19,11 +19,15 @@ import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.painter.Painter
+import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
 import org.mozilla.fenix.R
 import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.theme.FirefoxTheme
+
+const val DEFAULT_MAX_LINES = 2
 
 /**
  * Base component for buttons.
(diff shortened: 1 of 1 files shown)


rahulsainani:
@Piratesarereal can you please take a look at this PR when you have a moment.

Piratesarereal:
Yes, this makes sense.

DreVla:
According to last decision from UX team the button's maxLines attribute is now set by default to 2 and if font is increased due to accessibility reasons, the maxLines is set to no limit. The button now also has an additional boolean parameter, enableAccessibilityDetection which by default is set to true. This will allow us to control if we want the button to change size when a larger font is detected or not.

Screenshot 2023-05-15 at 15 13 20

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -65,9 +73,10 @@ private fun Button(
 
         Text(
             text = text,
+            textAlign = TextAlign.Center,
             color = textColor,
             style = FirefoxTheme.typography.button,
-            maxLines = 1,
+            maxLines = if (enableAccessibilityDetection && fontScale > 1.0f) Int.MAX_VALUE else 2,
```
rahulsainani, line 79: wondering if we should hoist the maxLines param in the button's params and use that instead of hardcoded 2. Maybe there are callers who would want to set it to 1.

MozillaNoah: I'd lean towards not making it a parameter (at least for now). There is no present need to be able to deviate from the default (now 2 lines), so I'd opt to not including it so there's no confusion as to whether to use this optional when these buttons are used in future screens by developers who don't have the context we're laying out here. If that need ever does arise, we'll hopefully be doing so via guidance from UX + design and it'll be a quick PR to make that change to the `Button` API.

rahulsainani: fair point, we can defer updating the api till we actually have a usecase. I guess gl is making the same point for `maxLines` below.

rahulsainani (approved):
looks good to me 👍

DreVla:
@mergify backport releases_v113

DreVla:
@mergify backport releases_v114

PR #1901, opened 2023-05-04
Bug 1831316 - Publish AC to Maven during the ship phase

rvandermeulen:
https://bugzilla.mozilla.org/show_bug.cgi?id=1831316
https://bugzilla.mozilla.org/show_bug.cgi?id=1831316
https://bugzilla.mozilla.org/show_bug.cgi?id=1831316


jcristau (changes requested):
Seems fine on principle.

One thing though: release-notify-push has a dependency on post-beetmover so that'll need to change.

jcristau:
> One thing though: release-notify-push has a dependency on post-beetmover so that'll need to change.

The transforms (https://github.com/mozilla-mobile/firefox-android/blob/main/taskcluster/android_taskgraph/transforms/release_deps.py specifically) end up removing that dependency here so things end up working out, but it's confusing.

jcristau (approved):
Thanks!

PR #1903, opened 2023-05-04
Bug 1834974 - run external-gradle-dependencies job on changes to dependencies

jcristau:
There's a tradeoff between running this job too often (and waiting for it to be done each time before anything else can run) vs not often enough and dealing with breakage when it should have run and didn't.  It's unclear whether this goes too far in the other direction, so leaving this as a draft for now.
https://bugzilla.mozilla.org/show_bug.cgi?id=1834974
https://bugzilla.mozilla.org/show_bug.cgi?id=1834974
https://bugzilla.mozilla.org/show_bug.cgi?id=1834974


rvandermeulen:
We just hit a scenario yesterday where a dependency update in a PR targeting the `releases_v114` branch triggered a cache rebuild for the first time since the branch was created off main. Along the way, we found out that there was a bug in the Application Services 114.0 release which caused the rebuilds to fail due to not being able to find the nimbus-fml plugin. This ended up being a blocking issue for the build which wasn't hit until after EOD for most people. The underlying problem was likely there for the entire cycle, but unnoticed due to the lack of anything in CI trying to fetch the nimbus-fml plugin up to that point.

IMO, this seems like a solid-enough reason to consider moving forward with this change. If this had already been the case, it would have meant a delay in building 114.0b1 on merge day as the A-S 114.0 release bump would have been blocked from landing, but it at least would have been during a time of day when we weren't scrambling to find someone to investigate.

rvandermeulen (approved):

gbrownmozilla (approved):

PR #1908, opened 2023-05-04
Bug 1822552: Prevent search group title from being cut

Alexandra-Virvara:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1822552

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
@@ -179,7 +179,7 @@ private fun RecentlyVisitedHistoryGroup(
             RecentlyVisitedTitle(
                 text = recentVisit.title,
                 modifier = Modifier
-                    .padding(top = 7.dp, bottom = 2.dp)
+                    .padding(top = 4.dp)
                     .weight(1f)
                     .semantics {
                         testTagsAsResourceId = true


fenix/app/src/main/java/org/mozilla/fenix/home/recentvisits/view/RecentlyVisited.kt
```diff
@@ -179,7 +179,7 @@ private fun RecentlyVisitedHistoryGroup(
             RecentlyVisitedTitle(
                 text = recentVisit.title,
                 modifier = Modifier
-                    .padding(top = 7.dp, bottom = 2.dp)
+                    .padding(top = 4.dp)
```
gabrielluong, line 182: @Alexandra-Virvara I am just curious how we arrived at these values? Typically, we want to follow the figma specs and be guided by UX. My concern here is that we are changing the spacing drastically which might go again the desired look at that UX is aiming for. That said we should ensure that the text aren't cut off as well.

gabrielluong: We should also have UX review the visual change and possibly updating the figma spec to follow the same padding values. This is to avoid undoing the changes if anyone had to work with the figma spec on updating anything related to this feature.

gabrielluong: Some additional thoughts about a different way to go about this. 

@MozillaNoah I believe this is probably caused by the hardcoded size on [L161](https://github.com/mozilla-mobile/firefox-android/pull/1908/files/60760a9fc524995f8e8b0607279e5b333a89e5b1#diff-c1cf213559053edc534f70494d9a5d2d9efad3be191377cddb73eef18f2c9521R161). I think we have already seen this happen elsewhere and we simply removed the hardcoded container size and simply made it "wrap"? Could you confirm?

That way we don't require a UX review while preserving the existing spacing.

MozillaNoah: It's likely caused by the fixed height, but a `Spacer` would need to be added above the `Divider`, otherwise the text may butt right up to the line. 

The text elements, going off of a figma I found, look like they should be using our `TextListItem` Composable, too, but not sure if that's out of the scope for this bug.

mavduevskiy (changes requested):
hey @Alexandra-Virvara I was wondering what's the status of this PR?

While your solution works and removes the cut off corner, hard-coding a new padding would actually stray from the [design](https://www.figma.com/file/HWTT1sbVrKtD2gxX3IzAM7/Home?type=design&node-id=2030-56391&mode=design&t=NKYlUaPgCQQzQfAy-4). I don't think we should aim for keeping the 7dp (which is an odd number for Android), but it looks like the text box is positioned center-vertically in the container (as is the icon).

Screenshot 2023-07-21 at 6 42 26 PM

It looks like rearranging the composable a bit will allow us to get rid of cut letters and keep the design.

gabrielluong:

PR #1936, opened 2023-05-08
Bug 1831352 - Target JVM 17 bytecode.

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1830872
https://bugzilla.mozilla.org/show_bug.cgi?id=1831352


rvandermeulen:
Would it make sense to factor the sourceCompatibility and targetCompatibility values into a common spot so it doesn't have to be changed all over the place every time we need to bump?

mcarare:
Maybe we could do that. Not expecting frequent changes, the last ones were due to us being kind of behind at some point.

taskcluster/docker/base/Dockerfile
```diff
@@ -30,16 +30,15 @@ RUN apt-get update -qq \
...
     # which we cannot navigate while building the Docker image.
     && apt-get install -y tzdata \
-    && apt-get install -y openjdk-11-jdk \
+    && apt-get install -y openjdk-17-jdk \
                           wget \
                           expect \
                           git \
                           curl \
-                          python \
-                          python-pip \
```
JohanLorenzo, line 39: Good catch 👍

taskcluster/docker/base/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-FROM ubuntu:18.04
+FROM ubuntu:22.04
```
JohanLorenzo, line 5: Nice!

focus-android/tools/docker/Dockerfile
JohanLorenzo, line 1: I wonder if we still use this file 🤔 For sure, it's not used by Taskcluster.

JohanLorenzo (approved):
The Taskcluster part looks good to me! 👍

focus-android/tools/docker/Dockerfile
mcarare: I remember forgetting to change this and all things still went on... So... not sure 😅

fenix/app/src/test/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsDetailsExceptionsFragmentTest.kt
```diff
@@ -64,7 +65,9 @@ class SitePermissionsDetailsExceptionsFragmentTest {
 
     @Test
     fun `WHEN initPhoneFeature is called THEN the feature label must be bind and a click listener must be attached`() {
-        val feature = spyk(PhoneFeature.CAMERA)
+        mockkObject(PhoneFeature.CAMERA)
```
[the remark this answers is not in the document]
jonalmeida: If we do need this, can we unmockk at the end of the test?

focus-android/app/src/main/java/org/mozilla/focus/utils/IntentUtils.kt
```diff
@@ -19,8 +19,8 @@ object IntentUtils {
      * This additional requirement improves your app's security.
      * FLAG_IMMUTABLE -> Flag indicating that the created PendingIntent should be immutable.
      */
-    val defaultIntentPendingFlags
-        get() = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+    fun defaultIntentPendingFlags(buildVersion: Int = Build.VERSION.SDK_INT): Int =
```
jonalmeida, line 22: Aside from the IntentUtilsTest, are there any other places we use this as a param? I don't follow the reason for this change as part of the JVM 17 upgrade. 🤔

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/test/ReflectionUtils.kt
```diff
@@ -4,18 +4,38 @@
...
 import java.lang.reflect.Field
 import java.lang.reflect.Modifier
+import java.security.AccessController
+import java.security.PrivilegedAction
 
 object ReflectionUtils {
     fun <T : Any> setField(instance: T, fieldName: String, value: Any?) {
-        val originField = instance.javaClass.getField(fieldName)
+        val mapField = AccessController.doPrivileged(
+            PrivilegedAction {
```
jonalmeida, line 16: The comment on this class says "Legacy security code; do not use", and "Android-changed: Stubbed the implementation.  Android doesn't support SecurityManager."

The android comment we can ignore since this is a test utils class that will not be run on Android, but I'm not sure about the first one.

The [javadocs also say](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/security/PrivilegedAction.html), we ought to use the `PrivilegedExceptionAction` instead for calling something that throws.

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/test/ReflectionUtils.kt
```diff
@@ -4,18 +4,38 @@
...
-        val modifiersField = Field::class.java.getDeclaredField("modifiers")
-        modifiersField.isAccessible = true
-        modifiersField.setInt(originField, originField.modifiers and Modifier.FINAL.inv())
+        mapField.set(instance, value)
+    }
+}
 
-        originField.isAccessible = true
-        originField.set(instance, value)
+private fun Field.removeFinal() {
```
jonalmeida, line 32: Stub comment: I don't understand this enough to review it yet but I don't want to hold up the rest of the review comments while I do some research.

android-components/components/feature/prompts/src/test/java/mozilla/components/feature/prompts/file/FilePickerTest.kt
```diff
@@ -112,22 +112,18 @@ class FilePickerTest {
 
     @Test
     fun `handleFilePickerRequest with the required permission will call startActivityForResult`() {
```
jonalmeida, line 114: Nice fixup. 🙂

focus-android/app/src/main/java/org/mozilla/focus/utils/IntentUtils.kt
mcarare: This is needed because we can no longer use reflection to change the `SDK_INT value`. Instead, if we can pass it as a param we can easily modify the value in tests. 
I would actually prefer to do this anywhere, even if we could have used reflection.

fenix/app/src/test/java/org/mozilla/fenix/settings/sitepermissions/SitePermissionsDetailsExceptionsFragmentTest.kt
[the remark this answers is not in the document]
mcarare: I refactored the test remove the need for it.

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/test/ReflectionUtils.kt
mcarare: seems this not even needed.

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/test/ReflectionUtils.kt
mcarare: I switch to using `PrivilegedExceptionAction`

mcarare:
@Mergifyio refresh

PR #1943, opened 2023-05-08
Bug 1831939 - Cloud Functions Test [WIP]

AaronMT:
Testing a `moz-focus-android` Cloud Function using `googleapis` and `firebase-functions` to read through a `testMatrix` object to obtain a list of `testCases` and read `testCase` data for on-event (e.g, `testMatrix().onComplete`) reading/ exploration.


boek:
@AaronMT Is this still something that we want to do?

AaronMT:
Not at the moment, was trying out Cloud Function deployments.

PR #1979, opened 2023-05-10
Bug 1832416 - Update Material Components to version 1.9.0

rvandermeulen:
See the bug for the reason behind adding the drawerlayout dependency as well. I targeted it to the component that was specifically failing to build (session-storage), but if it should be included at a higher level, let me know!
https://bugzilla.mozilla.org/show_bug.cgi?id=1832416
https://bugzilla.mozilla.org/show_bug.cgi?id=1832416


mcarare (approved):
Went through all the release notes (🤯 ) and nothing stands out as something that could regress UI/ behavior significantly.

See also bz comment on the issue.

PR #1988, opened 2023-05-11
Bug 1830895 - Add observeAsComposableState compose extension

rahulsainani:
### What
An `observeAsComposableState` that returns `ComposeState<R>`.

### Why?
The existing `observeAsComposableState` extension functions returns  `ComposeState<R?>` even though `R` might be non nullable in the Store State. This forces the callers to do the null check. 

This change aims to simplify the usage on the caller side, changing the usage from:
```
val addresses: State<List<Address>?> = store.observeAsComposableState { state -> state.addresses }
addresses.value ?: emptyList()
```
to
```
val addresses: State<List<Address>> = store.observeAsComposableState(store.state.addresses) { state -> state.addresses }
```

Note that [addresses is a non nullable list](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/settings/autofill/AutofillFragmentStore.kt#31) in the Store State and even then the caller had to take care of the case.

### How
Using `produceState`:  This returns an observable `ComposeState` that produces values over time. The lambda (`producer`) is called when  `produceState` enters the composition and is cancelled when it leaves the composition. 
`ProduceStateScope.value = map(browserState)` sets the value on the returned `ComposeState`.

Note: `ComposeState` is actually `androidx.compose.runtime.State`.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1830895


android-components/components/lib/state/src/main/java/mozilla/components/lib/state/ext/ComposeExtensions.kt
```diff
@@ -43,6 +44,35 @@ fun <S : State, A : Action, R> Store<S, A>.observeAsComposableState(map: (S) ->
...
+ * [LifecycleOwner] moves to the [Lifecycle.State.DESTROYED] state.
+ *
+ * @param initialMappedValue Initial value emitted.
+ * @param map The applied function to produced the mapped value [R] from [S].
+ * @return A non nullable [ComposeState], making the api more reasonable for callers where the
+ * state is non null.
+ */
+@Composable
+fun <S : State, A : Action, R> Store<S, A>.observeAsComposableState(
+    initialMappedValue: R,
```
gabrielluong, line 63:
```suggestion
    initialValue: R,
```

gabrielluong (approved):

android-components/components/lib/state/src/main/java/mozilla/components/lib/state/ext/ComposeExtensions.kt
```diff
@@ -43,6 +44,35 @@ fun <S : State, A : Action, R> Store<S, A>.observeAsComposableState(map: (S) ->
...
+ * The [Store] observer will automatically be removed when this composable disposes or the current
+ * [LifecycleOwner] moves to the [Lifecycle.State.DESTROYED] state.
+ *
+ * @param initialMappedValue Initial value emitted.
+ * @param map The applied function to produced the mapped value [R] from [S].
+ * @return A non nullable [ComposeState], making the api more reasonable for callers where the
+ * state is non null.
+ */
+@Composable
+fun <S : State, A : Action, R> Store<S, A>.observeAsComposableState(
```
gabrielluong, line 62: I am wondering if there's a reason we would want to keep the above `observeAsComposableState` that would return the nullable `ComposeState` after this?

rahulsainani: Yeah it's worth exploring whether we can pass in `store.state.r` as `initialValue` for all cases and then maybe remove the other variant.

MatthewTighe: I filed [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1832624) to track this as a follow-up, because I think it's worth doing.

MatthewTighe (approved):

android-components/components/lib/state/src/androidTest/java/mozilla/components/lib/state/ext/ComposeExtensionsKtTest.kt
```diff
@@ -55,6 +55,49 @@ class ComposeExtensionsKtTest {
         }
     }
 
+    @Test
+    fun usingInitialMappedValueWithUpdates() {
```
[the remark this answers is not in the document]
gabrielluong: I would also suggest moving this below `usingInitialValue` as well

android-components/components/lib/state/src/androidTest/java/mozilla/components/lib/state/ext/ComposeExtensionsKtTest.kt
```diff
@@ -55,6 +55,49 @@ class ComposeExtensionsKtTest {
...
+            assertEquals(listOf("Loading"), value)
+        }
+
+        store.dispatchBlockingOnIdle(TestAction.IncrementAction)
+        store.dispatchBlockingOnIdle(TestAction.IncrementAction)
+        store.dispatchBlockingOnIdle(TestAction.IncrementAction)
+        store.dispatchBlockingOnIdle(TestAction.IncrementAction)
+
+        // Value after 4 increments, aka counter is 4. Note that it doesn't recompose here
+        // as the mapped value has stayed the same. We have 1 item in the list and not 5.
```
gabrielluong, line 87: Perhaps move all of these comments before their assertion. This is simply follows the pattern that has already been set by the test case below.

android-components/components/lib/state/src/main/java/mozilla/components/lib/state/ext/ComposeExtensions.kt
rahulsainani: Thanks for creating that bug!

android-components/components/lib/state/src/main/java/mozilla/components/lib/state/ext/ComposeExtensions.kt
```diff
@@ -43,6 +44,35 @@ fun <S : State, A : Action, R> Store<S, A>.observeAsComposableState(map: (S) ->
...
+ * The [Store] observer will automatically be removed when this composable disposes or the current
+ * [LifecycleOwner] moves to the [Lifecycle.State.DESTROYED] state.
+ *
+ * @param initialValue Initial value emitted.
+ * @param map The applied function to produced the mapped value [R] from [S].
+ * @return A non nullable [ComposeState], making the api more reasonable for callers where the
+ * state is non null.
+ */
+@Composable
+fun <S : State, A : Action, R> Store<S, A>.observeAsComposableState(
```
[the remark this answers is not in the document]
rahulsainani: Hmm yeah i see your point and in most other projects `observeAsState` would read better as there would be only compose state. But since we also have `lib-state` it can be confusing for the reader. `ComposableState` suffix is just a bit more explicit.

But, to counter my own point above, since we are in the composable scope, it's clear, that here it means `androidx.compose.runtime.State`. I agree, we can shorten the name. I'll leave this open for a couple of days for more comments and hopefully we can take a call by Monday.

android-components/components/lib/state/src/main/java/mozilla/components/lib/state/ext/ComposeExtensions.kt
```diff
@@ -43,6 +44,35 @@ fun <S : State, A : Action, R> Store<S, A>.observeAsComposableState(map: (S) ->
...
+ */
+@Composable
+fun <S : State, A : Action, R> Store<S, A>.observeAsState(
+    initialValue: R,
+    map: (S) -> R,
+): ComposeState<R> {
+    val lifecycleOwner = LocalLifecycleOwner.current
+
+    return produceState(initialValue = initialValue) {
+        val subscription = observe(lifecycleOwner) { browserState ->
```
jonalmeida, line 69: 🚗 Drive-by nit: naming.

`browserState` -> `state` or alternatives.

PR #1997, opened 2023-05-11
Bug 1831602 - Convert two UI tests to support composified Tabs Tray

MozillaNoah:
Rather than rewrite the existing logic inside of [`TabDrawerRobot`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/TabDrawerRobot.kt) and [`TabbedBrowsingTest`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/TabbedBrowsingTest.kt), I created an intermediary `ComposeTabbedBrowsingTest` and `ComposeTabDrawerRobot` to perform the conversion, so that there aren’t breaking changes in 15+ files AND the two suites of tests can live in tandem until the switch-over happens.

In `ComposeTabDrawerRobot`, I opted for extension functions to obtain the Composables, as that would allow the logic of test tags, etc., to live in one unified place and could be used by both `ComposeTabDrawerRobot` and `ComposeTabDrawerRobot.Transition`. These extension functions, at-large, currently do not live in their own file, as calls to those Composables should, ideally, only happen within the `ComposeTabDrawerRobot` file.

The ignored tests will be handled in follow-up tickets. See this [meta](https://bugzilla.mozilla.org/show_bug.cgi?id=1810775), section 4.5, for the full list.

I also created a Kotlin object, `TabsTrayTestTag`, to contain all of the test tags in a central location and so the test tag naming can feed off of each other.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1831602

### DIFF
--- a/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ComposeTabbedBrowsingTest.kt
+++ b/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ComposeTabbedBrowsingTest.kt
@@ -1,0 +1,382 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.ui
+
+import androidx.compose.ui.test.junit4.AndroidComposeTestRule
+import androidx.test.platform.app.InstrumentationRegistry
+import androidx.test.uiautomator.UiDevice
+import okhttp3.mockwebserver.MockWebServer
+import org.junit.After
+import org.junit.Before
+import org.junit.Ignore
+import org.junit.Rule
+import org.junit.Test
+import org.mozilla.fenix.helpers.AndroidAssetDispatcher
+import org.mozilla.fenix.helpers.HomeActivityTestRule
+import org.mozilla.fenix.helpers.RetryTestRule
+import org.mozilla.fenix.helpers.TestAssetHelper
+import org.mozilla.fenix.ui.robots.homeScreen
+import org.mozilla.fenix.ui.robots.navigationToolbar
+
+/**
+ *  Tests for verifying basic functionality of tabbed browsing
+ *
+ *  Including:
+ *  - Opening a tab
+ *  - Opening a private tab
+ *  - Verifying tab list
+ *  - Closing all tabs
+ *  - Close tab
+ *  - Swipe to close tab (temporarily disabled)
+ *  - Undo close tab
+ *  - Close private tabs persistent notification
+ *  - Empty tab tray state
+ *  - Tab tray details
+ *  - Shortcut context menu navigation
+ */
+
+class ComposeTabbedBrowsingTest {
+    private lateinit var mDevice: UiDevice
+    private lateinit var mockWebServer: MockWebServer
+
+    @get:Rule(order = 0)
+    val composeTestRule =
+        AndroidComposeTestRule(
+            HomeActivityTestRule.withDefaultSettingsOverrides(
+                tabsTrayRewriteEnabled = true,
+            ),
+        ) { it.activity }
+
+    @Rule(order = 1)
+    @JvmField
+    val retryTestRule = RetryTestRule(3)
+
+    @Before
+    fun setUp() {
+        mDevice = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
+        mockWebServer = MockWebServer().apply {
+            dispatcher = AndroidAssetDispatcher()
+            start()
+        }
+    }
+
+    @After
+    fun tearDown() {
+        mockWebServer.shutdown()
+    }
+
+    @Test
+    fun openNewTabTest() {
+        val defaultWebPage = TestAssetHelper.getGenericAsset(mockWebServer, 1)
+
+        navigationToolbar {
+        }.enterURLAndEnterToBrowser(defaultWebPage.url) {
+            mDevice.waitForIdle()
+            verifyTabCounter("1")
+        }.openComposeTabDrawer(composeTestRule) {
+            verifyNormalBrowsingButtonIsSelected()
+            verifyExistingOpenTabs("Test_Page_1")
+            closeTab()
+        }
+        homeScreen {
+        }.openComposeTabDrawer(composeTestRule) {
+            verifyNoOpenTabsInNormalBrowsing()
+        }.openNewTab {
+        }.submitQuery(defaultWebPage.url.toString()) {
+            mDevice.waitForIdle()
+            verifyTabCounter("1")
+        }.openComposeTabDrawer(composeTestRule) {
+            verifyNormalBrowsingButtonIsSelected()
+            verifyExistingOpenTabs("Test_Page_1")
+        }
+    }
+
+    @Test
+    fun openNewPrivateTabTest() {
+        val defaultWebPage = TestAssetHelper.getGenericAsset(mockWebServer, 1)
+
+        homeScreen {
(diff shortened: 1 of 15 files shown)


fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTestTag.kt
```diff
@@ -0,0 +1,27 @@
+package org.mozilla.fenix.tabstray
```
MatthewTighe, line 1: nit: license is missing

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTestTag.kt
```diff
@@ -0,0 +1,27 @@
...
+
+internal object TabsTrayTestTag {
```
MatthewTighe, line 3: I wish Compose offered a better way of interacting with view objects so that we didn't have to do this, but I think this makes more sense than replicating these strings across the Composables and the test code.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ComposeTabbedBrowsingTest.kt
```diff
@@ -0,0 +1,380 @@
...
+            verifyPrivateTabsList()
+            verifyPrivateBrowsingButtonIsSelected()
+        }.toggleToNormalTabs {
+            verifyNoOpenTabsInNormalBrowsing()
+        }.toggleToPrivateTabs {
+            verifyPrivateTabsList()
+        }
+    }
+
+    @Ignore("Being converted in: https://bugzilla.mozilla.org/show_bug.cgi?id=1831602")
```
MatthewTighe, line 114: This matches the bug number of this patch. Should this test be included now?

MatthewTighe (commented):
I don't think I have the expertise to speak to these UI tests as well as our automation folks, so I will leave the bulk of reviewing up to them.

Is all this commented code actually finished and working? It didn't necessarily look like it at a glance but wanted to check. If so, I don't think I see much value in waiting to handle the rest of these tests in future tickets and instead think we could just close them all out with this patch.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ComposeTabbedBrowsingTest.kt
MozillaNoah: oh, that's a goof on my end. I'll fix this with the next push

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayTestTag.kt
MozillaNoah: I'm surprised the linter didn't catch this

MozillaNoah:
> Is all this commented code actually finished and working? It didn't necessarily look like it at a glance but wanted to check. If so, I don't think I see much value in waiting to handle the rest of these tests in future tickets and instead think we could just close them all out with this patch.

The commented code has not been touched yet. I wanted to see what made sense for this particular case, since it's not typical production code where we avoid pushing out unnecessary comments, but I can also still access the original version of this inside `TabDrawerRobot` and don't necessarily need to keep it here. It's not a huge hassle either way, so I'm fine with either approach.

With regards to the ignored tests, this was a discussion point we had offline, where we decided that it'd be easier to review the tests incrementally, rather than all 12 + **all** of the robot code.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/ComposeTabDrawerRobot.kt
```diff
@@ -0,0 +1,567 @@
...
+//            Until.findObject(By.res("$packageName:id/mozac_browser_tabstray_url")),
+//            waitingTime,
+//        )
+//        onView(withId(R.id.mozac_browser_tabstray_url))
+//            .check(matches(withText(containsString(url))))
+//    }
+//
+//    fun verifyNormalBrowsingButtonIsDisplayed() = assertNormalBrowsingButton()
+    fun verifyNormalBrowsingButtonIsSelected(isSelected: Boolean = true) =
+        assertNormalBrowsingButtonIsSelected(isSelected)
```
sv-ohorvath, line 40: We actually planned to do some cleanup in our robots and remove this kind of duplication. So, without expecting you to do it, only if it helps you, you can skip the extra "assert" method and add the implementation directly into these "verify" methods.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/ComposeTabDrawerRobot.kt
```diff
@@ -0,0 +1,567 @@
...
+    class Transition(private val composeTestRule: HomeActivityComposeTestRule) {
+//        fun openThreeDotMenu(interact: ThreeDotMenuMainRobot.() -> Unit): ThreeDotMenuMainRobot.Transition {
+//            mDevice.waitForIdle()
+//
+//            Espresso.openActionBarOverflowOrOptionsMenu(ApplicationProvider.getApplicationContext<Context>())
+//            ThreeDotMenuMainRobot().interact()
+//            return ThreeDotMenuMainRobot.Transition()
+//        }
+
+        fun openTabDrawer(interact: ComposeTabDrawerRobot.() -> Unit): Transition {
```
sv-ohorvath, line 360: See other comment on ComposeTabbedBrowsingTest

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/ComposeTabDrawerRobot.kt
MozillaNoah: Oh nice! Yeah, I'm down to do the clean-up in this file.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ext/ComposeTestRule.kt
```diff
@@ -0,0 +1,14 @@
...
+package org.mozilla.fenix.ui.ext
+
+import androidx.compose.ui.test.junit4.ComposeTestRule
+import androidx.compose.ui.test.onNodeWithTag
+import org.mozilla.fenix.tabstray.TabsTrayTestTag
+
+/**
+ * Obtains the root Tabs Tray.
+ */
+fun ComposeTestRule.tabsTray() = onNodeWithTag(TabsTrayTestTag.tabsTray)
```
sv-ohorvath, line 14: I don't think we need this extra step. You can just use composeTestRule.onNodeWithTag(TabsTrayTestTag.tabsTray) directly in the robot class.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/BrowserRobot.kt
```diff
@@ -919,6 +923,37 @@ class BrowserRobot {
...
+                        mDevice.findObject(
+                            UiSelector()
+                                .resourceId("$packageName:id/mozac_browser_toolbar_browser_actions"),
+                        ),
+                        waitingTime,
+                    )
+
+                    tabsCounter().click()
+
+                    composeTestRule.tabsTray().assertExists()
```
sv-ohorvath, line 939: See the previous comment.

sv-ohorvath (approved):
Just one comment, otherwise LGTM. Thank you!

PR #2040, opened 2023-05-15
Bug 1833093 - Add a Nimbus experiment for pdf.js

calixteman:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1833093


android-components/components/browser/engine-gecko/build.gradle
```diff
@@ -89,4 +90,20 @@ dependencies {
...
 apply from: '../../../android-lint.gradle'
 apply from: '../../../publish.gradle'
+apply plugin: "org.mozilla.appservices.nimbus-gradle-plugin"
+nimbus {
+    // The path to the Nimbus feature manifest file
+    manifestFile = "geckoview.fml.yaml"
+
+    channels = [
+            debug: "debug",
+            release: "release",
```
Amejia481, line 100: Should we also have nightly channel?

android-components/components/browser/engine-gecko/build.gradle
```diff
@@ -89,4 +90,20 @@ dependencies {
 apply plugin: "org.mozilla.telemetry.glean-gradle-plugin"
 apply from: '../../../android-lint.gradle'
 apply from: '../../../publish.gradle'
+apply plugin: "org.mozilla.appservices.nimbus-gradle-plugin"
+nimbus {
+    // The path to the Nimbus feature manifest file
+    manifestFile = "geckoview.fml.yaml"
```
Amejia481, line 96: I wonder if we should have `nimbus.fml.yaml`, which applies to the whole AC and not only to the `engine-gecko` module, @jhugman what do you think?

fenix/app/nimbus.fml.yaml
```diff
@@ -131,6 +131,14 @@ import:
...
             notification-config:
               refresh-interval: 120 # minutes (2 hours)
+  - path: ../../android-components/components/browser/engine-gecko/geckoview.fml.yaml
+    channel: release
+    features:
+      pdfjs:
+        - value: {
+          download-button: true,
+          open-in-app-button: true
+        }
```
Amejia481, line 141: Should we add a `default` property for other channels?

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -1098,6 +1098,22 @@ class GeckoEngineSession(
         override fun onShowDynamicToolbar(geckoSession: GeckoSession) {
             notifyObservers { onShowDynamicToolbar() }
         }
+
+        override fun onGetNimbusFeature(session: GeckoSession, featureId: String): JSONObject? {
```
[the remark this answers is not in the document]
calixteman: I added this method in GeckoSession.java:
https://phabricator.services.mozilla.com/D178045#C6430005NL3628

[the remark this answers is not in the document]
Amejia481: 👍🏽

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -1098,6 +1098,22 @@ class GeckoEngineSession(
...
+                    json.put("download-button", pdfjs.downloadButton)
+                    json.put("open-in-app-button", pdfjs.openInAppButton)
+                }
+                else -> {
+                    return null
+                }
+            }
+
+            return json
+        }
```
jhugman, line 1116: Note to future: [this PR](https://github.com/mozilla/application-services/pull/5574) adds `getFeature(featureId: String)` and `toJSONObject()` for usage here.

android-components/components/browser/engine-gecko/build.gradle
jhugman: No, for a number of reasons:

- the generated code is per app, and not every app needs every component. You could use a FML component per app, and share FML files by including them in to each of the app's component, however…
- the generated code needs to be accessible from the component itself. If you had a per-app FML component, then the components (e.g. `engine-gecko`) would need to know which app it's in at compile time.

I have thought about having a per app FML file in AC, which is just `imports` of the component FML files, which then gets included by the apps: this might be work; I'd suggest thinking about this refactor when the third component grows an FML file.

android-components/components/browser/engine-gecko/build.gradle
jhugman: Imported channels are less useful, and more confusing than I thought they were going to be. We may need to add channels here, depending on how the build is setup.

jhugman (approved):
Looking good so far! You may want to hold off for [`toJSONObject` to land](https://github.com/mozilla/application-services/pull/5574) before landing this, or not.

Approving to keep you unblocked.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
calixteman: I updated the patch to use this amazing new feature: @jhugman thank you very much.

Amejia481:
While testing I noticed by default in Fenix/Focus the features are disabled by default,  we'll need to investigate this further, @jhugman do you have an idea of why this could be happening ? Noticing we are already setting the default values to be `true` on `geckoview.fml.yaml` but for some reason we are getting `false`.

Amejia481:
Thanks @calixteman for letting my know offline that I have to remove the `FxNimbus.kt` file to see the feature changes!

android-components/components/browser/engine-gecko/build.gradle
```diff
@@ -56,6 +57,7 @@ dependencies {
     implementation project(':concept-fetch')
     implementation project(':support-ktx')
     implementation project(':support-utils')
+    implementation project(':service-nimbus')
```
badboy, line 60: What about this here:

```suggestion
    implementation(project(':service-nimbus')) {
        exclude group: 'org.mozilla.telemetry', module: 'glean-native'
    }
```

The issue is (I think) that engine-gecko depends on geckoview, which itself provides the glean-native part.
But then nimbus also depends on `glean-native` transitively, but doesn't know about geckoview, so it's pulled in again.
By excluding that explicitly it should fall back to the geckoview-provided one when it's all compiled together.

I ran `./gradlew lint` on this change locally and it seemed to work, but I didn't run any tests.
We do something similar here: https://github.com/mozilla-mobile/firefox-android/blob/8dc9f859a4f75050c2d5f6b020a54963211aadcf/android-components/components/service/sync-logins/build.gradle#L35-L37

PR #2064, opened 2023-05-16
Bug 1833404 - Update Sentry to version 6.19.0

rvandermeulen:
As noted in the bug, this release brings some changes to ANR reporting, so we'll want to keep an eye on that after it lands (but also not panic if we see big shifts from what we used to see!)
https://bugzilla.mozilla.org/show_bug.cgi?id=1833404
https://bugzilla.mozilla.org/show_bug.cgi?id=1833404


Amejia481 (approved):
✨

PR #2130, opened 2023-05-19
Bug 1822448 – Prevent tabs tray from being animated multiple times on older android versions

mavduevskiy:
### Release vs Patch

https://github.com/mozilla-mobile/firefox-android/assets/92760693/20e0716c-fb0c-419a-ac76-344463f52906

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1822448


fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -194,6 +194,15 @@ class TabsTrayFragment : AppCompatDialogFragment() {
         return tabsTrayDialog
     }
 
+    override fun onPause() {
+        super.onPause()
+
+        // On some android versions (known about OneUI 2.5) entering animation is played every time
+        // DialogFragment goes though an onStart() lifecycle call. We want to prevent animation from
+        // being repeated after app returns from background.
+        dialog?.window?.setWindowAnimations(R.style.DialogFragmentRestoreAnimation)
```
mavduevskiy, line 203: If it's a good approach (and I couldn't find a different one, to be honest), we might want to disable enter animation after onPause() on all of the dialog fragments, since they all have that problem on samsung android 10 (maybe others as well, but the reporter had that particular device and I happen to have it as well)

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt
```diff
@@ -194,6 +194,15 @@ class TabsTrayFragment : AppCompatDialogFragment() {
         return tabsTrayDialog
     }
 
+    override fun onPause() {
+        super.onPause()
+
+        // On some android versions (known about OneUI 2.5) entering animation is played every time
+        // DialogFragment goes though an onStart() lifecycle call. We want to prevent animation from
+        // being repeated after app returns from background.
```
MatthewTighe, line 202: nit: 
```suggestion
        // On some android versions (known about OneUI 2.5) entering animation is played every time
        // DialogFragment goes though an onStart() lifecycle call. We want to prevent animations from
        // being repeated after app returns from background.
```

I also think that this comment could live next to the style you've defined instead. That way, if we end up re-using it in other dialogs, the comment will be in the common area.

MatthewTighe (approved):
Outside of a small nit, this LGTM! Good work discovering a solution to this problem :shipit:

PR #2165, opened 2023-05-22
Bug 1830401 - add Adjust event for Activated Users

HarrisonOg:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1830401

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/AdjustMetricsService.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/AdjustMetricsService.kt
@@ -88,9 +88,13 @@ class AdjustMetricsService(
     override fun track(event: Event) {
         CoroutineScope(dispatcher).launch {
             try {
-                if (event is Event.GrowthData && storage.shouldTrack(event)) {
-                    Adjust.trackEvent(AdjustEvent(event.tokenName))
-                    storage.updateSentState(event)
+                if (event is Event.GrowthData) {
+                    if (storage.shouldTrack(event)) {
+                        Adjust.trackEvent(AdjustEvent(event.tokenName))
+                        storage.updateSentState(event)
+                    } else {
+                        storage.updatePersistentState(event)
+                    }
                 }
             } catch (e: Exception) {
                 crashReporter.submitCaughtException(e)
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
@@ -46,5 +46,12 @@ sealed class Event {
          * Event recording the first time a URI is loaded in Firefox in a 24 hour period.
          */
         object FirstUriLoadForDay : GrowthData("ja86ek")
+
+        /**
+         * Event recording when User is "activated" in first week of usage.
+         * Activated = if the user is active 3 days in their first week and
+         * if they search once in the latter half of that week (days 4-7).
+         */
+        data class UserActivated(val fromSearch: Boolean) : GrowthData("imgpmr")
     }
 }
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
@@ -34,6 +34,11 @@ interface MetricsStorage {
     suspend fun updateSentState(event: Event)
 
     /**
+     * Updates locally-stored data related to an [event] that has just been sent.
+     */
+    suspend fun updatePersistentState(event: Event)
+
+    /**
      * Will try to register this as a recorder of app usage based on whether usage recording is still
      * needed. It will measure usage by to monitoring lifecycle callbacks from [application]'s
      * activities and should update local state using [updateUsageState].
@@ -60,6 +65,7 @@ internal class DefaultMetricsStorage(
     /**
      * Checks local state to see whether the [event] should be sent.
      */
+    @Suppress("CyclomaticComplexMethod")
     override suspend fun shouldTrack(event: Event): Boolean =
         withContext(dispatcher) {
             // The side-effect of storing days of use always needs to happen.
@@ -91,6 +97,9 @@ internal class DefaultMetricsStorage(
                         currentTime.duringFirstMonth() &&
                         settings.uriLoadGrowthLastSent.hasBeenMoreThanDaySince()
                 }
+                is Event.GrowthData.UserActivated -> {
+                    hasUserReachedActivatedThreshold()
+                }
             }
         }
 
@@ -114,6 +123,23 @@ internal class DefaultMetricsStorage(
             Event.GrowthData.FirstUriLoadForDay -> {
                 settings.uriLoadGrowthLastSent = System.currentTimeMillis()
             }
+            is Event.GrowthData.UserActivated -> {
+                settings.growthUserActivatedSent = true
+            }
+        }
+    }
+
+    override suspend fun updatePersistentState(event: Event) {
+        when (event) {
+            is Event.GrowthData.UserActivated -> {
+                if (event.fromSearch && shouldUpdateSearchUsage()) {
+                    settings.growthEarlySearchUsed = true
+                } else if (!event.fromSearch && shouldUpdateUsageCount()) {
+                    settings.growthEarlyUseCount.increment()
+                    settings.growthEarlyUseCountLastIncrement = System.currentTimeMillis()
+                }
+            }
+            else -> Unit
         }
     }
 
@@ -176,12 +202,33 @@ internal class DefaultMetricsStorage(
 
     private fun Long.duringFirstDay() = this < getInstalledTime() + dayMillis
 
+    private fun Long.afterThirdDay() = this > getInstalledTime() + threeDayMillis
+
     private fun Long.duringFirstWeek() = this < getInstalledTime() + fullWeekMillis
 
     private fun Long.duringFirstMonth() = this < getInstalledTime() + shortestMonthMillis
 
     private fun Calendar.createNextDay() = (this.clone() as Calendar).also { calendar ->
         calendar.add(Calendar.DAY_OF_MONTH, 1)
+    }
+
+    private fun hasUserReachedActivatedThreshold(): Boolean {
+        return !settings.growthUserActivatedSent &&
+            settings.growthEarlyUseCount.value >= daysActivatedThreshold &&
+            settings.growthEarlySearchUsed
+    }
+
+    private fun shouldUpdateUsageCount(): Boolean {
+        val currentTime = System.currentTimeMillis()
+        return currentTime.afterFirstDay() &&
+            currentTime.duringFirstWeek() &&
+            settings.growthEarlyUseCountLastIncrement.hasBeenMoreThanDaySince()
+    }
+
+    private fun shouldUpdateSearchUsage(): Boolean {
+        val currentTime = System.currentTimeMillis()
+        return currentTime.afterThirdDay() &&
+            currentTime.duringFirstWeek()
     }
 
     /**
@@ -208,6 +255,7 @@ internal class DefaultMetricsStorage(
 
     companion object {
         private const val dayMillis: Long = 1000 * 60 * 60 * 24
+        private const val threeDayMillis: Long = 3 * dayMillis
         private const val shortestMonthMillis: Long = dayMillis * 28
 
         // Note this is 8 so that recording of FirstWeekSeriesActivity happens throughout the length
@@ -216,6 +264,9 @@ internal class DefaultMetricsStorage(
 
         // The usage threshold we are interested in is currently 340 seconds.
         private const val usageThresholdMillis = 1000 * 340
+
+        // The usage threshold for "activated" growth users.
+        private const val daysActivatedThreshold = 3
 
         /**
          * Determines whether events should be tracked based on some general criteria:
--- a/fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -1735,4 +1735,33 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         key = appContext.getPreferenceKey(R.string.pref_key_enable_tabs_tray_to_compose),
         default = FeatureFlags.composeTabsTray,
     )
+
+    /**
+     * Adjust Activated User sent
+     */
+    var growthUserActivatedSent by booleanPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_growth_user_activated_sent),
+        default = false,
+    )
+
+    /**
+     * Indicates how many days in the first week user opened the app.
+     */
+    val growthEarlyUseCount = counterPreference(
+        appContext.getPreferenceKey(R.string.pref_key_growth_early_browse_count),
+        maxCount = 3,
+    )
+
+    var growthEarlyUseCountLastIncrement by longPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_growth_early_browse_count_last_increment),
+        default = 0L,
+    )
+
+    /**
+     * Indicates how many days in the first week user searched in the app.
+     */
+    var growthEarlySearchUsed by booleanPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_growth_early_search),
+        default = false,
+    )
 }
--- a/fenix/app/src/test/java/org/mozilla/fenix/components/metrics/DefaultMetricsStorageTest.kt
+++ b/fenix/app/src/test/java/org/mozilla/fenix/components/metrics/DefaultMetricsStorageTest.kt
@@ -6,7 +6,9 @@ package org.mozilla.fenix.components.metrics
 
 import android.app.Activity
 import android.app.Application
+import io.mockk.Runs
 import io.mockk.every
+import io.mockk.just
 import io.mockk.mockk
 import io.mockk.slot
 import io.mockk.verify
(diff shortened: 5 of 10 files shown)


fenix/app/src/main/java/org/mozilla/fenix/components/metrics/AdjustMetricsService.kt
```diff
@@ -91,6 +91,8 @@ class AdjustMetricsService(
                 if (event is Event.GrowthData && storage.shouldTrack(event)) {
                     Adjust.trackEvent(AdjustEvent(event.tokenName))
                     storage.updateSentState(event)
+                } else if (event is Event.ActivatedUserData && storage.shouldTrack(event)) {
+                    storage.updateSentState(event)
```
MatthewTighe, line 95: I don't think this branch will get hit, because there are no concrete types that implement `ActivatedUserData`.

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
```diff
@@ -46,5 +46,27 @@ sealed class Event {
...
+         * Activated = if the user is active 3 days in their first week and
+         *                  if they search once in the latter half of that week (days 4-7).
+         */
+        object UserActivated : GrowthData("13mh5x")
+    }
+
+    /**
+     * Events for helping track user data for growth events
+     */
+    sealed class ActivatedUserData(val tokenName: String) : Event() {
```
MatthewTighe, line 61: Since the `object`s below are only namespaced by `ActivatedUserData` and do not implement it, we should be able to remove this unused `tokenName` property.

It also looks like they were potentially meant to implement `ActivatedUserData`, in which case we should be able to remove the property as well as the filler values that are used below.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/AdjustActivatedUserWorker.kt
```diff
@@ -0,0 +1,74 @@
...
+
+        @VisibleForTesting
+        internal fun shouldSendAdjustActivatedEvent(settings: Settings): Boolean {
+            if (!settings.adjustUserActivatedSent &&
+                settings.adjustUseCount.value >= DAYS_IN_FIRST_WEEK_ACTIVATED_THRESHOLD &&
+                settings.adjustEarlySearchCount.value >= SEARCHES_IN_FIRST_WEEK_ACTIVATED_THRESHOLD
+            ) {
+                return true
+            }
+            return false
```
MatthewTighe, line 71: This function looks to check the same conditions that the `Event.GrowthData.UserActivated` branch in `shouldTrack` does. `shouldTrack`will run during `track` anyway, so I think this function can be removed.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/AdjustActivatedUserWorker.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.google.android.gms.common.util.VisibleForTesting
+import org.mozilla.fenix.components.metrics.Event
+import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Worker that tracks conditions for "Activated" event for new users.
+ */
+class AdjustActivatedUserWorker(
```
MatthewTighe, line 22: This `Worker` looks to be doing pretty similar things to what is happening in the `MetricsMiddleware`. I don't necessarily have a strong opinion on using one over the other, but I think we should avoid splitting how we handle events like these into multiple types of solutions. I think it would probably be easier to remove this and use the `MetricsMiddleware`, but I am also open to having the events in the Middleware moved into the Worker instead. Since `WorkManager`'s work queue persists across things like process death, this would improve our data integrity, though I am not sure whether we know that that is actually needed or not.

If we do want to move the other events here, I think it would be fine to do that as a follow-up.

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
```diff
@@ -46,5 +46,27 @@ sealed class Event {
          * Event recording the first time a URI is loaded in Firefox in a 24 hour period.
          */
         object FirstUriLoadForDay : GrowthData("ja86ek")
+
+        /**
+         * Event recording when User is "activated" in first week of usage.
+         * Activated = if the user is active 3 days in their first week and
+         *                  if they search once in the latter half of that week (days 4-7).
+         */
+        object UserActivated : GrowthData("13mh5x")
```
MatthewTighe, line 55: FYI It looks like this is still using the dev token 😅

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
```diff
@@ -46,5 +46,27 @@ sealed class Event {
...
+    sealed class ActivatedUserData(val tokenName: String) : Event() {
+        /**
+         * Event recording search usage during days 2-7 of first week of install
+         */
+        object FirstWeekSearchActivity : GrowthData("yeahnope")
+
+        /**
+         * Event recording app usage during days 2-7 of first week of install
+         */
+        object FirstWeekUsageActivity : GrowthData("nopeyeah")
```
MatthewTighe, line 70: Instead of the above, you could also consider using only the `UserActivated` event, with some kind of flag to indicate where it was being triggered from. If it was a search event, you could update that state and otherwise update the usage. Then, the search state and usage state could be checked simultaneously to see if we should forward the event to Adjust. For example:

```kotlin
data class UserActivated(val fromSearch: Boolean) : GrowthData()

// shouldSend()
                Event.GrowthData.UserActivated -> {
                    currentTime.duringFirstWeek() &&
                        settings.adjustUseCount.value >= daysActivatedThreshold &&
                        settings.adjustEarlySearchCount.value >= searchesActivatedThreshold &&
                        !settings.adjustUserActivatedSent
                }

// updatePersistentState (new function, see review comment about `updateSentState`)
when (event) {
    is UserActivated -> {
         if (fromSearch) {
             // update pref about search
         } else {
            // update pref about usage
         }
    }
}
```

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -114,6 +131,17 @@ internal class DefaultMetricsStorage(
...
+            Event.GrowthData.UserActivated -> {
+                settings.adjustUserActivatedSent = true
+            }
+            Event.ActivatedUserData.FirstWeekUsageActivity -> {
+                settings.adjustUseCount.increment()
+                settings.adjustUseCountLastIncrement = System.currentTimeMillis()
+            }
+            Event.ActivatedUserData.FirstWeekSearchActivity -> {
+                settings.adjustEarlySearchCount.increment()
+                settings.adjustEarlySearchCountLastIncrement = System.currentTimeMillis()
```
MatthewTighe, line 143: Instead of updating this kind of persistent state here, perhaps we should either introduce another function like `updateDaysOfUse` or change that function to track any persistent state we are interested in. Admittedly the example I set with `FirstAppOpenForDay` and `FirstUriLoadForDay` was a little hacky, but I was unsure at the time how long this code would live.

MatthewTighe (changes requested):
Thanks for taking this on! I had some comments addressing some potentially larger updates that I'd like to get your opinion on 🙂

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
HarrisonOg: whoops! removed!

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
HarrisonOg: Ah thank you!
I made some changes, this is a much more elegant implementation.

using `shouldTrack` as intended for checking the threshold necessary for sending the "activated" signal while taking advantage of a flag to update the local data for the usage threshold in `updatePersistentState` if the `shouldTrack` call returns `false`

I also updated the test cases to reflect this new paradigm.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/AdjustActivatedUserWorker.kt
HarrisonOg: I agree, I removed this worker and moved some of the logic to `MetricsMiddleware` for the "usage" counter.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/AdjustActivatedUserWorker.kt
HarrisonOg: Removed this file completely (see below comment).

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -60,6 +60,7 @@ internal class DefaultMetricsStorage(
     /**
      * Checks local state to see whether the [event] should be sent.
      */
+    @Suppress("CyclomaticComplexMethod")
```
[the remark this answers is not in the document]
HarrisonOg: I removed `ActivatedUserData`, see below comment on `Event.kt`

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
HarrisonOg: I removed `ActivatedUserData`, see below comment on `Event.kt`

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/AdjustMetricsService.kt
HarrisonOg: I removed `ActivatedUserData`, see below comment on `Event.kt`

there's new logic to either send the event or update local data for determining if `AdjustMetricsService` should send event.

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/AdjustMetricsService.kt
```diff
@@ -99,7 +101,7 @@ class AdjustMetricsService(
     }
 
     override fun shouldTrack(event: Event): Boolean =
-        event is Event.GrowthData
+        event is Event.GrowthData || event is Event.ActivatedUserData
```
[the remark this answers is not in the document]
HarrisonOg: I removed `ActivatedUserData`, see below comment on `Event.kt`

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
HarrisonOg: changed how this works with new `updatePersistentState` function for events that require counting data before sending.

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/AdjustMetricsService.kt
```diff
@@ -88,9 +88,13 @@ class AdjustMetricsService(
...
-                if (event is Event.GrowthData && storage.shouldTrack(event)) {
-                    Adjust.trackEvent(AdjustEvent(event.tokenName))
-                    storage.updateSentState(event)
+                if (event is Event.GrowthData) {
+                    if (storage.shouldTrack(event)) {
+                        Adjust.trackEvent(AdjustEvent(event.tokenName))
+                        storage.updateSentState(event)
+                    } else {
+                        storage.updatePersistentState(event)
+                    }
```
MatthewTighe, line 97: Yeah, I'm loving how clear this is now 🙌 thanks for updating!

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/Event.kt
```diff
@@ -46,5 +46,12 @@ sealed class Event {
          * Event recording the first time a URI is loaded in Firefox in a 24 hour period.
          */
         object FirstUriLoadForDay : GrowthData("ja86ek")
+
+        /**
+         * Event recording when User is "activated" in first week of usage.
+         * Activated = if the user is active 3 days in their first week and
+         *                  if they search once in the latter half of that week (days 4-7).
+         */
```
MatthewTighe, line 54: nit: Awesome comment, but I'm not sure we need all that extra space in the start of the last line

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -60,6 +65,7 @@ internal class DefaultMetricsStorage(
...
```
MatthewTighe, line 68: Is this suppression still needed with your latest changes? NP if it is, but it does seem like something we could try to improve in a follow-up

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -184,6 +215,23 @@ internal class DefaultMetricsStorage(
         calendar.add(Calendar.DAY_OF_MONTH, 1)
     }
 
+    /**
+     *
+     */
```
MatthewTighe, line 220: nit: some leftover kdoc comment lines

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1735,4 +1735,38 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+     */
+    val adjustEarlySearchCount = counterPreference(
+        appContext.getPreferenceKey(R.string.pref_key_adjust_early_search),
+        maxCount = 1,
+    )
+
+    var adjustEarlySearchCountLastIncrement by longPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_adjust_early_search_count_last_increment),
+        default = 0L,
+    )
```
MatthewTighe, line 1771: I'm not sure if this is still the case, but I was cautioned to use `GrowthData` language instead of `Adjust` language when this project first began, which is why the existing pref names are following that convention. I don't think I care enough to block the PR on it, but it might help with consistency to follow the existing naming if you're so inclined.

fenix/app/src/test/java/org/mozilla/fenix/components/metrics/DefaultMetricsStorageTest.kt
```diff
@@ -382,6 +384,120 @@ class DefaultMetricsStorageTest {
...
+        storage.updatePersistentState(Event.GrowthData.UserActivated(fromSearch = true))
+
+        assertTrue(captureSent.captured > 0L)
+    }
+
+    @Test
+    fun `WHEN first week search activity is sent in first half of first week THEN settings will not be updated`() = runTest(dispatcher) {
+        val captureSent = slot<Long>()
+        val currentTime = System.currentTimeMillis()
+        installTime = currentTime - (dayMillis * 2) + 1
```
MatthewTighe, line 475: Clock consistency might make this impossible, but I'm wondering if this could be (dayMillis * 3) - 1 to align with the test above it.

fenix/app/src/test/java/org/mozilla/fenix/components/metrics/DefaultMetricsStorageTest.kt
```diff
@@ -382,6 +384,120 @@ class DefaultMetricsStorageTest {
         assertTrue(updateSlot.captured > 0)
     }
 
+    @Test
+    fun `GIVEN first week activated threshold reached THEN will be sent`() = runTest(dispatcher) {
```
MatthewTighe, line 388: It'll be pretty lengthy, but I think we should update this test name to make it clear that both conditions (days of use & search) are required.

fenix/app/src/test/java/org/mozilla/fenix/components/metrics/DefaultMetricsStorageTest.kt
```diff
@@ -382,6 +384,120 @@ class DefaultMetricsStorageTest {
...
+        every { settings.adjustEarlySearchCount.value } returns 1
+        every { settings.adjustUserActivatedSent } returns false
+
+        val result = storage.shouldTrack(Event.GrowthData.UserActivated(fromSearch = false))
+
+        assertTrue(result)
+    }
+
+    @Test
+    fun `GIVEN first week activated threshold not reached THEN will not be sent`() = runTest(dispatcher) {
```
MatthewTighe, line 401: I think we should update this test name to make it a little more clear that the "days of use" condition is the reason that the event isn't being sent. 

Additionally, I think we should add one more test where the days of use condition is met, but the search condition prevents the event from being sent.

MatthewTighe (changes requested):
I think this is looking fantastic now! I think we should add one more test and I have a few other surface-level comments, but this is seeming very close to me.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
HarrisonOg: gotcha, I'll make the adjustment (*buh-dum-tss*)

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
HarrisonOg: unfortunately it's still triggering the warning.
I was thinking of doing a follow up where I move some of the conditions to separate methods where it makes sense.

fenix/app/src/test/java/org/mozilla/fenix/components/metrics/DefaultMetricsStorageTest.kt
HarrisonOg: updated the title to reflect both conditions are met.

fenix/app/src/test/java/org/mozilla/fenix/components/metrics/DefaultMetricsStorageTest.kt
HarrisonOg: I had actually constructed these tests a little wrong and aligning the `installTime` to a similar length revealed my error.
Fixed the tests to be more stable.

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
HarrisonOg: removed thanks!

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/AdjustMetricsService.kt
HarrisonOg: yeah this new flow is much better and easier to expand on later if need be. thank you for your advice!

MatthewTighe (approved):
This looks great! 🚢 
it's all coming together

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1735,4 +1735,38 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+
+    var growthEarlyUseCountLastIncrement by longPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_growth_early_browse_count_last_increment),
+        default = 0L,
+    )
+
+    /**
+     * Indicates how many days in the first week user searched in the app.
+     */
+    val growthEarlySearchCount = counterPreference(
```
rocketsroger, line 1763: Do we need to know how many times the user searched?

HarrisonOg:
> I see there are some more changes coming in so I am removing my approval for now in case I forget 😅
> 
> Please let me know when this is ready to review again!

ah I have one final change coming and then we can re-review, thank you!
Roger reminded me that we need to account for the edge case where the user achieves the threshold for "activated" but doesn't open the app again.  We should still receive the "activated" signal on the 7th day.
The last change coming is to change the search threshold tracking to be more simple as we don't need a count or a `lastIncremented` tracking number.

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
```diff
@@ -0,0 +1,74 @@
...
+        if (!System.currentTimeMillis().isEndOfFirstWeek(applicationContext) ||
+            settings.growthUserActivatedSent
+        ) {
+            return Result.success()
+        }
+
+        applicationContext.metrics.track(Event.GrowthData.UserActivated(fromSearch = false))
+
+        return Result.success()
+    }
```
MatthewTighe, line 34: Would you mind filing a follow-up to add the rest of the metrics that are currently tracked through `MetricsMiddleware` here as well? This seems like a useful fallback for all of those

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.work.WorkerParameters
+import mozilla.components.support.utils.ext.getPackageInfoCompat
+import org.mozilla.fenix.ext.metrics
+import org.mozilla.fenix.ext.settings
+import java.util.concurrent.TimeUnit
+
+/**
+ * Worker that will send the User Activated event at the end of the first week.
+ */
+class GrowthActivatedUserWorker(
```
MatthewTighe, line 17: Maybe `GrowthDataWorker` would keep this slightly more decoupled, allowing us to add more events to it over time?

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
```diff
@@ -0,0 +1,74 @@
...
+
+        applicationContext.metrics.track(Event.GrowthData.UserActivated(fromSearch = false))
+
+        return Result.success()
+    }
+
+    companion object {
+        private const val GROWTH_USER_ACTIVATED_WORK_NAME = "org.mozilla.fenix.growth.work"
+        private const val dayMillis: Long = 1000 * 60 * 60 * 24
+        private const val fullWeekMillis: Long = dayMillis * 7
```
MatthewTighe, line 39: I think having a mixed capitalization styling for these constants is a little bit weird. It looks like the [convention guide](https://kotlinlang.org/docs/coding-conventions.html#property-names) suggests using screaming case, so I guess my understanding when I named the similar constants in `MetricsStorage` was out-of-date.

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun doWork(): Result {
+        val settings = applicationContext.settings()
+
+        if (!System.currentTimeMillis().isEndOfFirstWeek(applicationContext) ||
+            settings.growthUserActivatedSent
+        ) {
+            return Result.success()
+        }
+
+        applicationContext.metrics.track(Event.GrowthData.UserActivated(fromSearch = false))
```
MatthewTighe, line 31: Do we potentially need another flag for `UserActivated` that signals it comes from here? It seems like we could get in a situation where the worker activating from the background would count as user activity and could mess with our data integrity

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
```diff
@@ -0,0 +1,74 @@
...
+                GROWTH_USER_ACTIVATED_WORK_NAME,
+                ExistingWorkPolicy.KEEP,
+                growthSignalWork,
+            ).enqueue()
+        }
+
+        /**
+         * Returns [Boolean] value signaling if current time is after the first week after install.
+         */
+        private fun Long.isEndOfFirstWeek(context: Context): Boolean {
```
MatthewTighe, line 65: nit, maybe:
```suggestion
        private fun Long.isAfterFirstWeekFromInstall(context: Context): Boolean {
```

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -184,6 +212,25 @@ internal class DefaultMetricsStorage(
         calendar.add(Calendar.DAY_OF_MONTH, 1)
     }
 
+    private fun userActivatedThresholdReached(): Boolean {
```
MatthewTighe, line 215: nit:
```suggestion
    private fun hasUserReachedActivatedThreshold(): Boolean {
```

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
```diff
@@ -217,6 +265,9 @@ internal class DefaultMetricsStorage(
         // The usage threshold we are interested in is currently 340 seconds.
         private const val usageThresholdMillis = 1000 * 340
 
+        // The usage threshold for "activated" growth users.
+        private const val daysActivatedThreshold = 3
```
MatthewTighe, line 269: Since these names are already incorrectly cased (my bad), let's just leave these ones that you're adding as they are for now

MatthewTighe (commented):
Have just round more round of nits but this otherwise LGTM! I'll circle back around for the approval after these get addressed 🙂

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
HarrisonOg: gotcha, I'll change the name

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
HarrisonOg: updated function name

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
HarrisonOg: updated function name to be more descriptive

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
HarrisonOg: fixed

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsStorage.kt
HarrisonOg: I'll leave the ones in `MetricsStorage` for now, might change them in the follow up ticket

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/GrowthActivatedUserWorker.kt
MatthewTighe: Ah, that's a good point. I don't think we need to worry about it then. Thinking about it more, we'd probably want a generalized solution to that problem rather than for this specific event in any case.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1735,4 +1735,33 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+        key = appContext.getPreferenceKey(R.string.pref_key_growth_user_activated_sent),
+        default = false,
+    )
+
+    /**
+     * Indicates how many days in the first week user opened the app.
+     */
+    val growthEarlyUseCount = counterPreference(
+        appContext.getPreferenceKey(R.string.pref_key_growth_early_browse_count),
+        maxCount = 2,
```
MatthewTighe, line 1752: Should this `maxCount` be 3 or perhaps just use the default argument? Just feels a little weird to have a discrepancy between this and `daysActivatedThreshold`. It doesn't look like this property is actually used for anything in this case so I don't think it matters much, but it either update might be a more clear signal to future readers.

MatthewTighe (approved):
One small optional comment but this otherwise looks ready to land from my perspective. Good work and thanks for your patience in cycling on it a couple times!

rocketsroger (commented):
Looks good to me as well.  I'll let Matt give the final approval.  Thanks!

MatthewTighe (approved):

HarrisonOg:
@Mergifyio backport releases_v114

PR #2197, opened 2023-05-24
Bug 1822349 - monorepo-migration: replace commit references with URLs

jcristau:
Update convert-to-hg script to run a custom git-filter-repo extension to replace references to commit hashes in commit/tag messages with the commit's github URL, to avoid dangling references after the rewrite and conversion to hg.

This assumes that `git_filter_repo` is on the python import path which I'm not sure can be relied on, but I guess we can always fix that up if/when someone other than me runs the script.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1822349


jcristau:
As an example this turns https://github.com/mozilla-mobile/firefox-android/commit/f14ccc5b1c4ed2d112c1fd9e2ba8773fc2fefed5 into https://hg.mozilla.org/users/jcristau_mozilla.com/firefox-android-hg/rev/61eea0b35f6b (vs https://hg.mozilla.org/users/jcristau_mozilla.com/firefox-android-hg/rev/a079bbe52c33 without the extra rewrite).

A bonus is that some commit messages were rewritten in previous migration steps so that a reference to `https://github.com/mozilla-mobile/fenix/commit/aaaaaaaa` became `https://github.com/mozilla-mobile/fenix/commit/bbbbbbbb` when fenix code landed in firefox-android, causing a dead link; we now turn that reference into `https://github.com/mozilla-mobile/firefox-android/commit/bbbbbbbb`.

gbrownmozilla (approved):

PR #2210, opened 2023-05-25
Bug 1833757 - Merge `onTabMultiSelectClick` and `onTabClick` lambdas

MozillaNoah:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1833757


fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt
```diff
@@ -541,10 +519,17 @@ class DefaultTabsTrayController(
     }
 
     override fun handleTabSelected(tab: TabSessionState, source: String?) {
-        TabsTray.openedExistingTab.record(TabsTray.OpenedExistingTabExtra(source ?: "unknown"))
-        tabsUseCases.selectTab(tab.id)
-        browsingModeManager.mode = BrowsingMode.fromBoolean(tab.content.private)
-        handleNavigateToBrowser()
+        val selected = tabsTrayStore.state.mode.selectedTabs
```
MatthewTighe, line 522: It's a bit sneaky, but could we update the name of this function to use `Clicked` instead of `Selected` while we're here? I think that lines up a bit better with the rest of the methods, and I think it'll make the path more clear when we refactor this to an action.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt
```diff
@@ -541,10 +519,17 @@ class DefaultTabsTrayController(
...
+        when {
+            selected.isEmpty() && tabsTrayStore.state.mode.isSelect().not() -> {
+                TabsTray.openedExistingTab.record(TabsTray.OpenedExistingTabExtra(source ?: "unknown"))
+                tabsUseCases.selectTab(tab.id)
+                browsingModeManager.mode = BrowsingMode.fromBoolean(tab.content.private)
+                handleNavigateToBrowser()
+            }
+            tab.id in selected.map { it.id } -> handleTabUnselected(tab)
```
MatthewTighe, line 530: I'm wondering whether it's worth moving `handleTabUnselected` inline here as well. I see it's also used in the `Interactor` currently but I'm not sure if that's still necessary

MatthewTighe, line 530: This seems like a big improvement to me 😄

MatthewTighe (approved):
This all looks good to me, but I think we should get QA on this one since it's touching live code and we're approaching the end of this nightly cycle.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt
MozillaNoah: I'll take a look

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt
MozillaNoah: I'm on the same wavelength as you. This interface function is from an [AC interface](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/tabstray/src/main/java/mozilla/components/browser/tabstray/TabsTray.kt), and I actually have a [ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1814243) to decouple us from that and port the consumers to the interactor.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt
MozillaNoah: So it looks like `handleTabUnselected` is used exclusively for the [XML version of the tab reordering](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/browser/AbstractBrowserTabViewHolder.kt#251-257). We'll be able to clean this up after the XML gets removed though, so I'll file a follow-up.

PR #2230, opened 2023-05-25
Bug 1835107 - Attaches a Mechanism to the SentryException before sending it to Sentry

boek:
# Before
image

# After
image

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


android-components/components/lib/crash-sentry/src/main/java/mozilla/components/lib/crash/sentry/SentryService.kt
```diff
@@ -144,6 +148,20 @@ class SentryService(
...
+        // Sentry now uses the `Mechanism` to determine whether or not an exception is
+        // handled. Any exception sent with `Sentry.captureException` is assumed to be handled
+        // by Sentry. We can attach a `UncaughtExceptionHandler` mechanism to the `SentryException`
+        // to correctly signal to Sentry that this is an uncaught exception.
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=1835107
+        event.exceptions?.firstOrNull()?.let { sentryException ->
+            val mechanism = Mechanism()
+            mechanism.type = "UncaughtExceptionHandler"
+            mechanism.isHandled = false
+            sentryException.mechanism = mechanism
```
rahulsainani, line 161: 💡 
```suggestion
            sentryException.mechanism = Mechanism().apply {
                type = "UncaughtExceptionHandler" // maybe this can be extracted as a constant
                isHandled = false
            }
```

android-components/components/lib/crash-sentry/src/main/java/mozilla/components/lib/crash/sentry/SentryService.kt
```diff
@@ -136,6 +137,9 @@ class SentryService(
             options.environment = environment
             options.beforeSend = BeforeSendCallback { event, _ ->
                 val throwable = event.throwable
+                if (event.level == SentryLevel.FATAL) {
+                    alterEventToBeUnhandled(event)
+                }
```
rahulsainani, line 142: ❓ Should we add some tests for this?

boek: Yes, definitely! I pulled out the `BeforeSendCallback` and wrote some tests for it :)

titooan (approved):
LGTM!

android-components/components/lib/crash-sentry/src/test/java/mozilla/components/lib/crash/sentry/SentryServiceTest.kt
```diff
@@ -263,4 +269,48 @@ class SentryServiceTest {
         verify(service).prepareReport(breadcrumbs, SentryLevel.INFO)
         verify(service).reportToSentry(exception)
     }
+
+    @Test
+    @Suppress("UnstableApiUsage")
```
rahulsainani, line 274: ❓ JOOC is there a way to not add this, I see it's used for `SentryExceptionFactory`. 

Could we hook into the sent events and just assert that the sent event had the expected mechanism? That way it would be more of and input/output test and we won't have to call `execute` on `beforeSendCallback`, we could init Sentry and report the event which would trigger the beforeSendCallback like in production.

android-components/components/lib/crash-sentry/src/main/java/mozilla/components/lib/crash/sentry/SentryService.kt
```diff
@@ -189,6 +210,9 @@ class SentryService(
     }
 }
 
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+internal const val UNCAUGHT_EXCEPTION_TYPE = "UncaughtExceptionHandler"
```
rahulsainani, line 214: Is this supposed to be used in test? 🤔 Don't think it's used right now.

boek: good catch! I was going to assert on this as well but lost track.

android-components/components/lib/crash-sentry/src/test/java/mozilla/components/lib/crash/sentry/SentryServiceTest.kt
boek: @rahulsainani spent some time digging through the library and I'm not sure there would be a good way to test the sentry library in its entirety. I think I did find a better way to implement both this and the rust crash modifier though!

rahulsainani (approved):
🚀

PR #2263, opened 2023-05-29
Bug 1822349 - tweak commit message replacements

jcristau:
Change the git-filter-repo commit message rewrite step to avoid breaking existing URLs referencing commits, and to remediate some issues caused by rewrites during the previous migrations.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1822349


monorepo-migration/generate-replace-message-expressions.py
```diff
@@ -57,6 +57,37 @@ def main():
                 )
                 regexes.append(regex)
 
+    # fix up wrong replacements from previous migrations
```
gbrownmozilla, line 60: This is an impressive list!

gbrownmozilla (approved):

monorepo-migration/generate-replace-message-expressions.py
JohanLorenzo: +1 Thank you for finding and fixing these edge cases we missed in the previous migrations!

JohanLorenzo (approved):
Looks great to me. I look forward to seeing the new history 😃

jcristau:
New rewrite pushed: https://hg.mozilla.org/users/jcristau_mozilla.com/firefox-android-hg/pushloghtml?changeset=8fe4821c3bceb9f7a02a541f9dc751ea9862e873

PR #2274, opened 2023-05-30
Bug 1826473 – add new search engine settings screens

mavduevskiy:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1826473


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/SearchState.kt
```diff
@@ -38,6 +38,7 @@ data class SearchState(
     val additionalSearchEngines: List<SearchEngine> = emptyList(),
     val additionalAvailableSearchEngines: List<SearchEngine> = emptyList(),
     val hiddenSearchEngines: List<SearchEngine> = emptyList(),
+    val disabledSearchEngineShortcutsIds: List<String> = emptyList(),
```
jonalmeida, line 41: nit: Let's remove "shortcuts" from the naming. Shortcuts means something different in both of our supported browsers and they are also an application-level concept that does not need to be here.
```suggestion
    val disabledSearchEngineIds: List<String> = emptyList(),
```

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/SearchState.kt
```diff
@@ -51,6 +52,18 @@ data class SearchState(
 val SearchState.searchEngines: List<SearchEngine>
     get() = (regionSearchEngines + additionalSearchEngines + customSearchEngines + applicationSearchEngines)
 
+/**
+ * The list of search engine shortcuts to be available for quick search menu.
+ * // TODO maybe move it into ext file
+ */
+val SearchState.searchEngineShortcuts: List<SearchEngine>
```
jonalmeida, line 59: For similar reasons above, let's move this extension function to the fenix layer.

Meta: From a discussion with Mike offline, we also realized that sorting of this list is also an application-level customization for displaying in the app.

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/middleware/SearchMiddleware.kt
```diff
@@ -177,6 +183,18 @@ class SearchMiddleware(
         )
     }
 
+    private fun updateDisabledSearchEngineShortcutsIds(
+        action: SearchAction.UpdateDisabledSearchEngineShortcutsIdsAction,
+    ) = scope.launch {
+        val disabledIds = metadataStorage.getDisabledSearchEngineShortcutsIds()
```
jonalmeida, line 189: We have already loaded the most recent state from `loadSearchEngines` so we don't need to query the storage layer again. Instead, we can use the most version from the SearchState we have already fetched.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/RadioSearchEngineListPreference.kt
```diff
@@ -82,19 +82,28 @@ class RadioSearchEngineListPreference @JvmOverloads constructor(
             it.isGeneral
         }.size == 1
         state.searchEngines.filter { engine ->
-            engine.type != SearchEngine.Type.APPLICATION
+            if (context.settings().enableNewSearchSettings) {
```
jonalmeida, line 85: Meta: Mike and I spoke offline about this block and the one below - one feature flag is dependant on another, so we agreed that we should probably make sure that the dependant logic is true at the flag level

jonalmeida (approved):
The AC changes and logic looks good to me. I did not take an in-depth look into the UI, so I'd recommend someone else review that bit.

Nice work! ✨

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,273 @@
...
+import mozilla.components.lib.state.ext.observeAsComposableState
+import org.mozilla.fenix.R
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Top-level UI for search shortcuts settings.
+ */
+@Suppress("LongParameterList")
+@Composable
+fun SearchEngineShortcuts(
```
gabrielluong, line 54: I took a quick scan of the UI side of things. I think there's gonna be some extensive feedback (eg lots of papercut nits). Could you ping myself or @MozillaNoah when you are ready for a formal review?

I can point out a few small things in the meantime just so we have less feedback going forward.

gabrielluong (commented):
Provided some quick takeaways. I think one of my biggest takeaway here is to consider adding the Radio Button design component separately so that it could be reviewed independently and reused in this PR.

gabrielluong:
One additional thought since Jon already gave his approval on the AC side of things. Perhaps we should also split up the AC side of things to be landed ahead of the UI.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,273 @@
...
+    categoryTitle: String,
+    store: BrowserStore,
+    onCheckboxClicked: (SearchEngine, Boolean) -> Unit,
+    onEditEngineClicked: (SearchEngine) -> Unit,
+    onDeleteEngineClicked: (SearchEngine) -> Unit,
+    onAddEngineClicked: () -> Unit,
+) {
+    val searchState = store.observeAsComposableState { it.search }.value ?: SearchState()
```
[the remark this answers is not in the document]
MozillaNoah: components.store breaks previews, unfortunately :/

I thought Rahul landed a non-null version of `observeAsComposableState` though.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,273 @@
...
+}
+
+@Composable
+private fun SearchEngineDropdownMenu(
+    engine: SearchEngine,
+    isMenuExpanded: MutableState<Boolean>,
+    onEditEngineClicked: (SearchEngine) -> Unit,
+    onDeleteEngineClicked: (SearchEngine) -> Unit,
+) {
+    DropdownMenu(
```
MozillaNoah, line 203: we have a `ContextualMenu` composable component, which should save you some time here.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,273 @@
...
+    engine: SearchEngine,
+    name: String,
+    isEnabled: Boolean,
+    onCheckboxClicked: (SearchEngine, Boolean) -> Unit,
+    onEditEngineClicked: (SearchEngine) -> Unit,
+    onDeleteEngineClicked: (SearchEngine) -> Unit,
+) {
+    val isMenuExpanded: MutableState<Boolean> = remember { mutableStateOf(false) }
+
+    Row(
```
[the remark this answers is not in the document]
MozillaNoah: Or just extend the existing [`RadioButton`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/button/RadioButton.kt). If there isn't a figma, we can shoot Crystal a quick DM to figure this one out. 

I would definitely encourage breaking this PR down into smaller chunks though. It'll help with the reviewing, obviously, but it will also make any round of discussion and iteration that much quicker.

mavduevskiy:
hey, @gabrielluong, thanks for the early feedback, I have addressed it (except the "Go to main component" thing, editor mode isn't available for me).

I _really_ want to land it as soon as possible and ship unified-search in 115, that's the only blocker. And, generally speaking, I don't think nits should be blocking a feature. Let's do things like extracting a general compose component to AC layer as follow ups, and now focus only on blocking changes. 

Once we uplift it to beta and gain confidence, I will be doing some refactoring – removing duplicated setting screen, etc. It looks like a good opportunity to extract a basic `@Compose `component as well. Writing it now doesn't really help with landing the feature and is more of scope creeping.

I wanted to hide the new UI behind the feature flag, so I copy-pasted a lot of existing code with slight modifications to match the new UI to provide an alternative path for up to show the settings screen. Changing initial designed would make the patch smaller, but more difficult to roll-back.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,254 @@
...
+    Box(
+        modifier = Modifier
+            .height(48.dp)
+            .fillMaxWidth(),
+        contentAlignment = Alignment.CenterStart,
+    ) {
+        Text(
+            text = title,
+            color = FirefoxTheme.colors.textAccent,
+            fontWeight = FontWeight.W400,
```
gabrielluong, line 110: > hey, @gabrielluong, thanks for the early feedback, I have addressed it (except the "Go to main component" thing, editor mode isn't available for me).
> 
> I _really_ want to land it as soon as possible and ship unified-search in 115, that's the only blocker. And, generally speaking, I don't think nits should be blocking a feature. Let's do things like extracting a general compose component to AC layer as follow ups, and now focus only on blocking changes.
> 
> Once we uplift it to beta and gain confidence, I will be doing some refactoring – removing duplicated setting screen, etc. It looks like a good opportunity to extract a basic `@Compose `component as well. Writing it now doesn't really help with landing the feature and is more of scope creeping.
> 
> I wanted to hide the new UI behind the feature flag, so I copy-pasted a lot of existing code with slight modifications to match the new UI to provide an alternative path for up to show the settings screen. Changing initial designed would make the patch smaller, but more difficult to roll-back.

Could you ask @MozillaNoah for the UI review since Jon only approved the non-UI bits? I don't have the bandwidth for this. As any code reviewer, the only takeaway here is that we don't necessarily want to do +1189 lines of code review when we could break it down and land incrementally, and to be fair, it's harder to give constructive feedback when there's just so many things to comment on with any large PR.

I have 2 areas that I would ask that we focus on:
- Making sure we are aligned with the figma designs. In the highlighted code here, I found that the title is using W500 that already exists in the headline8 text style. ([figma for the title](https://www.figma.com/file/WzMZduDsU0dJUj8P3gqiPf/Search-engine-settings?type=design&node-id=40-22583&t=BAdltawcpOKBZbfs-0)). 
- Addressing the existing feedback. I would suggest resolving the existing review comments with a comment about the actions taken or decision made so that we're closing the loop. I know you said you addressed all the feedback, but it's kinda hard to tell since I could still see paddings with only "start" instead of "horizontal". If the changes were intentional, then a comment would clarify that. Even if the comment only says that you would file a follow up.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,273 @@
...
+                engine.name,
+            ),
+        )
+
+        Spacer(modifier = Modifier.width(16.dp))
+
+        Text(
+            modifier = Modifier
+                .align(Alignment.CenterVertically)
+                .padding(top = 8.dp, bottom = 8.dp)
```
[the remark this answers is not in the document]
mavduevskiy: implemented

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchShortcutsFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+                        requireContext().components.useCases.searchUseCases.updateDisabledSearchEngineShortcutsIds(
+                            engine.id,
+                            isEnabled,
+                        )
+                    },
+                    onDeleteEngineClicked = {
+                        deleteSearchEngine(requireContext(), it)
+                    },
+                    onAddEngineClicked = {
+                        navigateToSaveEngineFragment(null)
```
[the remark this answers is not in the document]
mavduevskiy: implemented

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,273 @@
...
+                }
+            }
+        }
+    }
+}
+
+@Composable
+private fun SearchEngineDropdownMenu(
+    engine: SearchEngine,
+    isMenuExpanded: MutableState<Boolean>,
```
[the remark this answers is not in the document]
mavduevskiy: implemented

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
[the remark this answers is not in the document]
mavduevskiy: created a [follow up](https://bugzilla.mozilla.org/show_bug.cgi?id=1837113)

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
mavduevskiy: implemented

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/SearchState.kt
mavduevskiy: implemented

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/SearchState.kt
mavduevskiy: implemented

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/middleware/SearchMiddleware.kt
mavduevskiy: implemented

fenix/app/src/main/java/org/mozilla/fenix/settings/search/RadioSearchEngineListPreference.kt
mavduevskiy: implemented

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,273 @@
...
+}
+
+@Composable
+private fun AddEngineButton(
+    onAddEngineClicked: () -> Unit,
+) {
+    Row(
+        modifier = Modifier
+            .defaultMinSize(minHeight = 56.dp)
+            .padding(start = 4.dp)
```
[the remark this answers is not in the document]
mavduevskiy: 4dp is here to match the spacing of radio button elements. Added end 16dp spacing

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchShortcutsFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+        }
+
+        return tabsTrayDialogBinding.root
+    }
+
+    private fun navigateToSaveEngineFragment(engine: SearchEngine?) {
+        val directions = SearchShortcutsFragmentDirections
+            .actionSearchShortcutsFragmentToSaveSearchEngineFragment(engine?.id)
+
+        Navigation.findNavController(requireView()).navigate(directions)
```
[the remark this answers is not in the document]
mavduevskiy: fixed

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
mavduevskiy: @gabrielluong 
I did, but he didn't have time. Somebody else will take a look, no worries.

This design has some inconsistencies with the rest of the app that I decided not to introduce. `PreferenceCategory` class is using 400, and it is very obsious coming from the [searching_preference](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/xml/search_preferences.xml) screen.
 400 | 500
 --- | --- 
 |  

The design also doesn't have top spacing between the label and the top of the screen we have all over the settings screens. I kept the spacing for consistency.

gabrielluong:
@mavduevskiy I will find some time tonight to help you review this.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -1409,6 +1409,7 @@ sealed class SearchAction : BrowserAction() {
         val regionSearchEngines: List<SearchEngine>,
         val customSearchEngines: List<SearchEngine>,
         val hiddenSearchEngines: List<SearchEngine>,
+        val disabledSearchEngineShortcutsIds: List<String>,
```
jonalmeida, line 1412:
```suggestion
        val disabledSearchEngineIds: List<String>,
```

Drive-by: you missed one more reference to update. 🙂

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineFragment.kt
```diff
@@ -22,8 +24,15 @@ import org.mozilla.gecko.search.SearchWidgetProvider
 
 class SearchEngineFragment : PreferenceFragmentCompat() {
 
+    private var newUI: Boolean = false
```
MatthewTighe, line 27:
```suggestion
    private val shouldUseNewUI: Boolean by lazy {
        requireContext().settings().enableNewSearchSettings
    }
```

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchShortcutsFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+            false,
+        )
+
+        tabsTrayDialogBinding.root.setContent {
+            FirefoxTheme(theme = Theme.getTheme(allowPrivateTheme = false)) {
+                SearchEngineShortcuts(
+                    getString(R.string.preferences_category_engines_in_search_menu),
+                    requireComponents.core.store,
+                    onEditEngineClicked = {
+                        deleteSearchEngine(requireContext(), it)
```
MatthewTighe, line 53: I think this should be navigate instead of delete

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchShortcutsFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+import org.mozilla.fenix.theme.Theme
+import org.mozilla.fenix.utils.allowUndo
+
+/**
+ * A [Fragment] that allows user to select what search engine shortcuts will be visible in the quick
+ * search menu.
+ */
+class SearchShortcutsFragment : Fragment(R.layout.fragment_search_shortcuts) {
+
+    lateinit var tabsTrayDialogBinding: FragmentSearchShortcutsBinding
```
MatthewTighe, line 34: I believe this will want to be nullable and handled by `onDestroyView` as is done with the binding in `SaveSearchEngineFragment`.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SaveSearchEngineFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+import org.mozilla.fenix.settings.SupportUtils
+
+/**
+ * A [Fragment] that allows user to add a new search engine.
+ */
+@SuppressWarnings("LargeClass", "TooManyFunctions")
+class SaveSearchEngineFragment : Fragment(R.layout.fragment_save_search_engine) {
+
+    private val args by navArgs<SaveSearchEngineFragmentArgs>()
+    private var searchEngine: SearchEngine? = null
```
MatthewTighe, line 38:
```suggestion
    private var searchEngine: SearchEngine? by lazy {
        requireComponents.core.store.state.search.customSearchEngines.find { engine ->
            engine.id == args.searchEngineIdentifier
        }
    }
```

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SaveSearchEngineFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+            )
+        }
+    }
+
+    override fun onResume() {
+        super.onResume()
+        if (args.searchEngineIdentifier != null) {
+            showToolbar(getString(R.string.search_engine_edit_custom_search_engine_title))
+        } else {
+            showToolbar(getString(R.string.search_engine_add_custom_search_engine_title))
```
MatthewTighe, line 105: Does this toolbar code need to run in `onResume` specifically? I think we could avoid introducing another function and do it in in `onViewCreated`

MatthewTighe (changes requested):
This is looking good to me, though I did catch one bug in the overflow menu of custom search engines. I did not really review the Compose code because I think GL will do a better job of that, but I will cycle back for it tomorrow morning if he isn't able to get to it by then.

I'm sympathetic to the time pressures here, but I do wanna re-iterate a bit of what others have mentioned in terms of PR size. It would be easier (and probably faster, ultimately) to land + review changes that were a bit more incremental IMO. 

I would also generally prefer that we try to push back against adding features that knowingly introduce tech debt. In this case I mean things like adding new XML layouts instead of Compose. Again though I'm sympathetic to the time pressures here so I can understand why it was prudent to borrow and update existing code, and I'm happy to move forward with landing those pieces as is.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchShortcutsFragment.kt
MatthewTighe: I also see the name is `tabsTrayDialogBinding` so that should be updated

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchShortcutsFragment.kt
```diff
@@ -0,0 +1,104 @@
...
+        tabsTrayDialogBinding.root.setContent {
+            FirefoxTheme(theme = Theme.getTheme(allowPrivateTheme = false)) {
+                SearchEngineShortcuts(
+                    getString(R.string.preferences_category_engines_in_search_menu),
+                    requireComponents.core.store,
+                    onEditEngineClicked = {
+                        deleteSearchEngine(requireContext(), it)
+                    },
+                    onCheckboxClicked = { engine, isEnabled ->
+                        requireContext().components.useCases.searchUseCases.updateDisabledSearchEngineShortcutsIds(
```
MatthewTighe, line 56: nit: this line is a bit too long. a newline after `searchUseCases` would be visually useful I think

fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -67,6 +67,12 @@ object FeatureFlags {
      */
     const val saveToPDF = true
 
+    /**
+     * Enables new search settings UI with two extra fragments, for managing the default engine
+     * and managing search shortcuts in the quick search menu.
+     */
+    val newSearchSettings = Config.channel.isNightlyOrDebug
```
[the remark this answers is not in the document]
mavduevskiy: I went a step further and created a [ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1837649) to remove all code related to the old flow, if it aligns with your intention

fenix/app/src/main/res/color/toggle_positive_button_text_color.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
+    <item android:state_enabled="false" android:alpha="0.50" android:color="?attr/textPrimary" />
+    <item android:color="?attr/textOnColorPrimary"/>
+</selector>
```
MatthewTighe, line 8: Rather than introduce a new color here, I think I would prefer to just change the alpha of the text color through the code in `SaveSearchEngineFragment` where we are already updating the `saveButton.isEnabled` property

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1762,4 +1762,9 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         key = appContext.getPreferenceKey(R.string.pref_key_growth_early_search),
         default = false,
     )
+
+    /**
+     * Indicates if the new Search settings UI is enabled.
+     */
+    var enableNewSearchSettings: Boolean = showUnifiedSearchFeature && FeatureFlags.newSearchSettings
```
MatthewTighe, line 1769: I have the same nit about the `NewSearchSettings` naming here. Would appreciate it if we could find something slightly more descriptive

MatthewTighe (changes requested):
Cycled back through this morning, thanks for working through these reviews with us!

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SaveSearchEngineFragment.kt
mavduevskiy: I was just copying `AddSearchEngineFragment`, but not taking a closer look `showToolbar` is mostly called during `onResume` call throughout the app.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/Extensions.kt
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * The list of search engine shortcuts to be available for quick search menu.
+ */
+val SearchState.searchEngineShortcuts: List<SearchEngine>
+    get() = (
+        regionSearchEngines + additionalSearchEngines + availableSearchEngines +
+            customSearchEngines + applicationSearchEngines
+        ).filter {
+        !disabledSearchEngineIds.contains(it.id)
+    }
```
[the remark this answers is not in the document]
MatthewTighe: Moving the extension function to A-C wouldn't prevent Focus from defining their own ordering, and I think A-C already has some opinionated functionality exposed in terms of ordering, like in `SearchState.searchEngines`. 

I also see that there is already new code in A-C that references `shortcutIds` like `BrowserState.updateDisabledSearchEngineShortcutsIds` and `class UpdateDisabledSearchEngineShortcutsIdsUseCase`, which is another part of my reasoning.

That said, I'm not overly concerned either way so I'll leave it up to whatever you want to do.

fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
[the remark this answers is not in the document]
MatthewTighe: Yeah that's great, thanks!

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SaveSearchEngineFragment.kt
MatthewTighe: Okay, no need to reinvent the wheel then

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,254 @@
...
+    store: BrowserStore,
+    onCheckboxClicked: (SearchEngine, Boolean) -> Unit,
+    onEditEngineClicked: (SearchEngine) -> Unit,
+    onDeleteEngineClicked: (SearchEngine) -> Unit,
+    onAddEngineClicked: () -> Unit,
+) {
+    val searchState = store.observeAsComposableState { it.search }.value ?: SearchState()
+    val searchEngines = with(searchState) {
+        regionSearchEngines + additionalSearchEngines + availableSearchEngines + customSearchEngines
+    }
```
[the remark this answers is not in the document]
mavduevskiy: `searchEngines + availableSearchEngines` gives the wrong order, we want to have available search engines (wiki and youtube) to be above the custom search engines. And yes, we also are [not showing](https://www.figma.com/file/WzMZduDsU0dJUj8P3gqiPf/Search-engine-settings?type=design&node-id=243-16216&t=bnt7IoL86eHywJmy-4) `applicationEngines` on this screen.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,254 @@
...
+                        showMenu = isMenuExpanded.value,
+                        onDismissRequest = { isMenuExpanded.value = false },
+                        menuItems = listOf(
+                            MenuItem(
+                                stringResource(R.string.search_engine_edit),
+                                color = FirefoxTheme.colors.textWarning,
+                            ) {
+                                onEditEngineClicked(
+                                    engine,
+                                )
```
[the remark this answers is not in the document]
mavduevskiy: I actually talked with @jeffreygee about it, he was going to talk to the design system team about that

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,254 @@
...
+                        tint = FirefoxTheme.colors.iconPrimary,
+                    )
+
+                    ContextualMenu(
+                        showMenu = isMenuExpanded.value,
+                        onDismissRequest = { isMenuExpanded.value = false },
+                        menuItems = listOf(
+                            MenuItem(
+                                stringResource(R.string.search_engine_edit),
+                                color = FirefoxTheme.colors.textWarning,
```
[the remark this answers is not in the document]
mavduevskiy: I just followed the existing implementation of the menu. But I agree, it's a bit confusing, and having only delete item in red colour sounds more reasonable. Should we sneak in the change?

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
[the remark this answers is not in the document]
MatthewTighe: Ah, that makes sense. I wasn't really sure what `applicationEngines` were but the design clarifies it. Thanks!

fenix/app/src/main/res/color/toggle_positive_button_text_color.xml
mavduevskiy: Would it be a better way though? Here we just allow system to handle it, as we manage the enabled state. The alternative would be something like

```
 val textColor = binding.saveButton.currentTextColor
 val textColorWithAlpha = ColorUtils.setAlphaComponent(textColor, if (bothFieldsHaveInput) 255 else 127)
 binding.saveButton.setTextColor(textColorWithAlpha)
```
to be triggered on create and on text changed?

fenix/app/src/main/res/xml/search_preferences_new.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gabrielluong, line 1: **nit:** Let's just this file to `search_settings_preferences` since it will align with the fragment toolbar name and it avoids having to rename this later on.

fenix/app/src/main/res/xml/default_engine_preferences.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gabrielluong, line 1: **nit:** Let's rename this to be `default_search_engine_preferences` so that the name will map to what we see when viewing the fragment on the device.

fenix/app/src/main/res/xml/search_preferences_new.xml
```diff
@@ -0,0 +1,76 @@
...
+    <PreferenceCategory
+        android:title="@string/preferences_search_address_bar"
+        android:selectable="false"
+        app:iconSpaceReserved="false"
+        android:layout="@layout/preference_category_no_icon_style">
+        <SwitchPreference
+            app:iconSpaceReserved="false"
+            android:defaultValue="true"
+            android:key="@string/pref_key_show_search_engine_shortcuts"
+            android:title="@string/preferences_show_search_engines" />
```
gabrielluong, line 32: **Blocker:** I am seeing the visibility of this preference disappearing when the fragment is being loaded on the device. This is happening as a result of https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineFragment.kt#53. I would suggest either removing this or setting this to be default invisible if we don't expect to see this for the new search view.

Otherwise, this will cause some movement when the screen is being created and the preference is being hidden.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
mavduevskiy: Changed it to `enableNewSearchSettingsUI`, to say roughly the same as the comment. I can't come up with a better naming, but open for suggestions :)

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,254 @@
...
+import org.mozilla.fenix.compose.MenuItem
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Top-level UI for search shortcuts settings.
+ */
+@Suppress("LongParameterList")
+@Composable
+fun SearchEngineShortcuts(
+    categoryTitle: String,
```
gabrielluong, line 55: **nit:** It seems like we are passing the same string here all the time. This probably makes sense to just inline.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,254 @@
...
+import org.mozilla.fenix.compose.ContextualMenu
+import org.mozilla.fenix.compose.MenuItem
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Top-level UI for search shortcuts settings.
+ */
+@Suppress("LongParameterList")
+@Composable
+fun SearchEngineShortcuts(
```
gabrielluong, line 54: **Blocker:** I didn't see any @Preview for these composables. I would suggest adding one for the top level composable.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
gabrielluong: +1 with keeping consistent with W400 for now. Let's file a followup bug to see if we should make sure all the other headline should be using W500.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,254 @@
...
+    onDeleteEngineClicked: (SearchEngine) -> Unit,
+) {
+    val isMenuExpanded: MutableState<Boolean> = remember { mutableStateOf(false) }
+
+    Row(
+        modifier = Modifier
+            .defaultMinSize(minHeight = 56.dp)
+            .padding(start = 4.dp),
+    ) {
+        Checkbox(
```
gabrielluong, line 134: **Blocker:** I believe the uncheck color for these Checkbox needs to be adjusted for the dark theme. They should appear white for the dark theme.

gabrielluong (commented):
This is looking good to me. I am sorry for piling more comments here. I have indicated which ones that I felt are blockers and others that could be followed up upon. I would suggest filing a new bug for any followup that we can agree upon.

fenix/app/src/main/res/color/toggle_positive_button_text_color.xml
MatthewTighe: I personally prefer to read code rather than need to search through XML files, but I don't care enough to block on this. FWIW, I think the above could also be

```
            if (bothFieldsHaveInput) {
                binding.saveButton.isEnabled = true
                binding.saveButton.alpha = 1f
            } else {
                binding.saveButton.isEnabled = false
                binding.saveButton.alpha = 0.5f
            }
```

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineFragment.kt
mavduevskiy: It breaks `SearchEngineFragmentTest`, as requireContext() would be null. And with the way the test is written I don't think we make it work.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/Extensions.kt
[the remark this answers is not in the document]
mavduevskiy: Actually, keeping those was an overlook on my part, as @jonalmeida asked to get rid of the shortcut name.
Since Jonathan originally asked me to move the extension away from the `SearchState`, I wonder if he could share his opinion on the subject :)

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/storage/SearchMetadataStorage.kt
```diff
@@ -78,6 +79,18 @@ internal class SearchMetadataStorage(
             ?.toList() ?: emptyList()
     }
 
+    override suspend fun getDisabledSearchEngineIds(): List<String> {
+        return preferences.value
+            .getStringSet(PREFERENCE_KEY_DISABLED_SEARCH_ENGINE_ID, setOf("reddit", "youtube"))
```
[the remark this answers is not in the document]
mavduevskiy: I actually was debating on that myself. Changed `SearchMetadataStorage` to accept the disabled set in the constructor, so the `SearchMiddleware` could pass on `additionalBundledSearchEngineIds` to the storage.

fenix/app/src/main/res/color/toggle_positive_button_text_color.xml
mavduevskiy: Oh, I see what you mean. We could even move it into two lines.
```
binding.saveButton.isEnabled = bothFieldsHaveInput
binding.saveButton.alpha = if (bothFieldsHaveInput) 1f else 0.5f
```

But this file doesn't just solve the `alpha` problem, it also adds a colour state for the button to be [white](https://www.figma.com/file/WzMZduDsU0dJUj8P3gqiPf/Search-engine-settings?type=design&node-id=262-27194&t=rUBEH49xfwNHzIHm-4) in enabled mode. 

Which made me realize that what we actually need is another state list color. Added `text_on_color_state_list_text_color.xml`

fenix/app/src/main/res/xml/search_preferences_new.xml
mavduevskiy: While I don't like it myself, it's not something that was introduced with the patch. It was copied from `search_preferences.xml` and we have this behaviour in nightly and it has been around for 9 months. 
Created a [follow up](https://bugzilla.mozilla.org/show_bug.cgi?id=1837864).

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/SearchState.kt
```diff
@@ -38,6 +38,7 @@ data class SearchState(
     val additionalSearchEngines: List<SearchEngine> = emptyList(),
     val additionalAvailableSearchEngines: List<SearchEngine> = emptyList(),
     val hiddenSearchEngines: List<SearchEngine> = emptyList(),
+    val disabledSearchEngineIds: List<String> = emptyList(),
```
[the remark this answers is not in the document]
mavduevskiy: I find this suggestion to be very debatable. 

We already are using`hiddenSearchEngines`, that is an individual list of ids to handle UI state. Also, `isEnabled` should be persisted, and persisting it in `xml` is too costly. Calculating it in the builder, as we do with `isGeneral`, based on list of disabled ids we stored separately doesn't seem like an improvement to me. And it goes against the established design in `hiddenSearchEngines`.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
```diff
@@ -0,0 +1,316 @@
...
+                IconButton(
+                    onClick = {
+                        isMenuExpanded.value = true
+                    },
+                ) {
+                    Icon(
+                        painter = painterResource(id = R.drawable.ic_menu),
+                        contentDescription = stringResource(id = R.string.content_description_menu),
+                        tint = FirefoxTheme.colors.iconPrimary,
+                    )
```
MatthewTighe, line 194: optional nit: I think this conditional would be easier to parse if we extracted the true case `engine.type == SearchEngine.Type.CUSTOM` into a separate Composable. So the `Box` and everything in it, basically

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1762,4 +1762,9 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         key = appContext.getPreferenceKey(R.string.pref_key_growth_early_search),
         default = false,
     )
+
+    /**
+     * Indicates if the new Search settings UI is enabled.
+     */
+    var enableNewSearchSettingsUI: Boolean = showUnifiedSearchFeature && FeatureFlags.unifiedSearchSettings
```
MatthewTighe, line 1769: nit: would it make sense to use the `unifiedSearch` language here as well to match the feature flag naming?

MatthewTighe (commented):
I have a few more nits, but I also noticed another bug while testing it this morning.

STR:
click search selector while keyboard is collapsed and pick youtube or reddit

expected:
youtube or reddit is used as engine for that search

actual:
switches back to default search engine

Video demonstrates it working correctly when the toolbar is already focused, and working incorrectly when the toolbar is not focused.

IMO, this could potentially be a small enough bug to handle in a follow-up, but I am not really sure what's causing it.


https://github.com/mozilla-mobile/firefox-android/assets/10427470/89bf17bb-c973-46d7-88f4-bab104fb8164

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
mavduevskiy: I think I want to revisit that comment during the compose refactoring [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1837113#c2). Made a note about your comment.

mavduevskiy:
@MatthewTighe thanks, Matt! 

fixed the bug. the search engine isn't passed directly to the dialog fragment, but just an id. the fragment then looks for the matching search engine, but was doing it in the wrong place and defaulting when it failed to find the engine.

fixed the nits, except the Compose refactoring one! 
Do you think we can land now? :)

MatthewTighe (approved):
This LGTM given the scope and the follow-ups that were filed. I do see that GL had another [unaddressed nit](https://github.com/mozilla-mobile/firefox-android/pull/2274#discussion_r1224810127) about renaming the extension function file. I think it is worth doing for search-ability in the IDE, since the growing number of files named "Extensions.kt" will make it hard to open the correct one when doing things like CMD + shift + O.

Thanks again for your patience cycling on these reviews! I would prefer to make sure we aren't landing something @gabrielluong would consider a blocker, so would appreciate a quick sign-off from him as well. For example, I'm happy with the resolution of [this comment](https://github.com/mozilla-mobile/firefox-android/pull/2274#discussion_r1224828916) for now

mavduevskiy:
> This LGTM given the scope and the follow-ups that were filed. I do see that GL had another [unaddressed nit](https://github.com/mozilla-mobile/firefox-android/pull/2274#discussion_r1224810127) about renaming the extension function file. I think it is worth doing for search-ability in the IDE, since the growing number of files named "Extensions.kt" will make it hard to open the correct one when doing things like CMD + shift + O.

I treat nits as optional, as since there were a lot of people participating didn't want to close the thread. And since there was no reasoning behind that suggestion, I decided to follow the existing pattern of `Extention.kt`. It is a confusing when copying an existing pattern leads to change requests.

gabrielluong:
> This LGTM given the scope and the follow-ups that were filed. I do see that GL had another [unaddressed nit](https://github.com/mozilla-mobile/firefox-android/pull/2274#discussion_r1224810127) about renaming the extension function file. I think it is worth doing for search-ability in the IDE, since the growing number of files named "Extensions.kt" will make it hard to open the correct one when doing things like CMD + shift + O.
> 
> Thanks again for your patience cycling on these reviews! I would prefer to make sure we aren't landing something @gabrielluong would consider a blocker, so would appreciate a quick sign-off from him as well. For example, I'm happy with the resolution of [this comment](https://github.com/mozilla-mobile/firefox-android/pull/2274#discussion_r1224828916) for now

No extra review necessary from me. I did take a scan of everything. I would ideally like to see a comment to indicate the result for all the resolved conversations for those that don't have a comment. 

The only notable thing I want to see some follow up actions on is https://github.com/mozilla-mobile/firefox-android/pull/2274#discussion_r1225886471. While I understand the visibility is turned off by the existing code, nonetheless we are displaying a new search preference which will be affected. I just want to make sure that this is something that we will address. This could mean simply adding it to the "Up next" in Jira or just having it assigned so it doesn't seem like we filed a followup and forgotten since this is usually something we fix before shipping a feature which has preferences.

One last thing, I do believe there has been some feedback form the Design System team (see [Slack](https://mozilla.slack.com/archives/CNP7Q51KN/p1686592786439929)) regarding https://github.com/mozilla-mobile/firefox-android/pull/2274#discussion_r1224839099. I just want to make sure we have a follow up ticket that indicates that we want to implement the W500 for the headlines in the settings.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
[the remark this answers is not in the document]
mavduevskiy: ignored

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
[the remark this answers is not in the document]
mavduevskiy: ignored

fenix/app/src/main/java/org/mozilla/fenix/settings/search/Extensions.kt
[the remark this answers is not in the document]
mavduevskiy: moved to a separate file, but kept inside Fenix

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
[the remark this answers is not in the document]
mavduevskiy: implemented

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SaveSearchEngineFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+                true
+            }
+
+            !searchString.contains("%s") -> {
+                binding.customSearchEngineSearchStringField.error =
+                    resources.getString(R.string.search_add_custom_engine_error_missing_template)
+                true
+            }
+
+            else -> false
```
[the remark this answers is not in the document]
mavduevskiy: implemented

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/SearchState.kt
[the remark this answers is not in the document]
mavduevskiy: ignored

fenix/app/src/main/java/org/mozilla/fenix/settings/search/DefaultEngineFragment.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.preference.PreferenceFragmentCompat
+import mozilla.components.support.ktx.android.view.hideKeyboard
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.getPreferenceKey
+import org.mozilla.fenix.ext.showToolbar
+
+/**
+ * A [Fragment] that allows user to set the default search engine.
+ */
+class DefaultEngineFragment : PreferenceFragmentCompat() {
```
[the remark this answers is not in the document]
mavduevskiy: implemented

fenix/app/src/main/res/xml/search_preferences_new.xml
mavduevskiy: implemented

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
mavduevskiy: ignored

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
mavduevskiy: implemented

mavduevskiy:
@gabrielluong 
Thanks for keeping an eye on it. I sometimes omit leaving comments and just live a thumbs-up if I like the suggestion and implement it.

I have assigned the animation bug to myself. Still waiting for more clarity from the UX team before I can file a bug.

mavduevskiy:
@MatthewTighe @gabrielluong @jonalmeida 
Just wanted to say thanks for helping me out with this. Your feedback is (most of the time) really appreciated!

gabrielluong:
> > This LGTM given the scope and the follow-ups that were filed. I do see that GL had another [unaddressed nit](https://github.com/mozilla-mobile/firefox-android/pull/2274#discussion_r1224810127) about renaming the extension function file. I think it is worth doing for search-ability in the IDE, since the growing number of files named "Extensions.kt" will make it hard to open the correct one when doing things like CMD + shift + O.
> 
> I treat nits as optional, as since there were a lot of people participating didn't want to close the thread. And since there was no reasoning behind that suggestion, I decided to follow the existing pattern of `Extention.kt`. It is a confusing when copying an existing pattern leads to change requests.
> 
> Anyway, thanks for the review, Matt!

Rather than treating nits as optional, I think there is always a more effective way to respond to these review comments when it seems debatable by simply responding about the action and rationale you have chosen to take given the feedback instead of leaving it as unaddressed. We can be more effective when we remove some of the guess work in the follow up reviews when we are looking to see whether or not an issue has been addressed. 

Aside, having an `ext` package is totally reasonable within the existing patterns - I just don't want this to seem like my comment came out of the air with no rationale behind it. See https://searchfox.org/mozilla-mobile/search?q=.ext.&path= for examples. Typically you will see `ext/<blah>.kt` when we are only adding extension functions for `blah`. I don't see a problem with either direction taken, but let's try to be more conscientious about closing the loop in these review comments so reviewers don't have to ask why a nit was unaddressed.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineShortcuts.kt
gabrielluong: @mavduevskiy Any particular reason for ignoring this one? Ideally, we are also explaining why we are ignoring something.

mavduevskiy:
@gabrielluong 
Fair enough! I will be more transparent with reacting to comments from now on👍

mavduevskiy:
@Mergifyio backport releases_v115

PR #2276, opened 2023-05-30
Bug 1807154: Persist desktop mode option for tabs when Fenix is reopened

Alexandra-Virvara:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1807154


android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -767,10 +767,11 @@ abstract class EngineSession(
      * will be restored.
      *
      * @param state A saved session state.
+     * @param enableDesktopMode True, if desktop mode option is enabled, false otherwise
      * @return true if the engine session has successfully been restored with the provided state,
      * false otherwise.
      */
-    abstract fun restoreState(state: EngineSessionState): Boolean
+    abstract fun restoreState(state: EngineSessionState, enableDesktopMode: Boolean): Boolean
```
jonalmeida, line 774: We shouldn't be passing the flag individually as a param here. What happens when we want to add more things? That shouldn't mean n parameters.

We should get the toggle state from the `EngineSessionState` which each of the engine's implement.

jonalmeida (changes requested):
This looks good! There is one change that I think we should make though. Let me know if my comment isn't clear and I can try to do better.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -336,6 +340,11 @@ class GeckoEngineSession(
             return false
         }
 
+        // TODO fetch the correct value for enableDesktopMode
```
jonalmeida, line 343: Argh, this comment is the place where you've discovered that this solution isn't enough. It looks like we should have GeckoView save and restore the state in their `EngineSessionState` (`GeckoSessionState`) which we can then apply here.

Please file a bug in the GeckoView component to add support for restoring desktop mode. I think your change today can still work with some tweaks.

A possible solution is to set the desktop mode _before_ the `restoreState` is called. This happens in the `CreateEngineSessionMiddleware` - specifically above [this line](https://searchfox.org/mozilla-mobile/rev/9f5ba3d5f8fb052d07447bb3b406d4a3b8ca2d82/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/CreateEngineSessionMiddleware.kt#119) where we can set the value before instructing the `EngineSession` to load.

That might suffice for now.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/Engine.kt
```diff
@@ -114,7 +114,11 @@ interface Engine : WebExtensionRuntime, DataCleanable {
      * @return the newly created [EngineSession].
      */
     @MainThread
-    fun createSession(private: Boolean = false, contextId: String? = null): EngineSession
+    fun createSession(
+        private: Boolean = false,
+        enableDesktopMode: Boolean = false,
```
jonalmeida, line 119: General advice: This is another place to stop and consider if we should be passing this as a separate parameter. We're making a big API change here.

I see from the `GeckoEngineSession` implementation that we're using this value to set the `viewPortMode`, but this isn't the only value needed in order to change into desktop mode (we need to set the `userAgentMode` too). If we were to add the user-agent as another parameter that sounds like we are starting a pattern where we could lead to new params for each new setting in the future.

For the cases of `private` and `contextId` these are explicitly needed internally to create a GeckoSession (basically, they are special). Is enabling desktop mode one of those? I'm not sure.

We might need to consider if this change is necessary or the strategy. See the comment above where I think you've identified the problem already with a possible workaround.

Alexandra-Virvara:
bug for GeckoView: https://bugzilla.mozilla.org/show_bug.cgi?id=1838851

gabrielluong:

PR #2288, opened 2023-05-31
Bug 1796331 - Added @Throws for AutofillCreditCardsAddressesStorage 

sharmaachintya08:
In this issue they requested us to add throws annotation to the required methods in AutofillCreditCardsAddressesStorage.

There are only 2 methods on which throws annotation could be applied, those methods already had try and catch block in it.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1796331

### DIFF
--- a/android-components/components/service/sync-autofill/src/main/java/mozilla/components/service/sync/autofill/AutofillCreditCardsAddressesStorage.kt
+++ b/android-components/components/service/sync-autofill/src/main/java/mozilla/components/service/sync/autofill/AutofillCreditCardsAddressesStorage.kt
@@ -10,6 +10,7 @@ import androidx.annotation.VisibleForTesting
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.withContext
+import kotlin.jvm.Throws
 import mozilla.appservices.autofill.AutofillApiException.NoSuchRecord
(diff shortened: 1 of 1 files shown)


MatthewTighe:
@gabrielluong I'm happy to review this, but was wondering about the original context. Was the goal to expose the error types through annotations like this, or were we hoping to capture the errors in a crash reporter/logger?

MatthewTighe:
@gabrielluong 
I see that the [patch](https://github.com/gabrielluong/android-components/commit/c9cdc7ec44d25980cd81135f33d03b565639472b) introducing the try/catch was actually landed several days before the [original issue](https://github.com/mozilla-mobile/android-components/issues/10275) was opened, which is why I was a little confused 😅 

Kotlin does not have checked exceptions, so the only reason to include the `@Throws` annotation is for Java compatibility since that language does have checked exceptions. That said, the try/catch block that was introduced will explicitly capture `NoSuchRecord` exceptions such that they aren't ever rethrown from this method. 

As such I am wondering if perhaps the original intent of the issue was to do more robust error handling or logging here.

Either way, I think if we need to do anything here we should file a new issue. I will close the old one.

@sharmaachintya08 We really appreciate your initiative on contributing, but as explained above I think this patch is unneeded. Closing this PR

PR #2301, opened 2023-06-01
Bug 1815579 - Improve performance of image loading in tab items

MozillaNoah:
#### Thumbnail bitmap retention
Thumbnail bitmaps were seemingly fetched, retained, and re-fetched. They were garbage collected when the Tabs Tray was dismissed, but violently scrolling up and down would explode the in-app memory.
  - My solution for this was to add a bitmap recycle call to `ThumbnailImage` in a `DisposableEffect` block.

#### The Favicon composable
Relying on the `Favicon` Composable when the icon was available via `TabsessionState.content.icon` was causing unnecessary icon fetches inside `TabGridItem` AND `ThumbnailCard`. 
  - I ended up needing to differentiate use cases for the original `ThumbnailCard`; one via Recent Synced Tabs (where `TabSessionState` isn't available) and one where `TabSessionState` is the driving data object. `TabThumbnail` is driven via `TabSessionState` so that the passed-in bitmap of `tab.content.icon` can be used instead of needlessly re-fetching the favicon.
  - The favicon (whether via `Favicon` or `TabsessionState.content.icon`) was also being loaded & rendered before the thumbnail was loaded AND they were existing simultaneously as the thumbnail (the thumbnail was being painted on top of the favicon). I refactored `ThumbnailImage` to have a `fallbackContent` Composable to render the favicon for a tab ONLY when the thumbnail could not be fetched. 
  - I also added an optional painter to `FaviconListitem` to cover this use case. This will aid in the performance of loading inactive tabs that have their icons still on disk.

#### Lazy list/grid composables
I noticed that some of the thumbnail fetching was built around the idea that the `LazyColumn` and `LazyGrid` Composables functioned 1:1 as a `RecyclerView`. I discovered this is not the case; Composables are NOT reused like Views are in the XML analog. We should not be coding our item Composables to take lazy loading into account but instead let their parent Composables handle this behind the scenes. I refactored the logic around fetching the thumbnails around a changing tab key string, since the key was never supposed to be changing in the first place. I also changed the bitmap `remember` to `rememberSaveable` as that will allow it to be preserved via app configuration changes. 

#### Bitmap fetching + side effect coroutines
One thing I'm not 100% certain on is using the coroutines from [side effects](https://developer.android.com/jetpack/compose/side-effects) to perform intense, time sensitive operations, like bitmap fetching in this situation. One hypothesis we had during investigation was the bitmap fetching coroutine in `LaunchedEffect` was launching _several seconds_ from when it should be. I opted to directly use `rememberCoroutineScope` instead to ensure the bitmap fetching coroutines are launched and de-scoped correctly.  

[One resource I used](https://developer.android.com/jetpack/compose/graphics/images/optimization)

### Profiles
- [Perf changes w/ 50 tabs, big fling down, big fling up](https://share.firefox.dev/3Jz8niy) (taken directly from the Firefox profiler)
- [Perf changes - open tabs tray w/ 50 tabs](https://share.firefox.dev/3NlD2RE) (taken via Android Studio profiler)

I tried to take more extensive profiles, but the tool in Android Studio was crashing in this use flow when capturing the trace.

### UI test comparison on Pixel 2
| XML time (run time, total) | Compose time before changes (run time, total) | Compose time after changes (run time, total) |
| -- | -- | -- |
| 3m 44s, 6m 51s | 3m 30s, 7m 13s | 3m 31s, 6m 39s |

### Videos (captured on Pixel 5, non-release build)
| XML (main) | Compose (main) | Compose (with changes) |
| -- | -- | -- |
| https://github.com/mozilla-mobile/firefox-android/assets/87384386/6febaae5-c6fd-4b4f-8989-11ae2e85a2a2 | https://github.com/mozilla-mobile/firefox-android/assets/87384386/64185033-4fc6-4a04-a5e1-9436c92e2a0e | https://github.com/mozilla-mobile/firefox-android/assets/87384386/87cf04ec-6923-47f3-997d-d9e82647dfa6 |

This is visual anecdotal evidence, but I still wanted to provide a "look and feel" comparison.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1815579

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/compose/TabThumbnail.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/compose/TabThumbnail.kt
@@ -5,7 +5,6 @@
 package org.mozilla.fenix.compose
 
 import androidx.compose.foundation.Image
-import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.RoundedCornerShape
@@ -15,25 +14,23 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.asImageBitmap
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
-import mozilla.components.browser.icons.compose.Loader
-import mozilla.components.browser.icons.compose.Placeholder
-import mozilla.components.browser.icons.compose.WithIcon
-import org.mozilla.fenix.components.components
+import mozilla.components.browser.state.state.TabSessionState
+import mozilla.components.browser.state.state.createTab
 import org.mozilla.fenix.theme.FirefoxTheme
 
 private const val THUMBNAIL_SIZE = 108
 private const val FALLBACK_ICON_SIZE = 36
 
 /**
- * Card which will display a thumbnail. If a thumbnail is not available for [url], the favicon
+ * Thumbnail belonging to a [tab]. If a thumbnail is not available, the favicon
  * will be displayed until the thumbnail is loaded.
  *
- * @param url Url to display thumbnail for.
- * @param key Key used to remember the thumbnail for future compositions.
+ * @param tab The given [TabSessionState] to render a thumbnail for.
  * @param size [Dp] size of the thumbnail.
  * @param backgroundColor [Color] used for the background of the favicon.
  * @param modifier [Modifier] used to draw the image content.
@@ -43,12 +40,12 @@ private const val FALLBACK_ICON_SIZE = 36
  * @param alignment [Alignment] used to draw the image content.
  */
 @Composable
-fun ThumbnailCard(
-    url: String,
-    key: String,
+@Suppress("LongParameterList")
+fun TabThumbnail(
+    tab: TabSessionState,
+    modifier: Modifier = Modifier,
     size: Dp = THUMBNAIL_SIZE.dp,
     backgroundColor: Color = FirefoxTheme.colors.layer2,
-    modifier: Modifier = Modifier,
     contentDescription: String? = null,
     contentScale: ContentScale = ContentScale.FillWidth,
     alignment: Alignment = Alignment.TopCenter,
@@ -58,31 +55,32 @@ fun ThumbnailCard(
         backgroundColor = backgroundColor,
     ) {
         ThumbnailImage(
-            key = key,
+            key = tab.id,
             size = size,
             modifier = modifier,
             contentScale = contentScale,
             alignment = alignment,
         ) {
-            components.core.icons.Loader(url) {
-                Placeholder {
-                    Box(modifier = Modifier.background(color = FirefoxTheme.colors.layer3))
-                }
-
-                WithIcon { icon ->
-                    Box(
-                        modifier = Modifier.size(FALLBACK_ICON_SIZE.dp),
-                        contentAlignment = Alignment.Center,
-                    ) {
-                        Image(
-                            painter = icon.painter,
-                            contentDescription = contentDescription,
-                            modifier = Modifier
-                                .size(FALLBACK_ICON_SIZE.dp)
-                                .clip(RoundedCornerShape(8.dp)),
-                            contentScale = contentScale,
-                        )
-                    }
+            Box(
+                modifier = Modifier.size(FALLBACK_ICON_SIZE.dp),
+                contentAlignment = Alignment.Center,
+            ) {
+                val icon = tab.content.icon
+                if (icon != null) {
+                    icon.prepareToDraw()
+                    Image(
+                        bitmap = icon.asImageBitmap(),
+                        contentDescription = contentDescription,
+                        modifier = Modifier
+                            .size(FALLBACK_ICON_SIZE.dp)
+                            .clip(RoundedCornerShape(8.dp)),
+                        contentScale = contentScale,
+                    )
+                } else {
+                    Favicon(
+                        url = tab.content.url,
+                        size = FALLBACK_ICON_SIZE.dp,
+                    )
                 }
             }
         }
@@ -93,11 +91,10 @@ fun ThumbnailCard(
 @Composable
 private fun ThumbnailCardPreview() {
     FirefoxTheme {
-        ThumbnailCard(
-            url = "https://mozilla.com",
-            key = "123",
+        TabThumbnail(
+            tab = createTab(url = "www.mozilla.com", title = "Mozilla"),
             modifier = Modifier
-                .size(THUMBNAIL_SIZE.dp)
+                .size(THUMBNAIL_SIZE.dp, 80.dp)
                 .clip(RoundedCornerShape(8.dp)),
         )
     }
--- a/fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
@@ -1,0 +1,127 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.compose
+
+import android.graphics.Bitmap
+import android.os.Parcelable
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.asImageBitmap
+import androidx.compose.ui.graphics.painter.BitmapPainter
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.platform.LocalDensity
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+import kotlinx.coroutines.launch
+import kotlinx.parcelize.Parcelize
+import mozilla.components.concept.base.images.ImageLoadRequest
+import org.mozilla.fenix.components.components
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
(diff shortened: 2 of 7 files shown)


jonalmeida:
> I performed the tests above by running the Espresso tests for the XML and Compose Tabs Trays. For the XML runs, I disabled the two tests that have not been converted to Compose yet.

Are we only comparing performance to the XML version? Shouldn't we also run these tests without your changes as well to see the difference with and without the diff?

Are there any links to the profiles that were gathered in the investigation? In the top-right corner, you can click the 'Upload local profile' to get a shareable link which can be added to the bugzilla or this PR.

MozillaNoah:
> Are we only comparing performance to the XML version? Shouldn't we also run these tests without your changes as well to see the difference with and without the diff?

I think that makes sense. Matt and I are gonna pair to figure out some more wins here, but I'll have to rebase this branch and re-grab results anyway, so I'll add another column for `Compose before changes`.

> Are there any links to the profiles that were gathered in the investigation? In the top-right corner, you can click the 'Upload local profile' to get a shareable link which can be added to the bugzilla or this PR.

Top-right corner of which application? I'll upload some profiles after the next round of testing.

fenix/app/src/main/java/org/mozilla/fenix/compose/TabThumbnail.kt
```diff
@@ -0,0 +1,95 @@
...
+    contentDescription: String? = null,
+    contentScale: ContentScale = ContentScale.FillWidth,
+    alignment: Alignment = Alignment.TopCenter,
+) {
+    Card(
+        modifier = modifier,
+        backgroundColor = backgroundColor,
+    ) {
+        Box(
+            modifier = Modifier.size(36.dp),
```
gabrielluong, line 54: Let's move the "36", "108", and "80" values into constants.

MozillaNoah:
@MatthewTighe this is ready for re-review

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailCard.kt
```diff
@@ -62,72 +57,35 @@ fun ThumbnailCard(
...
-        Image(
-            painter = painter,
-            contentDescription = null,
+        ThumbnailImage(
+            key = key,
+            size = size,
             modifier = modifier,
             contentScale = contentScale,
             alignment = alignment,
+            fallbackContent = {
```
MatthewTighe, line 66: optional nit: I almost feel like we could move this argument out of the list here and treat it the same way the that, for example, the `content` parameter for `Box`es are usually treated.

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
```diff
@@ -0,0 +1,106 @@
...
+    alignment: Alignment,
+    fallbackContent: @Composable () -> Unit,
+) {
+    if (inComposePreview) {
+        Box(modifier = Modifier.background(color = FirefoxTheme.colors.layer3))
+    } else {
+        val thumbnailSize = LocalDensity.current.run { size.toPx().toInt() }
+        val request = ImageLoadRequest(key, thumbnailSize)
+        val storage = components.core.thumbnailStorage
+        var state by rememberSaveable { mutableStateOf<Pair<Bitmap?, Boolean>>(Pair(null, false)) }
```
MatthewTighe, line 57: optional nit: I prefer private data classes to `Pair`s, because it adds more semantic value to the type. That is, it requires more attention to remember what `.first` and `.second` are compared to `.bitmap` and `.hasLoaded`. It might also save the re-declaration  of `val (bitmap, hasLoaded) = state`

fenix/app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt
```diff
@@ -107,11 +110,21 @@ fun FaviconListItem(
...
+                        .padding(horizontal = 16.dp)
+                        .size(ICON_SIZE),
+                )
+            } else {
+                Favicon(
+                    url = url,
+                    size = ICON_SIZE,
+                    modifier = Modifier.padding(horizontal = 16.dp),
+                )
+            }
```
MatthewTighe, line 127: I'm wondering if it's worth propagating this pattern all the way down into `Favicon` in a follow-up. I don't really know enough about how often they get invalidated to know if it's safe to rely on properties cached in memory.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
```diff
@@ -105,6 +107,7 @@ fun InactiveTabsList(
                         FaviconListItem(
                             label = tab.toDisplayTitle(),
                             description = tabUrl,
+                            faviconPainter = tab.content.icon?.run { BitmapPainter(this.asImageBitmap()) },
```
MatthewTighe, line 110: Related to my comment above about Favicon painters:

I'm just discovering this, but it seems odd to me that we are keeping this `tab.content.icon` in memory _and_ we have a separate memory cache for icons through the `components.core.icons.Loader()` function tree. I think it could be worth investigating in a follow-up if we're doing unnecessary work.

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
```diff
@@ -0,0 +1,106 @@
...
+    contentScale: ContentScale,
+    alignment: Alignment,
+    fallbackContent: @Composable () -> Unit,
+) {
+    if (inComposePreview) {
+        Box(modifier = Modifier.background(color = FirefoxTheme.colors.layer3))
+    } else {
+        val thumbnailSize = LocalDensity.current.run { size.toPx().toInt() }
+        val request = ImageLoadRequest(key, thumbnailSize)
+        val storage = components.core.thumbnailStorage
```
MatthewTighe, line 56: nit: I only see storage used once, so I don't think we need a separate variable declaration for it.

fenix/app/src/main/java/org/mozilla/fenix/compose/tabstray/TabGridItem.kt
```diff
@@ -146,13 +149,17 @@ fun TabGridItem(
...
+
+                        tab.content.icon?.let {
+                            Image(
+                                bitmap = it.asImageBitmap(),
+                                contentDescription = null,
+                                modifier = Modifier
+                                    .align(Alignment.CenterVertically)
+                                    .size(16.dp),
+                            )
+                        }
```
MatthewTighe, line 162: Does this mean that we won't load a favicon for the corner if it isn't already in memory?

MatthewTighe (commented):
This is looking good to me, just wanted to comment cycle on a few things real quick.

I did want to summarize my offline findings in terms of testing this patch for other readers:
To realize the performance gains of this patch, it will need to be built as a non-debuggable release variant. Compose performance is still hurt dramatically by debug builds, which means that the performance is noticeably slower than XML in those cases.. 

Since release builds seem comparable from my own testing, I would suggest that the performance issues should no longer be a blocker to putting this on nightly.

fenix/app/src/main/java/org/mozilla/fenix/compose/TabThumbnail.kt
```diff
@@ -0,0 +1,95 @@
...
+            contentAlignment = Alignment.Center,
+        ) {
+            if (tab.content.icon != null) {
+                Image(
+                    bitmap = tab.content.icon!!.asImageBitmap(),
```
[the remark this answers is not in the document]
MozillaNoah: a let does not, so I'll just make a local variable.

fenix/app/src/main/java/org/mozilla/fenix/tabstray/inactivetabs/InactiveTabs.kt
MozillaNoah: yeah, it's a good question of whether or not they are living in two places simultaneously. from a cursory search, it looks like the tab icons are being set via an "Icons" extension callback and then browser icons maintains a cache but because of JVM magic, they could be the same thing in memory with two references? Not entirely sure, but this would require some more thorough investigation,.

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
MozillaNoah: Unfortunately, components can't be accessed from within side effect lambdas, so we have to declare it and keep it outside in the root Composable block.

MozillaNoah: Kinda more just makes us need to, long term, figure out a potentially something better than singleton-accessed storage. 

Maybe this will be something else, like the Stores, that we can do the [CompositionLocal stuff](https://developer.android.com/jetpack/compose/compositionlocal) with.

fenix/app/src/main/java/org/mozilla/fenix/compose/tabstray/TabGridItem.kt
MozillaNoah: That's correct, and this is how the [XML implementation does it](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/browser/AbstractBrowserTabViewHolder.kt#122).

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
MatthewTighe: Ah gotcha, that's interesting. Thanks!

fenix/app/src/main/java/org/mozilla/fenix/compose/tabstray/TabGridItem.kt
MatthewTighe: Oh okay cool. Thanks

MatthewTighe (approved):
Based on our conversations I'm satisfied with all this. Good work on these investigations!

MozillaNoah:
I may have found another performance win. [According to this article](https://developer.android.com/jetpack/compose/graphics/images/optimization#preparetodraw), we need to call `Bitmap`'s `prepareToDraw` function in order to signal to the GPU ahead of time that some images need to be rendered. A call like this would normally be handled by image loading libraries, but since we're working with our own homebrew situation here, we need to call it ourselves.

The most recent push contains these changes for `ThumbnailImage`, `TabGridItem`'s corner favicon, and the inactive tabs' use of `FaviconListitem`

MatthewTighe (approved):
Looks like the linter is complaining, but this otherwise LGTM

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
```diff
@@ -0,0 +1,127 @@
...
+                )
+            }
+        }
+
+        if (state.bitmap == null && state.hasLoaded) {
+            fallbackContent()
+        } else {
+            state.bitmap?.let { bitmap ->
+                Image(
+                    painter = BitmapPainter(bitmap.asImageBitmap()),
```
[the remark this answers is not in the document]
MozillaNoah: I am above in the coroutine. I agree though that we will probably need an abstraction for this, as long as we maintain our homebrewed thumbnail loader.

PR #2314, opened 2023-06-02
Bug 1836022 - catch NS_ERROR_HOST_IS_IP_ADDRESS when querying cookie banners exception

jackyzy823:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1836022


android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/cookiebanners/GeckoCookieBannersStorage.kt
```diff
@@ -105,8 +105,9 @@ class GeckoCookieBannersStorage(
                 )
             }
         } catch (e: Exception) {
-            if ((e.message ?: "").contains("NS_ERROR_INSUFFICIENT_DOMAIN_LEVELS")) {
-                // This normally happen on internal sites like about:config
+            // This normally happen on internal sites like about:config or ip sites.
+            val geckoError = listOf("NS_ERROR_INSUFFICIENT_DOMAIN_LEVELS", "NS_ERROR_HOST_IS_IP_ADDRESS")
+            if (geckoError.any { (e.message ?: "").contains(it) }) {
```
MatthewTighe, line 110: nit:
```suggestion
            val disabledErrors = listOf("NS_ERROR_INSUFFICIENT_DOMAIN_LEVELS", "NS_ERROR_HOST_IS_IP_ADDRESS")
            if (disabledErrors.any { (e.message ?: "").contains(it) }) {
```

MatthewTighe (commented):
Besides my nit, this LGTM and solves the crash. I'll cycle back for approval once the name is updated 🙂

MatthewTighe:
I filed [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1837082) to see if we can get product input on further improvements here, but I think this patch will be worth landing as is in order to fix the crasher.

jackyzy823:
@MatthewTighe  The `always off even manually enable` problem could be solved here 

change
https://github.com/mozilla-mobile/firefox-android/blob/dcc1820500e117c7fea94a0d7868a62235ea71d2/fenix/app/src/main/java/org/mozilla/fenix/settings/quicksettings/protections/cookiebanners/CookieBannersStorageExt.kt#L40-L43
to 
```kotlin
 if (hasException) { 
     CookieBannerUIMode.HIDE
 }
```
which I think is reasonable. If you feel ok , i will add it to commit.

MatthewTighe (approved):

MatthewTighe:
@jackyzy823 This does seem like a reasonable approach, but I would prefer to let some UX or product people weigh in before we move forward with a solution. I'll add your suggestion to bug report I filed above, though, thanks!

Let's land this patch as is to address the crash, and then we can see about moving forward. Thanks again for another contribution!

MatthewTighe:
bors try

PR #2483, opened 2023-06-15
Bug 1838662 - cleanup push-apk.py in firefox-android

AaronMT:
https://github.com/mozilla-mobile/firefox-android/blob/348dc4badab80135f016076eb7c071d478afc96d/taskcluster/android_taskgraph/transforms/push_apk.py#L45 is not needed as it is replaced in https://github.com/mozilla-mobile/firefox-android/blob/348dc4badab80135f016076eb7c071d478afc96d/taskcluster/android_taskgraph/transforms/push_apk.py#L48 as the `startup-test` config has a Fenix and Focus task in it now and those tasks are new/renamed.
https://bugzilla.mozilla.org/show_bug.cgi?id=1838662


jcristau (approved):
Thank you!

PR #2487, opened 2023-06-15
Bug 1837864 – fix search settings opening visual glitch

mavduevskiy:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1837864

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineFragment.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineFragment.kt
@@ -32,6 +32,12 @@ class SearchEngineFragment : PreferenceFragmentCompat() {
             if (unifiedSearchUI) R.xml.search_settings_preferences else R.xml.search_preferences,
             rootKey,
         )
+
+        // Visibility should be set before the view has been created, to avoid visual glitches.
+        requirePreference<SwitchPreference>(R.string.pref_key_show_search_engine_shortcuts).apply {
+            isVisible = !context.settings().showUnifiedSearchFeature
+        }
+
         view?.hideKeyboard()
     }
 
@@ -65,7 +71,6 @@ class SearchEngineFragment : PreferenceFragmentCompat() {
         val showSearchShortcuts =
             requirePreference<SwitchPreference>(R.string.pref_key_show_search_engine_shortcuts).apply {
                 isChecked = context.settings().shouldShowSearchShortcuts
-                isVisible = !context.settings().showUnifiedSearchFeature
             }
 
         val showHistorySuggestions =


fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineFragment.kt
```diff
@@ -32,6 +32,12 @@ class SearchEngineFragment : PreferenceFragmentCompat() {
             if (unifiedSearchUI) R.xml.search_settings_preferences else R.xml.search_preferences,
             rootKey,
         )
+
+        // Visibility should be set before the view has been created, to avoid visual glitches.
+        requirePreference<SwitchPreference>(R.string.pref_key_show_search_engine_shortcuts).apply {
+            isVisible = !context.settings().showUnifiedSearchFeature
+        }
```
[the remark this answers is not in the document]
mavduevskiy: We want to keep the option of turning it off in case it affects search revenue. So unless we are ready to remove all non unified search code, we should support both behaviours. The estimate for the 
"probation" period would be around two release, from what I have heard.

If we disable it in XML, then we will get the glitch again if `showUnifiedSearchFeature == false`. 
I would set the visibility before the view is created, and then remove the `SwitchPreference` completely once `UnifiedSearch` is verified not to be a regression.

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineFragment.kt
```diff
@@ -32,6 +32,12 @@ class SearchEngineFragment : PreferenceFragmentCompat() {
             if (unifiedSearchUI) R.xml.search_settings_preferences else R.xml.search_preferences,
             rootKey,
         )
+
+        // Visibility should be set before the view has been created, to avoid visual glitches.
```
gabrielluong, line 36:
```suggestion
        // Visibility should be set before the view has been created to avoid visual glitches.
```

gabrielluong (approved):

fenix/app/src/main/java/org/mozilla/fenix/settings/search/SearchEngineFragment.kt
mavduevskiy: Could you explain the reasoning behind the suggestion? Grammatically it is fine to add a comma in there, and it emphasizes the reason more.

gabrielluong: I don't really want to have to do this all the time, but I would suggest throwing it up to a checker like https://quillbot.com/grammar-check. 

Specifically, you don't need a comma before the preposition here. https://prowritingaid.com/comma-after-prepositional-phrase

mavduevskiy: Is it a mistake though? Comma usage can be very flexible. In this case, it is an infinitive phrase, which can be separated with commas depending on the meaning the author wants to emphasize. I know I am not a native English speaker, but let's not assume a mistake on my part if rules are ambiguous.

I want to keep the comment, as those lines look a bit out of place (since a lot of `SwitchPreference` are accessed in `onResume`) and the comment explains not what the code does, but why it has to be in the place it is.

mavduevskiy:
@Mergifyio backport releases_v115

DonalMe:
@mergify backport releases_v115

PR #2510, opened 2023-06-16
Bug 1838903 - Update ktlint to version 0.49.1

rvandermeulen:
https://bugzilla.mozilla.org/show_bug.cgi?id=1838903
https://bugzilla.mozilla.org/show_bug.cgi?id=1838903
https://bugzilla.mozilla.org/show_bug.cgi?id=1838903

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/TrackingProtectionPolicyFactory.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/TrackingProtectionPolicyFactory.kt
@@ -106,9 +106,9 @@ class TrackingProtectionPolicyFactory(
     }
 }
 
+@Suppress("MaxLineLength")
 @VisibleForTesting
-internal fun TrackingProtectionPolicyForSessionTypes.applyTCPIfNeeded(settings: Settings):
-    TrackingProtectionPolicyForSessionTypes {
+internal fun TrackingProtectionPolicyForSessionTypes.applyTCPIfNeeded(settings: Settings): TrackingProtectionPolicyForSessionTypes {
     val updatedCookiePolicy = if (settings.enabledTotalCookieProtection) {
         CookiePolicy.ACCEPT_FIRST_PARTY_AND_ISOLATE_OTHERS
     } else {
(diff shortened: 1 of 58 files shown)


fenix/app/src/main/java/org/mozilla/fenix/components/TrackingProtectionPolicyFactory.kt
```diff
@@ -106,9 +106,9 @@ class TrackingProtectionPolicyFactory(
     }
 }
 
+@Suppress("MaxLineLength")
 @VisibleForTesting
-internal fun TrackingProtectionPolicyForSessionTypes.applyTCPIfNeeded(settings: Settings):
-    TrackingProtectionPolicyForSessionTypes {
+internal fun TrackingProtectionPolicyForSessionTypes.applyTCPIfNeeded(settings: Settings): TrackingProtectionPolicyForSessionTypes {
```
gabrielluong, line 111: Could we undo this change to not hit the `MaxLineLength`?

gabrielluong (approved):

gabrielluong:
This will need to be rebased when https://github.com/mozilla-mobile/firefox-android/pull/2462 lands

jonalmeida (approved):

fenix/app/src/main/java/org/mozilla/fenix/components/TrackingProtectionPolicyFactory.kt
rvandermeulen: @gabrielluong The alternative is suppressing a `Single space expected between colon and return type (standard:function-return-type-spacing)` error as I don't see a good way of appeasing that rule without exceeding the maximum line length. IMO, I found the line length suppression to be the reasonable trade-off for better readability and consistency throughout the file. Let me know, though!

gabrielluong: Sounds reasonable. I guess this is more of an unfortunate issue with the how we have named things here.

PR #2537, opened 2023-06-19
Bug 1823367 - Show snackbar instead of toast when save to PDF error appears.

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


ohall-m:
Thank you for working on this bug! Could you update the bug ID to 1813428?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -1102,6 +1102,13 @@ sealed class EngineAction : BrowserAction() {
         val throwable: Throwable,
     ) : EngineAction(), ActionWithTab
 
+    /**
+     * State of save to pdf error snackBar has changed
+     */
+    data class ShowSaveToPDFErrorSnackBarChange(
```
ohall-m, line 1108:
```suggestion
    data class ShowStandardSnackbarErrorAction(
```
I think this could be useful to have around as a generic component. What do you think? 

Could you add a parameter of `val errorMsg: String` to the signature and rename this and the associated functions to something like `Standard` error?

(I plan on reusing this UI element for print and I expect it could be useful for other unrecoverable errors too. It would be great to call something like: `ShowStandardSnackbarErrorAction(true, R.string.unable_to_print_error)`)

fenix/app/src/main/res/values/strings.xml
```diff
@@ -1159,8 +1159,10 @@
     <string name="share_button_content_description">Share</string>
     <!-- Text for the Save to PDF feature in the share menu -->
     <string name="share_save_to_pdf">Save as PDF</string>
-    <!-- Text for error message when generating a PDF file Text for error message when generating a PDF file. -->
+    <!-- Text for snackbar error message when generating a PDF file. -->
     <string name="unable_to_save_to_pdf_error">Unable to generate PDF</string>
+    <!-- Text for error snackbar dismiss button when generating a PDF file. -->
+    <string name="unable_to_save_to_pdf_error_dismiss_button">Dismiss</string>
```
ohall-m, line 1165:
```suggestion
    <string name="standard_snackbar_error_dismiss">Dismiss</string>
```
Maybe something like this phrasing, if we move to `ShowStandardSnackbarErrorAction`? I think the dismiss string would be universal and we just would need to pass an error message.

ohall-m (commented):
Thanks for working on this and adding these snackbar error states!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
Amejia481: Yeah, completely agree, we will need to create an state that is generic enough that other similar use cases could use, as this is something that maybe app specific in this case Fenix we could put it under [AppState](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt), also tagging @jonalmeida in case he could have other input :)

jonalmeida: This sounds like a good idea!

> Could you add a parameter of val errorMsg: String to the signature and rename this and the associated functions to something like Standard error?

Consider how this message coming from the component is going to be localized. I'm not sure yet, what's the best direction for this if it was the `errorMessage` that we display.

> this is something that maybe app specific in this case Fenix we could put it under [AppState](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt),

This might not work. If we are dispatching an action from the BrowserStore at the engine-level we would have have to chain it to the AppStore before we can get it to the UI. This _might_ make it better for the point I mentioned above, but again I'm not sure, because now we are daisy-chaining two stores together.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -339,13 +349,61 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
 
         subscribeToTabCollections()
         updateLastBrowseActivity()
+        observeSaveToPDFErrorSnackBar()
+    }
+
+    @VisibleForTesting
+    internal fun observeSaveToPDFErrorSnackBar() {
```
jonalmeida, line 356: Before landing, please move code like this into their own classes so they are isolated and testable without growing large classes like the BrowserFragment one. There are examples like the "feature" component or the `AbstractBinding`-based classes.

If this is just for testing since the change is just a draft, please ignore this comment. 🙂

jonalmeida (commented):
Nice idea, Gabriel! This looks like a solid direction.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
iorgamgabriel: Done

fenix/app/src/main/res/values/strings.xml
iorgamgabriel: Done

fenix/app/src/test/java/org/mozilla/fenix/share/SaveToPDFMiddlewareTest.kt
```diff
@@ -39,190 +48,233 @@ class SaveToPDFMiddlewareTest {
...
     @Test
-    fun `GIVEN a save to pdf request WHEN it fails unexpectedly THEN unknown failure telemetry is sent`() = runTestOnMain {
-        val exceptionToThrow = RuntimeException("reader save to pdf failed")
-        val middleware = SaveToPDFMiddleware(testContext)
-        val mockEngineSession: EngineSession = mockk<EngineSession>().apply {
-            every {
-                checkForPdfViewer(any(), any())
-            } answers {
-                secondArg<(Throwable) -> Unit>().invoke(exceptionToThrow)
+    fun `GIVEN a save to pdf request WHEN it fails unexpectedly THEN unknown failure telemetry is sent`() =
```
ohall-m, line 58: nit: 
```suggestion
    fun `GIVEN a save to pdf request WHEN it fails unexpectedly THEN unknown failure telemetry is sent AND a snackbar error is shown`() =
```
(A few other tests could be renamed too.)

ohall-m (commented):
Thanks for all the work on this component and making it reusable! I had just a few minor renaming suggestions.

I tested by causing save to PDF to always fail in GV. The component worked as expected in light/night mode and also localized the error messages.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
ohall-m: Thanks for the update! I tested by changing system languages and the message correctly localized.

fenix/app/src/main/res/values/strings.xml
ohall-m: Thanks!

ohall-m:
Testing screenshots:

localize-fr
night
light

fenix/app/src/main/java/org/mozilla/fenix/browser/StandardSnackbarErrorBinding.kt
```diff
@@ -0,0 +1,97 @@
...
+                        )
+                        snackBar.setAction(
+                            text = activity.getString(R.string.standard_snackbar_error_dismiss),
+                            action = {
+                                view.context.components.appStore.dispatch(
+                                    AppAction.ShowStandardSnackbarErrorAction(
+                                        StandardSnackbarError(
+                                            false,
+                                            activity.getString(
+                                                R.string.unable_to_save_to_pdf_error,
```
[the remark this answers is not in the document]
iorgamgabriel: The snackbar text should be "Unable to generate PDF" in this case.

fenix/app/src/main/java/org/mozilla/fenix/browser/StandardSnackbarErrorBinding.kt
```diff
@@ -0,0 +1,97 @@
...
+                if (standardSnackbarError?.standardSnackbarErrorVisibility == true) {
+                    activity.getRootView()?.let { view ->
+                        val snackBar = FenixSnackbar.make(
+                            view = view,
+                            duration = Snackbar.LENGTH_INDEFINITE,
+                            isDisplayedWithBrowserToolbar = true,
+                        )
+                        snackBar.setText(
+                            standardSnackbarError.standardSnackbarErrorMessage
+                                ?: "",
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/browser/StandardSnackbarErrorBinding.kt
```diff
@@ -0,0 +1,97 @@
...
+                        snackBar.setButtonTextColor(
+                            ContextCompat.getColor(
+                                activity,
+                                R.color.fx_mobile_text_color_primary,
+                            ),
+                        )
+                        snackBar.setBackground(
+                            AppCompatResources.getDrawable(
+                                activity,
+                                R.drawable.save_to_pdf_error_snackbar_background,
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/test/java/org/mozilla/fenix/share/SaveToPDFMiddlewareTest.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/browser/StandardSnackbarErrorBinding.kt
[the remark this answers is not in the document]
ohall-m: Thanks for double checking!

ohall-m (approved):
Thanks, LGTM!

@jonalmeida or @Amejia481, I think the App State part is working as expected based on testing, but maybe one of you would like to take a look too?

fenix/app/src/main/java/org/mozilla/fenix/browser/StandardSnackbarErrorBinding.kt
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Standard Snackbar Error data class.
+ *
+ * @param standardSnackbarErrorMessage that will appear on the snackbar.
+ * @param standardSnackbarErrorVisibility of the snackbar.
+ */
+data class StandardSnackbarError(
+    val standardSnackbarErrorVisibility: Boolean = false,
+    val standardSnackbarErrorMessage: String? = null,
```
Amejia481, line 96: nit: as these are properties of `StandardSnackbarError` we could just remove the prefix `standardSnackbarError.*`, for example `standardSnackbarErrorVisibility` could be just visibility and `standardSnackbarErrorMessage` could be `message`.

Amejia481, line 96: I think `messsage` should be non-null, as not sure an use case where we could have a `SnackbarError` without a message, checking closely I'm not sure if we need `standardSnackbarErrorVisibility` could you help me to understand why do we need to include a `visibility` field ?

Amejia481: I was imagining that we would show the `Snackbar` and after sometime it will go away, as topical Snackbars

iorgamgabriel: The snackbar will stay on the screen until the user presses the dismiss button. I show and hide the snack based on visibility flag.

Amejia481: Thanks for the clarification, that's a bit odd behaviour for a `Snackbar`, is this a product/UI/UX suggestion?

iorgamgabriel: Done

Amejia481: Perfect, thanks for the clarification!

fenix/app/src/main/java/org/mozilla/fenix/browser/StandardSnackbarErrorBinding.kt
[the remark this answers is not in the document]
Amejia481: If the idea is to remove the snackbar should we dispatch  `dispatch(AppAction.ShowStandardSnackbarErrorAction(null))`? As just want to clear up the state.  And instead of calling  `ShowStandardSnackbarErrorAction` we could have `UpdateStandardSnackbarErrorAction`?

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -208,4 +209,11 @@ sealed class AppAction : Action {
          */
         object PauseAction : AppLifecycleAction()
     }
+
+    /**
+     * State of standard error snackBar has changed.
+     */
+    data class ShowStandardSnackbarErrorAction(
```
Amejia481, line 216: Maybe `UpdateStandardSnackbarErrorAction`? As the description mentions we are only update the value of the standard error snackBar.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -208,4 +209,11 @@ sealed class AppAction : Action {
...
+        val standardSnackbarError: StandardSnackbarError,
```
Amejia481, line 217: Maybe this could accept `null`.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
iorgamgabriel: Done

Amejia481 (commented):
Thanks for all the changes @iorgamgabriel !
In general this looks good, the last part that I think we need to take into consideration, it's that we are not dismissing dialog if the apps gets kill, or switched away, if that happens and the user re-enter that app we will see the error message.

iorgamgabriel:
Done , test added.

Amejia481 (approved):
Great work!
LGTM!

PR #2566, opened 2023-06-21
Bug 1810630 - add 'manage account' to the sync menu.

mhammond:
This seems to work well - for reasons I don't fully understand, the `FxaWebChannelFeature` automagically kicks in when opening the view, so we get the correct exchange of the FxA account information between the server and the browser "for free".

I put the menu item at the top of the menu and made the English version of the string "Manage account" to match desktop.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1810630

### DIFF
--- a/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
+++ b/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
@@ -157,6 +157,14 @@ interface OAuthAccount : AutoCloseable {
     suspend fun getTokenServerEndpointURL(): String?
 
     /**
+     * Fetches the URL for the user to manage their account
+     *
+     * @param entryPoint A string which will be included as a query param in the URL for metrics.
+     * @return The URL which should be opened in a browser tab.
+     */
+    suspend fun getManageAccountURL(entryPoint: FxAEntryPoint): String
+
+    /**
      * Get the pairing URL to navigate to on the Authority side (typically a computer).
      *
      * @return The URL to show the pairing user
--- a/fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
@@ -41,12 +41,14 @@ import org.mozilla.fenix.GleanMetrics.SyncAccount
 import org.mozilla.fenix.R
 import org.mozilla.fenix.components.FenixSnackbar
 import org.mozilla.fenix.components.StoreProvider
+import org.mozilla.fenix.components.accounts.FenixFxAEntryPoint
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.getPreferenceKey
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.secure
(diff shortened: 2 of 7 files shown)


android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
```diff
@@ -156,6 +156,13 @@ interface OAuthAccount : AutoCloseable {
      */
     suspend fun getTokenServerEndpointURL(): String?
 
+    /**
+     * Fetches the URL for the user to manage their account
+     *
+     * @return The URL which should be opened in a browser tab.
+     */
+    suspend fun getManageAccountURL(entryPoint: String): String
```
jonalmeida, line 164: Could we add an `@param` KDoc here?

fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -362,6 +367,21 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
         return true
     }
 
+    private fun getClickListenerForManageAccount(): Preference.OnPreferenceClickListener {
+        return Preference.OnPreferenceClickListener {
+            viewLifecycleOwner.lifecycleScope.launch(Main) {
+                context?.let {
+                    var url = accountManager.authenticatedAccount()?.getManageAccountURL("menu")
```
jonalmeida, line 374:
```suggestion
                    var url = accountManager.authenticatedAccount()?.getManageAccountURL("menu")
```

"menu" is a bit generic, maybe you want to add something more specific like "account-menu" or reuse one of  the other values in `FenixFxAEntryPoint`?

android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
jonalmeida, line 164:
```suggestion
    suspend fun getManageAccountURL(entryPoint: FxAEntryPoint): String
```

Is this `entryPoint` similar to the other `FenixFxAEntryPoint` implementation values? We perhaps want our API to use that instead. This should give a simple compile-time check for other users of this API to apply an explicit and known entry point value when using the account manager in different parts of the application.

jonalmeida (commented):
This looks nice! Left some minor comments about the entryPoint, but this looks good to land after those are addressed.

> This seems to work well - for reasons I don't fully understand, the FxaWebChannelFeature automagically kicks in when opening the view, so we get the correct exchange of the FxA account information between the server and the browser "for free".

The `FxaWebChannelFeature` is added to the [`BaseBrowserFragment`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#889-900) that all our different browser views extend from, so a Custom Tab would get this feature enabled in it for free. From there, the extension is attached and runs like a regular webext would on [the domains that match](https://github.com/mozilla-mobile/firefox-android/blob/05064e101794abdc9fb5a717169a1bd02e7aacc6/android-components/components/feature/accounts/src/main/assets/extensions/fxawebchannel/manifest.template.json#L10-L22) within the manifest.

jonalmeida (approved):
lgtm!

PR #2344, opened 2023-06-06
Bug 1835665 - Unify FocusDependencies with AC's DependenciesPlugin

gabrielluong:
### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1835665

PR #2574, opened 2023-06-22
Bug 1837107 - Unify FenixDependencies with AC's DependenciesPlugin

gabrielluong:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1837107

### DIFF
--- a/fenix/app/build.gradle
+++ b/fenix/app/build.gradle
@@ -248,7 +248,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion = FenixVersions.androidx_compose_compiler
+        kotlinCompilerExtensionVersion = Versions.compose_compiler
     }
 
     namespace 'org.mozilla.fenix'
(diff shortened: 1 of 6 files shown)


mcarare (approved):
LGTM!

mcarare:
As a follow-up, I was wondering if it makes sense to keep any library versioning in Fenix / Focus, even if they are not used in AC. Maybe we could move at least the ones that are common in Fenix and Focus. ( test orchestrator for example).

gabrielluong:
> As a follow-up, I was wondering if it makes sense to keep any library versioning in Fenix / Focus, even if they are not used in AC. Maybe we could move at least the ones that are common in Fenix and Focus. ( test orchestrator for example).

Based on my prior conversation with Christian and Jon in https://github.com/mozilla-mobile/firefox-android/pull/2344#issuecomment-1581409116, we want to treat AC as a separate project so it should only know about the dependencies that are used for AC. The alternative would be to create a new Android Dependencies Plugin that serves as a root. Currently, I am treating the Components Dependencies Plugin as that root, but it might make sense to reconsider that.

PR #2578, opened 2023-06-22
Bug 1838140 - Update composed tabs tray media button ripple effect to match the xml version

Alexandru2909:
Confirmed with Verdi we should keep the whole tab ripple effect (brightening for light theme) and remove any media button ripple effect.

| XML version | Compose version |
| --- | --- |
| <video src="https://github.com/mozilla-mobile/firefox-android/assets/35462038/5a51d6c1-d1d5-4b35-a57b-5ac6c8bfdb97" /> | <video src="https://github.com/mozilla-mobile/firefox-android/assets/35462038/740a7e86-d889-4364-964b-c0331628de57" /> |

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1838140


fenix/app/src/main/java/org/mozilla/fenix/theme/FirefoxTheme.kt
```diff
@@ -151,6 +151,7 @@ private val darkColorPalette = FirefoxColors(
     borderAccent = PhotonColors.Violet40,
     borderDisabled = PhotonColors.LightGrey05A40,
     borderWarning = PhotonColors.Red40,
+    rippleEffect = PhotonColors.Black,
```
gabrielluong, line 154: Piggybacking on @MozillaNoah's comment, this doesn't seem like a color variable that is provided by the design system so I wouldn't want to see this here.

gabrielluong: I believe the correct answer here is to fallback to the MaterialTheme.

fenix/app/src/main/java/org/mozilla/fenix/compose/tabstray/TabGridItem.kt
```diff
@@ -131,6 +139,13 @@ fun TabGridItem(
                     .then(tabBorderModifier)
                     .padding(4.dp)
                     .combinedClickable(
+                        interactionSource = interactionSource,
+                        indication = rememberRipple(
+                            color = when (isSystemInDarkTheme()) {
+                                true -> PhotonColors.White
+                                false -> PhotonColors.Black
```
MozillaNoah, line 146: I think the underlying question is do we even need to override the color here? does it provide a default value?

Alexandru2909: The default value provided by Android would be [black](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/Indication.kt;l=159) for both light and dark themes. From my discussion with Verdi it was clear we want the tab to get brighter for light theme at least.

cc @Verdi

MozillaNoah: gotcha gotcha. I think this is because we don't provide our own `RippleTheme`. I'd say for now, let's pass in a color directly and file a follow-up to write [our own custom override](https://stackoverflow.com/questions/67265345/modify-ripple-color-of-iconbutton-in-jetpack-compose) for our `FirefoxTheme`.

Alexandru2909: Opened [1840020](https://bugzilla.mozilla.org/show_bug.cgi?id=1840020) in this scope.

fenix/app/src/main/java/org/mozilla/fenix/theme/FirefoxTheme.kt
Alexandru2909: The `RippleTheme` provided by `MaterialTheme` also seems to only darken the view, similar to what the default Compose ripple effect does.

fenix/app/src/main/java/org/mozilla/fenix/compose/tabstray/TabGridItem.kt
gabrielluong: This sounds like the right idea, thanks!

MozillaNoah (approved):

PR #2594, opened 2023-06-23
Bug 1837526 - Call StartupTimeline in the Compose Top Sites

DreVla:
Because we are migrating Top Sites to Compose, we want to ensure we are still recording when top sites are loaded, therefore the call for StartupTimeline is now called after creating the top sites compose elements. Calling it here will ensure it will be called even after we replace the home page recycler with a Compose view.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1837526

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/home/topsites/TopSites.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/home/topsites/TopSites.kt
@@ -104,6 +104,7 @@ fun TopSites(
     onRemoveTopSiteClicked: (topSite: TopSite) -> Unit,
     onSettingsClicked: () -> Unit,
     onSponsorPrivacyClicked: () -> Unit,
+    onTopSitesItemBound: () -> Unit,
 ) {
     val pageCount = ceil((topSites.size.toDouble() / TOP_SITES_PER_PAGE)).toInt()
 
@@ -154,6 +155,7 @@ fun TopSites(
                                     topSiteColors = topSiteColors,
                                     onTopSiteClick = { item -> onTopSiteClick(item) },
                                     onTopSiteLongClick = onTopSiteLongClick,
+                                    onTopSitesItemBound = onTopSitesItemBound,
                                 )
                             }
                         }
@@ -251,6 +253,7 @@ private fun TopSiteItem(
     topSiteColors: TopSiteColors,
     onTopSiteClick: (TopSite) -> Unit,
     onTopSiteLongClick: (TopSite) -> Unit,
+    onTopSitesItemBound: () -> Unit,
 ) {
     var menuExpanded by remember { mutableStateOf(false) }
 
@@ -338,6 +341,10 @@ private fun TopSiteItem(
                 submitTopSitesImpressionPing(topSite = topSite, position = position)
             }
         }
+    }
+
+    LaunchedEffect(Unit) {
+        onTopSitesItemBound()
     }
 }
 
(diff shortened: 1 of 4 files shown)


fenix/app/src/main/java/org/mozilla/fenix/home/topsites/TopSites.kt
```diff
@@ -312,6 +314,14 @@ private fun TopSiteItem(
             }
         }
     }
+
+    val context = LocalContext.current
+    val activity = context as? HomeActivity
+    LaunchedEffect(Unit) {
+        activity?.let {
+            StartupTimeline.onTopSitesItemBound(it)
+        }
```
MozillaNoah, line 323: drive by: this will very likely break previews, and obtaining the activity reference in Compose code isn't ideal. It might make more sense to pass in a lambda to handle this instead.

gabrielluong: Agreed, @DreVla let's move this into lambda that is passed into our @Composables from the fragment. We can call it `onTopSitesItemBound`

fenix/app/src/main/java/org/mozilla/fenix/home/topsites/TopSites.kt
```diff
@@ -100,6 +101,7 @@ fun TopSites(
     onRemoveTopSiteClicked: (topSite: TopSite) -> Unit,
     onSettingsClicked: () -> Unit,
     onSponsorPrivacyClicked: () -> Unit,
+    onTopSitesItemBound: (HomeActivity) -> Unit,
```
MozillaNoah, line 104: I think what @gabrielluong and I were trying to push you towards was not having _any_ reference to or direct usage of an activity inside a Compose file. Let's instead drop `HomeActivity` from this file, making the Compose agnostic to the underlying activity/fragment architecture, and have it call a dumb-as-possible lambda that then leads to `TopSitesViewHolder` where `StartupTimeline.onTopSitesItemBound` is called inside an inline lambda and the activity reference is obtained there.

```suggestion
    onTopSitesItemBound: () -> Unit,
```

```
    onTopSitesItemBound = {
        StartupTimeline.onTopSitesItemBound(activity = composeView.context as HomeActivity)
    }
```

gabrielluong (dismissed):
Setting this as "needs change" so that this is off the "Needs Review" filter for now

fenix/app/src/main/java/org/mozilla/fenix/home/topsites/TopSites.kt
```diff
@@ -315,6 +319,14 @@ private fun TopSiteItem(
             }
         }
     }
+
+    val context = LocalContext.current
+    val activity = context as? HomeActivity
+    LaunchedEffect(Unit) {
```
gabrielluong, line 325: I don't know too much about side effects in Compose, but I am wondering if using `Unit` here might be too relaxed? Would `onTopSitesItemBound` be called more than necessary if there's a recomposition after the home screen is fully drawn? This might not be an actual problem, but I just want to make sure we understand this well.

I am trying to also determine if calling `reportFullyDrawn` multiple times has any consequences. I want to assume it will just log the first time it is called and the subsequent calls will do nothing, but want to double check my understanding.

MozillaNoah: `LaunchedEffect(Unit)` will only be called at first composition

gabrielluong: > `LaunchedEffect(Unit)` will only be called at first composition

Perfect!

PR #2625, opened 2023-06-26
Bug 1840483 - 'manage account' is only available on Nightly/Debug builds

mhammond:
We need the UX of this to improve before we should unleash this as a new feature on our users. Flagging @jonalmeida as he approved the bug which first landed the new option.
https://bugzilla.mozilla.org/show_bug.cgi?id=1840483
https://bugzilla.mozilla.org/show_bug.cgi?id=1840483


jonalmeida (approved):
Cheers, thanks for putting up this patch!

jonalmeida:
I'm going to add the landing label since it's simple enough that we should have sooner than later.

PR #2632, opened 2023-06-27
Bug 1840578 - Add shopping experience entry point

rahulsainani:
### What
– Adds the shopping cart icons and entry point on the url bar if the feature flag is enabled. 
– The check for product page url will be done as part of [another bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1840580) when the API for that lands in toolkit.
– Hides reader view on shopping experience enabled urls

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1840578


fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
```diff
@@ -0,0 +1,28 @@
...
+ * Feature implementation that provides review quality check information for supported product
+ * pages.
+ */
+class ReviewQualityCheckFeature(
+    private val onAvailabilityChange: (isAvailable: Boolean) -> Unit,
+) : LifecycleAwareFeature {
+
+    override fun start() {
+        val isFeatureEnabled = FxNimbus.features.shoppingExperience.value().enabled
+        // Update to use product page detector api in Bug 1840580
```
calumozilla, line 20: @jonalmeida we talked about removing the extra step of communication of: first fetching "is this a product page", if so also fetch the analysis. what would happen here instead?

(Rahul, Jon might be out for PTO the next couple weeks, not a blocker)

calumozilla (approved):

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
jonalmeida: Here, is where we would use the `EngineSession.registerDelegate` API and when the page load is complete, we (GV) internally do the "is supported check" and chain that up to fire this delegate when we get a response. That would feed into the callback on the line below.

The WebPushIntegration works the same way (on the engine/runtime level).

calumozilla: gotcha, thanks for clarifying!

rahulsainani: The idea here is to just perform the `isSupportedProductPage` check, which is a local regex check AFAIU, and show the shopping experience icon or not. This would return quite fast. Jared suggested something like this [here](https://phabricator.services.mozilla.com/D180978#6014444).

If you're discussing fetching the product details here as well, I'd like to make a case that we should fetch the product details and recommendations when the user actually opens the bottom sheet, or at least delay it, so we don't fetch super eagerly, as we don't know how many people would eventually open the bottom sheet per "unique supported product page load".

I'm curious what are the advantages of removing the extra step of communication of `isSupportedProductPage`?

jonalmeida: > I'm curious what are the advantages of removing the extra step of communication of `isSupportedProductPage`?

@rahulsainani I don't see any response here, so just to close the loop: we assumed that we wanted to request product information regardless of the embedder checking. Having seen your demo in Slack recently, it looks like having the separate `isSupportedProductPage` makes sense for an indicator. A product check after that then makes sense. 🙂

PR #2640, opened 2023-06-28
Bug 1837446 - Fix tabs tray pager rotation bug

MozillaNoah:
I spent well over a week on this, and unfortunately, all I could find that worked was a bit of a hack. I legitimately don't remember this being a bug a couple a months ago but 🤷‍♀️ 

`HorizontalPager` functions correctly when rotated in a device-deployed preview via `TabsTray` AND when the configuration flags are removed from the manifest, so the best hypothesis I have for a true root cause is that something in the bones of the app is causing this bug. 

Aside from the historical instability and volatility of managing web view components on Android, I'd be curious to learn about Fenix's history with the manifest flags.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1837446

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
@@ -21,16 +21,19 @@ import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.toMutableStateList
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.input.nestedscroll.nestedScroll
+import androidx.compose.ui.platform.LocalConfiguration
 import androidx.compose.ui.platform.rememberNestedScrollInteropConnection
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
+import kotlinx.coroutines.delay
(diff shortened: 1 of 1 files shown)


fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -49,6 +52,8 @@ import org.mozilla.fenix.tabstray.syncedtabs.SyncedTabsListItem
 import org.mozilla.fenix.theme.FirefoxTheme
 import mozilla.components.browser.storage.sync.Tab as SyncTab
 
+private const val PAGER_DELAY = 100L
```
MatthewTighe, line 55: In my testing, I'm not seeing that this delay is required to fix the bug. Is that different than what you've been seeing? What motivated it originally? IMO if we can get this in without the delay it will feel a lot less hacky.

MozillaNoah: WAT

So, I tried a NUMBER of different things, and in the beginning of debugging potential solutions, I tried running a `LaunchedEffect` with the forced re-scroll to the correct page, but that didn't seem to do anything. Fast forward to last night, it worked. Out of nowhere. I _thought_ I had tested without the delay, but I'll double check and report back.

MozillaNoah: OK, so taking away the delay made it broken for _every_ rotation to landscape 🤦‍♀️ 

As a footnote, I've never reproduced this in portrait (either when rotating back to portrait, or opening the tabs tray in landscape and then rotating).

MozillaNoah: hrmm, a delay of 50 instead of 100 seems to also work

MozillaNoah:
I'm going to close this PR for now.

See this [BZ comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1837446#c3) for more details.

PR #2743, opened 2023-07-06
Bug 1839984 - Update architecture example to use lib-state

MatthewTighe:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1839984


fenix/docs/architectureexample/HistoryStore.kt
```diff
@@ -0,0 +1,52 @@
...
+    val selectedItems: List<History>,
+    val itemsBeingDeleted: List<History>,
+    companion object {
+        val initial = HistoryState(
+            items = listOf(),
+            selectedItems = listOf(),
+            itemsBeingDeleted = listOf(),
+        )
+    }
+)
```
rahulsainani, line 32: Can create a derived property of this state, so this logic is removed from the UI and when the state is updated, the displayItems are present as it's based on two other existing state values. 

```
    val displayItems = items.filter { item ->
        item !in itemsBeingDeleted
    }
```

fenix/docs/architectureexample/HistoryFragment.kt
```diff
@@ -0,0 +1,73 @@
...
+
+@Composable
+private fun HistoryScreen(store: HistoryStore) {
+    val state = store.observeAsState(
+        initialValue = HistoryState.initial
+    ) { state -> state }
+    val displayItems = state.items.filter { item ->
+        item !in state.itemsBeingDeleted
+    }
+    LazyColumn(
```
rahulsainani, line 41: I know this is an example but I think to have the previews working and for the composables to be testable, we can extract a `HistoryContent` composable that takes in the state and callbacks with no reference to the store. The `HistoryScreen` composable acts as a scaffold for it where it observes the state.

fenix/docs/architectureexample/HistoryNavigationMiddleware.kt
```diff
@@ -0,0 +1,27 @@
...
+    private val navController: NavController,
+) : Middleware<HistoryState, HistoryAction> {
+    override fun invoke(
+        context: MiddlewareContext<HistoryState, HistoryAction>,
+        next: (HistoryAction) -> Unit,
+        action: HistoryAction,
+    ) {
+        when(action) {
+            is HistoryAction.OpenItem -> {
+                next(action)
```
rahulsainani, line 17: Seems like we can extract `next(action)` above the when, similar to `TelemetryMiddleware` ? 
Same point for `StorageMiddleware`.

rahulsainani (commented):
This looks good overall. I've left some comments for your consideration, I think we should address/discuss the comments and then this is good to go!

fenix/docs/architectureexample/HistoryFragment.kt
MatthewTighe: I'm not sure I fully agree that we need an additional top-level composable here - IIUC with the only purpose to observe state and define callbacks to dispatch actions. That's essentially what the `HistoryScreen` here is already doing. I think what you're suggesting is the following:

```
@Composable
private fun HistoryScreen(store: HistoryStore) {
    val state = store.observeAsState(initialValue = HistoryState.initial) { state -> state }
    val onMultiSelectHeaderDeleteClick = { store.dispatch(HistoryAction.DeleteItems(state.selectedItems)) }
    val onDeleteHeaderClick = { store.dispatch(HistoryAction.DeleteItems(state.items)) }
    val onItemClick = { item: History ->
        store.dispatch(HistoryAction.OpenItem(item))
    }
    val onItemLongClick = { item: History ->
        store.dispatch(HistoryAction.ToggleItemSelection(item))
    }
    val onItemDeleteClick = { item: History ->
        store.dispatch(HistoryAction.DeleteItems(listOf(item)))
    }
    HistoryContent(
        state,
        onMultiSelectHeaderDeleteClick,
        onDeleteHeaderClick,
        onItemClick,
        onItemLongClick,
        onItemDeleteClick,
    )
}

@Composable
private fun HistoryContent(
    state: HistoryState,
    onMultiSelectHeaderDeleteClick: () -> Unit,
    onDeleteHeaderClick: () -> Unit,
    onItemClick: (History) -> Unit,
    onItemLongClick: (History) -> Unit,
    onItemDeleteClick: (History) -> Unit,
) {
    LazyColumn(
        if (state.selectedItems.isNotEmpty()) {
            HistoryMultiSelectHeader(onDeleteSelectedClick = ::onMultiSelectHeaderDeleteClick)
        } else {
            HistoryHeader(onDeleteAllClick = ::onDeleteHeaderClick)
        }
        items(items = displayItems, key = { item -> item.id } ) { item ->
            val isSelected = state.selectedItems.find { selectedItem ->
                selectdItem == item
            }
            HistoryItem(
                item = item,
                isSelected = isSelected,
                onClick = ::onItemClick,
                onLongClick = ::onItemLongClick,
                onDeleteClick = ::onDeleteItemClick,
            )
        }
    )
}
```

@MozillaNoah and I have discussed this in some length during the tabs tray rewrite. I'm happy to discuss this more offline as well, but both of the above Composables, and the current incarnation of `HistoryScreen`, all feel like "container" Composables (vs "presentational" Composables). [Here is an article by Dan Abramov](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0) with some prior art on the subject.

Ultimately, I think one of the benefits to switching to a Redux model with Compose is that we can have more localized logic. I think extracting all the state observations and dispatches at a top-level Composable and then drilling those props through the whole Compose tree will cause a lot of context to be lost. Additionally, I find the above example to be pretty unnecessarily repetitive. 

As we switch to a Redux-ish model, I am expecting that we will find ourselves passing stores directly through container Composables throughout our UI trees, in order to avoid the kind of parameter explosions seen above. Since Stores and their reducers are pure, this shouldn't largely shouldn't impact previews or testing. We've even discussed adding Stores to `LocalContext`s but that would take some more thought.

fenix/docs/architectureexample/HistoryFragmentExample.kt
```diff
@@ -0,0 +1,60 @@
...
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        return ComposeView(requireContext()).apply {
+            setContent {
+                HistoryScreen(store)
+            }
+        }
+    }
+}
+
+Composable
```
MozillaNoah, line 31:
```suggestion
@Composable
```

fenix/docs/architectureexample/HistoryStorageMiddlewareExample.kt
```diff
@@ -0,0 +1,40 @@
...
+                scope.launch {
+                    val history = storage.load()
+                    context.store.dispatch(HistoryAction.ItemsChanged(history))
+                }
+            }
+            is HistoryAction.DeleteItems -> {
+                scope.launch {
+                    val currentItems = context.state.items
+                    if (storage.delete(action.items) is HistoryStorage.Success) {
+                        val updatedItems = currentItems - action.items
```
MozillaNoah, line 30: `updatedItems` is never used. should another action be fired, or should the item subtraction occur in the reducer for `DeleteFinished`?

fenix/docs/architectureexample/HistoryFragmentExample.kt
```diff
@@ -0,0 +1,60 @@
...
+        } else {
+            HistoryHeader(
+                onDeleteAllClick = { store.dispatch(HistoryAction.DeleteItems(state.items)) }
+            )
+        }
+        items(items = state.displayItems, key = { item -> item.id } ) { item ->
+            val isSelected = state.selectedItems.find { selectedItem ->
+                selectdItem == item
+            }
+            HistoryItem(
```
MozillaNoah, line 51: just to make sure, was an actual `HistoryItem` and some of the other composables intentionally omitted?

MozillaNoah (commented):
quick pass, but I like how this looks. just left a comment for clarification about your intent of this being an example

fenix/docs/architectureexample/HistoryFragmentExample.kt
MatthewTighe: Yeah, in keeping with the original examples I thought it would be fine to leave detailed Composable implementations to the imagination of the reader. I would consider adding them, but I worry that it would just add noise

fenix/docs/architectureexample/HistoryFragmentExample.kt
MatthewTighe: I am nothing without IDE safeguards

fenix/docs/architectureexample/HistoryStorageMiddlewareExample.kt
MatthewTighe: moved it to the reducer, thanks!

fenix/docs/architectureexample/HistoryNavigationMiddleware.kt
MatthewTighe: Yeah, added some comments to explain the thinking there as well. Thanks!

fenix/docs/architectureexample/HistoryStore.kt
MatthewTighe: Nice 😎

fenix/docs/architectureexample/HistoryFragment.kt
```diff
@@ -0,0 +1,73 @@
...
+        StoreProvider.get(this) {
+            HistoryStore(
+                initialState = HistoryState.initial,
+                middleware = listOf(
+                    HistoryNavigationMiddleware(findNavController())
+                    HistoryStorageMiddleware(HistoryStorage()),
+                    HistoryTelemetryMiddleware(),
+                )
+            ).apply {
+                dispatch(HistoryAction.Init)
```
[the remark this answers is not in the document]
MatthewTighe: My initial thinking was for cases where a Store was created but not immediately used, but I think your suggestion is better

fenix/docs/architectureexample/HistoryFragment.kt
rahulsainani: In the code you posted, `HistoryContent` can be easily tested based on the state, whereas the current example cannot be. To make it work in preview, we only need to pass the state and not the store (which might not be possible in all cases).

Compose recommends hoisting state upwards, so much so there's a dedicated section in the docs about it -> https://developer.android.com/jetpack/compose/state#state-hoisting 

Also, twitter has a dedicated lint rule for exactly this 😄  --> https://twitter.github.io/compose-rules/rules/#hoist-all-the-things

MatthewTighe: I'm not really willing to die on this hill especially for purposes of an example so am still willing to update this if you'd prefer. That said, I'm finding the discussion helpful for my own understanding, and am hoping future readers might be get something out of it as well 🙂 

These lint rules (as clarified offline are rehosted [here](https://slackhq.github.io/compose-lints/)) about hoisting seem to largely be intended to remove programmer error by removing mutability from the depths of the Compose tree. That makes sense - we all felt the pain of stateful View objects in the XML Android world (if I never touch another `EditText` again it will be too soon). However, an immutable, unidirectional data flow is also the purpose of the Redux model. Though technically possible, we don't run the same risk of accidental/harmful state mutations that something like the ViewModel pattern does since Stores don't expose mutable properties and instead require dispatches to update their state. In that way, a Store sits "above" a Compose tree already - we don't necessarily need to define all our state observations and callbacks at the root. IMO Google offering advice like their [state hoisting](https://developer.android.com/jetpack/compose/state#state-hoisting) is tailored to the rest of their architectural guidance. 

There are other industrial learnings that point to the pattern of hoisting being unfriendly from a developer ergonomics stance. Without being a React expert, I'd assume these deep parameter trees are also why [Hooks](https://legacy.reactjs.org/docs/hooks-custom.html) were invented and are now commonplace. It allows reading properties and dispatching actions directly to the application state from a lot more places in a codebase.

Frankly, I'm pulling my viewpoint directly from experience that we're already gaining here as well. The [TabsTray rewrite](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt#91) follows hoisting advice and IMO I do not find a parameter list like that useful. Not to mention, it requires reading all the way down the tree to find where a callback actually occurs and then all the way back up the tree to find the actual definition of the callback. That amount of indirection and code-clickthroughs are part of the problem I was intending to solve by removing interactors and controllers.

Finally, as to testing and previews. I am firstly interested in when this

> not the store (which might not be possible in all cases).

would be the case. For previews and for tests, there's no requirement to include the usual middleware attached to a Store. Since without them, Stores are just a collection of data and pure functions then it should be possible to use them directly in both tests and previews. In the case of tests, I would argue that using an actual Store (instead of defining a bunch of fake callbacks to pass to the Composable) is actually preferable because it is closer to a real-world usecase. In the case of previews, using an actual Store saves a huge amount of typing in terms of defining fake callbacks as well - and it means that the previews are automatically functioning in interactive mode. The [TabsTray Preview](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt#440) shows some examples of dispatches directly, but that could be simplified even further if those dispatches were made inline. It also shows some of the difficulty in not using Stores inline - it will end up being just a lot of extra boilerplate.

I think an argument against passing Stores directly would be that it more strongly ties us to the architectural pattern. Although a risk, IMO the benefits we would receive in terms of ease of reading and writing code would outweigh that. 

TL;DR Stores are inherently unidirectional, so IMO we don't need to be as concerned about hoisting.

MatthewTighe: Also, I am not advocating for passing the Store to every Composable. Generally speaking, I think the only hard rule we should derive is that leaf nodes of the Compose tree should have all their state and actions hoisted. Otherwise, I think we can use our best judgement as to the depth that we want to propagate Stores.

rahulsainani: Thanks for sharing your thoughts with such thoroughness, I really appreciate it. 🙌 

So we're on the same page that stateful UI is not something we want. By passing down the store, we create more stateful UI and by hoisting as much above as possible, we create more stateless UI which can be tested/previewed by just passing the state (content, loading, error, different content). 

To the comment saying that we can exercise judgement calls as to the depth that Stores get propagated down the tree, my judgement and I agree with the docs here, that depth is at the root level :) The root level composable or fragment that observes the stores, acts as a controller/screen and just passes the state down. That way there's only one place the state is being observed. Imagine if parent and child are both observing state and parent also passes a mapped state down to the child, what's the source of truth here? 🤔 

> Finally, as to testing and previews. I am firstly interested in when this
> 
>     not the store (which might not be possible in all cases).
When the middlewares are fetching/deleting/updating data and then dispatching actions so the state is updated. If we run the preview with a store without this middleware, these interactions won't work. We'd have to pass in an initial state while constructing that store which is what we just wanted to do with `HistoryContent`.


> Frankly, I'm pulling my viewpoint directly from experience that we're already gaining here as well. The [TabsTray rewrite](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt#91) follows hoisting advice and IMO I do not find a parameter list like that useful. Not to mention, it requires reading all the way down the tree to find where a callback actually occurs and then all the way back up the tree to find the actual definition of the callback. That amount of indirection and code-clickthroughs are part of the problem I was intending to solve by removing interactors and controllers.

I agree that the big parameter list is an issue and I'd argue that speaks about the complexity and the specific kind of feature that is. I'll be careful of creating generalized ideas from one feature.


Please let me know if you have other thoughts on it, I'm sure you'll do! Also, happy to discuss more on a call. :) 

**Bit tangential**
> IMO Google offering advice like their [state hoisting](https://developer.android.com/jetpack/compose/state#state-hoisting) is tailored to the rest of their architectural guidance.

I'll argue that this is more tailored to compose usage than their arch guideline. Saying that, `ViewModel` or any `StateHolder` can be, and is generally used with UDF pattern as well, specially with compose.

MatthewTighe: > When the middlewares are fetching/deleting/updating data and then dispatching actions so the state is updated. If we run the preview with a store without this middleware, these interactions won't work.

Alternatively, creating middleware fakes for tests would be convenient to re-use in previews:
```
class FakeScreenMiddleware(
    private val onNavItem: (item: Item) -> Unit,
) : Middleware<ScreenState, ScreenAction> {
    override fun invoke(
        context: MiddlewareContext<ScreenState, ScreenAction>,
        next: (ScreenAction) -> Unit,
        action: ScreenAction,
    ) {
        when (action) {
            is ScreenAction.Init -> context.store.dispatch(ScreenAction.ItemsLoaded(fakeItems)) // acts like we really fetched items
            is ScreenAction.ItemClicked -> onNavItem(action.item) // verify in test and ignore in preview
        }
    }
}
```

>  We'd have to pass in an initial state while constructing that store which is what we just wanted to do with HistoryContent.

I anticipate that most `Store`s will have an easy way to create an initial state. It should then be easy to adjust this based on the needs of the test or preview using the `copy` method:

```
@Preview
fun ChildComposablePreview {
    val state = State.initial.copy(changeMe = "newThing", middleware = listOf(FakeMiddleware())
    ChildComposable(Store(initialState = state, middleware = middleware, reducer = ::realReducer))
}

// vs

@Preview
fun ChildComposablePreview {
    val changedState = "newThing"
    val defaultState1 = "default1"
    val defaultState2 = "default2"
    // etc for a few more properties
    val clickHandler1 = { }
    val clickHandler2 = { }
    val clickHandler3 = { }
    val clickHandler4 = { }
    ChildComposable(
        changedState = changedState
        defaultState1 = defaultState1
        ...
   )
}
```

> Imagine if parent and child are both observing state and parent also passes a mapped state down to the child, what's the source of truth here?

The source of truth in all cases is the Store. For your concern to be harmful, a parent would need to be mutating the state it read before passing it down to the child, or generating a new property conditionally. I am also not sure why a child would expect a property as a parameter and then also be observing that property from the Store. 

The ultimate responsibility of state hoisting is to ensure that components that share state don't mutate that state in a way the others can't observe. Moving the observations of state and the declaration of actions to a root level Composable isn't actually hoisting it, but instead just changing what line of code the definition is on. 

Looking at the "Key Points" section of the [state hoisting link](https://developer.android.com/jetpack/compose/state#state-hoisting) again:

1. State should be hoisted to at least the lowest common parent of all composables that use the state (read).
2. State should be hoisted to at least the highest level it may be changed (write). 
3. If two states change in response to the same events they should be hoisted together.

We see that there is a focus on reading and writing state. Since this is handled entirely by the Store, we have effectively hoisted the state already. 

For another example:

```
@Composable
fun TopLevelScreen(store: ScreenStore) {
    val shouldDisplaySection = store.observeAsState() { state -> state.newSectionIsVisible }
    if (shouldDisplaySection) {
        NewSectionComposable(store)
    } else {
        OtherSectionComposable(store)
    }
}

@Composable
fun NewSectionComposable(store: ScreenStore) {
    val headerTitle = store.observeAsState() { state -> state.activeSectionHeader }
    val headerButtonClick = { store.dispatch { ScreenStoreAction.HeaderClicked } }
    ReusableHeaderComposable(headerTitle)
    BodyComposable(store)
}

@Composable
fun BodyComposable(store: ScreenStore) {
    val listState = val state = rememberLazyListState()
    val storeItems = store.observeAsState() { state -> state.items }
    val displayOnboardingSection = store.observeAsState() { state -> state.isOnboardingVisible }
    SectionOnboardingComposable(store)
    LazyColumn(listState) {
        items(storeItems) { item ->
            val numberItems = storeItems.size
            val onClick = { store.dispatch(ScreenStoreAction.ItemClicked(item)) }
            val onLongClick = { store.dispatch(ScreenStoreAction.ItemLongClicked(item)) }
            ItemComposable(item, onClick, onLongClick)
        }
    }
}

@Composable
fun SectionOnboardingComposable(store: ScreenStore) {
    val sectionDescription = store.observeAsState() { state -> state.activeSectionDescription }
    val learnMoreLink = store.observeAsState() { state -> state.learnMoreLink }
    val learnMoreLinkDescription = store.observeAsState { state -> learnMoreLinkDescription }
    val dismissOnboarding = { store.dispatch(ScreenStoreAction.OnboardingDismissed) }
    Column {
        Row {
            ReusableCancelButton(onClick = dismissOnboarding)
        }
        ReusableTextBody(text = sectionDescription)
        ReusableLink(description = learnMoreLinkDescription, link = learnMoreLink)
    }
}
```

This is a relatively small, contrived example still. There are 7 state observations and 4 dispatches. Should those all be lifted to a top-level Composable and turned into parameters to functions? To me, that creates a lot of noise in the intermediate levels.

Once we Composify the rest of the Home Screen, would there be one top-level Home Screen that defined every piece of state and every action that could be taken on the screen?

rahulsainani (approved):
After some rather lengthy discussions about state hoisting, similarities/differences between `ViewModels` and `Stores`, we've decided to agree to disagree in good spirit and move forward, leaving room for some software craftsmanship and yet not sacrificing a lot on consistency. We agree that stores should not be part of the leaf nodes and where the limit is left to the author to decide. If and when we incorporate the [compose lint rules](https://slackhq.github.io/compose-lints/), we will decide later if we'd like to enable the [hoist all the things](https://slackhq.github.io/compose-lints/rules/#hoist-all-the-things) rule as that rule will not be compatible with this approach. 

Thanks for the discussions! Hopefully, we can find some balance sooner in the future 🚀

PR #2776, opened 2023-07-10
Bug 1841639 - Add common Switch composable

rahulsainani:
This will be used in the shopping experience settings card and since it's a common component that'll be used across the app, it's in the compose package. [Figma](https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?type=design&node-id=1808-4077&mode=design&t=8cfht7Ec03SldCew-0)

### Preview
Screenshot 2023-07-10 at 15 26 34

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1841639


t-p-white:
Nice job, lgtm 👍 

One pedantic comment is that the implementation is _slightly_ different to the Figma design as the track is visible through the circle.

rahulsainani:
> Nice job, lgtm 👍
> 
> One pedantic comment is that the implementation is _slightly_ different to the Figma design as the track is visible through the circle when disabled.

Good catch! 🔍  Just updated to so the track is not visible through the thumb for the disabled state, material switch also uses the same method of compositing the color.

t-p-white (approved):

MozillaNoah:
I think this is 80% there, but it looks like this is derived from the root Switch component in Figma (which are on a hidden Figma page), vs the [Switch design component](https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?type=design&node-id=3031-17475&mode=design&t=K5uHMfcSVVKpI0jL-0) that includes the descriptor text. 

Can you update this to be an implementation of the complete design component instead of just the switch? 

Alternatively, you could include two APIs, one for just the radio button and a second for the radio button + the text label (and support for nested labels).

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchToggle.kt
```diff
@@ -0,0 +1,115 @@
...
+ * @param enabled whether the component is enabled or grayed out
+ */
+@Composable
+fun SwitchToggle(
+    checked: Boolean,
+    onCheckedChange: ((Boolean) -> Unit),
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+) {
+    Switch(
```
MozillaNoah, line 40:
```suggestion
fun Switch(
    checked: Boolean,
    onCheckedChange: ((Boolean) -> Unit),
    modifier: Modifier = Modifier,
    enabled: Boolean = true,
) {
    androidx.compose.material.Switch(
```

this mirrors what we've done in `Divider`, `TextButton`, `Button`, and a couple other places

gabrielluong: +1 to the using the `Switch` naming since this will mirror naming in the figma component.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchToggle.kt
```diff
@@ -0,0 +1,115 @@
...
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.theme.FirefoxTheme
+
+private const val DISABLED_ALPHA = 0.5f
+
+/**
+ * UI for a switch that can be on or off.
+ *
+ * @param checked whether or not this component is checked.
```
gabrielluong, line 27: For all the @param description, could we capitalize the start of the descriptions?

```suggestion
 * @param checked Whether or not this component is checked.
```

rahulsainani:
I see, thanks for the clarification! So I was referring to a component in Figma that's not really a component at all, it looked pretty formal since it had the spec defined. I've updated to use the label, the only concern I have is that the label is styled with `subtitle1` however in shopping experience it is styled with `body2`. I will sync with the design team about this. 

Anyway, this is implemented based on the android component spec, please let me know your thoughts.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
```diff
@@ -0,0 +1,179 @@
...
+            Row {
+                Spacer(Modifier.weight(1f))
+
+                SwitchWithLabel(
+                    checked = false,
+                    onCheckedChange = {},
+                    label = "Nested",
+                    modifier = Modifier.weight(1f),
+                )
+            }
```
MozillaNoah, line 176: it might make more sense to have an "isNested" boolean in the API for `SwitchWithLabel`, that way the spacing is consistent across consumers of it. I double checked the figma and our existing nested preferences, and the nested spacing should be 72.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
```diff
@@ -0,0 +1,179 @@
...
+        modifier = modifier,
+        enabled = enabled,
+        colors = SwitchDefaults.colors(
+            uncheckedThumbColor = FirefoxTheme.colors.formOff,
+            uncheckedTrackColor = FirefoxTheme.colors.formSurface,
+            checkedThumbColor = FirefoxTheme.colors.formOn,
+            checkedTrackColor = FirefoxTheme.colors.formSurface,
+            disabledUncheckedThumbColor = FirefoxTheme.colors.formOff
+                .copy(alpha = DISABLED_ALPHA)
+                .compositeOver(FirefoxTheme.colors.formSurface),
```
MozillaNoah, line 97: where is the alpha value derived from? 

We may not need to do this copy copying + alpha logic, as`MaterialSwitch` is already applying an alpha via `uncheckedTrackAlpha` and `checkedTrackAlpha` as part of the `SwitchDefaults.colors` API

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
```diff
@@ -0,0 +1,179 @@
...
+            disabledCheckedTrackColor = FirefoxTheme.colors.formSurface.copy(alpha = DISABLED_ALPHA),
+        ),
+    )
+}
+
+@LightDarkPreview
+@Composable
+private fun SwitchWithLabelPreview() {
+    FirefoxTheme {
+        Column(
```
MozillaNoah, line 111: time allowing, it'd be nice to have the Switches interactable in the preview here (i.e. able to have demo their toggles) and the disabled state could be validated.

MozillaNoah (commented):
We're super close with this! I liked your use of a type alias for the root material switch.

Thanks for holding up with the iterations on this. All of our top level components have received review churn, since we want to make sure they're all square on the first go, so I appreciate your patience here.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
```diff
@@ -0,0 +1,179 @@
...
+ * state is requested.
+ * @param label Text to be displayed next to the switch.
+ * @param modifier Modifier to be applied to the switch layout.
+ * @param enabled Whether the switch is enabled or grayed out.
+ */
+@Composable
+fun SwitchWithLabel(
+    checked: Boolean,
+    onCheckedChange: ((Boolean) -> Unit),
+    label: String,
```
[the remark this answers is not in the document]
rahulsainani: Yeah I thought the same, then reverted, and then double checked the `MaterialSwitch` API. We're following the same params order here already.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
rahulsainani: hmm, the way I understood the nested component is that there can be parents (Rows) which have their own content on the start (e.g. desktop site menu item) and then `SwitchWithLabel`, and not just a 72.dp padding. 

That seems very much like Settings Screen component and not a generic one, wdyt? The way I think it should be is to be wrapped in another Parent. Also reducing if logic from our components.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
rahulsainani: The alpha is from the [designs](https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?type=design&node-id=3031-17475&mode=design&t=liRd1rKuUxEhzCcZ-0). 

It seems we do, otherwise it composites over MaterialTheme.colors.surface, which results in a different color from the spec. 

Tangential, An idea would be to map the theme colors to material as well so when we use some material component like dialog/calendar it automatically uses our surface/primary colors, the basic ones which are easily mappable.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
rahulsainani: good idea, done!

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
MozillaNoah: Hrmm, yeah, I can't immediately think of another example of nested / indented switches besides Settings, so I'm cool with deferring making it part of the API until we have a better understanding of the use cases of this component.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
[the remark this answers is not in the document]
MozillaNoah: this isn't necessarily to match the order of `MaterialSwitch` (the `Switch` Composable correctly matches though, as you said), as this is for our wrapper Composable. This is more to match our other Composable parameter ordering. 

Maybe the label could be the first parameter, that way the `oncheckedchange` and `checked` are still visually grouped next to each other?

MozillaNoah (approved):
left a comment for the remaining nit, but this looks good and the preview is great! Thanks for making those changes to make the preview interactable.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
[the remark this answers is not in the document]
rahulsainani: Oh yeah makes, had a brain fart there when I read it the first time.

PR #2821, opened 2023-07-12
Bug 1843168 - Expose Mockito through test support component

MatthewTighe:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1843168


android-components/components/support/test/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     implementation ComponentsDependencies.kotlin_reflect
 
     implementation ComponentsDependencies.androidx_test_junit
-    implementation ComponentsDependencies.testing_mockito
+    api ComponentsDependencies.testing_mockito
```
rahulsainani, line 40: [api](https://developer.android.com/topic/modularization/patterns) leads to larger build times as multiple modules might have to be built, it's for testing but still something to consider. I see you want to expose verify functions, maybe we can create them in `support-test` similar to mocks or just add the `testImplementation` in app for mockito, similar to other common 3rd party dependencies.

rahulsainani: That's the general guidance. On further thought, `support-test` is intended to be a pass through module so let's move ahead with this. 👍 

Tangential
However, we should be mindful of `api`. Generally modules that need should directly `implement` 3rd party dependencies they need, as a result the modules remain fairly independent and can be compiled in parallel.

rahulsainani (approved):

android-components/components/support/test/build.gradle
MatthewTighe: Oh yeah, my intent was explicitly to use our existing support module as a passthrough. Apologies if that wasn't clear from the bug description.

On second thought though, should we remove the direct `implementation` usage throughout A-C as part of this patch?

rahulsainani: thanks for the clarification, sounds fair 👍

MatthewTighe:
Realized it would be best to wait until Monday to land this, if anybody sees this and is wondering

PR #2889, opened 2023-07-18
Bug 1842604 - Telemetry to measure Search results response time

t-p-white:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1842604


t-p-white:
Please reach out on Slack for any discussions on this PR

t-p-white:
⚠️ TODO Add telemetry comment

fenix/app/src/main/java/org/mozilla/fenix/telemetry/TelemetryMiddleware.kt
```diff
@@ -20,27 +20,37 @@ import mozilla.components.concept.base.crash.CrashReporting
...
+ * @property timerIdFunc Fun to provide the Metric [TimerId].
  */
 class TelemetryMiddleware(
     private val context: Context,
     private val settings: Settings,
     private val metrics: MetricController,
     private val crashReporting: CrashReporting? = null,
+    private val nimbusSearchEngine: String = FxNimbus.features.searchExtraParams.value().searchEngine,
+    private val searchState: MutableMap<String, TimerId> = mutableMapOf(),
+    private val timerIdFunc: () -> TimerId = Metrics.searchPageLoadTime::start,
```
boek, line 53: nit: Since we already know that this is a function, maybe instead of `timerIdFunc` we could use something like `createTimer` instead? Up to you :)

fenix/app/src/test/java/org/mozilla/fenix/telemetry/TelemetryMiddlewareTest.kt
```diff
@@ -60,12 +63,21 @@ class TelemetryMiddlewareTest {
...
         Clock.delegate = clock
         settings = Settings(testContext)
-        telemetryMiddleware = TelemetryMiddleware(testContext, settings, metrics)
+        telemetryMiddleware = TelemetryMiddleware(
+            context = testContext,
+            settings = settings,
+            metrics = metrics,
+            nimbusSearchEngine = SEARCH_ENGINE_NAME,
+            searchState = searchState,
+            timerIdFunc = { timerId },
```
boek, line 79: We might not even need this as a function injected in the constructor as we're using the same value in the tests 🤔

fenix/app/src/main/java/org/mozilla/fenix/telemetry/TelemetryMiddleware.kt
```diff
@@ -54,12 +64,31 @@ class TelemetryMiddleware(
...
                         Events.normalAndPrivateUriCount.add()
                     }
+
+                    // We'll ignore any type of consent page as this will interfere with loading times.
+                    val progressCompleted = tab.content.progress == PROGRESS_COMPLETE
+                    if (hasFinishedLoading && progressCompleted) {
+                        searchState[action.sessionId]?.let {
+                            Metrics.searchPageLoadTime.stopAndAccumulate(it)
+                        }
+                        searchState.remove(action.sessionId)
```
boek, line 90: Should we remove the entry from `searchState` when we've finished loading even if progress hasn't been completed?

boek (approved):
Just a couple of questions, but I think we're good for now

fenix/app/src/test/java/org/mozilla/fenix/telemetry/TelemetryMiddlewareTest.kt
t-p-white: As far as I can see, we need to pass the TimerId from the tests to be able to assert the correct value, but looking at it again it doesn't need to be a fun reference 👍

fenix/app/src/main/java/org/mozilla/fenix/telemetry/TelemetryMiddleware.kt
t-p-white: I'm thinking probably not, we could get a situation where loading has finished, but progress hasn't completed, then when it has completed it would no longer exist in the map to stop?

boek: @t-p-white In that case we still wouldn't be able to send the telemetry as we only run this code when the loading state changes. Maybe we should create a `tryToFinish` function that we call here as well as when progress changes on the content state

t-p-white: @boek good callout, thanks! I think given the use case, we can just simplify to just cancel the timer when progress is not completed

PR #2891, opened 2023-07-18
Bug 1843029 - Add preferences for shopping experience

rahulsainani:
### How
– Add preferences in `Settings.kt`
– Create `ReviewQualityCheckPreferences` as an abstraction to get/update preferences.
– Define store actions based on their property, whether they are handled by a middleware or cause an update to the state.
– Create `PreferencesMiddleware` that uses `ReviewQualityCheckPreferences`

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1843029

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
@@ -9,4 +9,48 @@ import mozilla.components.lib.state.Action
 /**
  * Actions for review quality check feature.
  */
-sealed interface ReviewQualityCheckAction : Action
+sealed interface ReviewQualityCheckAction : Action {
+
+    /**
+     * Actions that are observed by middlewares.
+     */
+    sealed interface MiddlewareAction : ReviewQualityCheckAction
+
+    /**
+     * Actions that cause updates to state.
+     */
+    sealed interface UpdateAction : ReviewQualityCheckAction
+
+    /**
+     * Actions related to preferences.
+     */
+    sealed interface PreferencesMiddlewareAction : MiddlewareAction
+
+    /**
+     * Triggered when the store is initialized.
+     */
+    object Init : PreferencesMiddlewareAction
+
+    /**
+     * Triggered when the user has opted in to the review quality check feature.
+     */
+    object OptIn : PreferencesMiddlewareAction
+
+    /**
+     * Triggered when the user has opted out of the review quality check feature.
+     */
+    object OptOut : PreferencesMiddlewareAction, UpdateAction
+
+    /**
+     * Triggered when the user has enabled or disabled product recommendations.
+     */
+    object ToggleProductRecommendation : PreferencesMiddlewareAction, UpdateAction
+
+    /**
+     * Triggered as a result of a [PreferencesMiddlewareAction] to update the state.
+     */
+    data class UpdateUserPreferences(
+        val hasUserOptedIn: Boolean,
+        val isProductRecommendationsEnabled: Boolean,
+    ) : UpdateAction
+}
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferences.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferences.kt
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.shopping.state
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Interface to get and set preferences for the review quality check feature.
+ */
+interface ReviewQualityCheckPreferences {
+    /**
+     * Returns true if the user has opted in to the review quality check feature.
+     */
+    suspend fun enabled(): Boolean
+
+    /**
+     * Returns true if the user has enabled product recommendations.
+     */
+    suspend fun productRecommendationsEnabled(): Boolean
+
+    /**
+     * Sets whether the user has opted in to the review quality check feature.
+     */
+    suspend fun setEnabled(isEnabled: Boolean)
+
+    /**
+     * Sets whether the user has enabled product recommendations.
+     */
+    suspend fun setProductRecommendationsEnabled(isEnabled: Boolean)
+}
+
+/**
+ * Implementation of [ReviewQualityCheckPreferences] that uses [Settings] to store/fetch
+ * preferences.
+ *
+ * @param settings The [Settings] instance to use.
+ */
+class ReviewQualityCheckPreferencesImpl(
+    private val settings: Settings,
+) : ReviewQualityCheckPreferences {
+
+    override suspend fun enabled(): Boolean = withContext(Dispatchers.IO) {
+        settings.isReviewQualityCheckEnabled
+    }
+
+    override suspend fun productRecommendationsEnabled(): Boolean = withContext(Dispatchers.IO) {
+        settings.isReviewQualityCheckProductRecommendationsEnabled
+    }
+
+    override suspend fun setEnabled(isEnabled: Boolean) {
+        settings.isReviewQualityCheckEnabled = isEnabled
+    }
+
+    override suspend fun setProductRecommendationsEnabled(isEnabled: Boolean) {
+        settings.isReviewQualityCheckProductRecommendationsEnabled = isEnabled
+    }
+}
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
@@ -1,0 +1,93 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.shopping.state
+
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareContext
+import mozilla.components.lib.state.Store
+
+/**
+ * Middleware for getting and setting review quality check user preferences.
+ *
+ * @param reviewQualityCheckPreferences The [ReviewQualityCheckPreferences] instance to use.
+ * @param scope The [CoroutineScope] to use for launching coroutines.
+ */
+class ReviewQualityCheckPreferencesMiddleware(
+    private val reviewQualityCheckPreferences: ReviewQualityCheckPreferences,
+    private val scope: CoroutineScope,
+) : Middleware<ReviewQualityCheckState, ReviewQualityCheckAction> {
+
+    override fun invoke(
+        context: MiddlewareContext<ReviewQualityCheckState, ReviewQualityCheckAction>,
+        next: (ReviewQualityCheckAction) -> Unit,
+        action: ReviewQualityCheckAction,
+    ) {
+        when (action) {
+            is ReviewQualityCheckAction.PreferencesMiddlewareAction -> {
+                processAction(context.store, action)
+            }
+
+            else -> {
+                // no-op
+            }
+        }
+        // Forward the actions
+        next(action)
+    }
+
+    private fun processAction(
+        store: Store<ReviewQualityCheckState, ReviewQualityCheckAction>,
+        action: ReviewQualityCheckAction.PreferencesMiddlewareAction,
+    ) {
+        when (action) {
+            is ReviewQualityCheckAction.Init -> {
+                scope.launch {
+                    val hasUserOptedIn = reviewQualityCheckPreferences.enabled()
+                    val isProductRecommendationsEnabled =
+                        reviewQualityCheckPreferences.productRecommendationsEnabled()
+                    store.dispatch(
+                        ReviewQualityCheckAction.UpdateUserPreferences(
+                            hasUserOptedIn = hasUserOptedIn,
+                            isProductRecommendationsEnabled = isProductRecommendationsEnabled,
+                        ),
+                    )
+                }
+            }
+
+            ReviewQualityCheckAction.OptIn -> {
+                scope.launch {
+                    val isProductRecommendationsEnabled =
+                        reviewQualityCheckPreferences.productRecommendationsEnabled()
+                    store.dispatch(
+                        ReviewQualityCheckAction.UpdateUserPreferences(
+                            hasUserOptedIn = true,
+                            isProductRecommendationsEnabled = isProductRecommendationsEnabled,
+                        ),
+                    )
+
+                    // Update the preference
+                    reviewQualityCheckPreferences.setEnabled(true)
+                }
+            }
+
+            ReviewQualityCheckAction.OptOut -> {
+                scope.launch {
+                    // Update the preference
+                    reviewQualityCheckPreferences.setEnabled(false)
+                }
+            }
+
+            ReviewQualityCheckAction.ToggleProductRecommendation -> {
+                scope.launch {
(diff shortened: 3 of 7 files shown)


fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
```diff
@@ -9,4 +9,47 @@ import mozilla.components.lib.state.Action
...
+
+    /**
+     * Actions that cause updates to state.
+     */
+    sealed interface UpdateAction : ReviewQualityCheckAction
+
+    /**
+     * Actions related to preferences.
+     */
+    sealed interface PreferencesMiddlewareAction : MiddlewareAction
```
MozillaNoah, line 27: I'm on the fence about having these sealed interfaces for the actions to inherit from. There are undoubtedly going to be situations where one user input will trigger something that needs to be handled by multiple middlewares. 

Looking at `OptInAction` below, we'll be dispatching this when the user clicks on the preference toggle, but it's a `PreferencesMiddlewareAction` and is therefore only ever going to be consumed by the `PreferencesMiddleware`. Based on the current state of this, if there needed to be any additional side effects as a result of that user click (such as telemetry), we'll then have to fire additional actions for every specific middleware that needs to react. 

Or was your intent for Actions to implement any number of these super types, and `OptInAction` would eventually be a `PreferencesMiddlewareAction` and something like`TelemetryMiddlewareAction`?

Ultimately, I'm not sure there's much to gain having the overhead of these three super types vs just following the existing pattern of having the middlewares receive every action and simply ignoring the irrelevant ones. The middleware in this PR is already using a `when` for parsing out the relevant actions, there's now just an additional interstitial short-circuit check before it.

This all seems like a pattern on top of Redux, but I'm curious what @MatthewTighe would think.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferences.kt
```diff
@@ -0,0 +1,59 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.shopping.state
+
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Interface to get and set preferences for the review quality check feature.
+ */
+interface ReviewQualityCheckPreferences {
```
MozillaNoah, line 12: is this intention here to abstract our `Settings` implementation away from the middleware?

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
```diff
@@ -0,0 +1,90 @@
...
+
+            is ReviewQualityCheckAction.OptInAction -> {
+                scope.launch {
+                    val isProductRecommendationsEnabled =
+                        reviewQualityCheckPreferences.productRecommendationsEnabled()
+                    next(
+                        ReviewQualityCheckAction.UpdateFeatureStateAction(
+                            hasUserOptedIn = action.hasUserOptedIn,
+                            isProductRecommendationsEnabled = isProductRecommendationsEnabled,
+                        ),
```
MozillaNoah, line 66: Going off of the previous comment, we don't need to dispatch separate actions to mutate the state in this middleware for all of these actions. The primary reducer could just handle `ReviewQualityCheckAction.OptInAction` and `ReviewQualityCheckAction.ProductRecommendationAction`.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
```diff
@@ -0,0 +1,90 @@
...
+                        hasUserOptedIn = true,
+                        isProductRecommendationsEnabled = action.isProductRecommendationEnabled,
+                    ),
+                )
+                scope.launch {
+                    // Update the preference
+                    reviewQualityCheckPreferences.setProductRecommendationsEnabled(
+                        action.isProductRecommendationEnabled,
+                    )
+                }
```
MozillaNoah, line 86: should the coroutine usage instead live in `ReviewQualityCheckPreferencesImpl`? is it necessary?

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
```diff
@@ -9,4 +9,47 @@ import mozilla.components.lib.state.Action
...
+
+    /**
+     * Triggered when the store is initialized.
+     */
+    object Init : PreferencesMiddlewareAction
+
+    /**
+     * Triggered when the user has opted in or out of the review quality check feature.
+     */
+    data class OptInAction(
```
MozillaNoah, line 37: Having `action` in the name is a bit redundant
```suggestion
    data class OptInChange(
```

Alternatively, we could have two actions
```suggestion
    object OptIn(
    object OptOut(
```

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
```diff
@@ -9,4 +9,47 @@ import mozilla.components.lib.state.Action
...
+        val hasUserOptedIn: Boolean,
+    ) : PreferencesMiddlewareAction
+
+    /**
+     * Triggered when the user has enabled or disabled product recommendations.
+     */
+    data class ProductRecommendationAction(
```
MozillaNoah, line 44: This might make the verbiage and logic flow a little more clear upstream and downstream. We have existing examples of using two actions to toggle on/off part of the state vs one action with a boolean.
```suggestion
    object EnableProductRecommendation(
    object DisableProductRecommendation(
```

MozillaNoah (commented):
first pass. I have some concerns about the sealed class overhead but, in general, I like the direction these changes are going.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
MatthewTighe: Alternatively, use something like `ToggleProductRecommendations` and let the reducer and middleware change it based on previous state.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
MatthewTighe: I left a comment in the middleware below that addresses my feelings on this a bit. In general, I agree with Noah. I think we risk introducing brittleness to this feature with coupling that is this specific.

MatthewTighe (commented):
🚗 Left a few comments for consideration. In general, I would encourage thinking about each component of the Redux model as a mostly independent actor. The reducer shouldn't have knowledge of what middleware have done to an action through the chain. Similarly, middleware shouldn't know about each other either.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
```diff
@@ -0,0 +1,90 @@
...
+            is ReviewQualityCheckAction.Init -> {
+                scope.launch {
+                    val hasUserOptedIn = reviewQualityCheckPreferences.enabled()
+                    val isProductRecommendationsEnabled =
+                        reviewQualityCheckPreferences.productRecommendationsEnabled()
+                    next(
+                        ReviewQualityCheckAction.UpdateFeatureStateAction(
+                            hasUserOptedIn = hasUserOptedIn,
+                            isProductRecommendationsEnabled = isProductRecommendationsEnabled,
+                        ),
```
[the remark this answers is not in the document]
rahulsainani: I agree, that's a good point! updated 👍

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
rahulsainani: Updated to make the reading prefs, suspend functions, so it's more clear now, the scope is still injected in the store/middleware.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferences.kt
rahulsainani: Yes

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1676,6 +1676,22 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         }
     }
 
+    /**
+     * Indicates if the review quality check feature is enabled by the user.
+     */
+    var isReviewQualityCheckEnabled by booleanPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_is_review_quality_check_enabled),
+        default = false,
```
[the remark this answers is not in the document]
rahulsainani: No, it should be false by default

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
rahulsainani: This is not too dissimilar to having `MessagingAction` sealed heirarchy in `AppAction`. The overall idea is to rely on the compiler and have all the cases covered and not use else. I've updated the actions so it will make more sense now. Let's get on a call if we want to discuss more.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferences.kt
```diff
@@ -0,0 +1,61 @@
...
+        settings.isReviewQualityCheckProductRecommendationsEnabled
+    }
+
+    override fun setEnabled(isEnabled: Boolean) {
+        settings.isReviewQualityCheckEnabled = isEnabled
+    }
+
+    override fun setProductRecommendationsEnabled(isEnabled: Boolean) {
+        settings.isReviewQualityCheckProductRecommendationsEnabled = isEnabled
+    }
```
MatthewTighe, line 60: These could `scope.launch` directly. IMO It would add value to this abstraction to have it enforce the threading rules expected of these calls

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
```diff
@@ -0,0 +1,91 @@
...
+
+                    // Update the preference
+                    reviewQualityCheckPreferences.setEnabled(true)
+                }
+            }
+
+            ReviewQualityCheckAction.OptOut -> {
+                // Update the preference
+                reviewQualityCheckPreferences.setEnabled(false)
+            }
```
MatthewTighe, line 80: If my above suggestion is not taken, this will need a `scope.launch` I think.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Middleware for getting and setting review quality check user preferences.
+ *
+ * @param reviewQualityCheckPreferences The [ReviewQualityCheckPreferences] instance to use.
+ * @param scope The [CoroutineScope] to use for launching coroutines.
+ */
+class ReviewQualityCheckPreferencesMiddleware(
+    private val reviewQualityCheckPreferences: ReviewQualityCheckPreferences,
+    private val scope: CoroutineScope,
```
MatthewTighe, line 21: I think we could create a default parameter for the scope here so that it didn't need to be propagated through the type hierarchy

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckStore.kt
```diff
@@ -4,12 +4,70 @@
...
+        ReviewQualityCheckAction.OptOut -> {
+            return ReviewQualityCheckState.NotOptedIn
+        }
+
+        ReviewQualityCheckAction.ToggleProductRecommendation -> {
+            return if (state is ReviewQualityCheckState.OptedIn) {
+                state.copy(productRecommendationsPreference = !state.productRecommendationsPreference)
+            } else {
+                state
+            }
```
MatthewTighe, line 70: Will it be possible to toggle product recommendations while not opted in? If so, could we get a test describing that case? If not, could we get a comment describing that this isn't a reachable state?

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
MatthewTighe: Would be happy to have this discussion in more depth offline, but thought I'd leave a comment since I'm already here.

The big difference I see with the `MessagingAction` is that it's shrinking specificity around a feature of a global Store. To me, it makes sense to add that level of granularity. That's mirrored in the fact that the state of the feature can be easily encapsulated in a `MessagingState` property on the greater `AppState`. There's also no tight coupling implied between `MessagingAction` and its middleware or reducer. Anything that interacts with `AppState` can potentially respond to `MessagingAction`s. 

My experience in prototyping to combine global Stores (and potentially move to a single `ScopedStore` model where feature stores could be temporarily combined) informed my opinion that in some cases we actually want to _broaden_ our specificity in order to create a more powerful tool. For example, I think we will need to remove the generic action bounds on pretty much every foundational type in `lib-state` in order to make those changes. We would definitely lose some compile-time exhaustiveness checking in our `when` expressions, but I think the overall benefits will be well worth it.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckStore.kt
rahulsainani: No, it won't be possible.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
MozillaNoah: To close the loop of this convo from my end, I think we acknowledged some of the "risks" of doing a state machine on top of Redux, so I'm fine with moving forward and seeing the final product of the code. I'm personally not 100% sold on the approach yet, but I want to be proven wrong and I trust you to do a good execution on this.

MozillaNoah (approved):

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckPreferencesMiddleware.kt
rahulsainani: That would not auto cancel the coroutines when the fragment is destroyed, the idea is to use [lifecycleScope](https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope) so any long running tasks are auto cancelled when the  fragment is destroyed.

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckAction.kt
rahulsainani: sure thing! Happy to learn and iterate as we move forward 👍

fenix/app/src/test/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckStoreTest.kt
```diff
@@ -0,0 +1,134 @@
...
+            tested.waitUntilIdle()
+
+            val expected = ReviewQualityCheckState.OptedIn(productRecommendationsPreference = false)
+            assertEquals(expected, tested.state)
+        }
+}
+
+private class FakeReviewQualityCheckPreferences(
+    private val isEnabled: Boolean = false,
+    private val isProductRecommendationsEnabled: Boolean = false,
```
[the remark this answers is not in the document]
rahulsainani: If we want to add more implementation based tests for this middleware, where we want to verify that the call was made, we can add that in that middleware test. For the behavior based Store test, where we are asserting the state, we don't need these.

PR #2910, opened 2023-07-19
Bug 1844317 - Target SDK 34 in Fenix

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1844317

### DIFF
--- a/android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
+++ b/android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
@@ -7,6 +7,7 @@ package mozilla.components.browser.toolbar.edit
 import android.content.Context
 import android.graphics.Typeface
 import android.graphics.drawable.Drawable
+import android.os.Build
 import android.view.KeyEvent
 import android.view.View
 import android.widget.ImageView
@@ -215,6 +216,10 @@ class EditToolbar internal constructor(
     fun focus() {
         views.url.run {
             if (!hasFocus()) {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+                    // On Android 14 this needs to be called before requestFocus() in order to receive focus.
+                    isFocusableInTouchMode = true
+                }
                 requestFocus()
             }
         }
--- a/android-components/.config.yml
+++ b/android-components/.config.yml
@@ -4,4 +4,4 @@ componentsGroupId: "org.mozilla.components"
 # Synchronized build configuration for all modules
 compileSdkVersion: 34
 minSdkVersion: 21
-targetSdkVersion: 33
+targetSdkVersion: 34
--- a/focus-android/app/build.gradle
+++ b/focus-android/app/build.gradle
@@ -24,7 +24,7 @@ android {
         applicationId "org.mozilla"
         minSdkVersion config.minSdkVersion
         compileSdk config.compileSdkVersion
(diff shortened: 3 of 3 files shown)


mcarare:
Reminder to also test https://github.com/mozilla-mobile/reference-browser/issues/2522.

Amejia481 (approved):

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
```diff
@@ -215,6 +215,7 @@ class EditToolbar internal constructor(
     fun focus() {
         views.url.run {
             if (!hasFocus()) {
+                isFocusableInTouchMode = true
```
Amejia481, line 218: Maybe we should add this change only on Android 14 and up to avoid any possible backward issues.

jonalmeida: This seems suspiciously close to a change that landed here not too long ago, so might be a red herring for this PR.

 @rocketsroger might have some context here while I am AFK to investigate this further.

mcarare: Yes, this is the change that causes the Focus issue. Adding back `showKeyboard` fixes the Focus issue with the keyboard not showing up when the toolbar is focused.

Amejia481: Should we land in file follow up ticket?

Amejia481: ideally we would like to get early feedback as soon as possible.

rocketsroger: This seems like a workaround for something that shouldn't happen.  Originally the show keyboard call was only for Fire TV, which device are you seeing this issue?  Also I think it shouldn't piggyback on this PR.  If there is an issue we should fix it separately (maybe with some investigation involved)

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
```diff
@@ -215,6 +217,10 @@ class EditToolbar internal constructor(
     fun focus() {
         views.url.run {
             if (!hasFocus()) {
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
```
rocketsroger, line 220: Please create an issue for this and separate this from Bug 1844317.  My thinking is if we need a fix here for devices that has focus but doesn't show keyboard, the better fix will only request show keyboard when we 1. have focus, 2. keyboard not visible. Let's not piggy back on this unrelated Bug's PR.

rocketsroger (changes requested):

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/edit/EditToolbar.kt
mcarare: It's not unrelated, tests fail without setting the isFocusableInTouchMode to true after targeting Android 14.

rocketsroger (approved):
🚢

PR #2973, opened 2023-07-24
Bug 1842245 - Refactor history fragment item click handlers to lib-state

MatthewTighe:
This is the first of the history fragment Redux refactors. It includes two commits - the first was just a simple feature flag introduction which I wanted to have a separate ticket in order to track better, but didn't feel worth a separate review cycle. 

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1842246
https://bugzilla.mozilla.org/show_bug.cgi?id=1842245


fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
```diff
@@ -92,12 +94,18 @@ class HistoryFragment : LibraryPageFragment<History>(), UserInteractionHandler,
...
+                middleware = listOf(
+                    HistoryNavigationMiddleware(
+                        navController = findNavController(),
+                        openToBrowser = ::openItem,
+                    ),
+                    HistoryTelemetryMiddleware(
+                        isInPrivateMode = {
+                            (activity as HomeActivity)
+                                .browsingModeManager.mode == BrowsingMode.Private
+                        },
```
MozillaNoah, line 107: you might be able to get away with this being a static boolean vs a lambda, since the whole activity is recreated anyway when toggling between normal/private. Though there also isn't a toggle to do so in the History flow anyway (right?).

fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistoryNavigationMiddleware.kt
```diff
@@ -0,0 +1,63 @@
...
+        next: (HistoryFragmentAction) -> Unit,
+        action: HistoryFragmentAction,
+    ) {
+        next(action)
+        scope.launch {
+            when (action) {
+                is HistoryFragmentAction.HistoryItemClicked -> {
+                    if (context.state.mode.selectedItems.isEmpty()) {
+                        when (val item = action.item) {
+                            is History.Regular -> openToBrowser(item)
```
MozillaNoah, line 43: rather than a lambda, should this be firing off a new Action instead for the telemetry and `openToBrowserAndLoad` logic?

fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragmentStore.kt
```diff
@@ -110,8 +111,11 @@ fun HistoryMetadata.toHistoryMetadata(position: Int): History.Metadata {
 /**
  * The [Store] for holding the [HistoryFragmentState] and applying [HistoryFragmentAction]s.
  */
-class HistoryFragmentStore(initialState: HistoryFragmentState) :
-    Store<HistoryFragmentState, HistoryFragmentAction>(initialState, ::historyStateReducer)
+class HistoryFragmentStore(
```
MozillaNoah, line 114: Should this eventually be renamed to `HistoryStore`?

fenix/app/src/test/java/org/mozilla/fenix/library/history/state/HistoryNavigationMiddlewareTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        )
+        val store =
+            HistoryFragmentStore(HistoryFragmentState.initial, middleware = listOf(middleware))
+
+        store.dispatch(HistoryFragmentAction.HistoryItemClicked(history)).joinBlocking()
+
+        verify(navController).navigate(
+            directions = any(),
+            navOptions = any(),
+        )
```
MozillaNoah, line 65: based on your test title, should this be actually verifying that its navigating to the metadata fragment?

fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragmentStore.kt
MatthewTighe: Potentially! There are other areas within the History feature - there is an `HistorySearchFragmentStore` and the `HistoryMetadataFragment` will need a Store eventually as well, for example. I would probably hold off on renaming until those could be all consolidated, if ever.

fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistoryNavigationMiddleware.kt
MatthewTighe: I do like the idea of firing an Action to the Browser Store to handle the page opening logic. I would prefer to do this in a follow-up though and continue to re-use the logic as it is still shared with the controller until the flag is removed. How does that sound?

The telemetry is already captured by the telemetry middleware, though I'm now realizing that means it will be recorded twice and so the flag should probably be used to prevent that.

fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
MatthewTighe: Ah, that's a good point! I'll change the lambda to just run during instantiation instead

fenix/app/src/test/java/org/mozilla/fenix/library/history/state/HistoryNavigationMiddlewareTest.kt
MatthewTighe: I ran into some issues using mockito to verify the right call with this, see [thread](https://mozilla.slack.com/archives/G01EXNFFAVC/p1689094086224469?thread_ts=1689093900.869579&cid=G01EXNFFAVC) for details. I didn't want to spin my wheels too hard on it now, but I could file a follow-up if you'd like. IMO this method is still preferable to capturing the `navController.navigate` call in a lamdba, passing that through to the middleware, and asserting that the lambda is called under test. This way the test at least captures that the navController is being interacted with, while the lambda solution would not. 

I can change the test name if it seems to mismatch, though!

MozillaNoah: I was more inquiring about the usage of `any`, but I'm cool with your approach since you've already tried to do it the "right" way, and I agree that doing the lambda check doesn't get us much more compared to this.

fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
```diff
@@ -92,12 +95,16 @@ class HistoryFragment : LibraryPageFragment<History>(), UserInteractionHandler,
...
-                    isDeletingItems = false,
+                initialState = HistoryFragmentState.initial,
+                middleware = listOf(
+                    HistoryNavigationMiddleware(
+                        navController = findNavController(),
+                        openToBrowser = ::openItem,
+                    ),
+                    HistoryTelemetryMiddleware(
+                        isInPrivateMode = (activity as HomeActivity)
+                            .browsingModeManager.mode == BrowsingMode.Private,
```
MozillaNoah, line 106: if you're leery about using the activity still, the mode is also available in the `AppStore`

MozillaNoah (approved):
LGTM :shipit:

fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistoryNavigationMiddleware.kt
MatthewTighe: https://bugzilla.mozilla.org/show_bug.cgi?id=1845409

PR #2978, opened 2023-07-24
Bug 1830348 - Use do not intercept cache in app links interceptor

rocketsroger:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1830348


android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksUseCases.kt
```diff
@@ -295,11 +295,46 @@ class AppLinksUseCases(
...
+            val currentTimeStamp = SystemClock.elapsedRealtime()
+
+            if (cacheTimeStamp != null &&
+                currentTimeStamp <= cacheTimeStamp + APP_LINKS_DO_NOT_OPEN_CACHE_INTERVAL
+            ) {
+                return true
+            }
+
+            return false
+        }
```
t-p-white, line 329: drive-by - could this be something like:
```suggestion
return cacheTimeStamp != null && currentTimeStamp <= (cacheTimeStamp + APP_LINKS_DO_NOT_OPEN_CACHE_INTERVAL)
```

rocketsroger: Good catch.  Updated, thanks.

android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksUseCases.kt
```diff
@@ -295,11 +295,41 @@ class AppLinksUseCases(
...
+            return cacheTimeStamp != null &&
+                currentTimeStamp <= (cacheTimeStamp + APP_LINKS_DO_NOT_OPEN_CACHE_INTERVAL)
+        }
+
+        internal fun addUserDoNotIntercept(url: String, appIntent: Intent?) {
+            val cacheKey = getCacheKey(url, appIntent)
+            cacheKey?.let {
+                userDoNotInterceptCache[it] = SystemClock.elapsedRealtime()
+            }
+        }
```
titooan, line 331: I think those functions can live directly into AppLinksUseCases instead of the Companion object, and be called from `AppLinksFeatures` and `AppLinksInterceptor` using the instance of `appLinksUseCases` they already have. What do you think?

rocketsroger: Unfortunately `userDoNotInterceptCache` needs to be in companion.  Since `AppLinksFeature` and `AppLinksInterceptor` doesn't use the same instance of `AppLinksUseCases`.  So moving these functions outside of the companion might be confusing, since we don't need an instance of `AppLinksUseCases` to access the cache.  We should allow the user to access it without creating an instance.

titooan (approved):
LGTM!

titooan (approved):

android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksUseCases.kt
rocketsroger: As discussed, I've moved the cache into `AppLinksInterceptor` . Thanks giving me this idea!

PR #3000, opened 2023-07-26
Bug 1845530 - Update AndroidX Paging to version 3.2.0

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1845530


rvandermeulen:
FYI, version 3.2.0 was released today.
https://developer.android.com/jetpack/androidx/releases/paging#3.2.0

android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
```diff
@@ -77,7 +77,7 @@ object Versions {
         const val espresso = "3.5.1"
         const val room = "2.4.3"
         const val savedstate = "1.2.1"
-        const val paging = "2.1.2"
+        const val paging = "3.1.1"
```
rvandermeulen, line 80: Can we remove the duplication from FenixDependencies now too?

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/permission/GeckoSitePermissionsStorage.kt
```diff
@@ -105,7 +105,7 @@ class GeckoSitePermissionsStorage(
 
         return onDiskStorage.getSitePermissionsPaged().map { onDiskPermission ->
             val geckoPermissions = geckoPermissionsByHost[onDiskPermission.origin].groupByType()
-            mergePermissions(onDiskPermission, geckoPermissions)
+            mergePermissions(onDiskPermission, geckoPermissions) ?: onDiskPermission
```
mcarare, line 108: @Amejia481 The alternative here was to use the not-null assertion operator. Is this mapping still correct?

Amejia481: Could you help to clarify why do we need to provide a not-null value?

mcarare: The signature for the map method in `DataSource` class has changed in the new library version, and the code was migrated to Kotlin. It now expects a non-null result for mapping function, and the mergePermissions method returns nullable `SitePermissions`

Amejia481: Yes, I think it's safe to add the Elvis operator!

rvandermeulen (approved):
Dependencies changes look good to me. I'll defer to Arturo for the code changes, however 😄

Amejia481 (approved):

PR #3012, opened 2023-07-27
Bug 1842249 - Refactor HistoryFragment sync handlers to lib-state

MatthewTighe:
It should be noted that the view often won't actually update after a Sync, but this is not a regression over the current main branch.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1842249


fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistorySyncMiddleware.kt
```diff
@@ -0,0 +1,51 @@
...
+        next(action)
+        when (action) {
+            is HistoryFragmentAction.StartSync -> {
+                scope.launch {
+                    accountManager.syncNow(
+                        reason = SyncReason.User,
+                        debounce = true,
+                        customEngineSubset = listOf(SyncEngine.History),
+                    )
+                    refreshView()
```
MozillaNoah, line 44: a lambda feels a little clunky here to update the view when we're firing an action immediately after it. should this just be a side effect of `HistoryFragmentAction.FinishSync`?

MozillaNoah (commented):
LGTM other than one question about the use of a lambda to update the UI via a middleware.

fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistorySyncMiddleware.kt
MatthewTighe: How do you mean? The View will observe State updates, not Actions being fired. It could observe `HistoryFragmentState.Mode` changes, but then would be triggered switching from `Mode.Editing` to `Mode.Normal` as well. The View could potentially track and compare the new mode to the old, but that feels a little clunky to me as well.

Alternatively, we could introduce an `isSyncing` property to the State? IMO it may end a fairly common pattern to define callbacks for middleware, since the Android framework tends to have a lot of control over interactions like navigation, showing dialogs, etc.

MatthewTighe: For a couple concrete examples, here are a couple more lambdas that were going to be incorporated into the `HistoryStorageMiddleware` in the next patch:

```
class HistoryStorageMiddleware(
    private val appStore: AppStore,
    private val browserStore: BrowserStore,
    private val historyProvider: DefaultPagedHistoryProvider,
    private val historyStorage: PlacesHistoryStorage,
    private val deleteSnackbar: (
        items: Set<History>,
        undo: suspend (Set<History>) -> Unit,
        delete: (HistoryFragmentStore, Set<History>) -> suspend (context: Context) -> Unit,
    ) -> Unit,
    private val onTimeFrameDeleted: () -> Unit
)
```

deleteSnackbar definition:
```
    private fun deleteSnackbar(
        items: Set<History>,
        undo: suspend (items: Set<History>) -> Unit,
        delete: (HistoryFragmentStore, Set<History>) -> suspend (context: Context) -> Unit,
    ) {
        CoroutineScope(IO).allowUndo(
            requireActivity().getRootView()!!,
            getMultiSelectSnackBarMessage(items),
            getString(R.string.snackbar_deleted_undo),
            {
                undo.invoke(items)
            },
            delete(historyStore, items),
        )
    }
```

onTimeFrameDeleted:
```
    private fun onTimeFrameDeleted() {
        runIfFragmentIsAttached {
            historyView.historyAdapter.refresh()
            showSnackBar(
                binding.root,
                getString(R.string.preferences_delete_browsing_data_snackbar),
            )
        }
    }
```

The undo snackbar could potentially be triggered by observing changes to the `isDeletingItems` or `pendingDeletionItems` properties of the state, but it seems potentially more error-prone. For time frames, we'd need to introduce new state.

MozillaNoah: I guess part of it is my not 100% understanding of how the account manager updates the local data and how the Views pulling it. 

If the order of operations is:
1. Fire the action to initiate the sync
2. Perform the sync
3. Update account manager's data
4. Tell the View to update

I'm more wondering if we can circumvent step 4, and have the View listening for the changes. 

Typing all of this out though, and given how complicated account manager is, that's probably way outside the scope of this PR and your overall redux work. The lambda also isn't too far off from our `use case` pattern, either. 

All this is to say, I'll approve it as-is. History + account manager may just need some TLC, in general.

MozillaNoah (approved):

fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistorySyncMiddleware.kt
MatthewTighe: Big agree on that last point! Actually, it does look like history acts differently than I assumed. I thought we were manually fetching items from storage, but it does look like our [paginated source](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt#75) should act as a collected flow. 

I would expect updates to Storage to propagate through the Paging library automatically, but I'm not entirely sure that's the case. TBH, there is quite a lot of complexity in the interplay between the adapter and the paging library. In the interest of not regressing anything (or continuing to investigate this too deeply for now) I think I will leave the lambda in the middleware, since it's effectively copied from the [one originally passed to the controller](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt#404).

MatthewTighe: I think we could do a lot of work cleaning this area up even further, especially if we convert it to Compose. I'll take some notes for follow-ups to file

fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistorySyncMiddleware.kt
```diff
@@ -0,0 +1,57 @@
...
+        context: MiddlewareContext<HistoryFragmentState, HistoryFragmentAction>,
+        next: (HistoryFragmentAction) -> Unit,
+        action: HistoryFragmentAction,
+    ) {
+        next(action)
+        if (!FeatureFlags.historyFragmentLibStateRefactor) {
+            // StartSync is also fired by the interactor, so we should avoid syncing twice in cases
+            // where it is still being used.
+            return
+        }
```
MozillaNoah, line 41: nit thought: alternatively, you could put the feature flag check in the store's middleware intialization

MatthewTighe: Oh, good call!

PR #3013, opened 2023-07-27
Bug 1842638 - Expose GeckoView isProductUrl API

owlishDeveloper:
To check if the current page is a product page

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1842638


android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -602,6 +607,26 @@ abstract class EngineSession(
         NO_DETECTED,
     }
 
+    /**
+     * Represents a shopping product status of currently displayed page.
+     */
+    enum class ProductUrlStatus {
```
owlishDeveloper, line 613: Note: At the moment I am not sure if I am being overly dramatic here with the enum, but there probably will be an option to disable the feature down the road, so I decided to add this just in case. It's easily removable ;)

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -602,6 +607,26 @@ abstract class EngineSession(
...
+        /**
+         * Indicates that the tab is currently displaying a product page
+         */
+        PRODUCT_URL,
+
+        /**
+         * Indicates that we are currently displaying a non-product page
```
calumozilla, line 620: `Indicates that the tab is currently displaying a non-product page`

docs/changelog.md
```diff
@@ -67,6 +67,11 @@ permalink: /changelog/
     * Rename mozac_ic_cookies to mozac_ic_cookies_24
     * Rename mozac_ic_sync to mozac_ic_sync_24
 
+# 117.0
+
+* **concept-engine**
+  * 🌟️️ Add `ProductUrlStatus` to `SessionState` instance to indicate the status of the given session state see more on [bug #1842638](https://bugzilla.mozilla.org/show_bug.cgi?id=1842638).
```
calumozilla, line 73: Could be more specific: `indicate whether or not a product page of a given session state is being displayed`

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
owlishDeveloper: Changed this to Boolean. I think the status of the feature and the status of the page are separate

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -1033,6 +1035,10 @@ class GeckoEngineSession(
             notifyObservers { onCookieBannerChange(CookieBannerHandlingStatus.HANDLED) }
         }
 
+        override fun onProductUrl(session: GeckoSession) {
```
rahulsainani, line 1038: For my own understanding, can this function actually get the boolean `isProductUrl` and that's passed on to `onProductUrlChange`. As we are not doing anything with the session, we might not need to pass that and instead just pass the boolean. As a result we could also remove the reseting of the status on lines 710-711, wdyt?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -860,6 +860,18 @@ sealed class CookieBannerAction : BrowserAction() {
         CookieBannerAction()
 }
 
+/**
+ * [BrowserAction] implementations related to updating the [SessionState.ShoppingProduct] of a single [SessionState] inside
+ * [BrowserState]
+ */
+sealed class ShoppingProductAction : BrowserAction() {
```
rahulsainani, line 867: Could this be internal?

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
owlishDeveloper: This is the GV API. The way it works, this is part of a `ContentDelegate` interface. GV calls this function once a product URL has been detected. The way to use the API is you override the interface method (as I am doing here) and you write your code, whatever you want to happen when a product URL is encountered.

We are passing in the GeckoSession to make it available to people in case they need it (we have multiple APIs like this, see `onCookieBannerHandled` API right above). If you don't need it - it's fine, you don't have to use it. ("You" as in "the consumer of GV API", so in this case the AC code)

owlishDeveloper: This is the GV API. The way it works, this is part of a `ContentDelegate` interface. GV calls this function once a product URL has been detected. The way to use the API is you override the interface method (as I am doing here) and you write your code, whatever you want to happen when a product URL is encountered.

We are passing in the GeckoSession to make it available to people in case they need it (we have multiple APIs like this, see `onCookieBannerHandled` API right above). If you don't need it - it's fine, you don't have to use it. ("You" as in "the consumer of GV API", so in this case the AC code)

owlishDeveloper: (to add to the explanation, GV is a general Android library, it's not Fenix-specific or AC-specific)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
owlishDeveloper: None of the other actions here are. Why do you suggest it should be internal?

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
```diff
@@ -294,6 +294,7 @@ class GeckoEngineSessionTest {
         var observedCanGoBack = false
         var observedCanGoForward = false
         var cookieBanner = CookieBannerHandlingStatus.HANDLED
+        var displaysProduct = false
```
calumozilla, line 297: Would this be better left as isProductUrl instead of making another variable name

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/EngineSession.kt
```diff
@@ -65,6 +65,11 @@ abstract class EngineSession(
         fun onTrackerLoaded(tracker: Tracker) = Unit
         fun onNavigateBack() = Unit
 
+        /**
+         * Event to indicate a product URI is currently open.
```
calumozilla, line 69: url*

android-components/components/browser/engine-gecko/src/test/java/mozilla/components/browser/engine/gecko/GeckoEngineSessionTest.kt
owlishDeveloper: Now that I see the automation failure I remembered that the reason for the rename was that otherwise it confuses this variable with the parameter name from the function. This is just a test, it's ok to have a different name. Reverting this

calumozilla (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
rahulsainani: As it's only used in the `browser/state` module, it can be internal and we don't need to expose this to Fenix or other consumers.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
rahulsainani: I understand that.

What I'm saying is if the GV API was something like this:

`fun onProductUrl(isProductUrl: Boolean)`

So we won't need to reset as we are doing above and the api returns what we needs, not `GeckoSession`.

This is an idea and with/without it doesn't impact the consumer of AC, happy to discuss more on slack.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
owlishDeveloper: Rahul, I followed the convention already established and wrote this class in exactly the same way as other classes are written. I think if the team that takes care of this repository now wants to change the conventions - that needs to be a separate bug. This is out of scope of this patch.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
owlishDeveloper: Sure, we can always discuss on Slack! I am resolving this for now.

PR #3018, opened 2023-07-27
Bug 1845825 - Update AndroidX Fragment to version 1.6.1

rvandermeulen:
https://bugzilla.mozilla.org/show_bug.cgi?id=1845825
https://bugzilla.mozilla.org/show_bug.cgi?id=1845825
https://bugzilla.mozilla.org/show_bug.cgi?id=1845825
https://bugzilla.mozilla.org/show_bug.cgi?id=1845825
https://bugzilla.mozilla.org/show_bug.cgi?id=1845825
https://bugzilla.mozilla.org/show_bug.cgi?id=1845825


mcarare:
Looks like we need to wait for ViewPager2's 1.1.0 release:

>Users of ViewPager2 must update to at least 1.1.0-beta02 when updating to this version of RecyclerView to avoid crashes.

mcarare (changes requested):
either we add it as a direct dependency when released or wait for the fragment library to update the dependency in a future release.

rvandermeulen:
Blech, I overlooked that because we use that implicitly. I'll drop the Recyclerview update for now and spin that off to a new bug.

mcarare (approved):

PR #3021, opened 2023-07-28
Bug 1845905 - Add a utility to create store lazily

rahulsainani:
### What
Similar to how `by viewModels()` work, this adds a utility to create stores in `ViewModelStoreOwner`s like `Fragment` or `Activity`.

### Why
– To not use late init var to have the property in a bad state early on with a promise that it will be created later.
– To make sure that `StoreProvider` is used and the `Store` will survive the configuration change.
– To prevent accidental use of `LazyThreadSafetyMode.SYNCHRONIZED` by default.

`lazy` uses `LazyThreadSafetyMode.SYNCHRONIZED` thread safety mode by default. While that may be fine for singleton (app scoped) stores, but for fragment scoped stores, as we're sure that the initialization (and access) will always happen in the same (main) thread, as the one where you use the property, we can use `LazyThreadSafetyMode.NONE`. It doesn't incur any thread-safety guarantees and related overhead.

### Usage 
From:
```
private val store by lazy(LazyThreadSafetyMode.NONE) {
    StoreProvider.get(this) {
        MyStore()
    }
}
```

The syntax gets simplified to:

```
val myStore by store { MyStore() }
```

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1845905


fenix/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt
```diff
@@ -42,3 +45,37 @@ class StoreProviderFactory<T : Store<*, *>>(
         return StoreProvider(createStore()) as VM
     }
 }
+
+/**
+ * Helper function for creation of [StoreLazy].
+ */
+@MainThread
+fun <T : Store<*, *>> ViewModelStoreOwner.store(
```
MozillaNoah, line 53: thoughts?
```suggestion
fun <T : Store<*, *>> ViewModelStoreOwner.lazyStore(
```

MozillaNoah (approved):
this looks great! just left one question about the naming

fenix/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt
rahulsainani: Yeah makes sense, update to this and simplified the implementation.

jonalmeida:
@rahulsainani @MozillaNoah hey folks this is a foundational change with how we will use the store that flew under the radar:

1. Let's not land a change like this without getting more eyes on it. Please feel free to cc other team members too that have worked extensively on the code base at this level.
2. We've landed utility code and not any usage of it. The closest I've seen is [this PR](https://github.com/mozilla-mobile/firefox-android/pull/3153/) but no usage of it. Adding things that we don't have clear understanding how they will be used is not great.

rahulsainani:
As discussed on the call about the need & utility for these changes, I'll create the parent bug with the story about these 2 changes. Appreciate your perspective. We can surely try to get more eyes on patches like this in the future.

PR #3031, opened 2023-07-28
Bug 1844967 - Improve performance of tab thumbnail loading in Compose

MozillaNoah:
Original performance work: [Bug 1815579](https://bugzilla.mozilla.org/show_bug.cgi?id=1815579) 

## Discoveries
- `ThumbnailStorage` was being fetched for _every_ tab.
- Thumbnail dimensions were getting calculated for _every_ item.
- The grid thumbnail size was being set to the _entire width of the screen_, and therefore fetching a thumbnail of that size for _every_ tab. 

## Remedies
- `ThumbnailStorage` is now being sent from the root Composable down to the leaf nodes.
- The thumbnail dimensions were uplifted to `TabGrid` and `TabList` composables.
- `TabGrid` and `TabList` now use the dimension resources used in `BrowserTabViewHolder` AND they get the pixel dimension directly, rather than getting the `DP` unit type and then converting it (this is what happens under the hood when fetching a `DP` dimension in Compose via `dimensionResouce()`.

# Videos Samsung Galaxy Tab S7+ with ~180 open tabs)

## Before

https://github.com/mozilla-mobile/firefox-android/assets/87384386/1dd7d3af-60bc-46c7-a0b4-7175d076b059

## After

https://github.com/mozilla-mobile/firefox-android/assets/87384386/5c6a335a-4669-4b0a-ad5d-e3ac367d7937

## XML version for comparison (it also struggles with tablet + this many tabs

https://github.com/mozilla-mobile/firefox-android/assets/87384386/c3804e19-3d6f-4612-b673-3150cde0fd1f

# Videos - Pixel 2 with ~90 tabs

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1844967


fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
```diff
@@ -20,40 +20,35 @@ import androidx.compose.ui.Modifier
...
-    size: Dp,
+    request: ImageLoadRequest,
+    storage: ThumbnailStorage?,
     modifier: Modifier,
     contentScale: ContentScale,
     alignment: Alignment,
     fallbackContent: @Composable () -> Unit,
 ) {
-    if (inComposePreview) {
+    if (storage == null) {
```
MatthewTighe, line 49: IMO if this is the only purpose the nullable parameter is serving it would be best to switch back to `inComposePreview`. The nullable parameter communicates that the storage component is optional, which seems like it could lead to unexpected behavior for consumers of the API

fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTray.kt
```diff
@@ -92,6 +94,7 @@ fun TabsTray(
     appStore: AppStore,
     browserStore: BrowserStore,
     tabsTrayStore: TabsTrayStore,
+    storage: ThumbnailStorage?,
```
MatthewTighe, line 97: I think there's an interesting conversation we could have about prop drilling utility classes like this down the whole tree vs fetching them from within it (at an appropriate level to avoid problems like the one you're solving). This ties in pretty closely to our ongoing conversation about state hoisting as well, since this problem would've been avoided if we followed that advice to the letter.

IMO it would still be fine to fetch it from `LocalContext` at a lower level in the tree than this, since it is a stateless utility class. For simplicity and consistency I think we should keep this patch as it is now, but this could be a good example to use in future conversations about hoisting.

MatthewTighe (commented):
Great job figuring these pieces out! I'll cycle back for an approval once this is undrafted. I'm also not at my home office today so I'll report in on Monday in regards to the performance on my low-end reference device

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
MozillaNoah: We discussed this offline, and I am pushing changes to add `inComposePreview` back in, after discovering I can create a dummy `ThumbnailStorage` inside of previews. So the nullability will also go away as a result.

MatthewTighe (approved):
LGTM 😎

PR #3053, opened 2023-07-31
Bug 1840692 - Expose Request Analysis Fakespot API

owlishDeveloper:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1840692

### DIFF
--- a/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
+++ b/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
@@ -20,6 +20,8 @@ import mozilla.components.browser.engine.gecko.media.GeckoMediaDelegate
 import mozilla.components.browser.engine.gecko.mediasession.GeckoMediaSessionDelegate
 import mozilla.components.browser.engine.gecko.permission.GeckoPermissionRequest
 import mozilla.components.browser.engine.gecko.prompt.GeckoPromptDelegate
+import mozilla.components.browser.engine.gecko.shopping.GeckoProductAnalysis
+import mozilla.components.browser.engine.gecko.shopping.Highlight
 import mozilla.components.browser.engine.gecko.window.GeckoWindowRequest
 import mozilla.components.browser.errorpages.ErrorType
 import mozilla.components.concept.engine.EngineSession
@@ -36,6 +38,7 @@ import mozilla.components.concept.engine.manifest.WebAppManifest
 import mozilla.components.concept.engine.manifest.WebAppManifestParser
 import mozilla.components.concept.engine.request.RequestInterceptor
 import mozilla.components.concept.engine.request.RequestInterceptor.InterceptionResponse
+import mozilla.components.concept.engine.shopping.ProductAnalysis
 import mozilla.components.concept.engine.window.WindowRequest
 import mozilla.components.concept.fetch.Headers.Names.CONTENT_DISPOSITION
 import mozilla.components.concept.fetch.Headers.Names.CONTENT_LENGTH
@@ -645,6 +648,70 @@ class GeckoEngineSession(
                 GeckoResult()
             },
         )
+    }
+
+    /**
+     * See [EngineSession.requestProductAnalysis]
+     */
+    @Suppress("ComplexCondition")
+    override fun requestProductAnalysis(
+        url: String,
+        onResult: (ProductAnalysis) -> Unit,
+        onException: (Throwable) -> Unit,
+    ) {
+        geckoSession.requestAnalysis(url).then({
+                response ->
+            if (response == null) {
+                logger.error(
+                    "Invalid value: unable to get analysis result from Gecko Engine.",
+                )
+                onException(
+                    java.lang.IllegalStateException(
+                        "Invalid value: unable to get analysis result from Gecko Engine.",
+                    ),
+                )
+                return@then GeckoResult()
+            }
+
+            val highlights: Highlight?
+            if (
+                response.highlights?.quality == null &&
+                response.highlights?.price == null &&
+                response.highlights?.shipping == null &&
+                response.highlights?.appearance == null &&
+                response.highlights?.competitiveness == null
+            ) {
+                highlights = null
+            } else {
+                highlights = Highlight(
+                    response.highlights?.quality?.toList(),
+                    response.highlights?.price?.toList(),
+                    response.highlights?.shipping?.toList(),
+                    response.highlights?.appearance?.toList(),
+                    response.highlights?.competitiveness?.toList(),
+                )
+            }
+
+            val analysisResult = GeckoProductAnalysis(
+                response.productId,
+                response.analysisURL,
+                response.grade,
+                response.adjustedRating,
+                response.needsAnalysis,
+                response.lastAnalysisTime,
+                response.deletedProductReported,
+                response.deletedProduct,
+                highlights,
+            )
+
+            onResult(analysisResult)
+            GeckoResult<ProductAnalysis>()
+        }, {
+                throwable ->
+            logger.error("Requesting product analysis failed.", throwable)
+            onException(throwable)
+            GeckoResult()
+        })
     }
 
     /**
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -10,6 +10,10 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/plugins/dependencies/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/.config.yml)
 
+* **concept-engine**
+  * 🌟️️ Add `ProductUrlStatus` to `SessionState` instance to indicate whether or not a product page of a given session state is being displayed. See more on [bug #1842638](https://bugzilla.mozilla.org/show_bug.cgi?id=1842638).
+  * 🌟️️ Add `requestProductAnalysis` API to `EngineSession` to allow request product analysis result from the engine. See more on [bug #1840692](https://bugzilla.mozilla.org/show_bug.cgi?id=1840692).
+
 * **feature-addons**
   * ⚠️ **This is a breaking change**: the method `getAvailableAddons()` in `AddonsProvider` has been renamed to `getFeaturedAddons()`.
   * ⚠️ **This is a breaking change**: the `AddonCollectionProvider` has been renamed to `AMOAddonsProvider`.
@@ -76,11 +80,6 @@ permalink: /changelog/
     * Rename mozac_ic_add_to_homescreen to mozac_ic_add_to_homescreen_24
     * Rename mozac_ic_cookies to mozac_ic_cookies_24
     * Rename mozac_ic_sync to mozac_ic_sync_24
-
-# 117.0
-
-* **concept-engine**
-  * 🌟️️ Add `ProductUrlStatus` to `SessionState` instance to indicate whether or not a product page of a given session state is being displayed. See more on [bug #1842638](https://bugzilla.mozilla.org/show_bug.cgi?id=1842638).
 
 # 116.0
 * [Commits](https://github.com/mozilla-mobile/firefox-android/compare/releases_v115..releases_v116)
--- a/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoProductAnalysis.kt
+++ b/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoProductAnalysis.kt
@@ -1,0 +1,49 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.engine.gecko.shopping
+
+import mozilla.components.concept.engine.shopping.ProductAnalysis
+
+/**
+ * Holds the result of the analysis of a shopping product.
+ *
+ * @property productId Product identifier (ASIN/SKU)
+ * @property analysisURL Analysis URL
+ * @property grade Reliability grade for the product's reviews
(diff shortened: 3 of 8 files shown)


android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysisResult.kt
```diff
@@ -0,0 +1,49 @@
...
+ * @property needsAnalysis Boolean indicating if the analysis is stale
+ * @property lastAnalysisTime Time since the last analysis was performed
+ * @property deletedProductReported Boolean indicating if reported that this product has been deleted
+ * @property deletedProduct Boolean indicating if this product is now deleted
+ * @property highlights Object containing highlights for product
+ */
+data class GeckoShoppingProductAnalysisResult(
+    override val productId: String,
+    val analysisURL: String,
+    val grade: String,
```
owlishDeveloper, line 25: to do: this probably needs to be an enum

owlishDeveloper:
Automated tests are coming after we have the test endpoints (the desktop team works on the patch)

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysisResult.kt
owlishDeveloper: update: no, it's better not, as we are not sure if the API is not going to change

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysisResult.kt
```diff
@@ -0,0 +1,49 @@
...
+    val adjustedRating: Float,
```
calumozilla, line 26: This is still a double in GV, do you want to change it here or in GV?

owlishDeveloper: It is still double, and should be double here, I'll correct this

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/shopping/ShoppingProductAnalysisResult.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Holds the result of the analysis of a shopping product.
+ */
+interface ShoppingProductAnalysisResult {
+
+    /**
+     * Product identifier (ASIN/SKU)
+     */
+    val productId: String
```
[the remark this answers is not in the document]
owlishDeveloper: This is the file in the `concept/engine`, so this is not the GV-specific area of the code. This is a generic interface, which you are supposed to implement if you are adding this functionality to a specific engine. To make this as generic as possible I left only the ID field which seems like the minimum essential thing.

In the GV-specific area I am implementing this interface and adding all our fields (it's the `engine/gecko/shopping/GeckoShoppingProductAnalysisResult.kt` file)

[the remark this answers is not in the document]
owlishDeveloper: My understanding is, we cannot simply use the GeckoView type, because this type should be present in the method 

```
abstract fun requestProductAnalysis(
  url: String, 
  onResult: (ShoppingProductAnalysisResult) -> Unit, 
  onException: (Throwable) -> Unit)

```

that we are defining in the concept Engine. The concept Engine is a general concept, so it cannot use Gecko-specific types, therefore, we need a concept type. I made it an concept interface which you are supposed to implement for your specific engine.

(I think this is how things work in AC)

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysis.kt
```diff
@@ -0,0 +1,49 @@
...
+ * Contains information about highlights of a product's reviews.
+ *
+ * @property quality Highlights about the quality of a product
+ * @property price Highlights about the price of a product
+ * @property shipping Highlights about the shipping of a product
+ * @property appearance Highlights about the appearance of a product
+ * @property competitiveness Highlights about the competitiveness of a product
+ */
+data class Highlight(
+    val quality: Array<String>?,
```
rahulsainani, line 44: think we can use `List<String>?` here

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysis.kt
```diff
@@ -0,0 +1,49 @@
...
+ * @property lastAnalysisTime Time since the last analysis was performed
+ * @property deletedProductReported Boolean indicating if reported that this product has been deleted
+ * @property deletedProduct Boolean indicating if this product is now deleted
+ * @property highlights Object containing highlights for product
+ */
+data class GeckoShoppingProductAnalysis(
+    override val productId: String,
+    val analysisURL: String?,
+    val grade: String?,
+    val adjustedRating: Double,
```
rahulsainani, line 26: Can this be null as well?

Same question for `lastAnalysisTime` and both `deletedX` properties

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysis.kt
```diff
@@ -0,0 +1,49 @@
...
+ * @property needsAnalysis Boolean indicating if the analysis is stale
+ * @property lastAnalysisTime Time since the last analysis was performed
+ * @property deletedProductReported Boolean indicating if reported that this product has been deleted
+ * @property deletedProduct Boolean indicating if this product is now deleted
+ * @property highlights Object containing highlights for product
+ */
+data class GeckoShoppingProductAnalysis(
+    override val productId: String,
+    val analysisURL: String?,
+    val grade: String?,
```
rahulsainani, line 25: think we should use an enum here

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -647,6 +650,56 @@ class GeckoEngineSession(
...
+                    "Invalid value: unable to get analysis result from Gecko Engine.",
+                )
+                onException(
+                    java.lang.IllegalStateException(
+                        "Invalid value: unable to get analysis result from Gecko Engine.",
+                    ),
+                )
+                return@then GeckoResult()
+            }
+            val highlights = Highlight(
```
rahulsainani, line 674: It would be nice to have a check if all the highlights are null, then this `highlights` is set to null and not 
Highlight(null, null, null, null, null).

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
```diff
@@ -647,6 +650,56 @@ class GeckoEngineSession(
...
+                response.highlights?.quality,
+                response.highlights?.price,
+                response.highlights?.shipping,
+                response.highlights?.appearance,
+                response.highlights?.competitiveness,
+            )
+            val analysisResult = GeckoShoppingProductAnalysis(
+                response.productId ?: "",
```
rahulsainani, line 682: If `productId` is null, we shouldn't return an empty string here. I think this case should correspond to returning a different object - this seems like a case where needsAnalysis would be true. Check [this doc section](https://docs.google.com/document/d/1eppEnQKrSKavM3qmlm36WG7y8o2ggm-oxugnoSr2tpw/edit#heading=h.ff818tiik47) for reference.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysis.kt
owlishDeveloper: In Java `Double` is a primitive value and is not nullable. GV is written in Java, so you will not get a `null` from GV.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysis.kt
owlishDeveloper: oh, I did not update you on the enum story.

So initially I agreed with you that since they use a limited set of values (American school grades), this can totally be an enum. But then I decided to make sure, just in case - I asked Jared if it's possible that their API is going to change in the future. His answer indicated that there is, in fact, such possibility (although not in MVP, but still).

So I decided that the initial design with String is more robust. Because if they decide to change the set of values, or make then numbers etc - we would have to change throughout the whole stack, and that would involve deprecating an API - which takes 2 release cycles in GV. I would much rather have this more vague design that doesn't have the potential to generate tech debt down the road.

When using this API in Fenix or Focus, you can make this an enum of `{"A", "B", "C", "D", "F"}` - but again, keep in mind the API is not stable yet, and we are not sure if any other characters (such as "?") would be or are used, etc.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysis.kt
owlishDeveloper: Sorry, I was wrong above - `Double` in Java is an object (not the primitive `double`), but in GV when we get it from the Gecko's response - we check for null, and substitute `0.0` instead of `null`. So the prop is annotated `@NonNull` in GV code. You cannot get `null` here

owlishDeveloper: (but the `lastAnalysisTime` and the `deleted` are indeed primitive types, so not nullable)

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
owlishDeveloper: I'll make this field nullable then.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
owlishDeveloper: Is there a nice idiomatic Kotlin way of doing that? When I do the simple thing (check properties sequentially), the linter is not happy.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysis.kt
rahulsainani: I think this would lead to double checking across the layers, GV is checking if `null` and then setting `0.0` and the AC is just passing it through and then Fenix UI layer does the same check but for `0.0`, lots of translations and it can be simplified with having it nullable. I think we should revisit that and have a discussion about it. 

By just keeping it null, it simplifies and we don't get into a state where we check for `0.0` or empty string ever.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/shopping/GeckoShoppingProductAnalysis.kt
owlishDeveloper: It's a tech debt because we would have to spend time updating GV API (writing a new one, deprecating old one, removing dead code), and AC. This time could be spent more productively. I think an enum in Fenix layer is fine - in that case there would be only one layer to change should the API change.

>  It should never return ? but null in that case IMO.

Hey, I did not design that API :) I'm with you on this one :) but it is what it is. I am referring to the Gecko API. (not GeckoView API)

docs/changelog.md
```diff
@@ -10,6 +10,10 @@ permalink: /changelog/
 * [Gecko](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/plugins/dependencies/src/main/java/Gecko.kt)
 * [Configuration](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/.config.yml)
 
+* **concept-engine**
+  * 🌟️️ Add `ProductUrlStatus` to `SessionState` instance to indicate whether or not a product page of a given session state is being displayed. See more on [bug #1842638](https://bugzilla.mozilla.org/show_bug.cgi?id=1842638).
```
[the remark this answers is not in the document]
owlishDeveloper: oh, this one is completely wrong, thank you for catching this!

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngineSession.kt
owlishDeveloper: As we discussed in Slack, I will make this a simple `if else` and check every prop sequentially, and suppress the lint warning.

docs/changelog.md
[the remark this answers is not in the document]
owlishDeveloper: ah, wait. This is a correction from previous patch - I put this in the wrong section of the document, that is why I am moving it to the correct place (which is the v. 118 section.). In any case, thank you for paying attention!

calumozilla (approved):

PR #3056, opened 2023-07-31
Bug 1838389 - Catch NullPointerException while printStackTrace

rocketsroger:
In some cases `printStackTrace` fails with `NullPointerException`.  
The value is direction retrieved through the throwable and there's nothing we can do about it.

Couple reasons why we should catch this exception:
1. In general crash reporter should avoid crashing.
2. We also attempts to report stack trace through `extractThrowableList`, we should catch the `NullPointerException` so the crash report can still be sent without the stack trace in string.  

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1838389

### DIFF
--- a/android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/MozillaSocorroService.kt
+++ b/android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/MozillaSocorroService.kt
@@ -552,10 +552,15 @@ class MozillaSocorroService(
         return resultMap
     }
 
-    private fun getExceptionStackTrace(throwable: Throwable, isCaughtException: Boolean): String {
-        return when (isCaughtException) {
-            true -> "$LIB_CRASH_INFO_PREFIX ${throwable.getStacktraceAsString()}"
-            false -> throwable.getStacktraceAsString()
+    @Suppress("TooGenericExceptionCaught")
+    private fun getExceptionStackTrace(throwable: Throwable, isCaughtException: Boolean): String? {
+        return try {
+            when (isCaughtException) {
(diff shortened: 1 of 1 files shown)


android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/MozillaSocorroService.kt
```diff
@@ -552,10 +552,15 @@ class MozillaSocorroService(
...
-            false -> throwable.getStacktraceAsString()
+    @Suppress("TooGenericExceptionCaught")
+    private fun getExceptionStackTrace(throwable: Throwable, isCaughtException: Boolean): String? {
+        return try {
+            when (isCaughtException) {
+                true -> "$LIB_CRASH_INFO_PREFIX ${throwable.getStacktraceAsString()}"
+                false -> throwable.getStacktraceAsString()
+            }
+        } catch (e: NullPointerException) {
+            null
```
MozillaNoah, line 563: optional nit: It might make sense to add a comment explaining why we had to catch a "generic" exception and what function call is causing it.

MozillaNoah (approved):

android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/MozillaSocorroService.kt
jonalmeida: +1

Maybe fix-up the phrasing for the first two lines in your PR and in-line that here.

I'm not sure what this sentence means: "The value is direction retrieved through the throwable and there's nothing we can do about it."

PR #3100, opened 2023-08-03
Bug 1840339 - Add review highlights card for shopping experience

rahulsainani:
### How
– Add composable UI
~– Add `highlightsForCompactMode` property that contains the first two highlights of the first highlight type to display in the highlights card in compact state.~
– Add `highlights.forCompactMode()` extension function that returns the first two highlights of the first highlight type to display in the highlights card in compact state.

### Previews
Screenshot 2023-08-03 at 13 58 27
Screenshot 2023-08-03 at 13 58 10

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1840339

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
@@ -4,19 +4,32 @@
 
 package org.mozilla.fenix.shopping.ui
 
+import androidx.compose.animation.Crossfade
+import androidx.compose.animation.animateContentSize
+import androidx.compose.animation.core.spring
+import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.ExperimentalLayoutApi
 import androidx.compose.foundation.layout.FlowRow
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.width
+import androidx.compose.material.Icon
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.unit.dp
@@ -25,7 +38,9 @@ import org.mozilla.fenix.compose.SwitchWithLabel
 import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.compose.button.SecondaryButton
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckState
+import org.mozilla.fenix.shopping.store.ReviewQualityCheckState.HighlightType
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckState.OptedIn.ProductReviewState.AnalysisPresent
+import org.mozilla.fenix.shopping.store.forCompactMode
 import org.mozilla.fenix.theme.FirefoxTheme
 
 /**
@@ -59,6 +74,20 @@ fun ProductAnalysis(
             rating = productAnalysis.adjustedRating,
             modifier = Modifier.fillMaxWidth(),
         )
+
+        if (productAnalysis.highlights != null) {
+            HighlightsCard(
+                highlights = productAnalysis.highlights,
+                modifier = Modifier.fillMaxWidth(),
+            )
+
+            Text(
+                text = stringResource(R.string.review_quality_check_highlights_caption),
+                color = FirefoxTheme.colors.textPrimary,
+                style = FirefoxTheme.typography.caption,
+                modifier = Modifier.padding(horizontal = 16.dp),
+            )
+        }
 
         SettingsCard(
             modifier = Modifier.fillMaxWidth(),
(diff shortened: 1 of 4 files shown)


fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -151,6 +302,24 @@ private fun SettingsCard(
     }
 }
 
+@Composable
+private fun highlightIcon(): Painter {
+    // As part of Bug 1841600, update this to use the correct icon based on the highlight type.
```
t-p-white, line 307: 👍

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckState.kt
```diff
@@ -75,7 +75,21 @@ sealed interface ReviewQualityCheckState : State {
...
-            ) : ProductReviewState
+            ) : ProductReviewState {
+
+                /**
+                 * Highlights to display in compact mode that contains first 2 highlights of the
+                 * first highlight type.
+                 */
+                val highlightsForCompactMode: Map<HighlightType, List<String>>? =
+                    highlights?.entries?.firstOrNull()?.let { entry ->
+                        mapOf(entry.key to entry.value.take(NUMBER_OF_HIGHLIGHTS_FOR_COMPACT_MODE))
```
t-p-white, line 86: out of interest, for the highlights:

1. do we have a preference on how these are ordered, e.g. alphabetically
2. if so, are we expecting the api will have already sorted this?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -121,6 +151,127 @@ private fun AdjustedProductRatingCard(
     }
 }
 
+@Composable
+private fun HighlightsCard(
+    highlights: Map<HighlightType, List<String>>,
+    highlightsForCompactMode: Map<HighlightType, List<String>>,
```
t-p-white, line 157: I wonder if instead of passing the 2 maps as params we just pass the original `highlights` map and have some kind of `getCompactHighlights` fun we can call from here on `highlights` - this would mean we're working from a single source of truth within this func

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -151,6 +302,24 @@ private fun SettingsCard(
...
 }
 
+@Composable
+private fun highlightIcon(): Painter {
+    // As part of Bug 1841600, update this to use the correct icon based on the highlight type.
+    return painterResource(id = R.drawable.ic_shopping_cart)
+}
+
+@Composable
+private fun highlightTitleText(highlightType: HighlightType): String {
```
t-p-white, line 312: The string res could be part of the HighlightType enum which would remove the need for this

fenix/app/src/main/java/org/mozilla/fenix/shopping/state/ReviewQualityCheckState.kt
rahulsainani: Good questions! I will have to get back to you on those. Once we know the order, we should update the fake data to follow the same order though. 📓 

Since this is the UIState, it doesn't matter here but it would, for the mapper that will map the object exposed via AC to this State.

t-p-white (commented):
Looks great! Just a couple of comments from me

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
rahulsainani: Ideally would try to avoid android/R imports in `State`. Maybe we could map them earlier similar to [ReviewGrade](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ReviewGrade.kt#51) and how it's mapped from [Grade](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ReviewGrade.kt#181).  The upcoming icon mapping logic would be the same, so maybe it makes more sense, wdyt?

t-p-white: Sure, makes sense to be consistent with the existing method 👍

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -121,6 +150,129 @@ private fun AdjustedProductRatingCard(
...
+    ReviewQualityCheckCard(modifier = modifier) {
+        var isExpanded by remember { mutableStateOf(false) }
+        val highlightsForCompactMode = remember(highlights) { highlights.forCompactMode() }
+        val highlightsToDisplay = remember(isExpanded, highlights) {
+            if (isExpanded) {
+                highlights
+            } else {
+                highlightsForCompactMode
+            }
+        }
```
gabrielluong, line 167: 👍 Using `remember` to calculate and storing these states for recomposition like this kinda blew my mind. Very cool!

rahulsainani: Nice! it's good to remember for cpu "intensive" operations as we don't want to run them on each composition. I learnt that from this [talk - this section talks about this exactly](https://youtu.be/6BRlI5zfCCk?t=1318), the whole talk is really insightful about how compose recomposes and the data structure it uses internally.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
rahulsainani: ✅

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -151,6 +303,42 @@ private fun SettingsCard(
...
+    when (this) {
+        HighlightType.QUALITY -> Highlight.QUALITY
+        HighlightType.PRICE -> Highlight.PRICE
+        HighlightType.SHIPPING -> Highlight.SHIPPING
+        HighlightType.PACKAGING_AND_APPEARANCE -> Highlight.PACKAGING_AND_APPEARANCE
+        HighlightType.COMPETITIVENESS -> Highlight.COMPETITIVENESS
+    }
+
+// As part of Bug 1841600, update iconResourceId for each highlight type.
+private enum class Highlight(
```
t-p-white, line 316: Appreciate you updating this 👍 non-blocking question, should we call this something like HighlightUiData?

t-p-white (approved):
Looking good! Just the one question, but it's non-blocking

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
rahulsainani: for sure! thanks for the review! 

hmm I think even the `State` objects are ui `data/model/state`. Being a private class for a composable does say that it's a `UIData`, it think maybe we don't need the extra verbosity as it's clear it's a ui data. wdyt?

t-p-white: That's a fair point - happy to keep as-is :shipit:

PR #3140, opened 2023-08-07
Bug 1840580 - Use BrowserState to detect product page for shopping experience

rahulsainani:
### How
– Observe `browserState` in `ReviewQualityCheckFeature` and expose the detection of the product page using the `onAvailabilityChange` callback. 
– Cancel the observation and scope on `stop()`.
– Add `isProductUrl` param in the convenient function `createTab()`

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
```diff
@@ -4,28 +4,48 @@
...
+    private var scope: CoroutineScope? = null
 
     override fun start() {
-        val isFeatureEnabled = FxNimbus.features.shoppingExperience.value().enabled
-        // Update to use product page detector api in Bug 1840580
-        val isSupportedProductPage = false
-        onAvailabilityChange(isFeatureEnabled && isSupportedProductPage)
+        if (shoppingExperienceFeature.isEnabled().not()) {
+            return onAvailabilityChange(false)
+        }
```
t-p-white, line 36: would it be beneficial to call this prior to calling this fun?

t-p-white (approved):
nice work! lgtm 🙌

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
rahulsainani: As discussed offline, this `Feature` is lifecycle aware and is set on a toggle button in `BrowserFragment`. Being the entry point to the feature, it made sense to keep the shopping related logic in one place and not have `BrowserFragment` deal with the shopping feature flag. 

Good catch on the return, it should be below the callback, updated!

fenix/app/src/main/java/org/mozilla/fenix/shopping/ShoppingExperienceFeature.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * An abstraction for shopping experience feature flag.
+ */
+class ShoppingExperienceFeature {
+
+    /**
+     * Returns true if the shopping experience feature is enabled.
+     */
+    fun isEnabled() = FxNimbus.features.shoppingExperience.value().enabled
```
MozillaNoah, line 17: drive by: if you prefer `val` usages over functions, you could instead make this a val with a `get` override

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
```diff
@@ -4,28 +4,49 @@
...
+            return
+        }
+
+        scope = browserStore.flowScoped { flow ->
+            flow.mapNotNull { it.selectedTab }
+                .map { it.isProductUrl }
+                .distinctUntilChanged()
+                .collect {
+                    onAvailabilityChange(it)
+                }
```
MozillaNoah, line 45: drive by shower thought: curious if this works
```suggestion
                .collect(onAvailabilityChange)
```

MozillaNoah (commented):
left some non-blocking drive by thoughts

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
rahulsainani: Nice! 👍

Alexandru2909 (approved):

fenix/app/src/main/java/org/mozilla/fenix/shopping/ShoppingExperienceFeature.kt
rahulsainani: It's fair, we don't need any params and both compile to the same code, updated!

PR #3146, opened 2023-08-07
Bug 1847645 - Changed the Confirm PromptFeature to correctly use the …

zmckenney:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
+++ b/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/PromptFeature.kt
@@ -864,10 +864,10 @@ class PromptFeature private constructor(
                     } else {
                         positiveButtonTitle
                     }
-                    val negativeButton = if (positiveButtonTitle.isEmpty()) {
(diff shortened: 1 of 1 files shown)


gabrielluong (approved):
Do you have an example test page?

zmckenney:
> Do you have an example test page?

We currently don't have any uses where there is a Negative label being used for the button (all uses are default values for labels). There is some testing for the prompt feature but mostly for constructor argument order. We could create a new test for prompt features where we create the prompt and dispatch it, then check the resulting dialog for the correct labels but it seemed out of scope for a quick fix on something I noticed when using this for a prototype.

Here are 2 screenshots of the fix, the first without the code changes and the second with:
Screenshot_20230808-160325

Screenshot_20230808-160446

gabrielluong:
@zmckenney Thanks for the explanation! The changes looks good to me. I was more curious about how to test the changes in this case and visually see what the prompt looks like. Since we don't have any usage of the negative label, then I think it's all good to me. I appreciate seeing your test example.

PR #3147, opened 2023-08-07
Bug 1840338 - Add review quality info card to review product analysis.

MozillaNoah:
As per an offline discussion with UX, we do not currently want a click ripple on the learn more link.

Things I'm intentionally omitting from this PR for follow-up work
- Opening the URL (I'm assuming this will first require a navigation middleware)
- Styling the primary body of copy. (The copy is not finalized yet, so I'm holding off taking the time to create different `Text` composables and doing the substring formatting until the copy is ready.)

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1840338


fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -303,6 +311,91 @@ private fun SettingsCard(
     }
 }
 
+@Composable
+private fun ReviewQualityInfoCard(
+    onLearnMoreClick: () -> Unit,
```
rahulsainani, line 316: think we can have a modifier param here and the parent/caller can set the max width for it

fenix/app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
```diff
@@ -32,6 +32,7 @@ import org.mozilla.fenix.theme.FirefoxTheme
 @Composable
 fun ClickableSubstringLink(
     text: String,
+    textStyle: TextStyle = FirefoxTheme.typography.caption,
```
rahulsainani, line 35: I think this is fine for now, 

but going forward, we should use `LinkText` in onboarding after migrating it here or use the same api so every caller is not calculating the index and add accessibility support. There's a separate bug for that, that also needs to support multiple LinkTexts in one Text.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -303,6 +311,91 @@ private fun SettingsCard(
...
+    Row(
+        modifier = Modifier.fillMaxWidth(),
+        verticalAlignment = Alignment.Top,
+        horizontalArrangement = Arrangement.spacedBy(8.dp),
+    ) {
+        reviewGrades.forEach { grade ->
+            ReviewGradeCompact(grade = grade)
+        }
+
+        if (reviewGrades.size == 1) {
```
rahulsainani, line 387: ❓ Curious why this was chosen over setting the Row width to 64.dp?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ReviewQualityCheckBottomSheet.kt
```diff
@@ -54,6 +54,9 @@ fun ReviewQualityCheckBottomSheet(
                     onProductRecommendationsEnabledStateChange = {
                         store.dispatch(ReviewQualityCheckAction.ToggleProductRecommendation)
                     },
+                    onReviewGradeLearnMoreClick = {
+                        // Bug TODO
```
rahulsainani, line 58: Just created [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1847740) to handle this.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -303,6 +311,91 @@ private fun SettingsCard(
...
+    }
+}
+
+@Composable
+private fun ReviewGradingScaleInfo(
+    reviewGrades: List<ReviewQualityCheckState.Grade>,
+    info: String,
+) {
+    Row(
+        modifier = Modifier.fillMaxWidth(),
```
rahulsainani, line 379: 💡 maybe we can merge the contents for talk back here.
or maybe the whole section from "Review Grading scale:" and all the `ReviewGradingScaleInfo`s.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
MozillaNoah: When it's a child composable that's only ever used in one manner, I'm honestly never completely sure if it's worth adding, out of caution of an accidental modifier regression, but it might make more sense since it _is_ a child composable to give the parent caller the responsibility of giving it the correct sizing modifiers, so I'll add the parameter here and to `ReviewGradingScaleInfo`

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
MozillaNoah: so this was primarily me just exploring an alternative approach. I honestly wasn't 100% how the 64 in your patch was calculated, so I opted for using the conditional spacer with 24, since that's the size of the letter grades (which we could then refactor into a dimen).

how does that sound?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ReviewQualityCheckBottomSheet.kt
MozillaNoah: thanks! I'll plug in the bug number

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
MozillaNoah: I think merging here makes sense! thanks for catching this!

As for merging the entire section, I don't have a super thought out answer. Based on the kdocs screenshot below, I guess it depends if we want the entire section to be read together or in their individual components. Logically  and visually, they're one content section, so I see the reasoning behind making them one object for screen readers. On the other hand, by having the title and grade sections four distinct groupings, people using screen readers can skip through a section they don't care about or have already listened to. 

So it kinda boils down to: one group and listen to a comma-concatenated string and have to listen to the entire thing _every_ time, or have four elements that can be listened to or skipped over based on user preference.

I'll push changes to accomplish the latter, but I'm curious what your thoughts are.

Screenshot 2023-08-08 at 2 00 31 PM

fenix/app/src/main/java/org/mozilla/fenix/compose/ClickableSubstringLink.kt
MozillaNoah: ok, that makes sense to me.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
rahulsainani: Yeah I think the latter makes sense to me as my guess is that the users would read/hear it once or a few times and then know what the grades mean, but I could be wrong.
This is also a [UX question](https://www.figma.com/file/VE5xayz3yOSB5qOATUU9Im/Mobile-Shopping-Experience-MVP?type=design&node-id=2608-60360&mode=design&t=o0AjoHaGSm9l5MbI-0) and we can always change it to the other way if that's the preference. 
There is a separate [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1844732) for handling missing a11y cases so not a blocker at all. Thanks for updating it!

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -303,6 +316,94 @@ private fun SettingsCard(
...
+@Composable
+private fun ReviewQualityInfo(
+    modifier: Modifier = Modifier,
+    onLearnMoreClick: () -> Unit,
+) {
+    Column(
+        modifier = modifier,
+        verticalArrangement = Arrangement.spacedBy(24.dp),
+    ) {
+        // Any and all text formatting (bullets, inline substring bolding, etc.) will be handled as
```
rahulsainani, line 328: sorry I missed this earlier, can you please create a bug for this?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
rahulsainani: It was basically (size of letter grade + spacer) * 2.

I think it's fine either way, only minor concern was with the if logic, but even 64 is kind of an implicit logic which will break if the size of letter grade changes. Here we are being explicit, so this is better. 👍 

The dimens approach sounds good, I did try to introduce that early on but we need more ux sync there.

rahulsainani:
This is looking good! Just a comment about creating a bug for updating the text styles when the final copy is available.

rahulsainani (approved):
✅

PR #3153, opened 2023-08-08
Bug 1846517 - Expose coroutine scope through lazyStore

rahulsainani:
### What
– Creates and exposes a `CoroutineScope` scoped to the lifecycle of the `ViewModel` and the `Store`.
– Adds usage of the api.

### Why
– Use `viewModelScope` as it gets cancelled when the `ViewModel` is destroyed.
– The idea is to use this scope for all middleware operations so when the owner (fragment, activity) is popped off, the operations cancel. 
– As/if we remove the `configChange` manifest flag, this will be beneficial as the operations will survive configuration change gracefully, whereas `lifesycleScope` will be cancelled on configuration change.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1846517


HarrisonOg:
test

fenix/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt
```diff
@@ -37,23 +48,36 @@ class StoreProvider<T : Store<*, *>>(
...
+ * @sample
+ * val store by lazy { scope ->
+ *   MyStore(
+ *     middleware = provideMyMiddleware(
+ *     settings = requireComponents.settings,
+ *     ...
+ *     scope = scope,
+ *     ),
+ *   )
+ * }
```
jonalmeida, line 76: It looks like `@sample` requires an identifier to a function. Instead, let's just use a regular code block for now.

Separate nit: use a `listOf(middleware)` in the sample.

```suggestion
 * Example:
 * ```kotlin
 * val store by lazy { scope ->
 *   MyStore(
 *     middleware = listOf(
 *       MyMiddleware(
 *         settings = requireComponents.settings,
 *         ...
 *         scope = scope,
 *       ),
 *     )
 *   )
 * }
 * ```
```

fenix/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt
```diff
@@ -10,19 +10,30 @@ import androidx.lifecycle.ViewModel
...
+ * that's tied to the lifecycle of the [StoreProvider] i.e it will cancel when
+ * [StoreProvider.onCleared] is called.
  */
 class StoreProvider<T : Store<*, *>>(
-    val store: T,
+    createStore: (CoroutineScope) -> T,
 ) : ViewModel() {
 
+    @VisibleForTesting
+    val store: T = createStore(viewModelScope)
```
jonalmeida, line 29: Since you're adding `@VisibileForTesting` here, presumably because this is used only for tests so far aside from the return in `get()`, we should add the `internal` visibility qualifier too then.

```suggestion
    @VisibleForTesting
    internal val store: T = createStore(viewModelScope)
```

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFragment.kt
```diff
@@ -40,14 +40,14 @@ class ReviewQualityCheckFragment : BottomSheetDialogFragment() {
     private var behavior: BottomSheetBehavior<View>? = null
     private val bottomSheetStateFeature =
         ViewBoundFeatureWrapper<ReviewQualityCheckBottomSheetStateFeature>()
-    private val store by lazy {
+    private val store by lazyStore { scope ->
```
jonalmeida, line 43: As we're using this as an example on how to use the scope, let's make it clear which scope it is that we are replacing the `lifecycleScope` with.
```suggestion
    private val store by lazyStore { viewModelScope ->
```

fenix/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt
```diff
@@ -10,19 +10,30 @@ import androidx.lifecycle.ViewModel
...
 import androidx.lifecycle.get
+import androidx.lifecycle.viewModelScope
+import kotlinx.coroutines.CoroutineScope
 import mozilla.components.lib.state.Store
 
 /**
  * Generic ViewModel to wrap a State object for state restoration.
  *
- * @property store [Store] instance attached to [ViewModel].
+ * @param createStore Function that creates a [Store] instance that receives [CoroutineScope] param
```
jonalmeida, line 20: +1 for updating `property` to `param` here.

fenix/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt
```diff
@@ -10,19 +10,30 @@ import androidx.lifecycle.ViewModel
...
 ) : ViewModel() {
 
+    @VisibleForTesting
+    val store: T = createStore(viewModelScope)
+
     companion object {
-        fun <T : Store<*, *>> get(owner: ViewModelStoreOwner, createStore: () -> T): T {
+        /**
+         * Returns an existing [Store] instance or creates a new one scoped to a
+         * [ViewModelStoreOwner]. @see [ViewModelProvider.get].
```
jonalmeida, line 34: nit: formatting.

```suggestion
         * [ViewModelStoreOwner].
         * @see [ViewModelProvider.get].
```

jonalmeida (commented):
This looks a lot better! Only a few doc updates and nits below.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFragment.kt
rahulsainani: Since we have abstracted away `ViewModel` from Store and features, it made sense to not reference it. I don't have a strong opinion here, but it's a bit confusing for the caller to reference `viewModelScope` when using Store, thoughts?

jonalmeida: We're still using the provider to get the ViewModel whose scope we are using implicitly. It seems more clear to give the caller some clarity whose scope this is when they decide to use it.

Since we will use this PR as reference when you share it to the team in our mailing list and groups, so I heed on the side of clarity for that case. i.e. I don't need to look at the implementation, to follow the usage and know which scope this is.

jonalmeida: Sure! I am biased to putting `Provider` in there too because it gives you a bit more context that this will give you a `Store`. Let's do it in a separate change though?

rahulsainani: yeah, unrelated to this change. Again, no strong opinion here, discussing more from the perspective of increasing clarity.

jonalmeida (approved):

PR #3159, opened 2023-08-08
Bug 1796189 - List extensions installed outside the list of featured extensions

willdurand:
~~:warning: Depends on https://github.com/mozilla-mobile/firefox-android/pull/3157~~

This PR allows to list all the installed extensions in the Fenix add-on
manager, in addition to the extensions that are featured (i.e. part of
the AMO collection).

When the `mozAddonManager` is enabled, this actually fixes an issue with
how we use the AMO collection under the hood. When a user configures a
custom collection, then add-ons listed on AMO (the list of officially
supported add-ons on Android) are no longer installable. The web API now
allows to install these add-ons, and this patch will list them as
installed like any other, instead of marking them as "currently
unsupported".

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1796189


android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonsProvider.kt
```diff
@@ -10,19 +10,34 @@ package mozilla.components.feature.addons
...
         allowCache: Boolean = true,
         readTimeoutInSeconds: Long? = null,
         language: String? = null,
     ): List<Addon>
+
+    /**
+     * Provides a list of [Addon] based on the list of GUIDs passed to it.
+     *
+     * @param readTimeoutInSeconds optional timeout in seconds to use when fetching the add-ons from
+     * a remote endpoint.
```
Amejia481, line 33: Maybe  we could omit the `remote endpoint` part so the interface will be agnostic from the source, and the implementation could indicate further details from where is fetching :)

fenix/app/src/main/java/org/mozilla/fenix/extension/WebExtensionPromptFeature.kt
```diff
@@ -56,9 +55,22 @@ class WebExtensionPromptFeature(
...
+                        id = promptRequest.extension.id,
+                        downloadUrl = promptRequest.extension.url,
+                        version = m.version,
+                        permissions = m.permissions + m.hostPermissions,
+                        translatableName = mapOf(Addon.DEFAULT_LOCALE to m.name.orEmpty()),
+                        translatableDescription = mapOf(Addon.DEFAULT_LOCALE to m.description.orEmpty()),
+                        // We don't have a summary at this point so let's reuse the description.
+                        translatableSummary = mapOf(Addon.DEFAULT_LOCALE to m.description.orEmpty()),
+                    )
                 }
```
Amejia481, line 72: Nice!!

Amejia481: We will need to test by changing a different locale from the default one and see how it plays. I'm not sure if we may loose translations in permission popup?

willdurand: yeah, this is still a WIP.  I guess writing some tests might help us uncover bugs (I hope so..)

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AMOAddonsProvider.kt
```diff
@@ -152,9 +154,12 @@ class AMOAddonsProvider(
     @Throws(IOException::class)
     override suspend fun getAddonsByGUIDs(
         guids: List<String>,
+        allowCache: Boolean,
         readTimeoutInSeconds: Long?,
         language: String?,
     ): List<Addon> {
+        // We should probably introduce some sort of caching to avoid calling the AMO API all the time.
```
willdurand, line 161: @Amejia481 I _think_ this will be needed eventually but I am not sure how we'll want to implement that... I don't think we necessarily need a cache file stored on disk, maybe we can have in-memory caching? I think we need caching because this method is called in `getAddons()`, which is invoked every time a user opens the Fenix add-ons manager. WDYT?

android-components/components/feature/addons/src/test/java/AddonManagerTest.kt
```diff
@@ -70,11 +70,20 @@ class AddonManagerTest {
...
+        // addon3 (ext3) is a featured extension that is marked as disabled.
         val addon3 = Addon(id = "ext3")
+        // addon4 (ext4) and addon5 (ext5) are not featured extensions but they are installed.
+        val addon4 = Addon(id = "ext4")
+        val addon5 = Addon(id = "ext5")
         val addonsProvider: AddonsProvider = mock()
         whenever(addonsProvider.getFeaturedAddons(anyBoolean(), eq(null), language = anyString())).thenReturn(listOf(addon1, addon2, addon3))
+        // addon4 (ext4) and addon5 (ext5) should already be installed, which is why we expect a call
+        // to `getAddonsByGUIDs()` with a list of two GUIDs.
+        whenever(addonsProvider.getAddonsByGUIDs(eq(listOf("ext5", "ext4")), anyBoolean(), eq(null), language = anyString())).thenReturn(listOf(addon4, addon5))
```
willdurand, line 86: I am honestly not sure why `getAddonsByGUIDs()` receives a list with `"ext5"` first. I would have expected `"ext4"` first, and then `"ext5"`.

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -70,55 +70,50 @@ class AddonManager(
...
+            //
+            // NB: We're keeping translations only for the default locale.
+            val installedAddonGUIDs = otherInstalledExtensions
+                .filterValues { it.getMetadata()?.temporary == false }
+                .map { extensionEntry -> extensionEntry.value.id }
+            val installedAddons = addonsProvider.getAddonsByGUIDs(installedAddonGUIDs, language = userLanguage)
+                .map { addon -> addon.filterTranslations(locales) }
+                .map { addon -> addon.copy(installedState = installedExtensions[addon.id]?.toInstalledState()) }
+            // Last but not least, we build the list of temporarily loaded extensions.
+            val temporarilyInstalledAddons = otherInstalledExtensions
```
Amejia481, line 104: As we are going towards to general availability I think we may a good idea to revise if we still need temporarily extensions. No action needed now, just bring this up :)

willdurand: ah yes, we do because that's how developers can test their extension (e.g. `web-ext` loads add-ons temporarily)

Amejia481:
Sorry, didn't have enough time yesterday to work on the cache, I think we could reuse the same concept that have for featured add-ons, we could have a separated file which will contain the result of `getAddonsByGUIDs` may called `non-feature add-ons`, we could use the same expiration time but with an special case that when the number non-featured installed add-on is greater than the one in our cache, we could assumed the cache is expired and fetch again overriding the exiting cache, if we can't fetch we should not override and just use the metadata that we have of the installed add-on from Gecko.

I will work on the cache today :), if you have some available time we could pair on it.

Amejia481:
Added a extra commit for adding a cache for non-feature add-ons, it's taking shape I'll be adding test :)

willdurand:
> Based on these points, I think instead of using a single file for non-featured add-on, it maybe better to have separated single files for each add-on metadata, that will give use flexibility when fetch for presenting the permissions dialog and uninstalling.

Yeah, that sounds reasonable.

Amejia481:
Pushed the latest WIP, and added an extra lager of in memory caching.
https://github.com/mozilla-mobile/firefox-android/pull/3159/commits/119c176d43c65549ef2da6f3d9fc272579a899b9

Amejia481:
Removed the on disk cache for now, and added a in memory cache I'll open a follow up.

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/amo/AMOAddonsProviderTest.kt
```diff
@@ -819,6 +822,33 @@ class AMOAddonsProviderTest {
...
+        assertEquals(addon, addons.first())
+
+        verify(mockedClient, times(0)).fetch(
+            Request(
+                url = "https://services.addons.mozilla.org/api/v4/addons/search/?guid=uBlock0@raymondhill.net",
+                readTimeout = Pair(DEFAULT_READ_TIMEOUT_IN_SECONDS, TimeUnit.SECONDS),
+            ),
+        )
+        provider.installedAddons.clear()
+    }
```
Amejia481, line 850: This tested the in memory cache.

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/amo/AMOAddonsProviderTest.kt
```diff
@@ -795,6 +796,8 @@ class AMOAddonsProviderTest {
         assertEquals(4.4096F, addon.rating!!.average, 0.4096F)
         assertEquals(1233, addon.rating!!.reviews)
 
+        assertEquals(addon, provider.installedAddons[addon.id])
```
Amejia481, line 799: This test that, we are adding info to the memory cache, after we fetch.

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AMOAddonsProvider.kt
```diff
@@ -150,11 +155,21 @@ class AMOAddonsProvider(
...
         readTimeoutInSeconds: Long?,
         language: String?,
     ): List<Addon> {
+        if (allowCache && installedAddons.isNotEmpty()) {
+            val cacheAddons =
+                installedAddons.findAddonsBy(guids, language ?: Locale.getDefault().language)
+            if (cacheAddons.isNotEmpty()) {
+                return cacheAddons
+            }
+        }
```
Amejia481, line 171: Returning early in case if have the requested `guids` in the in-memory cache.

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/amo/AMOAddonsProvider.kt
```diff
@@ -176,7 +191,11 @@ class AMOAddonsProvider(
                 if (response.isSuccess) {
                     val responseBody = response.body.string(Charsets.UTF_8)
                     return try {
-                        JSONObject(responseBody).getAddonsFromSearchResults(language)
+                        val addons = JSONObject(responseBody).getAddonsFromSearchResults(language)
```
Amejia481, line 194: Adding add-ons to the cache.

Amejia481:
Sorry, I just realized that I should pushed an extra commit with my changes instead of squashing, I added some minimal changes added some comments to guided you what changed.

android-components/components/feature/addons/src/test/java/AddonManagerTest.kt
Amejia481: That's really odd!

Amejia481 (approved):
✅ !!

PR #3190, opened 2023-08-09
Bug 1847996 - Migrate from kapt to KSP

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1847996

### DIFF
--- a/android-components/components/lib/crash/build.gradle
+++ b/android-components/components/lib/crash/build.gradle
@@ -12,7 +12,6 @@ buildscript {
 
         dependencies {
             classpath "org.mozilla.telemetry:glean-gradle-plugin:${Versions.mozilla_glean}"
-            classpath ComponentsDependencies.plugin_ksp
             classpath ComponentsDependencies.plugin_serialization
         }
     }
--- a/android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
+++ b/android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
@@ -50,6 +50,7 @@ object Versions {
 
     const val material = "1.9.0"
     const val ksp = "1.0.11"
+    val ksp_plugin = "$kotlin-$ksp"
 
     // see https://android-developers.googleblog.com/2022/06/independent-versioning-of-Jetpack-Compose-libraries.html
     // for Jetpack Compose libraries versioning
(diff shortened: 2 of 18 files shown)


android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
```diff
@@ -50,6 +50,7 @@ object Versions {
 
     const val material = "1.9.0"
     const val ksp = "1.0.11"
+    val ksp_plugin = "$kotlin-$ksp"
```
jonalmeida, line 53: Drive-by: This is interesting.. does this mean we are now tied to specific versions of kotlin that have ksp plugin support? What happens if we need a patch version of `kotlin` and there is no `ksp` version?

mcarare: That should not happen, according to the docs:

> Usually a corresponding KSP release is available within a couple of days after a new Kotlin version is released, including the [pre-releases (Beta or RC)](https://kotlinlang.org/docs/eap.html).

source: https://kotlinlang.org/docs/ksp-faq.html#how-often-do-you-update-ksp

titooan (approved):
Looks good to me!

android-components/components/lib/crash/build.gradle
```diff
@@ -12,7 +12,6 @@ buildscript {
 
         dependencies {
             classpath "org.mozilla.telemetry:glean-gradle-plugin:${Versions.mozilla_glean}"
-            classpath ComponentsDependencies.plugin_ksp
```
jonalmeida, line 15: Typo or we really do not need the plugin here? I think we have room in `lib-crash` too, so it would be needed, but I don't see CI failing.

jonalmeida (approved):
Sorry for the slow review churn. 🙇

fenix/build.gradle
```diff
@@ -164,7 +166,7 @@ allprojects {
         kotlinOptions.jvmTarget = "17"
         kotlinOptions.allWarningsAsErrors = true
         kotlinOptions.freeCompilerArgs += [
-            "-opt-in=kotlin.RequiresOptIn", "-Xjvm-default=all"
+            "-opt-in=kotlin.RequiresOptIn", "-Xjvm-default=all-compatibility"
```
jonalmeida, line 169: Self-note: this is fine to land because in the follow-up PR, we remove this because we use the new Room version that [has the fixes](https://android-review.googlesource.com/c/platform/frameworks/support/+/2673596).

android-components/components/lib/crash/build.gradle
mcarare: This is the build script block and we do not need it for the build itself. 
We still have it below in the top-level dependencies block.

PR #3104, opened 2023-08-03
Bug 1834842 - Replace Hardcoded SERP Telemetry

sarah541:
This is in regards with how our SERP Telemetry defining classes will update when we will be pulling in the config from remote source instead of hard coded elements. The RemoteSettingsDownloader.kt is a replica of how the remote config looks like but will eventually be getting removed and instead there will be a helper class which will be fetching this config .
This will be captured [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1834820) . As of the task related to this PR, everything is ready to be reviewed but is dependent on the task mentioned before hence not ready to be landed.
### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1834842
https://bugzilla.mozilla.org/show_bug.cgi?id=1870066

PR #3203, opened 2023-08-10
Bug 1834820 - Add RemoteSettings downloader class

sarah541:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1834820


android-components/components/feature/search/src/main/java/mozilla/components/feature/search/telemetry/BaseSearchTelemetry.kt
```diff
@@ -21,13 +25,30 @@ import mozilla.components.support.base.facts.Fact
...
+    init {
+        myCoroutineScope.launch {
+            val remoteSettingsDownloader = RemoteSettingsClient(
+                serverUrl = "https://firefox.settings.services.mozilla.com",
+                bucketName = "main",
+                collectionName = "search-telemetry-v2"
+            )
+
+            val something = remoteSettingsDownloader.fetchRemoteServerRecords()
+            Log.d("sarah",something.toString())
```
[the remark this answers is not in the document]
sarah541: Yes

MatthewTighe (commented):
Leaving a comment after our offline talk earlier today - hopefully it'll help unblock you 🙂

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,52 @@
...
+        bucketName = bucketName,
+        collectionName = collectionName,
+    )
+
+    /**
+     * Fetches all records for a collection that can be found in the server.
+     **/
+    suspend fun fetchRemoteServerRecords(): JSONObject = withContext(Dispatchers.IO) {
+        try {
+            val client = RemoteSettings(config)
```
MatthewTighe, line 40: It's worth noting that we should be calling this with [use](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/use.html) since it is an [interface in the UniFFI UDL definition](https://searchfox.org/mozilla-mobile/source/application-services/components/remote_settings/src/remote_settings.udl#48). [This conversation](https://mozilla.slack.com/archives/C0559DDDPQF/p1694553760906609) has a bit more detail

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,109 @@
...
+        serverUrl = serverUrl,
+        bucketName = bucketName,
+        collectionName = collectionName,
+    )
+
+    /**
+     * Fetches all records for a collection that can be found in the local storage.
+     **/
+    suspend fun fetch(): JSONObject? = withContext(Dispatchers.IO) {
+        val fileName = "list.json"
```
MatthewTighe, line 47: We should investigate a less opinionated file caching mechanisms for these records. I think we will at least want a way for consumers to be able to specify the name of the file they are saving a collection or record to. Happy to discuss design in more depth when I get back next week 🙂

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,109 @@
...
+            updateLocalFile(records)
+            return@withContext records
+        } catch (e: RemoteSettingsException.ResponseException) {
+            JSONObject()
+        } catch (e: Exception) {
+            Log.d("exception",e.message.toString())
+            JSONObject()
+
+        }
+    }
```
MatthewTighe, line 94: We should be careful about initiating disk writes without making that really clear what's happening in the API. We probably want to, at minimum, create separate `fetch` and `save` methods, though we could also consider a convenience function that did both.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
MatthewTighe, line 47: I think we'll probably want to figure out a robust strategy for saving different collections to different files. We may be able to do something as simple as prepending the collection name to the file, but I'm not sure if we'll need to worry about naming collisions. Can two collections on a server share a name? Two collections in a bucket?

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -252,12 +252,12 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
 
-            // We need to always initialize Glean and do it early here.
-            initializeGlean()
-
             // Attention: Do not invoke any code from a-s in this scope.
             val megazordSetup = finishSetupMegazord()
 
+            // We need to always initialize Glean and do it early here.
+            initializeGlean()
+
```
MatthewTighe, line 260: Can this change be safely reverted now?

MatthewTighe (commented):
🚗 some drive-by thoughts between conference talks. Super happy to talk more about this when I get back, but not sure how much bandwidth I'll have until then

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
sarah541: Yes, right! I missed that. It will be dealt most likely in this [PR](https://github.com/mozilla-mobile/firefox-android/pull/3104)

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,109 @@
...
+    private val config = RemoteSettingsConfig(
+        serverUrl = serverUrl,
+        bucketName = bucketName,
+        collectionName = collectionName,
+    )
+
+    /**
+     * Fetches all records for a collection that can be found in the local storage.
+     **/
+    suspend fun fetch(): JSONObject? = withContext(Dispatchers.IO) {
```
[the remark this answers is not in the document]
sarah541: Agreed!

android-components/components/support/remotesettings/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+dependencies {
+    implementation ComponentsDependencies.mozilla_remote_settings
+
+    implementation ComponentsDependencies.kotlin_coroutines
+    implementation ComponentsDependencies.androidx_annotation
+    implementation ComponentsDependencies.androidx_core
+    implementation ComponentsDependencies.androidx_core_ktx
+
+    // We expose the app-compat as API so that consumers get access to the Lifecycle classes automatically
+    api ComponentsDependencies.androidx_appcompat
```
MatthewTighe, line 37: Do we need this here?

android-components/.buildconfig.yml
```diff
@@ -2027,6 +2027,12 @@ projects:
     - support-utils
     - tooling-lint
     - ui-icons
+  support-remotesettings:
+    description: A component to help fetch data from remote servers.
+    path: components/support/remotesettings
+    publish: true
+    upstream_dependencies:
+      - support-base
```
MatthewTighe, line 2035: I'm not really familiar with when dependencies need to be upstreamed. I would assume that it's only required if we need to expose an API from the dependency. Is that correct? What API do we need exposed from `support-base`?

android-components/components/support/remotesettings/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+    namespace 'mozilla.components.support.remotesettings'
+}
+
+dependencies {
+    implementation ComponentsDependencies.mozilla_remote_settings
+
+    implementation ComponentsDependencies.kotlin_coroutines
+    implementation ComponentsDependencies.androidx_annotation
+    implementation ComponentsDependencies.androidx_core
+    implementation ComponentsDependencies.androidx_core_ktx
```
MatthewTighe, line 34: It doesn't look like any of these except `kotlin_coroutines` is needed. It'll help us keep the size of the component down to only add dependencies that will be used.

android-components/components/support/remotesettings/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+
+    // We expose the app-compat as API so that consumers get access to the Lifecycle classes automatically
+    api ComponentsDependencies.androidx_appcompat
+
+    testImplementation ComponentsDependencies.androidx_test_core
+    testImplementation ComponentsDependencies.androidx_test_junit
+    testImplementation ComponentsDependencies.testing_coroutines
+    testImplementation ComponentsDependencies.testing_robolectric
```
MatthewTighe, line 42: Speaking of test dependencies, it could be good to have some unit tests as well. Also, I am not sure whether we will need all of these - specifically `robolectric` seems like it may end up being unnecessary.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,109 @@
...
+        try {
+            val records = RemoteSettings(config).use {
+                it.getRecords().records.first().fields
+            }
+            updateLocalFile(records)
+            return@withContext records
+        } catch (e: RemoteSettingsException.ResponseException) {
+            JSONObject()
+        } catch (e: Exception) {
+            Log.d("exception",e.message.toString())
```
MatthewTighe, line 90: Generally I think we prefer using the `Logger` utility class if we are interested in keeping these log messages long-term.

MatthewTighe (changes requested):
👋 dropping a few more comments now that I'm back from PTO and have a chance to look at this in a little more depth. Just let me know if you want to talk more about any of this feedback offline!

android-components/.buildconfig.yml
sarah541: Using it for Logger now

android-components/components/support/remotesettings/build.gradle
sarah541: Sure I will, once we confirm the class needs no more changes?

android-components/components/feature/search/src/test/java/mozilla/components/feature/search/telemetry/BaseSearchTelemetryTest.kt
```diff
@@ -9,6 +9,7 @@ import mozilla.components.concept.engine.Engine
 import mozilla.components.support.test.any
 import mozilla.components.support.test.eq
 import mozilla.components.support.test.mock
+import mozilla.components.support.test.robolectric.testContext
```
t-p-white, line 12: unused import?

android-components/components/feature/search/src/main/java/mozilla/components/feature/search/telemetry/BaseSearchTelemetry.kt
```diff
@@ -6,7 +6,6 @@ package mozilla.components.feature.search.telemetry
 
 import androidx.annotation.VisibleForTesting
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.collect
```
t-p-white, line 9: 👌

android-components/components/support/remotesettings/README.md
```diff
@@ -0,0 +1,18 @@
+# [Android Components](../../../README.md) > Support > Remotesettings
+
+This component defines and installs an application-services `remotesettings` package that:
+- Fetches records/data from mozilla remote servers
```
t-p-white, line 4:
```suggestion
- Fetches data from mozilla remote servers
```

android-components/components/support/remotesettings/README.md
```diff
@@ -0,0 +1,18 @@
...
+
```
t-p-white, line 5: Just following other READMEs perhaps we should also include `## Usage`

android-components/components/support/remotesettings/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+    implementation project(':support-ktx')
+    implementation project(':support-base')
+    implementation ComponentsDependencies.mozilla_remote_settings
+
+    implementation ComponentsDependencies.kotlin_coroutines
+    implementation ComponentsDependencies.androidx_core
+
+    testImplementation ComponentsDependencies.androidx_test_core
+    testImplementation ComponentsDependencies.androidx_test_junit
+    testImplementation ComponentsDependencies.testing_coroutines
```
t-p-white, line 38: can we remove these test dependencies until they are needed?

android-components/components/support/remotesettings/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+    namespace 'mozilla.components.support.remotesettings'
+}
+
+dependencies {
+    implementation project(':support-ktx')
+    implementation project(':support-base')
+    implementation ComponentsDependencies.mozilla_remote_settings
+
+    implementation ComponentsDependencies.kotlin_coroutines
+    implementation ComponentsDependencies.androidx_core
```
t-p-white, line 34: are these required?

android-components/components/support/remotesettings/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+
+android {
+    compileSdkVersion config.compileSdkVersion
+
+    defaultConfig {
+        minSdkVersion config.minSdkVersion
+        targetSdkVersion config.targetSdkVersion
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+    }
```
t-p-white, line 16: can this be:
```suggestion
    defaultConfig {
        minSdkVersion config.minSdkVersion
        compileSdk config.compileSdkVersion
        targetSdkVersion config.targetSdkVersion
    }
```

android-components/components/support/remotesettings/build.gradle
```diff
@@ -0,0 +1,43 @@
...
+apply plugin: 'kotlin-android'
+
+android {
+    compileSdkVersion config.compileSdkVersion
+
+    defaultConfig {
+        minSdkVersion config.minSdkVersion
+        targetSdkVersion config.targetSdkVersion
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
```
t-p-white, line 15: can we remove this other test reference to`testInstrumentationRunner` until we have tests?

android-components/components/support/remotesettings/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,14 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this file,
+   - You can obtain one at http://mozilla.org/MPL/2.0/.  -->
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <!-- Adds the ability to discover apps that have matching <intent-filter> elements.
+    Needed for PackageManager.queryIntentActivities calls.
+    See https://developer.android.com/training/package-visibility/declaring for more details.-->
+    <queries>
```
t-p-white, line 8: OoI what are we using this for in RemoteSettings?

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+ *
+ * @property storageRootDirectory The top level app-local storage directory.
+ * @property serverUrl The optional url for the settings server. If not specified, the standard server will be used.
+ * @property bucketName The optional name of the bucket containing the collection on the server.
+ * If not specified, the standard bucket will be used.
+ * @property collectionName The name of the collection for the settings server.
+ */
+class RemoteSettingsClient(
+    private val storageRootDirectory: File,
+    val serverUrl: String = "https://firefox.settings.services.mozilla.com",
```
t-p-white, line 35: nit: perhaps `"https://firefox.settings.services.mozilla.com"` could be a const? same for `bucketName`

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+    val bucketName: String = "main",
+    val collectionName: String = "",
```
t-p-white, line 37: should this be defaulted?

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+) {
+
+    private val config = RemoteSettingsConfig(
+        serverUrl = serverUrl,
+        bucketName = bucketName,
+        collectionName = collectionName,
+    )
+
+
```
t-p-white, line 46:
```suggestion
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
t-p-white, line 37: think these can all be private?

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+import java.io.File
+import java.io.FileReader
+
+/**
+ * Helper class to download collections from remote settings in app services.
+ *
+ * @property storageRootDirectory The top level app-local storage directory.
+ * @property serverUrl The optional url for the settings server. If not specified, the standard server will be used.
+ * @property bucketName The optional name of the bucket containing the collection on the server.
+ * If not specified, the standard bucket will be used.
```
t-p-white, line 30:
```suggestion
 * If not specified, the name of the bucket "main" [or the const if updated]
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+    private val config = RemoteSettingsConfig(
+        serverUrl = serverUrl,
+        bucketName = bucketName,
+        collectionName = collectionName,
+    )
+
+
+    /**
+     * Fetches all records for the collection from remote servers and updates the local storage.
+     **/
```
t-p-white, line 49: nit: 
```suggestion
     */
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+            jsonObject.put("records", jsonArray)
+            AtomicFile(file).writeString { jsonObject.toString() }
+        } catch (e: Exception) {
+            e.printStackTrace()
+        }
+    }
+
+    /**
+     * Reads all records for a collection that can be found in the local storage.
+     **/
```
t-p-white, line 88: nit: 
```suggestion
     */
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+    suspend fun fetchRemoteServerRecords(): List<RemoteSettingsRecord> = withContext(Dispatchers.IO) {
+        try {
+            val serverRecords = RemoteSettings(config).use {
+                it.getRecords().records
+            }
+            return@withContext serverRecords
+        } catch (e: RemoteSettingsException.ResponseException) {
+            emptyList()
+        } catch (e: Exception) {
+            Logger(e.message.toString())
```
t-p-white, line 59: should this `Logger` be `Logger.error` or a warning `Logger.warn` ?

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+     **/
+    suspend fun fetchRemoteServerRecords(): List<RemoteSettingsRecord> = withContext(Dispatchers.IO) {
+        try {
+            val serverRecords = RemoteSettings(config).use {
+                it.getRecords().records
+            }
+            return@withContext serverRecords
+        } catch (e: RemoteSettingsException.ResponseException) {
+            emptyList()
+        } catch (e: Exception) {
```
t-p-white, line 58: not sure the catch all exception is needed? Perhaps the above catch could be `RemoteSettingsException` as it looks like it's always setting an empty list and possibly logging?

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+
+    private val config = RemoteSettingsConfig(
+        serverUrl = serverUrl,
+        bucketName = bucketName,
+        collectionName = collectionName,
+    )
+
+
+    /**
+     * Fetches all records for the collection from remote servers and updates the local storage.
```
t-p-white, line 48:
```suggestion
     * Fetches all records for the collection from remote servers and updates the local file.
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+            Logger(e.message.toString())
+            emptyList()
+
+        }
+    }
+
+    private fun writeRemoteRecordsToStorage(file: File, records: List<RemoteSettingsRecord>) {
+        try {
+            val jsonArray = JSONArray()
+            for (record in records) {
```
t-p-white, line 68: what you've got here is totally fine, I just wanted to mention you could also use the Kotlins Collection framework and do something like:
```
 records.forEach {
                val recordJson =
                    JSONObject().put("id", it.id)
...
}
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+                val recordJson = JSONObject()
+                recordJson.put("id", record.id)
```
t-p-white, line 70: nit: I _think_ you can chain the `put` calls on JSONObject

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+                recordJson.put("attachment", record.attachment)
+                recordJson.put("fields", record.fields)
+                jsonArray.put(recordJson)
+            }
+
+            val jsonObject = JSONObject()
+            jsonObject.put("records", jsonArray)
+            AtomicFile(file).writeString { jsonObject.toString() }
+        } catch (e: Exception) {
+            e.printStackTrace()
```
t-p-white, line 82: should we log here?

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+        }
+    }
+
+    /**
+     * Reads all records for a collection that can be found in the local storage.
+     **/
+    suspend fun readStorage(): List<RemoteSettingsRecord> = withContext(Dispatchers.IO) {
+        val folder = File(storageRootDirectory, bucketName)
+        val file =  File((folder), collectionName)
```
t-p-white, line 91:
```suggestion
        val file =  File(folder, collectionName)
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+        var remoteRecords: List<RemoteSettingsRecord>
+        var records = mutableListOf<RemoteSettingsRecord>()
+        if (file.exists()) {
+            try {
+                val fileReader = FileReader(file)
```
t-p-white, line 96: should we also be using `AtomicFile.readJSON`?

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
sarah541: bucketName can change

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+                            deleted = deleted,
+                            attachment = attachment as Attachment?,
+                            fields = fields as RsJsonObject
+                        )
+                    )
+                    fileReader.close()
+
+                }
+                launch(Dispatchers.IO) {
+                    remoteRecords = fetchRemoteServerRecords()
```
MatthewTighe, line 122: I think we should separate out fetching and writing from this function and rely on consumer to order their operations correctly. My main concern is that there is nothing in the name `readStorage` that suggests to me that this function will be initiating network IO and disk writes. 

Generally speaking, the advice that I try to follow for code (especially library code) is to try and avoid functions with the word "and" or "or" in them. This is the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle) in case you were not already aware of it. I think this function would need to be named something like `tryReadRecordsOrFetchAndWrite` to really capture all the behavior in it. 

This patch is most of the way there already anyway - I think we can just make `writeRemoteRecordsToStorage` public and do a few other tweaks. For example, we may want some kind of result type:

```kotlin
    sealed class RemoteSettingsResult {
        data class Success(val records: List<RemoteSettingsRecord>)
        data class DiskFailure(val error: Exception)
        data class NetworkFailure(val error: Exception)
    }
```

MatthewTighe: I do think it would be fine to _also_ expose a method that tries to read from disk and then fetches from the network if it needs to, but I think we should provide more fine-grained control for consumers as a first step. An extension function would be a more idiomatic way to compose the two functions (and excuse a long name), IMO:

```kotlin
// top-level function
suspend fun RemoteSettingsClient.readOrFetch(): RemoteSettingsResult =
    this.read().takeIf { it is RemoteSettingsResult.Success } ?: this.fetch()
```

MatthewTighe (commented):
This is really coming together, good work! I have a few more comments about the general structure of the API. I'm a little concerned about becoming overwhelming between myself and Tom, so I will probably try to let him handle the rest of the review, but please let me know if you want any more input from me!

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+            AtomicFile(file).writeString { jsonObject.toString() }
+        } catch (e: Exception) {
+            Logger.error(e.message.toString())
+        }
+    }
+
+    /**
+     * Reads all records for a collection that can be found in the local storage.
+     **/
+    suspend fun readStorage(): List<RemoteSettingsRecord> = withContext(Dispatchers.IO) {
```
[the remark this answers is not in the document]
sarah541: Is there a precedent for this? I mean do we offer the customers to have this flexibility every time we are loading files in the storage?

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,136 @@
...
+        val folder = File(storageRootDirectory, bucketName)
```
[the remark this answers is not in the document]
sarah541: Yes if we are offering the consumers to configure the path/name of the file, it could be hard to distinguish. Had we not done that, the same server and bucket names would still have contributed to different path , because the file name(collection name)  would have been different.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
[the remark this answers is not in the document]
MatthewTighe: After thinking about this more, I think you're right that's it's not necessary.
- There's no precedent that I can think of. 
- We have a general purpose HTTP client available anyway. I think my comment was colored with my experience trying to use Remote Settings for wallpapers - I was changing filepaths pretty regularly while experimenting.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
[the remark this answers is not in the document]
MatthewTighe: I think we should still use the server name as part of the directory path - otherwise 2 servers with the same bucket and collection names would overlap. I don't expect to actually run into that, but it seems worth updating this line to be something like:
```
val folder = File(storageRootDirectory, serverName.stripProtocol(), bucketName)
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
sarah541: I think the single responsibility principle makes very much sense and I sure will incorporate this in future too.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+            Logger.error(e.message.toString())
+        }
+    }
+
+    /**
+     * Reads all records for a collection  found in the local storage.
+     **/
+    suspend fun read(): RemoteSettingsResult = withContext(Dispatchers.IO) {
+        val path = "${storageRootDirectory.path}/$serverUrl/$bucketName/$collectionName"
+        val file = File(path)
```
MatthewTighe, line 95: Since we've decided to be opinionated about the filepath here, I think these could become class properties that can be shared between functions:

```kotlin
    private val config = RemoteSettingsConfig(
        serverUrl = serverUrl,
        bucketName = bucketName,
        collectionName = collectionName,
    )

    private val path = "${storageRootDirectory.path}/$serverUrl/$bucketName/$collectionName"
    private val file = File(path)
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+                            attachment = attachment as Attachment?,
+                            fields = fields as RsJsonObject
+                        )
+                    )
+                }
+
+                fileReader.close()
+                RemoteSettingsResult.Success(records)
+            } catch (e: Exception) {
+                e.printStackTrace()
```
MatthewTighe, line 126: We can remove this `printStackTrace` now

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+        } catch (e: RemoteSettingsException) {
+            Logger.error(e.message.toString())
+            RemoteSettingsResult.NetworkFailure(e)
+        }
+    }
+
+    /**
+     * Updates the local storage with [records] data.
+     */
+    fun write(file: File, records: List<RemoteSettingsRecord>) {
```
MatthewTighe, line 68: Since we're gonna be opinionated about filepath, I think we can remove this `file` parameter and instead use the properties as discussed below

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+        } else {
+            RemoteSettingsResult.DiskFailure(FileNotFoundException("File not found"))
+        }
+    }
+
+}
+/**
+ * Reads files from local storage. If file is empty, fetches teh file from the remote server.
+ **/
+suspend fun RemoteSettingsClient.readOrFetch(): ReadOrFetchResult {
```
MatthewTighe, line 138: I think a similar `fetchAndWrite` convenience method could be useful.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+sealed class RemoteSettingsResult {
+    data class Success(val records: List<RemoteSettingsRecord>) : RemoteSettingsResult()
+    data class DiskFailure(val error: Exception) : RemoteSettingsResult()
+    data class NetworkFailure(val error: Exception) : RemoteSettingsResult()
+}
+
+sealed class ReadOrFetchResult {
+    data class Success(val records: List<RemoteSettingsRecord>) : ReadOrFetchResult()
+    data class Failure(val error: Exception) : ReadOrFetchResult()
+}
```
MatthewTighe, line 163: This definitely works, but I think we could get away with just `RemoteSettingsResult`. That way, the return type from the convenience methods and the more focused methods can all be the same.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+                }
+                jsonArray.put(recordJson)
+            }
+
+            val jsonObject = JSONObject()
+            jsonObject.put("records", jsonArray)
+            AtomicFile(file).writeString { jsonObject.toString() }
+        } catch (e: Exception) {
+            Logger.error(e.message.toString())
+        }
```
MatthewTighe, line 87: I think it could be worth returning a `RemoteSettingsResult` from this function so that the exception isn't lost.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /**
+     * Reads all records for a collection  found in the local storage.
+     **/
+    suspend fun read(): RemoteSettingsResult = withContext(Dispatchers.IO) {
+        val path = "${storageRootDirectory.path}/$serverUrl/$bucketName/$collectionName"
+        val file = File(path)
+
+        if (file.exists()) {
```
MatthewTighe, line 97: nit:
```suggestion
        if (!file.exists()) {
            return@withContext RemoteSettingsResult.DiskFailure(FileNotFoundException("File not found"))
        }
        
        try {
            // ...
        }
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+                    put("deleted", it.deleted)
+                    put("attachment", it.attachment)
+                    put("fields", it.fields)
+                }
+                jsonArray.put(recordJson)
+            }
+
+            val jsonObject = JSONObject()
+            jsonObject.put("records", jsonArray)
+            AtomicFile(file).writeString { jsonObject.toString() }
```
MatthewTighe, line 84: We probably want to launch a coroutine on the IO dispatcher to handle the actual file write. Otherwise, we will probably hit a `StrictMode` violation when we go to use this in Fenix. We could alternatively make this function suspending.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,164 @@
...
+        is RemoteSettingsResult.DiskFailure -> {
+            when (val remoteResult = fetch()) {
+                is RemoteSettingsResult.Success -> ReadOrFetchResult.Success(remoteResult.records)
+                is RemoteSettingsResult.NetworkFailure -> ReadOrFetchResult.Failure(remoteResult.error)
+                is RemoteSettingsResult.DiskFailure -> ReadOrFetchResult.Failure(remoteResult.error)
+
+            }
+        }
+        is RemoteSettingsResult.NetworkFailure -> ReadOrFetchResult.Failure(storageResult.error)
+
```
MatthewTighe, line 151: nit: extra newline

MatthewTighe (changes requested):
This is looking great! I left a few more comments, and I think we should add some test cases for the client. Other than that, this seems pretty much ready to go to me!

MatthewTighe (commented):
The finish line is really in sight now, good work! Just a few more comments and this should be the last round of feedback 🙂 thanks for sticking with it!

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,181 @@
...
+                    put("fields", it.fields)
+                }
+                jsonArray.put(recordJson)
+            }
+
+            val jsonObject = JSONObject()
+            jsonObject.put("records", jsonArray)
+            AtomicFile(file).writeString { jsonObject.toString() }
+            RemoteSettingsResult.Success(records)
+        } catch (e: IOException) {
```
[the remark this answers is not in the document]
sarah541: Our lint does not allow us to catch Generic exceptions, does it?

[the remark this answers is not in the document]
MatthewTighe: Which linter is that? I am not seeing detekt complain when I try it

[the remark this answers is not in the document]
sarah541: Surprisingly it is not. But i did change it from Generic to IOException because of that. Anyway happy to change back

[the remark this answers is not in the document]
sarah541: It is detekt

Screen Shot 2023-10-10 at 2 57 39 PM

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -0,0 +1,181 @@
...
+            RemoteSettingsResult.DiskFailure(e)
+        }
+    }
+
+    /**
+     * Reads all records for a collection  found in the local storage.
+     **/
+    suspend fun read(): RemoteSettingsResult = withContext(Dispatchers.IO) {
+        val path = "${storageRootDirectory.path}/$serverUrl/$bucketName/$collectionName"
+        val file = File(path)
```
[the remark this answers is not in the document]
MatthewTighe: I think we will still want to make this change, otherwise the `read` path will be different than the `write` path.
```suggestion

```

android-components/components/support/remotesettings/src/test/java/mozilla/components/support/remotesettings/RemoteSettingsClientTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    private lateinit var mockFileReader: FileReader
+    private lateinit var mockJsonObject: JSONObject
+    private lateinit var remoteSettingsClient: RemoteSettingsClient
+
+    @Before
+    fun setUp() {
+        mockFile = mock(File::class.java)
+        mockFileReader = mock(FileReader::class.java)
+        mockJsonObject = mock(JSONObject::class.java)
+        mockRemoteSettings = mock(RemoteSettings::class.java)
```
MatthewTighe, line 33: Since the type of these mocks is declared above, you shouldn't need the explicit class arguments
```suggestion
        mockFile = mock()
        mockFileReader = mock()
        mockJsonObject = mock()
        mockRemoteSettings = mock()
```

android-components/components/support/remotesettings/src/test/java/mozilla/components/support/remotesettings/RemoteSettingsClientTest.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        require(result is RemoteSettingsResult.Success) { "Result should be Success" }
+        assertEquals(records, result.records)
+
+        tempFile.delete()
+    }
+
+    @Test
+    fun `test fetchAndWrite success`() {
+        val client = mock(RemoteSettingsClient::class.java)
```
MatthewTighe, line 63: nit: For inline mocks like these, I think the syntax that uses generics is more idiomatic
```suggestion
        val client = mock<RemoteSettingsClient>()
```

android-components/components/support/remotesettings/src/test/java/mozilla/components/support/remotesettings/RemoteSettingsClientTest.kt
```diff
@@ -0,0 +1,121 @@
...
+            `when`(client.read()).thenReturn(readResult)
+
+            val result = client.readOrFetch()
+
+            assertEquals(readResult, result)
+        }
+    }
+
+    @Test
+    fun `test readOrFetch read failure`() {
```
MatthewTighe, line 107: Thanks for writing all these test! Happy to see them 🙂

android-components/components/support/remotesettings/src/test/java/mozilla/components/support/remotesettings/RemoteSettingsClientTest.kt
```diff
@@ -0,0 +1,121 @@
...
+            `when`(client.write(records)).thenReturn(writeResult)
+
+            val result = client.fetchAndWrite()
+
+            assertEquals(writeResult, result)
+        }
+    }
+
+    @Test
+    fun `test fetchAndWrite fetch failure`() {
```
MatthewTighe, line 79: Could we rename these tests so that they follow the GIVEN WHEN THEN style that was established above?

MatthewTighe (approved):
Awesome work 🎉 ! Thanks for sticking with it through the long review

show time

PR #3211, opened 2023-08-10
Bug 1847923 - Part 1: Add ProductAnalysisMapper

rahulsainani:
Adding the `ReviewQualityCheckNetworkMiddleware` PR was getting too big this is broken down into two parts. This part adds the mapper that maps the object exposed from AC to the `ReviewQualityCheckState`s `ProductReviewState`.

Check usage in #3212

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1847923


fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+        ReviewQualityCheckState.HighlightType.PRICE to price,
+        ReviewQualityCheckState.HighlightType.SHIPPING to shipping,
+        ReviewQualityCheckState.HighlightType.COMPETITIVENESS to competitiveness,
+    )
+
+    return highlights.filterValues { it != null }.mapValues { it.value!! }.ifEmpty { null }
+}
+
+private fun Double.toFloatOrNull(): Float? =
+    if (this == 0.0) {
```
MozillaNoah, line 63: is this implying the service will never return 0.0?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+
+        if (mappedGrade == null && mappedRating == null && mappedHighlights == null) {
+            ProductReviewState.NoAnalysisPresent
+        } else {
+            ProductReviewState.AnalysisPresent(
+                productId = productId!!,
+                reviewGrade = mappedGrade,
+                needsAnalysis = needsAnalysis,
+                adjustedRating = mappedRating,
+                productUrl = analysisURL!!,
```
MozillaNoah, line 38: should `analysisURL` also be in the root `if` with `productID`, if they're both mandatory?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+    } catch (e: IllegalArgumentException) {
+        null
+    }
+
+private fun Highlight.toHighlights(): Map<ReviewQualityCheckState.HighlightType, List<String>>? {
+    val highlights: Map<ReviewQualityCheckState.HighlightType, List<String>?> = mapOf(
+        ReviewQualityCheckState.HighlightType.QUALITY to quality,
+        ReviewQualityCheckState.HighlightType.PRICE to price,
+        ReviewQualityCheckState.HighlightType.SHIPPING to shipping,
+        ReviewQualityCheckState.HighlightType.COMPETITIVENESS to competitiveness,
```
MozillaNoah, line 56: is the omission of `PACKAGING_AND_APPEARANCE` intentional?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
```diff
@@ -0,0 +1,67 @@
...
+    )
+
+    return highlights.filterValues { it != null }.mapValues { it.value!! }.ifEmpty { null }
```
MozillaNoah, line 59: is `mapValues` doing anything here? Is it just changing the value type from `List<String>?` to `List<String>`?

MozillaNoah (approved):
Left some non-blocking comments/questions

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
rahulsainani: Good question, yes exactly, it's basically to assert that the type is infact `List<String>`. Not sure if there's a better way to achieve this?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
rahulsainani: Nope, great catch! Fixed! Incidentally the test was also setting it to null to test the case for missing highlights, added another test with all highlights to make sure this is fixed.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
rahulsainani: Yes, my understanding is that rated products would have at least 1 as value, since people can't rate products as 0. 
Gecko layer is converting null values to `0.0` so till it's changed there back to null, we have to add this check unfortunately.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
rahulsainani: Good question, my understanding of the API is that when there's product ID and one of the 3 (grade, rating and highlights) present, there would be a fakespot analysis url present in the API.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
MozillaNoah: I couldn't think of any when I pulled down the code. We can typecast things, but that's still a transformation that has to occur in the `mapValues` block, so I'm not immediately aware of an "optimal" way to de-nullify a type without just force unwrapping it.

rahulsainani: I think it's fine as we know more that the compiler in this case that value is not null when `mapValues` lamda is triggered 👍

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
t-p-white: Sorry a little late on this. Wondering if it would be useful here to use the `HightlightType` enum to enforce all values are used e.g.

``` 
val highlights: Map<ReviewQualityCheckState.HighlightType, List<String>?> =
        ReviewQualityCheckState.HighlightType.values().associate { type ->
            when (type) {
                ReviewQualityCheckState.HighlightType.QUALITY -> type to quality
                ReviewQualityCheckState.HighlightType.PRICE -> type to price
                ReviewQualityCheckState.HighlightType.SHIPPING -> type to shipping
                ReviewQualityCheckState.HighlightType.PACKAGING_AND_APPEARANCE -> type to appearance
                ReviewQualityCheckState.HighlightType.COMPETITIVENESS -> type to competitiveness
            }
        }
```
then the IDE would show any missing types

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
```diff
@@ -0,0 +1,68 @@
+package org.mozilla.fenix.shopping.middleware
```
t-p-white, line 1: missing license

t-p-white:
Just a heads-up we're missing licences for the new files

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
t-p-white: might be worth doc'ing the 'why' here

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
rahulsainani: No worries, thanks! This is a good idea to make sure the IDE/compiler can assist us. I've updated this in #3212

PR #3227, opened 2023-08-11
Bug 1848188 - Remove Nimbus flag for Juno Onboarding

gabrielluong:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1848188

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -308,6 +308,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             if (savedInstanceState == null) {
                 navigateToHome()
             }
+
             if (!shouldStartOnHome() && shouldNavigateToBrowserOnColdStart(savedInstanceState)) {
                 navigateToBrowserOnColdStart()
             } else {
@@ -317,6 +318,7 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             if (settings().showHomeOnboardingDialog && components.fenixOnboarding.userHasBeenOnboarded()) {
                 navHost.navController.navigate(NavGraphDirections.actionGlobalHomeOnboardingDialog())
             }
+
             showNotificationPermissionPromptIfRequired()
         }
 
@@ -413,10 +415,6 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
      * Show the pre permission dialog to the user once if the notification are not enabled.
      */
     private fun showNotificationPermissionPromptIfRequired() {
-        if (settings().junoOnboardingEnabled) {
-            return
-        }
-
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU &&
             !NotificationManagerCompat.from(applicationContext).areNotificationsEnabledSafe() &&
             settings().numberOfAppLaunches <= 1
--- a/fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -1658,20 +1658,16 @@ class Settings(private val appContext: Context) : PreferencesHolder {
     )
 
     /**
-     * Indicates if juno onboarding feature is enabled.
-     */
-    val junoOnboardingEnabled: Boolean
-        get() = FxNimbus.features.junoOnboarding.value().enabled
-
-    /**
      * Returns whether juno onboarding should be shown to the user.
+     *
+     * @param hasUserBeenOnboarded Boolean to indicate whether the user has been onboarded.
      * @param isLauncherIntent Boolean to indicate whether the app was launched on tapping on the
      * app icon.
      */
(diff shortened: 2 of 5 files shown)


fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1656,21 +1656,17 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
     /**
      * Returns whether juno onboarding should be shown to the user.
+     *
+     * @param hasUserBeenOnboarded Boolean to indicate whether the user has been onboarded.
      * @param isLauncherIntent Boolean to indicate whether the app was launched on tapping on the
      * app icon.
      */
     fun shouldShowJunoOnboarding(hasUserBeenOnboarded: Boolean, isLauncherIntent: Boolean): Boolean {
         return if (!hasUserBeenOnboarded && isLauncherIntent) {
             FxNimbus.features.junoOnboarding.recordExposure()
```
gabrielluong, line 1668: Wondering if we will still need this call? 

The plan is to land this change and uplift to Release since Juno will be our default onboarding experience. Currently, we have a rolling experiment that prefs ON the Juno Onboarding in Release. When we land this in Nightly and uplift this, we can follow up with Product to disable the Nimbus experiment that is keeping this on.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1656,21 +1656,17 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = false,
     )
 
-    /**
-     * Indicates if juno onboarding feature is enabled.
-     */
-    val junoOnboardingEnabled: Boolean
```
t-p-white, line 1662: would this flag still be useful for dev purposes? E.g 
```
get() = !Config.channel.isDebug
```

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
t-p-white: Based on some of the info below, my take is probably not 🤔 however it may be useful to check in with the experiments team

    https://experimenter.info/jetstream/jetstream#enrollment-vs-exposure
    https://experimenter.info/telemetry#exposure
    https://experimenter.info/desktop/exposure

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -310,8 +311,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
             if (settings().showHomeOnboardingDialog && components.fenixOnboarding.userHasBeenOnboarded()) {
                 navHost.navController.navigate(NavGraphDirections.actionGlobalHomeOnboardingDialog())
+            } else {
+                showNotificationPermissionPromptIfRequired()
             }
-            showNotificationPermissionPromptIfRequired()
```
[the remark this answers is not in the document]
t-p-white: I think this is still required to prompt for notifications, as there is no option on the Upgrade Onboarding dialog? @rahulsainani may have some thoughts on this though

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
gabrielluong: I don't believe so since the juno onboarding is already the default experience (eg, there's no alternative pathway).

t-p-white: The behaviour I thought this would replicate (that is being removed) is how we currently disable it for dev with the manifest:
```
    defaults:
      - channel: developer
        value:
          enabled: false
```
Having to step through onboarding for each install may be a bit of a pain point?

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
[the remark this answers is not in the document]
rahulsainani: I think it's safe to keep it as it was as the conditions for when both the dialogs display do not intersect. Upgrade Onboarding is shown to existing users and Notification pre–permission prompt is shown to users on their first session (who don't have the permission enabled). 

Saying that, I will make a mental note to discuss deleting the notification pre permission prompt as it's disabled by default anyway, in favour of the new onboarding.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
gabrielluong: We spoke about this offline:

My take on this is that we should consider a secret setting pref that allows you to always see the juno onboarding on startup. My assumption is that it’s harder to get the juno onboarding showing consistently for debugging purposes compare to dismissing it entirely since you only ever have to run through the onboarding once. This variable where you have to manually modify, rebuild and launch doesn’t seem like the most developer/qa friendly approach. Having the secret settings would mean you would never have to rebuild and anybody including testers can always just toggle the juno onboarding without having to clear cache/storage/reinstall if they are testing the contents of the onboarding. You can still manually set the secret setting pref to false via as well.

t-p-white (approved):
lgtm 👍

PR #3265, opened 2023-08-14
Bug 1848605 - Implement similar functionality as loadIntoView for Compose Top Sites

t-p-white:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1848605


android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -255,6 +262,47 @@ class BrowserIcons constructor(
         }
     }
 
+    @Composable
+    fun LoadIntoViewInternal(
```
t-p-white, line 266: this fun is essentially a hybrid of the `loadIntoViewInternal` fun above that is used in non-Compose Top Sites and the `BrowserIcons.Loader` that was used to previously load the icon by FavIcon

t-p-white:
@gabrielluong @MozillaNoah interested to get your thoughts on this. Updating FavIcon to use the `LoadIntoViewInternal` will affect more than just Top Sites, however it seems to me that they will be updated to all be consistent

android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
MozillaNoah, line 266: Let's rename this to ditch the `View` and make it more clear that this is Composable UI and not just a business logic function with the `@Composable` tag.

```suggestion
    fun LoadableImage(
```

MozillaNoah (commented):
I think this is looking good. I briefly tested this locally with synced tabs (as those utilize `Favicon` via `FaviconListItem`, and it seemed to be running well and perhaps a bit better than before.

Nice work!

jonalmeida:
tenor-2801692567

android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
```diff
@@ -255,6 +262,57 @@ class BrowserIcons constructor(
...
+     * Loads an icon using [BrowserIcons] into the given Composable [content]. Synchronous loading
+     * via an in-memory cache is attempted first, followed by an asynchronous load as a fallback.
+     *
+     * @param url The URL of the website an icon should be loaded for.
+     * @param iconSize The preferred size of the icon that should be loaded.
+     * @param isPrivate Whether this request for this icon came from a private session.
+     * @param content The Composable content block to render the icon.
+     */
+    @Composable
+    fun LoadableImage(
```
MozillaNoah, line 275: does this need to still live inside the `BrowserIcons` class? it'll make it more encapsulated as its own thing but also clean up the usage inside of `Favicon`

MozillaNoah (approved):
LGTM. Just one question about potentially moving a Composable function outside of a class.

titooan (approved):
👏

android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/BrowserIcons.kt
t-p-white: @MozillaNoah my reasoning to keep it in this inside the `BrowserIcons` class is that the fun is using a number of functions provided in the class:

- `private fun desiredSizeForRequest`
- `private fun loadIconMemoryOnly`
- `internal fun loadIconInternalAsync`

By keeping this in the same file it means fewer changes to the existing code which is my preference, wdyt?

MozillaNoah: ahh gotcha gotcha. I totally missed that several class functions were being used here. I think in an ideal world we'd have the Composable in its own file, decoupled from some of the specific nuances of `BrowserIcons`, but that is definitely way out of scope here, having given a full pass on that class.

This is good to go 👍

PR #3270, opened 2023-08-15
Bug 1848377 - Add secret setting to control the visibility of the Shopping feature

MozillaNoah:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1848377

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/ShoppingExperienceFeature.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/ShoppingExperienceFeature.kt
@@ -4,16 +4,28 @@
 
 package org.mozilla.fenix.shopping
 
-import org.mozilla.fenix.nimbus.FxNimbus
+import org.mozilla.fenix.utils.Settings
 
 /**
  * An abstraction for shopping experience feature flag.
  */
-class ShoppingExperienceFeature {
+interface ShoppingExperienceFeature {
 
     /**
      * Returns true if the shopping experience feature is enabled.
      */
-    val isEnabled
-        get() = FxNimbus.features.shoppingExperience.value().enabled
+    val isEnabled: Boolean
+}
+
+/**
+ * The default implementation of [ShoppingExperienceFeature].
+ *
+ * @property settings Used to check if the feature is enabled.
+ */
+class DefaultShoppingExperienceFeature(
+    private val settings: Settings,
+) : ShoppingExperienceFeature {
+
+    override val isEnabled
+        get() = settings.enableShoppingExperience
 }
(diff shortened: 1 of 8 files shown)


fenix/app/src/main/java/org/mozilla/fenix/shopping/ShoppingExperienceFeature.kt
```diff
@@ -1,19 +0,0 @@
...
-/**
- * An abstraction for shopping experience feature flag.
- */
-class ShoppingExperienceFeature {
-
-    /**
-     * Returns true if the shopping experience feature is enabled.
-     */
-    val isEnabled
-        get() = FxNimbus.features.shoppingExperience.value().enabled
```
rahulsainani, line 18: Could we just use `settings.enableShoppingExperience` here so when `Settings` is removed we only have to change this reference and none of the feature and test code has to change since it's abstracted?

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1789,6 +1789,14 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = FeatureFlags.composeTopSites,
     )
 
+    /**
+     * Indicates if the shopping experience feature is enabled.
+     */
+    var enableShoppingExperience by booleanPreference(
```
rahulsainani, line 1795: Can this be a val? Seems `SharedPreferenceUpdater` just updates the underlying preference and not this field so this is a read only property

rahulsainani (approved):
Overall it looks good. Left one suggestion and one bigger abstraction question for you to consider.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ShoppingExperienceFeature.kt
MozillaNoah: We certainly could, but `Settings` would still need to be passed into here and the test, right? Either it being a property of `ShoppingExperienceFeature` or a parameter of `isEnabled`?  

Unless `ShoppingExperienceFeature` was made into an interface instead 🤔

rahulsainani: Yeah, settings would be injected here, having an interface would be great 👍 

Tangential but sharing as it's related, some tests are intermittent because of [mockito](https://firefoxci.taskcluster-artifacts.net/dNyDrlhfTzKsp3S4-b7HRQ/0/public/reports/test/testFenixDebugUnitTest/classes/org.mozilla.fenix.shopping.ReviewQualityCheckFeatureTest.html#WHEN%20feature%20is%20enabled%20and%20selected%20tab%20is%20switched%20to%20not%20a%20product%20page%20THEN%20callback%20returns%20false) usage. @gabrielluong and I were thinking of changing it back to a function to be sure mockito doesn't fail here mocking a val. Not 100% sure if mocking a val is an issue here. Or we can use a fake for tests so we don't need to mock.

PR #3281, opened 2023-08-15
Bug 1845613 – restore hidden engines for unified search settings

mavduevskiy:
The patch is pretty simple. The new unified search settings screen, introduced [previously](https://github.com/mozilla-mobile/firefox-android/pull/2274), removed the ability to delete and restore the default engines. It shouldn't create a problem for new users, but if the engine has been deleted on the previous version, there is no way to add it back in the newer one.

What the patch does is restoring the list of the default engines when the new unified search settings screen is enabled.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1845613


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/SearchReducer.kt
```diff
@@ -191,3 +192,9 @@ private fun BrowserState.updateDisabledSearchEngineIds(
         ),
     )
 }
+
+private fun BrowserState.restoreHiddenSearchEngines(): BrowserState {
+    return search.hiddenSearchEngines.fold(this) { state, engine ->
+        state.showSearchEngine(SearchAction.ShowSearchEngineAction(engine.id))
+    }
+}
```
jonalmeida, line 200: This confused me for a while, but I don't see any easy change that would make it clearer. We're iterating over the list of `hiddenSearchEngines` and then inside that, iterating over it again.

Why not just pull what's inside `showSearchEngine` out (without the `find` bit), and use that here? We're doing two loops here I think over the same list when we just need to pass in the `SearchEngine` to the `copy` function.

Does that make sense?

android-components/components/feature/search/src/test/java/mozilla/components/feature/search/middleware/SearchMiddlewareTest.kt
```diff
@@ -1243,6 +1243,42 @@ class SearchMiddlewareTest {
         assertFalse(store.state.search.disabledSearchEngineIds.contains("bing"))
     }
 
+    @Test
+    fun `WHEN restore hidden search engines action THEN hidden engines are added back to bundled engines list`() {
```
jonalmeida, line 1247: This test (or a separate new one) should assert that what we put into the BrowserStore via the `SearchReducer` matches with what we end passing to the `MetadataStorage`.

These are two systems that need to always be in sync, so we should ensure we do not regress that behaviour when we are going through two separate code paths for transforming our search engine lists.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -361,6 +361,10 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
             components.core.contileTopSitesUpdater.startPeriodicWork()
         }
 
+        if (settings().enableUnifiedSearchSettingsUI) {
+            components.useCases.searchUseCases.restoreHiddenSearchEngines.invoke()
```
jonalmeida, line 365: This restore use case will run every time a user starts Fenix, right? Why do we need this to run all the time? From what I can see, this is a one-time change that we need to apply to correct users whose search engines were disabled.

jonalmeida (changes requested):
The change looks good, but I have a few open questions before we can land this. Nice work!

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/action/SearchActionTest.kt
```diff
@@ -222,6 +222,75 @@ class SearchActionTest {
...
+        assertEquals(3, store.state.search.regionSearchEngines.size)
+        assertEquals(0, store.state.search.hiddenSearchEngines.size)
+
+        assertEquals("google", store.state.search.regionSearchEngines[0].id)
+        assertEquals("bing", store.state.search.regionSearchEngines[1].id)
+        assertEquals("duckduckgo", store.state.search.regionSearchEngines[2].id)
+    }
+
+    @Test
+    fun `WHEN restore hidden search engines action GIVEN there are no hidden engines THEN hidden engines are added back to the bundled engine list`() {
```
jonalmeida, line 261:
```suggestion
    fun `WHEN restore hidden search engines action GIVEN there are no hidden engines THEN there are no changes`() {
```

android-components/components/feature/search/src/test/java/mozilla/components/feature/search/middleware/SearchMiddlewareTest.kt
```diff
@@ -1243,6 +1243,50 @@ class SearchMiddlewareTest {
         assertFalse(store.state.search.disabledSearchEngineIds.contains("bing"))
     }
 
+    @Test
+    fun `WHEN restore hidden search engines action THEN hidden engines are added back to bundled engines list`() = runTestOnMain {
+        val metadataStorage = SearchMetadataStorage(testContext, preferences = lazy { FakeSharedPreferences() }) // SearchMetadataStorage(testContext, preferences = lazy { FakeSharedPreferences() })
```
jonalmeida, line 1248:
```suggestion
        val metadataStorage = SearchMetadataStorage(testContext, preferences = lazy { FakeSharedPreferences() })
```

Unnecessary comment? 🙂

android-components/components/feature/search/src/test/java/mozilla/components/feature/search/middleware/SearchMiddlewareTest.kt
```diff
@@ -1243,6 +1243,50 @@ class SearchMiddlewareTest {
...
+            customStorage = CustomSearchEngineStorage(testContext, dispatcher),
+            metadataStorage = metadataStorage,
+        )
+        val store = BrowserStore(middleware = listOf(searchMiddleware))
+
+        store.dispatch(
+            SearchAction.SetRegionAction(
+                RegionState("US", "US"),
+            ),
+        ).joinBlocking()
```
jonalmeida, line 1261: Why are we setting the region in this test?

jonalmeida (approved):
This looks good to me! Could you answer/fixup the comments below and it should be good to land this next week.

android-components/components/feature/search/src/test/java/mozilla/components/feature/search/middleware/SearchMiddlewareTest.kt
mavduevskiy: to populate `store.state.search.regionSearchEngines`! it's a common strategy throughout the `SearchMiddlewareTest` test.

mavduevskiy:
@mergifyio backport releases_v118

PR #3298, opened 2023-08-16
Bug 1849009 – extended fullscreen mode to utilize the notch area

mavduevskiy:
### Screenshots taken on Pixel7 
 Old | New
 --- | --- 
 |  

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1849009


android-components/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/view/ActivityTest.kt
```diff
@@ -34,6 +35,7 @@ class ActivityTest {
 
     private lateinit var activity: Activity
     private lateinit var window: Window
+    private lateinit var layoutParams: WindowManager.LayoutParams
```
t-p-white, line 38: `private lateinit var windowInsetsController: WindowInsetsController` & `windowInsetsController = mock()` can be removed now 🙂

t-p-white (commented):
Works nicely 🙌 just a couple of minor comments from me and some CI failures to check out

mavduevskiy:
Noticed that phrasing "on devices with Android X and above" is a bit misleading, because we were setting just the X target and never verify if it works on newer API's. Running `ActivityTest` on API 30 is broken, as inset controller implementation does the transition with [animation](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/InsetsController.java#1003), without changing window flags. So all verifications don't work.

Possible solution could be:
`verify(windowInsetsController).hide(anyInt())`

but I am still thinking what to do with the test. The default ask is not 33, but 28, and as the functionality is vert API specific, it might be worth rewriting all of the tests. Maybe as a follow up. But we have a lot of "check if api X and above" that actually don't check the higher versions. I wonder why the sdk is so low, a lot of API changes are invisible.

mavduevskiy:
Remove flags assertion, because it moved away from directly setting them, and it was working just because of the [implementation](https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-main/core/core/src/main/java/androidx/core/view/WindowInsetsCompat.java?source=post_page---------------------------%2F%2F%2F%2F%2F%2F%2F&autodive=0%2F%2F%2F%2F%2F%2F%2F%2F%2F%2F#89) inside `WindowInsetsCompat` for the sdk level out tests are running on. What we would want to do is verify that `windowInsetsController.hide()` was called, but mocking the [static](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/Window.kt#41) function isn't an option.

android-components/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/view/Activity.kt
```diff
@@ -16,7 +17,8 @@ import androidx.core.view.WindowInsetsControllerCompat
 import mozilla.components.support.base.log.logger.Logger
 
 /**
- * Attempts to enter immersive mode - fullscreen with the status bar and navigation buttons hidden.
+ * Attempts to enter immersive mode - fullscreen with the status bar and navigation buttons hidden,
+ * expending itself into the notch area for devices running Android 28+.
```
t-p-white, line 21:
```suggestion
 * expending itself into the notch area for devices running API 28+.
```
or 
```suggestion
 * expending itself into the notch area for devices running Android version 28+.
```

Preferably suggestion 1 as it's consistent with the below reference to API 30+

android-components/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/view/ActivityTest.kt
```diff
@@ -69,25 +70,20 @@ class ActivityTest {
...
-    fun `check setAsImmersive sets the correct flags on devices with Android Q and below`() {
+    @Config(sdk = [Build.VERSION_CODES.P])
+    fun `GIVEN Android version P WHEN setAsImmersive is called THEN notch flags are set`() {
         activity.setAsImmersive()
 
-        verify(decorView).systemUiVisibility = View.SYSTEM_UI_FLAG_FULLSCREEN
-        verify(decorView).systemUiVisibility = View.SYSTEM_UI_FLAG_HIDE_NAVIGATION
-        verify(window.decorView, never()).setOnSystemUiVisibilityChangeListener(any())
+        verify(window).setFlags(WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS, WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS)
+        Assert.assertEquals(WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES, layoutParams.layoutInDisplayCutoutMode)
```
t-p-white, line 78: can import the `Assert` 🙂

t-p-white:
Hey Mike, I've been thinking about this testing problem, as this also came up during a previous patch on this file. I think it's important we test the full function behaviour but appreciate given the current set up this may not be possible with mock. I have an idea that may be useful.

We create something like this for use in `Activity`:
```
interface WindowInsetsControllerWrapper{
 fun getWindowInsetsControllerStub(window : Window) : WindowInsetsControllerCompat
}

class WindowInsetsControllerWrapperImpl : WindowInsetsControllerWrapper{
    override fun getWindowInsetsControllerStub(window : Window) =
        window.getWindowInsetsController()
}
```

```
internal fun Activity.setAsImmersive(controller: WindowInsetsControllerWrapper = WindowInsetsControllerWrapperImpl()) {
```


And in the tests we can do something like:
```
    @Test
    @Config(sdk = [Build.VERSION_CODES.P])
    fun `GIVEN Android version P WHEN setAsImmersive is called THEN notch flags are set`() {
        activity.setAsImmersive(windowWrapper)

        verify(window).setFlags(WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS, WindowManager.LayoutParams.FLAG_LAYOUT_NO_LIMITS)
        assertEquals(WindowManager.LayoutParams.LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES, layoutParams.layoutInDisplayCutoutMode)
        verify(windowWrapper).getWindowInsetsControllerStub(window)
        verify(windowInsetsController).hide(WindowInsetsCompat.Type.systemBars())
        verify(windowInsetsController).systemBarsBehavior = WindowInsetsControllerCompat.BEHAVIOR_SHOW_TRANSIENT_BARS_BY_SWIPE
    }
``` 
wdyt?

mavduevskiy:
from a slack discussion with @t-p-white:

Wrapping could be solid solution, but it will be the first case to introduce one. We probably should do it in a separate PR and have a team discussion about the approach.

As a side note, I have talked no @jonalmeida about it, and his take was we should refactor the function instead. If we can't test it, it's not written well. Maybe we should consider refactoring instead, because the `setAsImmersive` isn't actually setting any flags any more. The content of the function changed completely but never did the naming and structure, to accommodate the new compat api.

t-p-white (approved):
lgtm 🙂 would you mind linking a bug for the refactor so we have a record of it?

mavduevskiy:
sure! added [it](https://bugzilla.mozilla.org/show_bug.cgi?id=1851079) as a regression to the original.

mavduevskiy:
@mergifyio backport releases_v118

PR #3307, opened 2023-08-17
Bug 1848791 - Use min/maxTime for TimeStepPickerDialog

jackyzy823:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1848791


android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/ext/Calendar.kt
```diff
@@ -60,6 +67,6 @@ internal fun Calendar.minMonth(): Int = getMinimum(Calendar.MONTH)
 internal fun Calendar.maxMonth(): Int = getActualMaximum(Calendar.MONTH)
 internal fun Calendar.minDay(): Int = getMinimum(Calendar.DAY_OF_MONTH)
 internal fun Calendar.maxDay(): Int = getActualMaximum(Calendar.DAY_OF_MONTH)
-internal fun Calendar.minYear(): Int = getMinimum(Calendar.YEAR)
-internal fun Calendar.maxYear(): Int = getActualMaximum(Calendar.YEAR)
+internal fun Calendar.minYear(): Int = DEFAULT_MIN_YEAR
+internal fun Calendar.maxYear(): Int = DEFAULT_MAX_YEAR
```
MatthewTighe, line 71: I think we should leave these as they are. Different types of `Calendar`s may be able to specify different minimums and maximums. Code for time is notoriously difficult to get correct, so in this case I think we should rely on the standard.

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/widget/MonthAndYearPicker.kt
```diff
@@ -155,23 +162,19 @@ internal class MonthAndYearPicker @JvmOverloads constructor(
...
-        @VisibleForTesting
-        internal const val DEFAULT_MIN_YEAR = 1
-
         internal fun getDefaultMinDate(): Calendar {
             return now().apply {
-                month = Calendar.JANUARY
-                year = DEFAULT_MIN_YEAR
+                month = minMonth()
+                year = minYear()
+                day = minDay()
```
MatthewTighe, line 169: I think these changes are a bit out of scope for the patch, since we are switching `TimePickerDialogFragment` to not use these functions. That said, these changes look more correct to me. However, I am wondering if perhaps these were intended as a fallback default for when `maximumDate?.toCalendar()` in `TimePickerDialogFragment` was null. Perhaps @Amejia481 could weigh in? It's been a long time but it looks like he was the original author

MatthewTighe (changes requested):
Couple comments. I also need to test this local, but am having trouble navigating the site in the bug as I am not familiar with the language. Do you happen to have another example page I could try, or could you prep a small video walkthrough for me from landing on the top-level URL? I will mention this in the bug as well

jackyzy823:
@MatthewTighe 

I see , so just keeping this change in  `TimePickerDialogFragment` and dropping other changes sounds better. If you agree that , i will make the change.

----------------------------------

Another tiny suggestion: we could change ` getDefaultMinDate/ getDefaultMaxDate`   to ` getDefaultMinYearAndMonth/ getDefaultMaxYearAndMonth`  to make the function name not ambiguous.

MatthewTighe:
> Another tiny suggestion: we could change ` getDefaultMinDate/ getDefaultMaxDate` to ` getDefaultMinYearAndMonth/ getDefaultMaxYearAndMonth` to make the function name not ambiguous.

I believe the naming convention here is referring to the return type `Date` and not to the common word "date" (as in a specific day/month/year). Given that, I think we should keep the name as is for now but thank you for keeping an attentive eye!

MatthewTighe:
👋 sorry for the delay! I'm just getting back from PTO. I will circle back to finish this review today or tomorrow

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/widget/MonthAndYearPicker.kt
Amejia481: That latest change looks good to me!!

MatthewTighe (approved):

MatthewTighe:
bors try

MatthewTighe:
bors retry

PR #3398, opened 2023-08-27
Bug 1850296 - Add a Firefox Suggest Android component.

linabutler:
This component wraps the Suggest Rust component from Application Services, and provides a way for applications to access Firefox Suggest search suggestions for web and sponsored content.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1850296


linabutler:
Hi Android folks! 👋🏼

Back in May, my team went through the mobile project intake process for Enhanced Cross-Platform Suggest. The goal of this project is to support contextual search suggestions in Firefox for Android and iOS, and bring our mobile Firefox Suggest feature up to parity with Desktop. We volunteered to take on the engineering work, with oversight and guidance from your team. This PR is the first step toward that 😊

This is our first time building a new Android component, so higher-level feedback on the design and architecture would be wonderful, in addition to comments on the implementation! There are several existing components that use `WorkManager` in a very similar way, so I borrowed heavily from them—hopefully, it doesn't look too out-of-place 😅

The basic design is:

* An Application Services Rust component that talks to Remote Settings, downloads new suggestions, stores them in an SQLite database, and makes them available for querying. The `mozilla.appservices.suggest.SuggestStore` class encapsulates all this.
* A scheduler (`FxSuggestIngestionScheduler`) that uses a `WorkManager` worker to periodically download and persist ("ingest") new suggestions into the store.
* A suggestion provider (`FxSuggestSuggestionProvider`) that queries the store for matching suggestions as the user types into the awesomebar.

I didn't integrate the component into Fenix in this PR, but [here's a proof-of-concept](https://github.com/linabutler/firefox-android/commit/bf5747c7359588943a6a6acb1e1206ea75af62d6) of what that might look like eventually.

Thank you so much for your time, and please let me know if you have questions, concerns, or comments!

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/GlobalFxSuggestDependencyProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Opens a Suggest store that persists suggestions, icons, and metadata in a database at the
+     * given path.
+     *
+     * Your application's [onCreate][android.app.Application.onCreate] method should call this
+     * method once.
+     *
+     * @param databasePath The absolute path of the Suggest database.
+     */
+    fun openSuggestStore(databasePath: String) {
```
jonalmeida, line 33: In your [WIP fenix patch](https://github.com/linabutler/firefox-android/commit/bf5747c7359588943a6a6acb1e1206ea75af62d6#diff-c07dbf42136256cfd1eea0fe95204636f9327f15255ec1fab0843e18bd2aaec7R309-R311), we suppress a strict mode violation which uses an IO API working on the main thread.

We may need either:
1. Moving the entire dependency provider initialization to happen on an IO dispatcher, and silently catch the `requireNotNull` exception that we may throw below (and this seems odd as I write it out)
2. Waiting until initialization is complete, and only then adding the AwesomeBar provider.

jonalmeida (commented):
Drive-by: This looks awesome, thanks for putting it together! 🙂 

> I didn't integrate the component into Fenix in this PR, but [here's a proof-of-concept](https://github.com/linabutler/firefox-android/commit/bf5747c7359588943a6a6acb1e1206ea75af62d6) of what that might look like eventually.

If it helps, you could integrate this into the :sample-browser project to play around with it there and put it up as part of this patch, but that's not a requirement.

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/GlobalFxSuggestDependencyProvider.kt
MatthewTighe: To add a bit of detail to the `2` option, I would suggest:
1. make this function suspending 
2. convert this to a class that receives a `Context`
3. initialize the `suggestStore` as a lazy deferred property
```
class GlobalFxSuggestDependencyProvider(
    context: Context, // val isn't required here, and will mean the class drops the reference to the god object after the property is computed
    private val coroutineScope = CoroutineScope(Dispatchers.IO),
) {

    private val suggestStore: Deferred<SuggestStore> = coroutineScope.async(start = CoroutineStart.Lazy) {
        SuggestStore(context.getDatabasePath(...)
    }

    suspend fun requireSuggestStore(): SuggestStore {
        suggestStore.await()
    }
}
```

It seems like every usage of the store in your WIP patch is either already in a suspending context, or we can simply launch a new coroutine and handle the result asynchronously as Jonathan is suggesting.

MatthewTighe (commented):
This is looking great, would like to re-iterate Jonathan's sentiment that we are very happy to have this contribution!

I have a few optional small nits, but otherwise this is looking good to me. I think you've run into an awkward design constraint since the main component needs an asynchronously computed property to initialize. AFAIK, we don't have (and could probably really use) an existing pattern or abstraction for that kind of thing, but I think what you've got here is very close to workable.

I'll circle back for another pass tomorrow! My brain is still catching up from last week 😅

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/GlobalFxSuggestDependencyProvider.kt
linabutler: Ooh, I like that approach, @MatthewTighe! It's a little unfortunate that we try to avoid main-thread I/O in `SuggestStore` by having it open the database connections lazily, but Android's `Context.getDatabasePath` still ends up doing I/O to ensure the databases directory exists 😞 Oh, well!

linabutler: @MatthewTighe Hmm, I just noticed that your suggestion makes `GlobalFxSuggestDependencyProvider` a class. Where would we store an instance of that class, so that both the background worker and suggestion provider can access it?

It looks like Fenix's `FenixApplication.components` would be a good place—except the worker and suggestion provider are in Android Components code, so I don't think we have a concrete application context here that we can use to store additional fields, right?

I wonder if we can work around this with double indirection, though: keep `GlobalFxSuggestDependencyProvider` an object, as now, but have it hold a reference to a `FxSuggestLazyStore` wrapper that does the lazy initialization of the `SuggestStore`.

Something like:

```kt
object GlobalFxSuggestDependencyProvider {
    internal var lazyStore: FxSuggestLazyStore? = null

    fun initialize(lazyStore: FxSuggestLazyStore) {
        self.lazyStore = lazyStore
    }

    fun requireLazyStore(): FxSuggestLazyStore {
        // ...
    }
}

class FxSuggestLazyStore(
    context: Context,
    private val coroutineScope = CoroutineScope(Dispatchers.IO),
) {
    private val store: Deferred<SuggestStore> = coroutineScope.async(start = CoroutineStart.Lazy) {
        SuggestStore(context.getDatabasePath(...))
    }

    suspend fun get(): SuggestStore = store.await()
}
```

...And then Fenix would do something like:

```kt
val lazyStore = FxSuggestLazyStore(context)
GlobalFxSuggestDependencyProvider.initialize(lazyStore)
```

WDYT? Is there a simpler way? 😅

linabutler: I pushed 4378d42468765b3a624070628f8ef085dba7790d to show how this might look; feedback welcome! The new `FxSuggestStorage` class (what I called `FxSuggestLazyStore` in my comment above) also manages separate coroutine scopes for reading and writing.

android-components/components/feature/fxsuggest/src/test/java/mozilla/components/feature/fxsuggest/FxSuggestIngestionWorkerTest.kt
```diff
@@ -21,64 +22,63 @@ import org.junit.Assert.assertThrows
...
 
-        verify(suggestStore).ingest(any())
+        verify(storage).ingest(any())
         assertEquals(ListenableWorker.Result.retry(), result)
     }
 
     @Test
-    fun workShouldRethrow() {
-        whenever(suggestStore.ingest(any())).then {
+    fun workShouldRethrow() = runTest {
```
linabutler, line 72: This test is failing now—`expected java.io.IOException to be thrown, but nothing was thrown`; the `IOException` either isn't thrown, or doesn't bubble up—and I can't figure out a way to make it pass! 😅

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
```diff
@@ -0,0 +1,68 @@
...
+    // We expect almost all Suggest storage operations to be reads, with infrequent writes. Our
+    // coroutine scopes reflect this workload: we use the dynamically scaled I/O dispatcher for
+    // reads, and a short-lived thread pool for writes.
+    internal var readScope = CoroutineScope(IO)
+    internal var writeScope = CoroutineScope(Cached)
+
+    /**
+     * Queries the store for suggestions.
+     */
+    suspend fun query(query: SuggestionQuery): List<Suggestion> = withContext(readScope.coroutineContext) {
```
linabutler, line 41: `PlacesStorage` goes a step further and redefines the data types, too, but I don't think we need to go that far. I think using these in our public API _does_ mean that `ComponentsDependencies.mozilla_suggest` has to become an `api` dependency, not an `implementation` one, though.

MatthewTighe (commented):
Alright, I think I've got a better understanding of the design space today than I did yesterday. Unfortunately I think that means I may have misled you a bit with my review yesterday, sorry about that! I am happy to chat or pair offline to help get this over the finish line if you are under any time constraints

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
```diff
@@ -0,0 +1,47 @@
...
+import mozilla.appservices.suggest.SuggestionQuery
+import mozilla.components.support.base.utils.NamedThreadFactory
+import java.util.concurrent.Executors
+
+class FxSuggestStorage(
+    context: Context
+) {
+    internal val store: Lazy<SuggestStore> = lazy {
+        SuggestStore(context.getDatabasePath(DATABASE_NAME).absolutePath)
+    }
```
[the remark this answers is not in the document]
MatthewTighe: Ignore this, I think this is all on the right path now that I'm looking at `PlacesStorage` in comparison

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
```diff
@@ -0,0 +1,68 @@
...
+ * @property context The Android application context.
+ */
+class FxSuggestStorage(
+    context: Context,
+) {
+    // Lazily initializes the store on first use. `Context.getDatabasePath()` does I/O, so this
+    // value must be accessed in the read or write scope.
+    internal val store: Lazy<SuggestStore> = lazy {
+        SuggestStore(context.getDatabasePath(DATABASE_NAME).absolutePath)
+    }
```
MatthewTighe, line 30: Using the `by` delegate keyword here should remove the need for all the explicit `.value` unwrapping
```suggestion
    internal val store: SuggestStore by lazy {
        SuggestStore(context.getDatabasePath(DATABASE_NAME).absolutePath)
    }
```

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
MatthewTighe: I would definitely prefer to not introduce additional types

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
MatthewTighe: Though I see that it might be necessary for `.cancelReads`

android-components/components/feature/fxsuggest/src/test/java/mozilla/components/feature/fxsuggest/FxSuggestIngestionWorkerTest.kt
MatthewTighe: I think this should do it:

```
        assertThrows(IOException::class.java) {
            runBlocking {
                worker.startWork().await()
            }
        }
```

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestIngestionWorker.kt
```diff
@@ -0,0 +1,42 @@
...
+internal class FxSuggestIngestionWorker(
+    context: Context,
+    params: WorkerParameters,
+) : CoroutineWorker(context, params) {
+    private val logger = Logger("FxSuggestIngestionWorker")
+
+    override suspend fun doWork(): Result {
+        logger.info("Ingesting new suggestions")
+        val storage = GlobalFxSuggestDependencyProvider.requireStorage()
+        try {
```
MatthewTighe, line 29: I think we'll need a `return try` here.

MatthewTighe (commented):
I think this is about ready for an approving review - would just like to take another look first thing tomorrow with fresh eyes and give you the opportunity to resolve the CI issues. It would be excellent if  your WIP patch could be updated as well to ensure everything is functioning as intended

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
jonalmeida: > I think using the Application Services Suggest types in our public API _does_ mean that `ComponentsDependencies.mozilla_suggest` has to become an `api` dependency, not an `implementation` one, though.

This is fine. Other examples in [Places](https://searchfox.org/mozilla-mobile/rev/e45f12112cf8f679691f83ac2dfa2b46919887f3/firefox-android/android-components/components/browser/storage-sync/build.gradle#30).

android-components/components/feature/fxsuggest/src/test/java/mozilla/components/feature/fxsuggest/FxSuggestIngestionWorkerTest.kt
linabutler: Omg, thank you! I completely forgot that `startWork` is a suspend function, but `assertThrows` isn't (in JUnit 4; it looks like JUnit 5 [supports suspend functions out of the box](https://github.com/junit-team/junit5/issues/1851)) 🎉

linabutler:
Here's the Compose sample browser showing a Firefox Suggest web suggestion! 🚀

Screenshot of the awesomebar showing a Wikipedia suggestion for 'cali'

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
linabutler: We chatted about this on Zoom yesterday, but yep, that's exactly why I used `Lazy<SuggestStore>` instead of delegation—I didn't want `cancelReads` to trigger initialization 😄 (In practice, though, I think it'll always be initialized by the time we call `cancelReads`).

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestIngestionWorker.kt
```diff
@@ -0,0 +1,42 @@
...
+ * @param params Parameters for this worker's internal state.
+ */
+internal class FxSuggestIngestionWorker(
+    context: Context,
+    params: WorkerParameters,
+) : CoroutineWorker(context, params) {
+    private val logger = Logger("FxSuggestIngestionWorker")
+
+    override suspend fun doWork(): Result {
+        logger.info("Ingesting new suggestions")
```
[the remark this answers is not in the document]
linabutler: Not in this case; I added this logging just to have some breadcrumbs in logcat! Our plan is to add proper Glean telemetry in a follow-up—how does that sound?

linabutler:
Alrighty, I think this is ready for capital-R Review! Thanks so much for all your guidance, @MatthewTighe! Here's a narrative summary of the changes since last time:

* Added an `FxSuggestStorage` class that wraps `mozilla.appservices.suggest.SuggestStore`. This wrapper takes care of initializing the store off the main thread, and wraps the synchronous `SuggestStore` in a nice coroutine API. It has two coroutine scopes (for reading and writing) that both target the I/O dispatcher, and lets us cancel reads without affecting writes. I did explore inlining the wrapper into `GlobalFxSuggestDependencyProvider`, but it made mocking in the tests so much harder.
* Folded in all your suggestions for more idiomatic Kotlin—thanks!
* Wired up the new component to the Compose sample browser, and verified that it can ingest and show suggestions in the awesomebar.

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
```diff
@@ -0,0 +1,93 @@
...
+                    it.toUByteArray().asByteArray().toBitmap()
+                },
+                title = "${details.fullKeyword} — ${details.title}",
+                description = if (details.isSponsored) {
+                    resources.getString(R.string.sponsored_suggestion_description)
+                } else {
+                    null
+                },
+                onSuggestionClicked = {
+                    loadUrlUseCase.invoke(details.url)
```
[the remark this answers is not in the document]
linabutler: Perfect, filed https://bugzilla.mozilla.org/show_bug.cgi?id=1851037 as a good first bug for a new contributor!

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestIngestionWorker.kt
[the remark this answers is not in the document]
MatthewTighe: Yep, sounds good!

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * An [AwesomeBar.SuggestionProvider] that returns Firefox Suggest search suggestions.
+ */
+class FxSuggestSuggestionProvider(
+    private val resources: Resources,
+    private val loadUrlUseCase: SessionUseCases.LoadUrlUseCase,
+    private val includeSponsoredSuggestions: Boolean,
+    private val includeNonSponsoredSuggestions: Boolean,
+    private val suggestionsHeader: String? = null,
+) : AwesomeBar.SuggestionProvider {
```
MatthewTighe, line 24: I think it would be helpful to add kdoc `@param` entries for all of these as well (or at least the non-boolean types)

android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
```diff
@@ -192,6 +192,7 @@ object ComponentsDependencies {
     val mozilla_sync_manager = "${ApplicationServicesConfig.groupId}:syncmanager:${ApplicationServicesConfig.version}"
     val mozilla_push = "${ApplicationServicesConfig.groupId}:push:${ApplicationServicesConfig.version}"
     val mozilla_remote_tabs = "${ApplicationServicesConfig.groupId}:tabs:${ApplicationServicesConfig.version}"
+    val mozilla_suggest = "${ApplicationServicesConfig.groupId}:suggest:${ApplicationServicesConfig.version}"
```
MatthewTighe, line 195: Not blocking on this, but curious about the naming convention difference between FxSuggest and MozillaSuggest. Assuming the name here is just following precedent, but it made me wonder if MozillaSuggest would make long-term sense for the code as well

MatthewTighe (approved):
This is massive, awesome work! Had just a couple small things but nothing blocking so stamping approval

android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
linabutler: Hmmm, I'd almost be tempted to go the other way. "Suggest" is the name of the Application Services component, and the feature is called "Firefox Suggest".

My preferred name would be `mozilla_appservices_suggest`, but none of our other Application Services components use that convention...yet? 😅 WDYT?

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
linabutler: Oops—done! 📝

linabutler:
Thank you so much for reviewing and working with me on this, @MatthewTighe! 🏆

android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
linabutler: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1851213.

PR #3414, opened 2023-08-29
Bug 1847031: add image mime types so that pngs can also be uploaded

Alexandra-Virvara:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1847031

### DIFF
--- a/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/MimeType.kt
+++ b/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/MimeType.kt
@@ -108,22 +108,35 @@ internal sealed class MimeType(
 
         override fun shouldCapture(mimeTypes: Array<out String>, capture: File.FacingMode) = false
 
+        @Suppress("NestedBlockDepth")
         override fun buildIntent(context: Context, request: File) =
             Intent(ACTION_GET_CONTENT).apply {
                 type = "*/*"
                 addCategory(CATEGORY_OPENABLE)
                 putExtra(EXTRA_LOCAL_ONLY, true)
                 if (request.mimeTypes.isNotEmpty()) {
+                    var shouldAddImageTypes = false
                     val types = request.mimeTypes
                         .map {
+                            shouldAddImageTypes = shouldAddImageTypes || it.startsWith("image")
+
                             if (it.contains("/")) {
                                 it
                             } else {
                                 mimeTypeMap.getMimeTypeFromExtension(it) ?: "*/*"
                             }
+                        }.toMutableList()
+
+                    if (shouldAddImageTypes) {
+                        val imageTypes =
+                            arrayOf("image/jpeg", "image/png", "image/gif", "image/bmp")
+                        for (imageType in imageTypes) {
+                            if (!types.contains(imageType)) {
+                                types += imageType
+                            }
                         }
-                        .toTypedArray()
-                    putExtra(EXTRA_MIME_TYPES, types)
+                    }
+                    putExtra(EXTRA_MIME_TYPES, types.toTypedArray())
                 }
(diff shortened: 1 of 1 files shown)


MatthewTighe (commented):
I see that there is a `MimeType.Image`. Why is that not working correctly?

Alexandra-Virvara:
> I see that there is a `MimeType.Image`. Why is that not working correctly?

The promptRequest file to be uploaded contains image/jpeg & video/* mime types (because of the video/* mime type, MimeType.Image won't be chosen when creating the intent). The intent that will be used when creating the file chooser will also accept files with the same mime types, so it will list only jpeg files or videos as possible files to upload

MatthewTighe:
Looking at the code a little more deeply, I see that `FilePicker::handleFileRequest` seems to be building separate intents for every mime type we get from the `promptRequest`, and then firing the [last intent that was built](https://searchfox.org/mozilla-mobile/rev/e45f12112cf8f679691f83ac2dfa2b46919887f3/firefox-android/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt#99). 

It seems like we may want to just build a single intent that combines each mime type, rather than building separate intents for each mime type and then inspecting the file to add additional data into every intent (as this patch is currently doing). I think that would add clarity to the use case by demonstrating that a user is intending to pick a file that could be one of several different types. 

How does that sound? Am I understanding the problem space correctly?

Alexandra-Virvara:
> Looking at the code a little more deeply, I see that `FilePicker::handleFileRequest` seems to be building separate intents for every mime type we get from the `promptRequest`, and then firing the [last intent that was built](https://searchfox.org/mozilla-mobile/rev/e45f12112cf8f679691f83ac2dfa2b46919887f3/firefox-android/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt#99).
> 
> It seems like we may want to just build a single intent that combines each mime type, rather than building separate intents for each mime type and then inspecting the file to add additional data into every intent (as this patch is currently doing). I think that would add clarity to the use case by demonstrating that a user is intending to pick a file that could be one of several different types.
> 
> How does that sound? Am I understanding the problem space correctly?

yes, it is correct

MatthewTighe:
I'm still not sure I'm understanding the underlying issue. It looks like we iterate through every mime type, which should include `MimeType.Image` and `MimeType.Video`. If the file has both image/jpeg & video/* mime types, I would expect both to be built at:
```
// in FilePicker::handleFileRequest
type.buildIntent(container.context, promptRequest)?.also {
    saveCaptureUriIfPresent(it)
    intents.add(it)
}

// and then combined here:
val lastIntent = intents.removeAt(intents.lastIndex)
val chooser = Intent.createChooser(lastIntent, null).apply {
    putExtra(EXTRA_INITIAL_INTENTS, intents.toTypedArray())
}
```

Why is that not the case? I am just a little concerned that we will end up creating lots of logical branches here if we need to keep adding special cases to `buildIntent`, and would prefer that we find a more generalized solution.

Alexandra-Virvara:
I did a deeper investigation on this issue, including tests on other browsers, as well.
The conclusion is that there isn't a bug, the behaviour is the expected one 

Usecase: upload a .png file on FB. When trying to upload an image, this is what happens on different browsers:

Mozilla: the .png files are greyed out as inactive and you can choose only .jpeg images
Opera: the .png files are shown as active, but after they are being chosen for upload, an error message appears saying that you should choose a file that is smaller in size or of another type
Chrome: the .png files are shown as active, you can choose them and upload them. But if you download them again after they're being uploaded, the file's extension is .jpeg. So Chrome internally converts the .pngs to .jpegs.

If we want to apply Chrome's approach, we should probably create another task and add the implementation.

PR #3459, opened 2023-09-01
Bug 1819737 - Added extensions process crash handling

zmckenney:
Requires the changes in GV for [D184071](https://phabricator.services.mozilla.com/D184071). This enables the GV extension process delegates to pass through to Fenix and the UI thread. The dialog UI shown to the user will be added in [Bug 1846979](https://bugzilla.mozilla.org/show_bug.cgi?id=1846979). Once both PR's are landed this will enable the user to decide whether to continue retrying  when extensions are crashing or to continue without them. 

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1819737


android-components/components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -174,6 +174,7 @@ object WebExtensionSupport {
         onSelectTabOverride: ((WebExtension?, String) -> Unit)? = null,
         onUpdatePermissionRequest: onUpdatePermissionRequest? = { _, _, _, _ -> },
         onExtensionsLoaded: ((List<WebExtension>) -> Unit)? = null,
+        onDisabledExtensionProcessSpawning: (() -> Unit)? = null,
     ) {
         this.onUpdatePermissionRequest = onUpdatePermissionRequest
         this.onExtensionsLoaded = onExtensionsLoaded
```
Amejia481, line 180: **No a blocker but It would be nice to investigate**: Noticed for all other callbacks we are also creating an [internal reference](https://github.com/mozilla-mobile/firefox-android/blob/9e5d0037b4e0afe5f02e5d56b0ac36e66c41ca62/android-components/components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt#L63-L65), don't remember exactly why,  I think it will be good to double if we don't need to follow the same pattern here.

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -670,6 +670,9 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
                     subscribeForNewAddonsIfNeeded(components.supportedAddonsChecker, extensions)
                 },
                 onUpdatePermissionRequest = components.addonUpdater::onUpdatePermissionRequest,
+                onDisabledExtensionProcessSpawning = {
+                    // https://bugzilla.mozilla.org/show_bug.cgi?id=1846979
+                },
```
Amejia481, line 675: Nit: As we are not providing any functionality here, maybe we could add this as part of [Bug 1846979](https://bugzilla.mozilla.org/show_bug.cgi?id=1846979). So that patch will contain everything related to its behaviours.

android-components/components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
```diff
@@ -294,6 +295,10 @@ object WebExtensionSupport {
                     store.dispatch(WebExtensionAction.UninstallAllWebExtensionsAction)
                     registerInstalledExtensions(store, runtime)
                 }
+
+                override fun onDisabledExtensionProcessSpawning() {
+                    onDisabledExtensionProcessSpawning?.invoke()
```
Amejia481, line 300: I think we could add a this on `WebExtensionSupportTest` to verify this behaviour :)

Amejia481 (approved):
Awesome job 🥇 !

I added some comments and suggestions but nothing blocking, feel free to land when you feel ready!

android-components/components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
zmckenney: Good catch! Now that I have (PR 3472)[https://github.com/mozilla-mobile/firefox-android/pull/3472] I can actually remove this since we'll just send a `BrowserAction` to the store. This will work the same as `onExtensionListUpdated()` in dispatching to the store from within this class.

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
zmckenney: agreed and also no longer needed after [PR 3472](https://github.com/mozilla-mobile/firefox-android/pull/3472) 😄

android-components/components/support/webextensions/src/main/java/mozilla/components/support/webextensions/WebExtensionSupport.kt
zmckenney: agreed, will add

willdurand (approved):
I think I am fine with merging this as is, no need to bundle the two PRs into a single one given that we have different bugs.

willdurand:
I went ahead and rebased this PR to get it landed soon (because of Labour Day)

willdurand:
@Mergifyio backport releases_v118

PR #3467, opened 2023-09-01
Bug 1844960 - Bookmark suggestions should appear unconcerned with the search suggestions setting.

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1844960


iorgamgabriel:
https://github.com/mozilla-mobile/firefox-android/assets/8118371/8a379145-1141-404c-a9d2-3978575530da

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -192,14 +192,16 @@ interface AwesomeBar {
      * @property providers The list of [SuggestionProvider]s in this group.
      * @property priority An optional priority for this group. Decides the order of this group
      * in the AwesomeBar suggestions. Group having the highest integer value will have the highest priority.
+     * By default, it should have -1 priority. If AwesomeBar.Suggestion list is not empty group.priority
+     * will be the first suggestion score from the list.
```
gabrielluong, line 196: To reviewers, I think we need another on this comment since I am not quite sure I am understanding it.

```suggestion
     * By default, it should have -1 priority. If [Suggestion] list is not an empty group, priority
     * will be the first suggestion score from the list.
```

iorgamgabriel: Done

rocketsroger (approved):

mavduevskiy:
I find this PR could benefit from providing a bit more context about the problem, explaining the solution (as gl, I am not quite understanding it) and also providing screen shorts of before/after – to show case a broken and the right behaviour.

android-components/components/compose/awesomebar/src/test/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionFetcherTest.kt
```diff
@@ -66,4 +66,21 @@ class SuggestionFetcherTest {
...
+        val fetcher = spy(SuggestionFetcher(listOf(providerGroup), null))
+
+        fetcher.processResultFrom(
+            group = providerGroup,
+            provider = provider,
+            suggestions = emptyList(),
+            profilerStartTime = null,
+        )
+
+        assertEquals(providerGroup.priority, -1)
```
mavduevskiy, line 84: expected and actual objects are swapped, should be `-1, providerGroup.priority`

mavduevskiy (changes requested):
I find this PR could benefit from providing a bit more context about the problem, explaining the solution (as gl, I am not quite understanding the comment) and also providing screen shorts of before/after – to show case a broken and the right behaviour.

mavduevskiy:
It looks like [using](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/AwesomeBar.kt#99) sorted map is problematic. It will be removing items from the list, if it finds equal while comparing, and I am not sure that was ever the intended behaviour. Pinging @sarah541 for more context on priority.

We either should guarantee unique priority values for suggestion provider groups we are using, or use a different sorting logic – so we won’t be risking throwing away elements. 

If we simply change the default value to -1, can we guarantee that we won't be loosing elements when two provider groups have the default priority?

As a side note, the bug that OP described is caused by the fact that `SotrtedMap` removes items it considers as duplicate in order of insertion. So if we had two keys `appProvider` (2 suggestions) and `searchProvider`(0 suggestions) in a sorted map, and `compareByDescending`  determines that providers have same priority, we will end up with a pair of `appProvider`(0 suggestions) :)

android-components/components/compose/awesomebar/src/test/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionFetcherTest.kt
iorgamgabriel: Done

iorgamgabriel:
Yes, the problem is from here 
val suggestions = derivedStateOf { fetcher.state.value }.value.toSortedMap(compareByDescending { it.priority }). 
If a provider group has priority -1, it shouldn't appear because the suggestions list is empty. The problem was that priority by default was 0 and  
 if (updatedSuggestions.isNotEmpty()) {
            group.priority = updatedSuggestions[0].score
 } 
updatedSuggestions[0].score was the same 0.

mavduevskiy:
Yeah, so we had two providers with the same priority 0 and `toSortedMap(compareByDescending)` would mess them up. If we change the default, can we guarantee that we won't ever have two provider groups with the same priority? Can we write tests for it? If more items are added to the search results, will we display it? Generally speaking, do we want to remove duplicates from this list of providers?

We might need to refactor [this](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/AwesomeBar.kt#99). Meanwhile, if the solution we are going for is to remove the empty lists from the equation, have you considered just filtering empty providers out from the state? It feels less hacky to me than changing the default :)

iorgamgabriel:
I still need to modify the default because updatedSuggestions[0].score starts from 0 , the same like the default one.

mavduevskiy (approved):
talked with @iorgamgabriel, agreed that the approach is good enough, but we should revisit [sorting](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/AwesomeBar.kt#99) as it is a potential source of issues.

mavduevskiy:
a shower thought: another way we could approach the problem is to stop removing duplicate priorities.

`val suggestions = derivedStateOf { fetcher.state.value }.value.toList().sortedByDescending { it.first.priority }.toMap()`

1. It looks like removing duplicates was never intended, there are no tests for that, no specific logic for handling duplicates, default handling of `SortedMap` is definitely not what we want.
2. We have this code in production for months, and the only edge case where groups have the same priority is with disabled search suggestions. 
3. The Awesome bar component doesn't have control over provider priorities, so it should be versatile in handling whatever is been thrown into it. I think, since Suggestions are designed to give the user an ability to order items in a desired way by providing proper priority, the Awesomebar should just display all the suggestions with the same priority in one bucket. Spontaneously removing suggestions in that bucket looks like a buggy behaviour. 

What do you think, @iorgamgabriel ?

iorgamgabriel:
@mavduevskiy Done. I made the changes.

mavduevskiy (commented):
except one small comment, lgtm!

iorgamgabriel:
@Mergifyio backport releases_v118

PR #3635, opened 2023-09-14
Bug 1853077 - Update Adjust to version 4.35.1

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1853077


jonalmeida:
I would be extra cautious about updating Adjust because of it's affect on our telemetry. I'll try to find another engineer also familiar with Adjust that would like to pair on reviewing this patch.

jonalmeida:
Non-blocker: I see that version [v4.35.0](https://github.com/adjust/android_sdk/releases/tag/v4.35.0) has the option for `setFinalAttributionEnabled` "to indicate if only final attribution is needed in attribution callback (by default attribution callback return intermediate attribution as well before final attribution if not enabled with this setter method)."

This is fine for now to upgrade, but I'm not sure if we need to adjust (hah!) our config to use this new behaviour for any experiment we are running?

cc: @MatthewTighe

jonalmeida (approved):
I think this is fine to land. Thanks!

PR #3650, opened 2023-09-14
Bug 1852802 - Prevents search suggestion prompt in private mode when felt privacy is enabled

boek:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1852802

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
@@ -163,15 +163,23 @@ class SearchDialogController(
                     settings.shouldShowSearchShortcuts,
             ),
         )
-        fragmentStore.dispatch(
-            SearchFragmentAction.AllowSearchSuggestionsInPrivateModePrompt(
-                text.isNotEmpty() &&
-                    activity.browsingModeManager.mode.isPrivate &&
-                    settings.shouldShowSearchSuggestions &&
-                    !settings.shouldShowSearchSuggestionsInPrivate &&
-                    !settings.showSearchSuggestionsInPrivateOnboardingFinished,
-            ),
-        )
+
+        // For felt private browsing mode we're no longer going to prompt the user to enable search
+        // suggestions while using private browsing mode. The preference to enable them will still
+        // remain in settings.
+        val isFeltPrivacyEnabled = settings.feltPrivateBrowsingEnabled
+
+        if (!isFeltPrivacyEnabled) {
+            fragmentStore.dispatch(
+                SearchFragmentAction.AllowSearchSuggestionsInPrivateModePrompt(
+                    text.isNotEmpty() &&
+                        activity.browsingModeManager.mode.isPrivate &&
+                        settings.shouldShowSearchSuggestions &&
+                        !settings.shouldShowSearchSuggestionsInPrivate &&
+                        !settings.showSearchSuggestionsInPrivateOnboardingFinished,
+                ),
+            )
+        }
     }
 
     override fun handleUrlTapped(url: String, flags: LoadUrlFlags) {
--- a/fenix/app/src/test/java/org/mozilla/fenix/search/SearchDialogControllerTest.kt
+++ b/fenix/app/src/test/java/org/mozilla/fenix/search/SearchDialogControllerTest.kt
@@ -426,6 +426,30 @@ class SearchDialogControllerTest {
     }
 
     @Test
+    fun `WHEN felt privacy is enabled THEN do not dispatch AllowSearchSuggestionsInPrivateModePrompt`() {
+        every { settings.feltPrivateBrowsingEnabled } returns true
+
+        val text = "mozilla"
+
+        createController().handleTextChanged(text)
+        val actionSlot = mutableListOf<SearchFragmentAction>()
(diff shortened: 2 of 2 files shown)


fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
```diff
@@ -153,15 +153,24 @@ class SearchDialogController(
...
-            ),
-        )
+
+        // For felt private browsing mode we're no longer going to prompt the user to enable search
+        // suggestions while using private browsing mode. The preference to enable them will still
+        // remain in settings.
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=1852802
+        val isFeltPrivacyEnabled = settings.feltPrivateBrowsingEnabled
+
+        if (!isFeltPrivacyEnabled) {
```
mavduevskiy, line 163: This got me thinking... We are putting `settings.feltPrivateBrowsingEnabled` outside of the `AllowSearchSuggestionsInPrivateModePrompt` constructor, where the rest of the settings go. Shouldn't all of the settings be outside then? Settings can't be changed between textChanges, the user has to leave the screen to change the settings. It looks like right now we are firing a lot of `action(show == false)` without ever firing `action(show == true)`. Possible follow up.

fenix/app/src/test/java/org/mozilla/fenix/search/SearchDialogControllerTest.kt
```diff
@@ -362,6 +363,116 @@ class SearchDialogControllerTest {
         verify { store.dispatch(SearchFragmentAction.ShowSearchShortcutEnginePicker(false)) }
     }
 
+    @Test
+    fun `WHEN not in private mode THEN do not show prompt`() {
```
mavduevskiy, line 367: While I do believe we should add tests for checking the flags inside the [action](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt#157), it might be out of scope of this PR. Generally speaking, when adding `val isFeltPrivacyEnabled = settings.feltPrivateBrowsingEnabled`, we would expect the store not to receive any of `AllowSearchSuggestionsInPrivateModePrompt` as the feature should be disabled, right? In the case, capturing the list of events that were dispatched and making sure that there are (or aren't) events of certain type would do the trick. Just wwo tests should be enough.

The next question should be, wouldn't we want to write the tests for other flags in the same manner? It looks like if the suggestions are disabled or we are not in the private mode, no events of `llowSearchSuggestionsInPrivateModePrompt` should be there. It looks like only the tests for empty and not empty text input would actually care about `AllowSearchSuggestionsInPrivateModePrompt` boolean value.

And that would require a bit of refactoring of the logic as well. A possible follow up?👀

fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
```diff
@@ -153,15 +153,24 @@ class SearchDialogController(
...
-                    settings.shouldShowSearchSuggestions &&
-                    !settings.shouldShowSearchSuggestionsInPrivate &&
-                    !settings.showSearchSuggestionsInPrivateOnboardingFinished,
-            ),
-        )
+
+        // For felt private browsing mode we're no longer going to prompt the user to enable search
+        // suggestions while using private browsing mode. The preference to enable them will still
+        // remain in settings.
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=1852802
```
mavduevskiy, line 160: the PR comment has the bug number, do you think it's not enough?

mavduevskiy (changes requested):
the change is looking good! some adjustments to the tests, and I think we can land it.

do you think we should create a follow up to refactor the condition inside the [action](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt#157)?

fenix/app/src/main/java/org/mozilla/fenix/search/SearchDialogController.kt
boek: I originally had it within but I felt that chaining 5 bools was already hard to understand and the fact that this was a feature flag is what led me to pull this one out. This setting should be getting removed in the next version after the feature rolls out.

fenix/app/src/test/java/org/mozilla/fenix/search/SearchDialogControllerTest.kt
boek: I was originally going to just write the initial test but I wanted a little more confidence that I wasn't breaking anything else so I wrote a couple more. Ultimately all of these tests will get removed once the feature rolls out as the prompt will no longer exist.

boek:
@mavduevskiy Responded to the comments! I think that you're right but since we're going to be dropping the prompt I'm not sure how much effort we want to put into refactoring the logic or tests.. In hindsight I maybe shouldn't have even added the extra ones that I did 🤔

boek:
Sync'd with @mavduevskiy and he had a much better approach to the tests :)

mavduevskiy (approved):
looks good, 🚢 it!

PR #3664, opened 2023-09-15
Bug 1853299 - Update review checker state for reanalysis

rahulsainani:
### What
There are two commits:
1. Prepares for the changes, includes minor improvement to the action and reducer.
2. Adds `AnalysisStatus` for `AnalysisPresent` case. 

### How
1. Now reducer is responsible for updating the loading state before the network side-effect takes place.
2. Add `mapIfOptedIn` function to reduce repeated state checking and update in reducer. 
3. Include 4 states of `AnalysisStatus` for the success case.
4. Add `isReanalyzing ` for `NoAnalysisPresent` case.
5. Add `ReanalyzeProduct` action, that triggers loading for reanalysis, the side-effect to signal reanalysis to the api and fetch the analysis state will be handled by the Network middleware with this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1853311). 

### What's next
– Include other error states from analysis_status api integration, along with no network and a generic error state. 

### Pull Request checklist

### After merge

6. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
7. Click on `View task in Taskcluster` in the new `DETAILS` section.
8. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1853299


fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -71,10 +72,22 @@ fun ProductAnalysis(
...
+                ReanalyzeCard(onReanalyzeClick = onReanalyzeClick)
+            }
+
+            AnalysisStatus.REANALYZING -> {
+                // TBD
+            }
+
+            AnalysisStatus.UP_TO_DATE -> {
+                // no-op
+            }
```
MozillaNoah, line 86: optional: make the no-op(s) last? (may just be a personal "feel" thing)

MozillaNoah (approved):
LGTM

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
```diff
@@ -89,6 +90,13 @@ sealed interface ReviewQualityCheckState : State {
                 val highlightsFadeVisible: Boolean =
                     highlights != null && showMoreButtonVisible &&
                         highlights.forCompactMode().entries.first().value.size > 1
+
+                /**
+                 * The status of the product analysis.
+                 */
+                enum class AnalysisStatus {
+                    NEEDS_ANALYSIS, REANALYZING, UP_TO_DATE, COMPLETED
```
calumozilla, line 98: Where are these statuses from? Should they follow the [Fakespot API Docs](https://docs.google.com/document/d/1eppEnQKrSKavM3qmlm36WG7y8o2ggm-oxugnoSr2tpw/edit#heading=h.84nyloh6qhar): `pending, in_progress, completed, not_analyzable, not_found, unprocessable`? Or is this referring to something else?

calumozilla: Or are the ones I'm referring to going to be translated to those error states?

rahulsainani: This is the UI state, only one of these 4 state is possible when analysis is present. The data (api response) layer is separated from UI so all the logic is abstracted in the state/store and not in the UI. 

`pending`, `in_progress` map to `REANALYZING`, the middleware will set it when the api calls successfully return with those statuses.
`completed` maps one to one.
The rest will map to different Error state(s), based on the status response or the logic, e.g. `notfound` has a specific logic based on review grade and other fields.

PR #3666, opened 2023-09-15
Bug 1848562 - Autocomplete for bookmarked website should appear when 'search browsing history' is disabled.

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1848562

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/search/toolbar/ToolbarView.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/search/toolbar/ToolbarView.kt
@@ -210,6 +210,10 @@ class ToolbarView(
                             true -> components.core.historyStorage
                             false -> null
                         },
+                        when (settings.shouldShowBookmarkSuggestions) {
+                            true -> components.core.bookmarksStorage
+                            false -> null
+                        },
                         components.core.domainsAutocompleteProvider,
                     ),
                 )
--- a/fenix/app/src/test/java/org/mozilla/fenix/search/toolbar/ToolbarViewTest.kt
+++ b/fenix/app/src/test/java/org/mozilla/fenix/search/toolbar/ToolbarViewTest.kt
@@ -144,6 +144,7 @@ class ToolbarViewTest {
     fun `GIVEN search term is set WHEN switching to edit mode THEN the cursor is set at the end of the search term`() {
         every { context.settings().showUnifiedSearchFeature } returns true
         every { context.settings().shouldShowHistorySuggestions } returns true
+        every { context.settings().shouldShowBookmarkSuggestions } returns true
         val view = buildToolbarView(false)
         mockkObject(FeatureFlags)
 
@@ -158,6 +159,7 @@ class ToolbarViewTest {
     fun `GIVEN no search term is set WHEN switching to edit mode THEN the cursor is set at the end of the search term`() {
         every { context.settings().showUnifiedSearchFeature } returns true
         every { context.settings().shouldShowHistorySuggestions } returns true
+        every { context.settings().shouldShowBookmarkSuggestions } returns true
         val view = buildToolbarView(false)
         mockkObject(FeatureFlags)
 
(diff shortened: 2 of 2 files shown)


iorgamgabriel:
https://github.com/mozilla-mobile/firefox-android/assets/8118371/b8c93750-45a0-4962-8f2a-5cf45e5c51a2

fenix/app/src/main/java/org/mozilla/fenix/search/toolbar/ToolbarView.kt
```diff
@@ -210,6 +210,10 @@ class ToolbarView(
                             true -> components.core.historyStorage
                             false -> null
                         },
+                        when (settings.shouldShowBookmarkSuggestions) {
```
gabrielluong, line 213: I am wondering if there is a spec somewhere which tells us how the suggestions should be ordered? Currently, I am reading this as the following order in which the suggestions should appear: History, Bookmarks, Domains. Just wanted to verify this ordering is correct.

iorgamgabriel: I didn't find any specifications about ordering.

gabrielluong: cc @rocketsroger to verify

mavduevskiy: @gabrielluong 
Just to clarify, you are asking about search suggestions order, not the autocomplete priority, right? Because the patch doesn't affect the order of suggestions. Both history and bookmarks autocomplete providers have have lesser priority than domain, so we are not affecting existing logic here.

mavduevskiy (changes requested):
looks good, works good! 

could we also add tests to this? to make sure that depending on the settings the right autocomplete providers are added

iorgamgabriel:
Done , tests added. @mavduevskiy

fenix/app/src/main/java/org/mozilla/fenix/search/toolbar/ToolbarView.kt
rocketsroger: AFAIK suggestion order is coming from A-C.  In unified search design there was no ask to change the order of those suggestions.

gabrielluong: I believe I am talking about the `autocompletePriority` here. Both history and bookmarks have an `autocompletePriority` of 0 whereas domain has 1. The addition of the bookmark autocomplete provider here is interesting because we can't [guarantee the ordering](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/AutocompleteProvider.kt#26) of the bookmark and history autocomplete suggestions. If we have "example.com" in both the history and bookmark storage, typing "example" one character at time and observing the suggestions, we will see that the ordering of the suggestions can flip in terms of the ordering. 

To reorganize my initial question, I was wondering if there should be a specific ordering - history before bookmarks or vice versa compare to this unorder state for bookmarks and history where the ordering isn't guaranteed. This goes back to whether or not we have a spec about these Firefox suggestion orderings. If we do not have a spec, the follow up question would be should we have a guaranteed ordering or are we okay with what it is. The point I would want to get across is around really understanding our search behaviour and making sure it's following a documented spec. If it's not documented, we should ideally document this. Given the number of different search modes and options, it's getting harder to tell if our search behaviour is aligned with what is expected at times. While I trust the changes are going in the right direction, I want to make sure we are also verifying that the changes are what we want and the behaviour is what we expect it to be with full understanding of how our search code behaves.

mavduevskiy: `autocompletePriority` affects the autocomplete priority, not the suggestion priority. In your example, once we start typing in "example" one character at a time – the autocomplete feature will pick up an autocomplete suggestion based on the input and `autocompletePriority`, and it will stay the same effectively until the input goes astray from it. But search suggestions will be provided based on the actual user input, and yes, they might jump up and down depending on the way `CombinedHistorySuggestionProvider` will order `Suggestion`s with the same [score](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt#107). But it isn't connected to the autocomplete feature, at least from my understanding.

I think we do need a consistent ordering, because the fast jumping of the items look like a glitch and sometimes results are shuffled – which is even more confusing. There is a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1852169) about that. Though, it doesn't seem like a major issue since it will be visible only when multiple app level storages have the same webpage and we are matching it. I don't think it's something users notice a lot if ever.

gabrielluong: Sorry, this was my fault for causing unnecessary confusion. I have been interpreting this as search suggestions when this is autocomplete suggestions. I got tangled up when looking at the code and seeing autocomplete suggestions and having the word "suggestions" used in both context made it all seem the same to me.

PR #3690, opened 2023-09-18
Bug 1853177 - Avoid multiple crash dialogs to be created and shown

willdurand:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1853177


fenix/app/src/main/java/org/mozilla/fenix/addons/ExtensionProcessDisabledController.kt
```diff
@@ -84,4 +39,67 @@ class ExtensionProcessDisabledController(
...
-)
+) {
+    override fun onDestroy(owner: LifecycleOwner) {
+        super.onDestroy(owner)
+        // In case the activity gets destroyed, we want to reset the dialog so that it gets re-created when needed.
+        dialog = null
+    }
+
+    companion object {
+        private var dialog: AlertDialog? = null
```
Amejia481, line 50: Ideally we shouldn't keep Android UI components as static references, it's. common Android code smell. As we could cause memory leaks. I noticed you're cleaning up the reference on `onDestroy`, but but we will be introducing brittle pattern. Let's try to think about other alternatives. I'll try to find a possible solution and let you know!

Amejia481: Have we tried this PR with "Don't keep activities" setting enabled ?

Amejia481: Another alternative instead of having a reference of `AlertDialog`, we could have a `boolean` which indicates if the dialog is already showing, what do you think?

willdurand: > have you already tried other alternatives?

No

> Have we tried this PR with "Don't keep activities" setting enabled ?

Yes, hence the `onDestroy()` call 🙈 

> Another alternative instead of having a reference of AlertDialog, we could have a boolean which indicates if the dialog is already showing, what do you think?

I suppose that could work yeah, let me try.

PR #3703, opened 2023-09-19
Bug 1853687 - Disable frequent failing test - get history metadata between

DonalMe:
https://bugzilla.mozilla.org/show_bug.cgi?id=1853687

This test started failing frequently since [A-S 119.20230915050340 merged](https://github.com/mozilla-mobile/firefox-android/pull/3659), it's impacting the stability of CI for GV bumps and nightly builds.

The test is racy, the current suspect is a [rust version bump in the A-S repo](https://github.com/mozilla/application-services/pull/5821) introduced a different performance profile, where a slower path is now faster.

### DIFF
--- a/android-components/components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesHistoryStorageTest.kt
+++ b/android-components/components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesHistoryStorageTest.kt
@@ -44,6 +44,7 @@ import org.junit.Assert.assertNull
 import org.junit.Assert.assertTrue
 import org.junit.Assert.fail
 import org.junit.Before
+import org.junit.Ignore
 import org.junit.Rule
 import org.junit.Test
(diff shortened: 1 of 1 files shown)


linabutler (approved):
Thanks!

AaronMT (approved):

linabutler (changes requested):
Sorry, actually, could you please disable `store can delete by 'range'`, too? I just noticed that this is a different test than the other failure you linked in Slack: https://firefoxci.taskcluster-artifacts.net/Xzgj706YQbOd-ta8jLEDhQ/0/public/reports/tests/testReleaseUnitTest/classes/mozilla.components.browser.storage.sync.PlacesHistoryStorageTest.html#store%20can%20delete%20by%20'range'

This one also looks racy, though; it looks like it’s relying on the system clock advancing by at least 1ms between metadata observations.

linabutler (approved):
Thank you so much, @DonalMe—and sorry I didn't catch that sooner! 🙏🏼

PR #3704, opened 2023-09-19
Bug 1853967 - Capture telemetry for the amount of RAM a device has.

Amejia481:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1853967

### DIFF
--- a/fenix/app/metrics.yaml
+++ b/fenix/app/metrics.yaml
@@ -2539,6 +2539,28 @@ metrics:
     metadata:
       tags:
         - Experiments
+  device_total_ram:
+    type: quantity
+    lifetime: application
+    description: >
+      The total amount of memory this device in bytes, when available will be
+      MemoryInfo.advertisedMem otherwise it will be MemoryInfo.totalMem.
+      This doesn't represent memory available to the application however.
+    send_in_pings:
+      - metrics
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1853967
+    data_reviews:
+      - https://github.com/mozilla-mobile/firefox-android/pull/2620
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: never
+    unit: integer
+    metadata:
+      tags:
+        - Performance
   search_page_load_time:
     type: timing_distribution
     time_unit: millisecond
--- a/fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
@@ -798,6 +798,7 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
             )
 
             ramMoreThanThreshold.set(isDeviceRamAboveThreshold)
+            deviceTotalRam.set(getDeviceTotalRAM())
         }
 
         with(AndroidAutofill) {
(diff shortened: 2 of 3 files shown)


fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -855,12 +856,29 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         }
     }
 
-    private fun deviceRamBytes(): Long {
+    @VisibleForTesting
+    @Suppress("MagicNumber")
+    internal fun getDeviceAvailableMemoryInGB(): Long {
+        val memoryInfo = getMemoryInfo()
+        val gigabytesDivider = (1e+9).toLong()
```
jonalmeida, line 863: This calculation will give a number that's slightly off, right? Don't we want `1024 * 1024 * 1024 * 1024`?

fenix/app/metrics.yaml
```diff
@@ -2539,6 +2539,28 @@ metrics:
     metadata:
       tags:
         - Experiments
+  device_total_ram:
+    type: quantity
+    lifetime: application
+    description: >
+      The total amount of memory this device has GBs, when available will be
+      MemoryInfo.advertisedMem otherwise it will be MemoryInfo.totalMem
```
jonalmeida, line 2547:
```suggestion
      The total amount of memory this device has in gigabytes. When available will be
      MemoryInfo.advertisedMem otherwise it will be MemoryInfo.totalMem. 
      This doesn't represent memory available to the application however.
```

Fixed a nit and also added a clarification (see overall note).

jonalmeida (commented):
@kycn and I were talking about this earlier today. This telemetry information is a bit flawed, because it's memory available to the entire device, and not for our application.

e.g. We can still have 8GB total memory available and still have our app crashing due to memory pressure.

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -855,12 +856,29 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+        return if (SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+            memoryInfo.advertisedMem / gigabytesDivider
+        } else {
+            memoryInfo.totalMem / gigabytesDivider
```
willdurand, line 867: I think I'd rather collect the "raw" number if possible.

willdurand:
> @kycn and I were talking about this earlier today. This telemetry information is a bit flawed, because it's memory available to the entire device, and not for our application.
> 
> e.g. We can still have 8GB total memory available and still have our app crashing due to memory pressure.

Yeah, definitely though we could quickly determine "low end devices" vs. others. Sure a 8GB device might have memory pressure but likely less than a 500MB or 1GB device, right?

gabrielluong:
@t-p-white This might be of interest since we were also looking at RAM.

willdurand:
> **Case**: A high end device that's actively running a lot of RAM consuming apps, and Fenix is launched.

That sounds like a different use case to me so maybe we need to collect both?

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
Amejia481: It was returning 7 for my pixel 6, which it's about the same that  it's showing on **Developer options** -> **Running services** :)

image

Anyway,  I think we will be moving to collect the raw data :)

fenix/app/metrics.yaml
Amejia481: 👍🏽

Amejia481:
> > **Case**: A high end device that's actively running a lot of RAM consuming apps, and Fenix is launched.
> 
> That sounds like a different use case to me so maybe we need to collect both?

Agree with @willdurand, we could collect both. this sounds like a different use case of what we are looking. Collecting the `availMem` sounds something more specific to a particular use case,  we could send the available memory when something important is happening. For example, we may want to collect the available memory when the extension process exceeded the auto restarts threshold. 

We could file a separated ticket?

Amejia481:
I updated PR with the suggested changes.

@kycn @jonalmeida we should we file a separated ticket to follow up the discussion?  At the moment, the physical total amount of RAM is good enough for our use case. Noticed we already have [Bug #1826718](https://bugzilla.mozilla.org/show_bug.cgi?id=1826718), but not sure if we need anything else :)

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -855,12 +856,27 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
         }
     }
 
-    private fun deviceRamBytes(): Long {
+    @VisibleForTesting
+    internal fun getDeviceAvailableMemoryInGB(): Long {
```
willdurand, line 860:
```suggestion
    internal fun getDeviceTotalRAM(): Long {
```

or something like that

Amejia481:
> r+wc from my side

Addressed!

Amejia481:
Sure @mergify 😅 !

jonalmeida (approved):
With the description addition this will at least make it clear that we can't rely on this telemetry to make concrete decisions on, so I am fine with this and following up in marco's telemetry request for our memory constraints.

PR #3713, opened 2023-09-20
Bug 1853113 - Monitor Search Widget presence based on activation

DreVla:
Previously, the way we registered if the user has installed the search widget was keeping a count that increments every time the widget was added and decrements everytime it was removed. The main problem here was the wrong usage of `onEnabled` and `onDeleted`. This patch aims to move towards a boolean based preference and use the provided `onDisabled` method together with `onEnabled`. The logic here is as follows: When the first instance of the widget has been added, `onEnabled` will be called and the preference set to true and no matter how many widgets are present on the homescreen, it will stay true. When the last widget is removed, `onDisabled` will be called and the pref will be set to false.

In order to also fix the scenarios where the metric is not recorded when the first instance of the widget was added or the last instance was removed, we also send the metric in `onEnabled` & `onDisabled`.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1853113


Alexandra-Virvara (approved):
LGTM

iorgamgabriel (approved):

DreVla:
@mergify backport releases_v118

PR #3737, opened 2023-09-21
Bug 1853311 - Add review checker reanalysis logic

rahulsainani:
### What
– Integrate `reanalyzeProduct` and `requestAnalysisStatus` APIs in `ReviewQualityCheckService` and middleware.
– Add polling logic for analysis status.
– Create a generic retry function.
– Add an edge case in mapper and test.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1853311

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/Retry.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/Retry.kt
@@ -1,0 +1,45 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.shopping.middleware
+
+import kotlinx.coroutines.delay
+
+private const val MAX_RETRIES = 10
+private const val INITIAL_DELAY_MS = 2000L
+private const val MAX_DELAY_MS = 20000L
+private const val FACTOR = 2.0
+
+/**
+ * Retry a suspend function until it returns a value that satisfies the predicate.
+ *
+ * @param maxRetries The maximum number of retries.
+ * @param initialDelayMs The initial delay in milliseconds.
+ * @param maxDelayMs The maximum delay in milliseconds.
+ * @param factor The factor to increase the delay by.
+ * @param predicate The predicate to check the result against.
+ * @param block The function to retry.
+ */
+suspend fun <T> retry(
+    maxRetries: Int = MAX_RETRIES,
+    initialDelayMs: Long = INITIAL_DELAY_MS,
+    maxDelayMs: Long = MAX_DELAY_MS,
+    factor: Double = FACTOR,
+    predicate: (T) -> Boolean,
+    block: suspend () -> T,
+): T {
+    var delayTime = initialDelayMs
+    var data: T = block()
+
+    repeat(maxRetries - 1) {
+        if (predicate(data)) {
+            delay(delayTime)
+            data = block()
+            delayTime = (delayTime * factor).toLong().coerceAtMost(maxDelayMs)
+        } else {
+            return data
+        }
+    }
+    return data
+}
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
@@ -6,14 +6,15 @@ package org.mozilla.fenix.shopping.middleware
 
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
-import mozilla.components.lib.state.Middleware
 import mozilla.components.lib.state.MiddlewareContext
 import mozilla.components.lib.state.Store
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckAction
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckAction.FetchProductAnalysis
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckAction.RetryProductAnalysis
+import org.mozilla.fenix.shopping.store.ReviewQualityCheckMiddleware
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckState
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckState.OptedIn.ProductReviewState
+import org.mozilla.fenix.shopping.store.ReviewQualityCheckState.OptedIn.ProductReviewState.AnalysisPresent.AnalysisStatus
 
 /**
  * Middleware that handles network requests for the review quality check feature.
@@ -26,7 +27,7 @@ class ReviewQualityCheckNetworkMiddleware(
     private val reviewQualityCheckService: ReviewQualityCheckService,
     private val networkChecker: NetworkChecker,
     private val scope: CoroutineScope,
-) : Middleware<ReviewQualityCheckState, ReviewQualityCheckAction> {
+) : ReviewQualityCheckMiddleware {
 
     override fun invoke(
         context: MiddlewareContext<ReviewQualityCheckState, ReviewQualityCheckAction>,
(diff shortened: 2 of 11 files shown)


fenix/app/src/test/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapperTest.kt
```diff
@@ -113,14 +113,29 @@ class ProductAnalysisMapperTest {
     }
 
     @Test
-    fun `WHEN product id is null THEN it is mapped to no analysis present`() {
```
calumozilla, line 116: do we also need tests to check AnalysisStatusDto and if the loading analysis card is present

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
```diff
@@ -34,15 +49,95 @@ class DefaultReviewQualityCheckService(
...
+    }
+
+    private inline fun <reified T : Enum<T>> String.asEnumOrDefault(defaultValue: T? = null): T? =
+        enumValues<T>().firstOrNull { it.name.equals(this, ignoreCase = true) } ?: defaultValue
+}
+
+/**
+ * Enum that represents the status of the product review analysis.
+ */
+enum class AnalysisStatusDto {
```
calumozilla, line 113: what does the `Dto` part of the name mean

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
```diff
@@ -46,21 +50,82 @@ class ReviewQualityCheckNetworkMiddleware(
...
+                    logger.info("Fetching product analysis")
+                    val productReviewState = fetchAnalysis()
+                    logger.debug("Product review state: $productReviewState")
+                    store.updateProductReviewState(productReviewState)
                 }
             }
 
             ReviewQualityCheckAction.ReanalyzeProduct -> {
-                // Bug 1853311 - Integrate analyze and analysis_status
+                scope.launch {
```
MozillaNoah, line 69: wondering aloud if the entire `when` should just be wrapped with `scope.launch{ }`

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
```diff
@@ -46,21 +50,82 @@ class ReviewQualityCheckNetworkMiddleware(
...
+                    val reanalysis = reviewQualityCheckService.reanalyzeProduct().also {
+                        logger.debug("Reanalysis: $it")
+                    }
+
+                    if (reanalysis == null) {
+                        store.updateProductReviewState(ProductReviewState.Error.GenericError)
+                        return@launch
```
MozillaNoah, line 76: what's the thought process of `return@launch` vs just `return` in this scenario? To ensure the coroutine terminates properly?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
```diff
@@ -46,21 +50,82 @@ class ReviewQualityCheckNetworkMiddleware(
...
+                                store.updateProductReviewState(ProductReviewState.NoAnalysisPresent())
+                            } else if (state.productReviewState is ProductReviewState.AnalysisPresent) {
+                                store.updateProductReviewState(
+                                    state.productReviewState.copy(
+                                        analysisStatus = AnalysisStatus.NEEDS_ANALYSIS,
+                                    ),
+                                )
+                            }
+                        }
+                        return@launch
```
MozillaNoah, line 98: is this line needed?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
```diff
@@ -46,21 +50,82 @@ class ReviewQualityCheckNetworkMiddleware(
...
+        retry(
+            predicate = { it == AnalysisStatusDto.PENDING || it == AnalysisStatusDto.IN_PROGRESS },
+            func = { reviewQualityCheckService.analysisStatus() },
+        ).also {
+            logger.debug("Status: $it")
+        }
+
+    private suspend fun AnalysisStatusDto.toProductReviewState(): ProductReviewState =
+        when (this) {
+            AnalysisStatusDto.COMPLETED -> fetchAnalysis()
```
MozillaNoah, line 121: can we maybe restructure this + `fetchAnalysis` slightly? they currently have a bit of a circular dependency 😨 

Maybe have the callers of `fetchAnalysis` call `toProductReviewState` instead, so the fetching is an isolated function and the transformation is a separate distinct step?

It feels off to have `toProductReviewState`, which feels like a state transformation/mapping function, firing off the network call, call itself, and return the result of calling itself back to `AnalysisStatusDto.COMPLETED`.

MozillaNoah (commented):
first pass

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/Retry.kt
```diff
@@ -0,0 +1,45 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.shopping.middleware
+
+import kotlinx.coroutines.delay
+
+private const val MAX_RETRIES = 10
+private const val INITIAL_DELAY_MS = 2000L
+private const val MAX_DELAY_MS = 20000L
+private const val FACTOR = 2.0
```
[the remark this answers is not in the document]
rahulsainani: Reanalysis takes 1-2 minutes and these delays work well for cases when the reanalysis is already wrapping up and then the initial delay is only 2 seconds, when a very patient user wants to see the whole loading, the max delay here would be 130 seconds, factor 2 being exponential backoff - double delay between each retry.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
rahulsainani: Data transfer object, we are separate the `State` which is a UI model and the objects exposed via toolkit/GV/AC which are regarded as DTOs as they match the rest api one to one.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/Retry.kt
```diff
@@ -0,0 +1,45 @@
...
+ * Retry a suspend function until it returns a value that satisfies the predicate.
+ *
+ * @param maxRetries The maximum number of retries.
+ * @param initialDelayMs The initial delay in milliseconds.
+ * @param maxDelayMs The maximum delay in milliseconds.
+ * @param factor The factor to increase the delay by.
+ * @param predicate The predicate to check the result against.
+ * @param func The function to retry.
+ */
+suspend fun <T> retry(
```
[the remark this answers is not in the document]
rahulsainani: I will talk about it in one of the engineering syncs, this and `NetworkChecker` has the potential to be more common functionality across the project.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
rahulsainani: No it is not, good catch!

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
rahulsainani: Since we're inside the a coroutine and this lamda/function is the `block` param of the `launch` function, we can only `return@launch`. The parent function has already executed the lines below `launch{}`,

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
rahulsainani: that's a good idea!

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
rahulsainani: They don't have a circular dependency – `fetchAnalysis` is calling `ProductAnalysis.toProductReviewState()` not this function. This function calls `fetchAnalysis`.

It's data logic that's acting on the response of one endpoint - deciding to fetch when the analysis is completed and returns errors in other cases. 

And yes, I think we agree that a mapper is a good place for functions that are just transformations, that's also another reason why this is not in `ProductAnalysisMapper`.

fenix/app/src/test/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapperTest.kt
rahulsainani: indeed, there are now tests in `ReviewQualityCheckStoreTest` for the analysis status. 

As for loading, I thought about it but it's not straightforward at this moment AFAIK. It's an intermediate state between the initial and the ending success/error states. It would involve observing the state in the test and asserting that there's loading in the middle. Not sure how to do that, let me know if you have any ideas. Also, depends on how pedantic we want to be about asserting the intermediate state.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/Retry.kt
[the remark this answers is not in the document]
MozillaNoah: Gotcha gotcha. makes sense to me

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
MozillaNoah: Ahh gotcha gotcha, sorry for the confusion on my end then. This looks good 👍

MozillaNoah (approved):
LGTM! Tested locally with a product listing for a microphone and it had an out of date review, switched to the analysis in progress UI, and dismissed the UI and updated the rating.

PR #3753, opened 2023-09-22
Bug 1854662 - Remove some unused AC strings

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1854662

### DIFF
--- a/android-components/components/browser/errorpages/src/main/res/values-az/strings.xml
+++ b/android-components/components/browser/errorpages/src/main/res/values-az/strings.xml
@@ -1,13 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <!-- The default document title of an error page (i.e., the contents of the `<title>` tag). -->
-    <string name="mozac_browser_errorpages_page_title">Səhifə yüklənmə xətası</string>
-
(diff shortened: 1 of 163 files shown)


rvandermeulen:
@Delphine I'm not sure about all the mechanics at play here. Is this safe to land as-is or do we need to make changes Somewhere Else to go along with it?

Delphine:
So this is touching only localized files, not the source ones. This means that if we merge this PR, the removed files will end up being imported back into the Firefox for Android repo, according to current [configuration](https://docs.google.com/document/d/1YlfECzZhGUV4pT10u6eC-egB9zOBB9GB9haKECDDdXE/edit#heading=h.6h5s1arniwqm) (note that this doc will end up in Confluence once it's final).
Since these strings are obsolete, we wouldn't want that.
Is it ok to ignore these warnings so that the above doesn't happen? (and so, not merge these kind of PRs). Unfortunately we will keep facing this, since in a-c some files are very rarely touched

rvandermeulen:
Per discussion with Delphine, removing it here will only bring them back with the next string sync. Removing them from android-l10n will make them go away, however.

PR #3757, opened 2023-09-22
Bug 1838541: Changed distinctUntilChanged for MediaSessionFullScreenFeature

titooan:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1838541

### DIFF
--- a/android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
+++ b/android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
@@ -35,57 +35,63 @@ class MediaSessionFullscreenFeature(
             flow.map {
                 it.tabs + it.customTabs
             }.map { tab ->
-                tab.filter { it.mediaSessionState != null && it.mediaSessionState!!.fullscreen }
-            }.distinctUntilChanged().collect { states ->
-                processFullscreen(states)
-                processDeviceSleepMode(states)
+                tab.firstOrNull { it.mediaSessionState?.fullscreen == true }
+            }.distinctUntilChanged { old, new ->
+                old.hasSameOrientationInformationAs(new)
+            }.collect { state ->
+                // There should only be one fullscreen session.
+                if (state == null) {
+                    activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
+                    activity.window?.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+                    return@collect
+                }
+
+                if (store.state.findCustomTabOrSelectedTab(tabId)?.id == state.id) {
+                    setOrientationForTabState(state)
+                }
+                setDeviceSleepModeForTabState(state)
             }
         }
     }
 
     @Suppress("SourceLockedOrientationActivity") // We deliberately want to lock the orientation here.
-    private fun processFullscreen(sessionStates: List<SessionState>) {
-        /* there should only be one fullscreen session */
-        val activeState = sessionStates.firstOrNull()
-        if (activeState == null || activeState.mediaSessionState?.fullscreen != true) {
-            activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
-            return
-        }
+    private fun setOrientationForTabState(activeTabState: SessionState) {
+        when (activeTabState.mediaSessionState?.elementMetadata?.portrait) {
+            true ->
+                activity.requestedOrientation =
+                    ActivityInfo.SCREEN_ORIENTATION_USER_PORTRAIT
 
-        if (store.state.findCustomTabOrSelectedTab(tabId)?.id == activeState.id) {
-            when (activeState.mediaSessionState?.elementMetadata?.portrait) {
-                true ->
+            false ->
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && activity.isInPictureInPictureMode) {
+                    activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
+                } else {
                     activity.requestedOrientation =
-                        ActivityInfo.SCREEN_ORIENTATION_USER_PORTRAIT
-                false ->
-                    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N && activity.isInPictureInPictureMode) {
-                        activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED
-                    } else {
-                        activity.requestedOrientation =
-                            ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
-                    }
-                else -> activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
-            }
+                        ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
+                }
+
+            null -> activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
         }
     }
 
-    private fun processDeviceSleepMode(sessionStates: List<SessionState>) {
-        val activeTabState = sessionStates.firstOrNull()
-        if (activeTabState == null || activeTabState.mediaSessionState?.fullscreen != true) {
-            activity.window?.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
-            return
-        }
+    private fun setDeviceSleepModeForTabState(activeTabState: SessionState) {
         activeTabState.mediaSessionState?.let {
             when (activeTabState.mediaSessionState?.playbackState) {
                 MediaSession.PlaybackState.PLAYING -> {
                     activity.window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
                 }
+
                 else -> {
                     activity.window.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
                 }
             }
         }
     }
+
+    private fun SessionState?.hasSameOrientationInformationAs(other: SessionState?): Boolean =
+        this?.mediaSessionState?.fullscreen == other?.mediaSessionState?.fullscreen &&
+            this?.mediaSessionState?.playbackState == other?.mediaSessionState?.playbackState &&
+            this?.mediaSessionState?.elementMetadata == other?.mediaSessionState?.elementMetadata &&
+            this?.content?.pictureInPictureEnabled == other?.content?.pictureInPictureEnabled
 
     override fun stop() {
         scope?.cancel()


android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,51 +35,53 @@ class MediaSessionFullscreenFeature(
...
-                tab.filter { it.mediaSessionState != null && it.mediaSessionState!!.fullscreen }
-            }.distinctUntilChanged().collect { states ->
-                processFullscreen(states)
-                processDeviceSleepMode(states)
+                tab.firstOrNull { it.mediaSessionState != null && it.mediaSessionState!!.fullscreen }
+            }.distinctUntilChanged { old, new ->
+                old?.mediaSessionState?.fullscreen == new?.mediaSessionState?.fullscreen &&
+                    old?.mediaSessionState?.playbackState == new?.mediaSessionState?.playbackState &&
+                    old?.content?.pictureInPictureEnabled == new?.content?.pictureInPictureEnabled &&
+                    old?.mediaSessionState?.elementMetadata == new?.mediaSessionState?.elementMetadata
```
MatthewTighe, line 43: I think it would add semantic value and be more extensible if we moved these comparisons into a function.

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,51 +35,53 @@ class MediaSessionFullscreenFeature(
...
+            }.collect { state -> /* there should only be one fullscreen session */
```
MatthewTighe, line 44: Is there anything to your knowledge that enforces the constraint mentioned in this comment? I'm not familiar with this domain space

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,51 +35,53 @@ class MediaSessionFullscreenFeature(
...
+                if (state == null || state.mediaSessionState?.fullscreen != true) {
```
MatthewTighe, line 45: I don't think the second half of this OR can be hit, since the tab will be mapped to `null` if `fullscreen` is false in `tab.firstOrNull { it.mediaSessionState != null && it.mediaSessionState!!.fullscreen }`

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,51 +35,53 @@ class MediaSessionFullscreenFeature(
...
-                    } else {
-                        activity.requestedOrientation =
-                            ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
-                    }
-                else -> activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
-            }
+                        ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE
+                }
+
+            else -> activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
```
MatthewTighe, line 74: nit: while we're changing this anyway, I personally would find `null` more clear than `else`. The `else` gives me a bit of anxiety wondering what other state a boolean could possibly be in besides `true` or `false` 😂

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,51 +35,53 @@ class MediaSessionFullscreenFeature(
...
 
     @Suppress("SourceLockedOrientationActivity") // We deliberately want to lock the orientation here.
-    private fun processFullscreen(sessionStates: List<SessionState>) {
-        /* there should only be one fullscreen session */
-        val activeState = sessionStates.firstOrNull()
-        if (activeState == null || activeState.mediaSessionState?.fullscreen != true) {
-            activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
-            return
-        }
+    private fun processFullscreen(activeTabState: SessionState) {
```
MatthewTighe, line 60: Also while we're here, would something like `setFullscreenOrientation` make more sense as a function name?

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,51 +35,53 @@ class MediaSessionFullscreenFeature(
             flow.map {
                 it.tabs + it.customTabs
             }.map { tab ->
-                tab.filter { it.mediaSessionState != null && it.mediaSessionState!!.fullscreen }
-            }.distinctUntilChanged().collect { states ->
-                processFullscreen(states)
-                processDeviceSleepMode(states)
+                tab.firstOrNull { it.mediaSessionState != null && it.mediaSessionState!!.fullscreen }
```
MatthewTighe, line 38: I'm not sure this is worth it, but we could technically get rid of the force unwrap (`!!`) with something like:

```
                    with (it.mediaSessionState) {
                        (this != null) && this.fullscreen
                    }   
```

or
```
                tab.firstOrNull {
                    val mediaState = it.mediaSessionState
                    mediaState != null && mediaState.fullscreen
                }
```

Up to you 🙂

MatthewTighe (commented):
First pass, hoping this finally knocks this out! I am still trying to get a good grip on the domain here. Could we add comments in the code or at least in this patch to describe how this is resolving the issue? 

Reading through the bug, I'm wondering if the underlying issue is that we are briefly leaking the `Flow` when the activity gets destroyed based on Zac's [comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1838541#c19):
> In release 117 on my Pixel Fold when attempting to go full this activity instance is null and will crash when in Private Browsing. I've stopped the configuration from continuing in the activity on my Pixel 7 as well (which is not crashing) and am able to replicate the screenshots posted to this and other tickets.

I think it could be worth investigating more deeply whether that is the case and how we can resolve it. That said, since we don't have reliable STR I think we can ship this patch soon and see how it fares in the wild.

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
titooan: Actually I think this could be even better? 
```suggestion
                tab.firstOrNull { it.mediaSessionState?.fullscreen == true }
```

MatthewTighe: Yes, that's perfect!

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,58 +35,64 @@ class MediaSessionFullscreenFeature(
...
                 }
+
                 else -> {
                     activity.window.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
                 }
             }
         }
     }
 
+    private fun SessionState?.hasSameOrientationInformationAs(other: SessionState?): Boolean =
```
MatthewTighe, line 90: Love this! I had been playing around with this function signature a bit myself and didn't find anything nearly this good ☺️

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,58 +35,64 @@ class MediaSessionFullscreenFeature(
...
+                if (state == null) {
+                    activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
+                    activity.window?.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+                    return@collect
+                }
+
+                if (store.state.findCustomTabOrSelectedTab(tabId)?.id == state.id) {
+                    setOrientationForTabState(state)
+                }
+                setDeviceSleepModeForTabState(state)
```
MatthewTighe, line 52: This whole block has a lot more clarity to me now, thanks for indulging me!

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,58 +35,64 @@ class MediaSessionFullscreenFeature(
...
+                /* there should only be one fullscreen session */
+                if (state == null) {
+                    activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
+                    activity.window?.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+                    return@collect
+                }
+
+                if (store.state.findCustomTabOrSelectedTab(tabId)?.id == state.id) {
+                    setOrientationForTabState(state)
+                }
```
MatthewTighe, line 51: AFAICT this condition should always be true, since a tab can't be full-screen unless it's the active tab, right? I'm wondering if perhaps we want to capture/log an error in the `else` case, or if there's anything else we can to do indicate that we've reach (what should be) an impossible state. 

Mostly just sharing as food-for-thought and to check my own understanding. If you feel like there's anything we can improve here I think it would be fine for us to do it in a follow-up.

MatthewTighe (approved):
LGTM! 🤞 that this resolves this issue in the wild

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
titooan: Thank you for noticing that it wasn't that clear and helping me improve it!

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
titooan: That seems totally true! Thanks for noticing that

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
titooan: It can be false, in the following case: 
Open Fenix, navigate to Youtube, play a video in fullscreen.
Put the video in PiP mode (pressing Home button or equivalent). 

Then open another app, that uses Custom Tabs. Fenix must be the default browser, and I think Chrome needs to be uninstalled (or desactivated). 
When the custom tab is opened, this code is reached and the condition is not met (the tab playing the video is not the active one).

I just checked because I was not totally sure, so thanks for raising this concern!

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
titooan: I'd say it's mainly because it is not technically possible to have two tabs in fullscreen at the same time. Does it seem to make sense to you?

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
MatthewTighe: Oh, good find! Yeah this logic definitely all checks out for me then

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
```diff
@@ -35,58 +35,64 @@ class MediaSessionFullscreenFeature(
...
             }.map { tab ->
-                tab.filter { it.mediaSessionState != null && it.mediaSessionState!!.fullscreen }
-            }.distinctUntilChanged().collect { states ->
-                processFullscreen(states)
-                processDeviceSleepMode(states)
+                tab.firstOrNull { it.mediaSessionState?.fullscreen == true }
+            }.distinctUntilChanged { old, new ->
+                old.hasSameOrientationInformationAs(new)
+            }.collect { state ->
+                /* there should only be one fullscreen session */
```
gabrielluong, line 42:
```suggestion
                // There should only be one fullscreen session.
```

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/fullscreen/MediaSessionFullscreenFeature.kt
MatthewTighe: Actually, I have one more question. I don't think we need to block on this, I'm just curious and having some trouble getting PiP to work in browser on youtube to check for myself.

> I'd say it's mainly because it is not technically possible to have two tabs in fullscreen at the same time. Does it seem to make sense to you?

Does a PiP video count as fullscreen? Would it then be possible to have a PiP video and a full screen video playing simultaneously? I'm mostly wondering where in the code we would actually prevent something like that

titooan:
@Mergifyio backport releases_v119

PR #3770, opened 2023-09-25
Bug 1854123 - Show the "find more add-ons" button unconditionally in Fenix

willdurand:
### Pull Request checklist

- ~~[ ] **Tests**: This PR includes thorough tests or an explanation of why it does not~~

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1854123


fenix/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt
```diff
@@ -274,6 +274,6 @@ class AddonsManagementFragment : Fragment(R.layout.fragment_add_ons_management)
         // This is locale-less on purpose so that the content negotiation happens on the AMO side because the current
         // user language might not be supported by AMO and/or the language might not be exactly what AMO is expecting
         // (e.g. `en` instead of `en-US`).
-        private const val AMO_HOMEPAGE_FOR_ANDROID = "https://addons.mozilla.org/android/"
+        private const val AMO_HOMEPAGE_FOR_ANDROID = BuildConfig.AMO_BASE_URL + "/android/"
```
willdurand, line 277: I should have done that in the previous patch, sorry...

fenix/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementView.kt
```diff
@@ -36,9 +35,7 @@ class AddonsManagementView(
         showNotYetSupportedAddonFragment(unsupportedAddons)
     }
 
-    override fun shouldShowFindMoreAddonsButton(): Boolean {
-        return !Config.channel.isRelease
-    }
+    override fun shouldShowFindMoreAddonsButton(): Boolean = true
```
Amejia481, line 38: If we are going to keep it enable, should we just remove `shouldShowFindMoreAddonsButton`?

Amejia481: Should we instead introduce a nimbus feature flag (`true` by default), just to be able to control this functionality via nimbus?

willdurand: > If we are going to keep it enable, should we just remove `shouldShowFindMoreAddonsButton`?

It's part of the public A-C API now (`AddonsManagerAdapterDelegate`) so that's why I kept the override. In Fenix, we want this button but in other apps? I am not sure.

> Should we instead introduce a nimbus feature flag (`true` by default), just to be able to control this functionality via nimbus?

I don't think we need that.

fenix/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementFragment.kt
Amejia481: No worries!
Hint: We also have strings interpolation in [Kotlin](https://kotlinlang.org/docs/java-to-kotlin-idioms-strings.html#concatenate-strings) too :) 

Not a suggestion to update the code, just making you aware in case you were not familiar :) 
This also could be  `"${BuildConfig.AMO_BASE_URL}/android/"`.

fenix/app/src/main/java/org/mozilla/fenix/addons/AddonsManagementView.kt
Amejia481: Got it 👍🏽 
Thanks for the clarification!

Amejia481 (approved):
✅

PR #3803, opened 2023-09-26
Bug 1853304 - Color the shopping icon while the bottom sheet is open

MozillaNoah:
Since the toolbar is in XML and depends upon the state of the shopping bottom sheet, using an `Action` became the only reasonable way to get the shopping icon to reset to its default, unselected state. If the toolbar is later converted to Compose, we'll be able to do more direct observations on this new state variable, rather than having to do it indirectly via `ReviewQualityCheckFeature`.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1853304

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
@@ -216,4 +216,9 @@ sealed class AppAction : Action {
     data class UpdateStandardSnackbarErrorAction(
         val standardSnackbarError: StandardSnackbarError?,
     ) : AppAction()
+
+    /**
+     * [AppAction] used to update the expansion state of the shopping sheet.
+     */
+    data class ShoppingSheetStateUpdated(val expanded: Boolean) : AppAction()
 }
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
@@ -7,28 +7,35 @@ package org.mozilla.fenix.shopping
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.drop
+import kotlinx.coroutines.flow.filterNot
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.mapNotNull
 import mozilla.components.browser.state.selector.selectedTab
 import mozilla.components.browser.state.store.BrowserStore
 import mozilla.components.lib.state.ext.flowScoped
 import mozilla.components.support.base.feature.LifecycleAwareFeature
+import org.mozilla.fenix.components.AppStore
 
 /**
  * Feature implementation that provides review quality check information for supported product
  * pages.
  *
+ * @property appStore Reference to the application's [AppStore].
  * @property browserStore Reference to the application's [BrowserStore].
  * @property shoppingExperienceFeature Reference to the [ShoppingExperienceFeature].
  * @property onAvailabilityChange Invoked when availability of this feature changes based on feature
  * flag and when the loaded page is a supported product page.
  */
 class ReviewQualityCheckFeature(
+    private val appStore: AppStore,
     private val browserStore: BrowserStore,
     private val shoppingExperienceFeature: ShoppingExperienceFeature,
     private val onAvailabilityChange: (isAvailable: Boolean) -> Unit,
+    private val onBottomSheetCollapsed: () -> Unit,
 ) : LifecycleAwareFeature {
     private var scope: CoroutineScope? = null
+    private var appStoreScope: CoroutineScope? = null
 
     override fun start() {
         if (!shoppingExperienceFeature.isEnabled) {
@@ -42,9 +49,20 @@ class ReviewQualityCheckFeature(
                 .distinctUntilChanged()
                 .collect(onAvailabilityChange)
         }
+
+        appStoreScope = appStore.flowScoped { flow ->
+            flow.mapNotNull { it.shoppingSheetExpanded }
+                .distinctUntilChanged()
+                .drop(1) // Needed to ignore the initial emission from the Store setup
+                .filterNot { it }
+                .collect {
+                    onBottomSheetCollapsed()
+                }
+        }
     }
 
     override fun stop() {
         scope?.cancel()
+        appStoreScope?.cancel()
     }
 }
--- a/fenix/app/src/main/res/drawable/ic_shopping_selected.xml
+++ b/fenix/app/src/main/res/drawable/ic_shopping_selected.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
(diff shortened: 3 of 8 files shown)


fenix/app/src/main/res/drawable/ic_shopping_selected.xml
```diff
@@ -0,0 +1,37 @@
...
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:aapt="http://schemas.android.com/aapt"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:pathData="M13.452,2H6v1.75h7.088l6.629,6.656 1.24,-1.234 -6.885,-6.914a0.875,0.875 0,0 0,-0.62 -0.258zM8.933,12.379l1.414,-1.415L8.933,9.55l-1.415,1.414 1.415,1.415z" >
```
MozillaNoah, line 12: the path data is copy/pasted from the main asset, `mozac_ic_shopping_24`

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -50,4 +51,5 @@ data class BrowserState(
     val restoreComplete: Boolean = false,
     val locale: Locale? = null,
     val showExtensionProcessDisabledPopup: Boolean = false,
+    val shoppingSheetExpanded: Boolean = false,
```
gabrielluong, line 54: I don't think this should be in the `BrowserState` since the state here shouldn't need to know anything about the application UI state. The Browser state should be limited to properties related to the browser features (used loosely here).

MozillaNoah: Fair enough. Do you have an alternative location in mind? AppState seemed too top-level, but seems more appropriate than `BrowserState` now.

gabrielluong: I think you're looking at `AppStore`. Here's a prior write-up of the rationale of the `AppStore` https://github.com/mozilla-mobile/fenix/issues/20704#issuecomment-956522471

MozillaNoah: Updated

MozillaNoah:
Fixing some leftover refactoring and adding a couple tests soon

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -50,4 +50,5 @@ data class BrowserState(
     val restoreComplete: Boolean = false,
     val locale: Locale? = null,
     val showExtensionProcessDisabledPopup: Boolean = false,
+
```
gabrielluong, line 53: nit

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -216,4 +216,9 @@ sealed class AppAction : Action {
     data class UpdateStandardSnackbarErrorAction(
         val standardSnackbarError: StandardSnackbarError?,
     ) : AppAction()
+
+    /**
+     * [AppAction] used to update the expansion state of the shopping sheet.
+     */
+    data class ShoppingSheetExpandedUpdated(val expanded: Boolean) : AppAction()
```
rahulsainani, line 223: non blocking naming suggestion  -->  `ShoppingSheetStateUpdated` or  `ReviewQualityCheckSheetStateUpdated`

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
```diff
@@ -13,22 +13,27 @@ import mozilla.components.browser.state.selector.selectedTab
...
  */
 class ReviewQualityCheckFeature(
+    private val appStore: AppStore,
     private val browserStore: BrowserStore,
     private val shoppingExperienceFeature: ShoppingExperienceFeature,
     private val onAvailabilityChange: (isAvailable: Boolean) -> Unit,
+    private val onBottomSheetCollapsed: () -> Unit,
 ) : LifecycleAwareFeature {
     private var scope: CoroutineScope? = null
+    private var expandedScope: CoroutineScope? = null
```
rahulsainani, line 36: ❓ Could we somehow use one same scope for collecting both flows?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
```diff
@@ -42,9 +47,18 @@ class ReviewQualityCheckFeature(
                 .distinctUntilChanged()
                 .collect(onAvailabilityChange)
         }
+
+        expandedScope = appStore.flowScoped { flow ->
+            flow.mapNotNull { it.shoppingSheetExpanded }
+                .distinctUntilChanged()
+                .collect { expanded ->
+                    if (!expanded) onBottomSheetCollapsed()
+                }
```
rahulsainani, line 56: An idea
```suggestion
                .filterNot { it }
                .collect {
                    onBottomSheetCollapsed()
                }
```

fenix/app/src/main/res/drawable/ic_shopping_selected.xml
rahulsainani: yeah makes sense, I'm assuming it's not straightforward to gradient tint programmatically with the APIs at hand?

fenix/app/src/test/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeatureTest.kt
```diff
@@ -205,6 +220,31 @@ class ReviewQualityCheckFeatureTest {
             assertEquals(1, availabilityCount)
             assertFalse(availability!!)
         }
+
+    @Test
+    fun `WHEN the shopping sheet is collapsed THEN the collapsed callback is called`() {
```
rahulsainani, line 225: Nice! 

Let's add another test case where we assert `onBottomSheetCollapsed` is not called when expanded is true. Right now, If we remove the `!expanded` check, none of the tests fail, the other test would catch that case.

rahulsainani (approved):
Overall this looks good!  Left a question about scope, a comment about test and other optional suggestions.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
MozillaNoah: I don't believe so in this case, since the scopes are being started from two different stores.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFeature.kt
```diff
@@ -42,9 +49,20 @@ class ReviewQualityCheckFeature(
                 .distinctUntilChanged()
                 .collect(onAvailabilityChange)
         }
+
+        appStoreScope = appStore.flowScoped { flow ->
+            flow.mapNotNull { it.shoppingSheetExpanded }
+                .distinctUntilChanged()
+                .drop(1) // Needed to ignore the initial emission from the Store setup
```
rahulsainani, line 56: Let's set null as default value for expanded property in `AppState` and remove the `drop`. This can lead to issues, we don't always know if flow will begin emitting from initial state.

PR #3814, opened 2023-09-27
Bug 1850610 - Fix review checker onboarding bottom sheet height

rahulsainani:
### How
When the state transitions from `Initial` to `NotOptedIn` aka onboarding opt in is dispayed, set the bottom sheet height to fully expanded, as we don't want to set it to fully expanded when state changes from OptedIn to NotOptedIn but only in this specific scenario.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1850610


fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFragment.kt
```diff
@@ -66,4 +70,18 @@ class ReviewQualityCheckFragment : BottomSheetDialogFragment() {
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        store.flowScoped(viewLifecycleOwner) { flow ->
+            val initial = Pair<ReviewQualityCheckState?, ReviewQualityCheckState?>(null, null)
+            flow.scan(initial) { acc, value ->
+                Pair(acc.second, value)
+            }.filter {
+                it.first is ReviewQualityCheckState.Initial && it.second is ReviewQualityCheckState.NotOptedIn
+            }.collect {
+                behavior?.state = BottomSheetBehavior.STATE_EXPANDED
```
Alexandru2909, line 83: Would it be worth to maybe cancel the scope after successfully expanding the bottom sheet?

rahulsainani: Not really. Since we want to set this state whenever we move from `Initial` to `NotOptedIn` state and that can occur multiple times if user doesn't opt in and keeps opening the bottom sheet. The flow is scoped to Fragment's [viewLifecycleScope](https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope) so it will auto cancel when the fragment is destroyed.

rahulsainani: Good point, we should cancel `onDestroy`, will look into it!

MozillaNoah:
We may want to double check with UX here, as it sounds like, based on the feedback doc, that this should actually be a fullscreen view and not simply a fullscreen bottom sheet with the dialog peek at the top.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ReviewQualityCheckFragment.kt
rahulsainani: Updated to use `AbstractBinding` which cancels the scope based on lifecycle.

rahulsainani:
> We may want to double check with UX here, as it sounds like, based on the feedback doc, that this should actually be a fullscreen view and not simply a fullscreen bottom sheet with the dialog peek at the top.

I think these are 2 different issues, this one solves the buttons not visible issue which is also an accessibility issue. So at least with this, we make the experience better till we figure out bottom sheet transforming into full screen experience. We can create a different bug for that, how does that sound?

Alexandru2909 (approved):
LGTM code-wise!

PR #3828, opened 2023-09-27
Bug 1855553 - Add missing copy & update existing review quality info copy

MozillaNoah:
Changes in this PR:
- Added `ProductVendor` to the `OptedIn` state, so it's now a pseudo top-level property for that entire flow. This required adding the necessary logic to `ReviewQualityCheckPreferencesMiddleware` and `ReviewQualityCheckVendorsService` 
- Added the missing text for two sections of the review quality info card. One was dependent on the first bullet point in order to include the product vendor text.
- Added the string ID to the `ProductVendor` enum.
- Added new strings for the AB and DF rating blocks.
- Removed the deprecated use of `ClickableText` for `LinkText`

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1855553

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckVendorsService.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckVendorsService.kt
@@ -25,6 +25,11 @@ interface ReviewQualityCheckVendorsService {
      * Returns the list of product vendors in order.
      */
     fun productVendors(): List<ProductVendor>
+
+    /**
+     * Returns the product's vendor.
+     */
+    fun productVendor(): ProductVendor
 }
 
 /**
@@ -53,6 +58,8 @@ class DefaultReviewQualityCheckVendorsService(
             }
         }
     }
+
+    override fun productVendor(): ProductVendor = productVendors().first()
 
     /**
      * Creates list of product vendors using the firstVendor param as the first item in the list.
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
@@ -46,6 +46,7 @@ sealed interface ReviewQualityCheckState : State {
     data class OptedIn(
         val productReviewState: ProductReviewState = ProductReviewState.Loading,
         val productRecommendationsPreference: Boolean?,
+        val productVendor: ProductVendor,
     ) : ReviewQualityCheckState {
 
         /**
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/NoAnalysis.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/NoAnalysis.kt
@@ -32,6 +32,7 @@ import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.unit.dp
 import org.mozilla.fenix.R
 import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.shopping.store.ReviewQualityCheckState
 import org.mozilla.fenix.theme.FirefoxTheme
 
 /**
@@ -39,6 +40,7 @@ import org.mozilla.fenix.theme.FirefoxTheme
  *
  * @param isAnalyzing Whether or not the displayed product is being analyzed.
  * @param productRecommendationsEnabled The current state of the product recommendations toggle.
+ * @param productVendor The vendor of the product.
  * @param onAnalyzeClick Invoked when the user clicks on the check review button.
  * @param onReviewGradeLearnMoreClick Invoked when the user clicks to learn more about review grades.
  * @param onOptOutClick Invoked when the user opts out of the review quality check feature.
@@ -51,6 +53,7 @@ import org.mozilla.fenix.theme.FirefoxTheme
 fun NoAnalysis(
     isAnalyzing: Boolean,
     productRecommendationsEnabled: Boolean?,
+    productVendor: ReviewQualityCheckState.ProductVendor,
     onAnalyzeClick: () -> Unit,
     onReviewGradeLearnMoreClick: () -> Unit,
     onOptOutClick: () -> Unit,
@@ -64,6 +67,7 @@ fun NoAnalysis(
         ReviewQualityNoAnalysisCard(isAnalyzing, onAnalyzeClick)
 
         ReviewQualityInfoCard(
+            productVendor = productVendor,
             onLearnMoreClick = onReviewGradeLearnMoreClick,
         )
 
@@ -173,6 +177,7 @@ private fun NoAnalysisPreview() {
         ) {
             NoAnalysis(
                 isAnalyzing = isAnalyzing,
+                productVendor = ReviewQualityCheckState.ProductVendor.AMAZON,
                 onAnalyzeClick = { isAnalyzing = !isAnalyzing },
                 productRecommendationsEnabled = false,
                 onReviewGradeLearnMoreClick = {},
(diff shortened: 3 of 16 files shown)


fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckVendorsService.kt
```diff
@@ -54,6 +59,8 @@ class DefaultReviewQualityCheckVendorsService(
         }
     }
 
+    override fun getProductVendor(): ProductVendor = productVendors().first()
```
calumozilla, line 62: Seems like we're defaulting to `AMAZON` for the previews. Should we use `productVendors().first()` there instead or change this to `AMAZON`?

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2120,17 +2120,21 @@
...
     <!-- Paragraph explaining the grading system we use to classify the reliability of a product's reviews. -->
-    <string name="review_quality_check_info_review_grade_header" tools:ignore="UnusedResources"><![CDATA[We assign each product’s reviews a <b>letter grade</b> from A to F.]]></string>
+    <string name="review_quality_check_info_review_grade_header"><![CDATA[We assign each product’s reviews a <b>letter grade</b> from A to F.]]></string>
     <!-- Description explaining grades A and B for review quality check adjusted grading. -->
-    <string name="review_quality_check_info_grade_info_AB">Reliable reviews. We believe the reviews are likely from real customers who left honest, unbiased reviews.</string>
+    <string name="review_quality_check_info_grade_info_AB" moz:RemovedIn="120" tools:ignore="UnusedResources">Reliable reviews. We believe the reviews are likely from real customers who left honest, unbiased reviews.</string>
+    <!-- Description explaining grades A and B for review quality check adjusted grading. -->
+    <string name="review_quality_check_info_grade_info_AB_2">We believe the reviews to be reliable.</string>
```
calumozilla, line 2127: Can we rename this string to be more descriptive instead of `_2`? What is this secondary info? same with `DF_2`

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
```diff
@@ -31,8 +33,10 @@ sealed interface ReviewQualityCheckState : State {
     /**
      * Supported product retailers.
      */
-    enum class ProductVendor {
-        AMAZON, BEST_BUY, WALMART,
+    enum class ProductVendor(@StringRes val stringId: Int) {
+        AMAZON(stringId = R.string.review_quality_check_retailer_name_amazon),
```
rahulsainani, line 37: Let's not add R imports in State, we can do this mapping on the UI layer similar to Onboarding.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckVendorsService.kt
rahulsainani, line 62: Do we need this? I think if we're just doing first, we could do that in the middleware and keep the `RQCVendorsService` API lean, wdyt?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ReviewQualityInfoCard.kt
```diff
@@ -44,27 +47,32 @@ fun ReviewQualityInfoCard(
...
     ) {
         ReviewQualityInfo(
+            vendor = vendor,
             modifier = Modifier.fillMaxWidth(),
             onLearnMoreClick = onLearnMoreClick,
         )
     }
 }
 
-@Suppress("Deprecation")
```
rahulsainani, line 53: Nice!

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
```diff
@@ -83,7 +83,10 @@ class ReviewQualityCheckStoreTest {
             tested.waitUntilIdle()
             dispatcher.scheduler.advanceUntilIdle()
 
-            val expected = ReviewQualityCheckState.OptedIn(productRecommendationsPreference = false)
+            val expected = ReviewQualityCheckState.OptedIn(
+                productRecommendationsPreference = false,
+                productVendor = FakeReviewQualityCheckVendorsService().getProductVendor(),
```
rahulsainani, line 88: Let's assert the value like `Amazon` here and not the function call, same for other tests

fenix/app/src/test/java/org/mozilla/fenix/shopping/fake/FakeReviewQualityCheckVendorsService.kt
```diff
@@ -15,4 +15,6 @@ class FakeReviewQualityCheckVendorsService(
     ),
 ) : ReviewQualityCheckVendorsService {
     override fun productVendors(): List<ProductVendor> = productVendors
+
+    override fun getProductVendor(): ProductVendor = productVendors().first()
```
rahulsainani, line 19: Yeah I don't think we need this function in the API as we are duplicating the logic in real and fake, the tests should pass in the `productVendors` when creating `FakeReviewQualityCheckVendorsService` and then assert expected to be the first item in the passed list.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/NoAnalysis.kt
```diff
@@ -51,6 +53,7 @@ import org.mozilla.fenix.theme.FirefoxTheme
 fun NoAnalysis(
     isAnalyzing: Boolean,
     productRecommendationsEnabled: Boolean?,
+    vendor: ReviewQualityCheckState.ProductVendor,
```
rahulsainani, line 56: 🔍 nit: how about naming this to `productVendor` in all composables?

fenix/app/src/main/res/values/strings.xml
MozillaNoah: This is part of the process in Fenix to update existing strings. As long as there aren't any substantive changes to the string or the UI its intended for, we just add a numbered suffix (we have instances of `_3`, adjust the comment as needed, and mark the original string as removed so we clean them up 3 releases after the number listed.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckVendorsService.kt
MozillaNoah: since it's just a preview, it doesn't matter a ton since it's just controlling which string appears

MozillaNoah: I guess my intent was to abstract the implementation away from the Middleware, since it likely shouldn't know that `productVendors()` is an ordered list with the product's vendor first. 

`productVendors().first()` wasn't the first iteration of this. I initially looked at doing something similar at the Gecko mapping process, but then discovered we need the vendor at the entire `OptedIn` section of the feature, so my options became:
1. Refactor `productVendors` and have some shared logic between two function calls
2. Write `getProductVendor` completely from scratch but with the assumption that it has a good-formed URL
3. Create `getProductVendor` and tack on `first` because of the assumptions we're making about correctness. 
4. Do `productVendors().first()` inside of Middleware and call it a day.

I opted for number 3 in order to keep the implementation simple and the Middleware blinded to the internals of either of the function.

MozillaNoah: We could also "guarantee" that the first is always the relevant one with tests and kdocs though

rahulsainani: 4 seems good to me as we use the existing API and that's tested in Store without faking another function but I also see your reasons for 3 where we hide the implementation details 👍

PR #3831, opened 2023-09-28
Bug 1855586 - Update VisitType enum from application-services

skhamis:
I wanted to completely get rid of the `VisitType` in `HistoryStorage.kt` but still want to respect that the paradigm of `swappable storage layer`. However if anyone feels strongly, I'd happily remove it as it would allow us to get rid of a bunch of transform code in a-c. 

Also I decided to fix the intermittent tests found in https://bugzilla.mozilla.org/show_bug.cgi?id=1853687 since it was mainly just race conditions and `thread.sleep` was already used in other parts of the test and fixes up the tests pretty good.

Lastly, this will need to land after https://github.com/mozilla/application-services/pull/5846

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1855586


skhamis:
Thanks to branch builds -> https://github.com/mozilla/application-services/pull/5846 this is verified building successfully with this patch. @MatthewTighe let me know if there is a "places" person on android but I'll defer to you for the review!

android-components/components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesHistoryStorageTest.kt
```diff
@@ -545,15 +543,14 @@ class PlacesHistoryStorageTest {
...
     fun `store can delete by 'range'`() = runTestOnMain {
         history.recordVisit("http://www.mozilla.org/1", PageVisit(VisitType.TYPED))
-        advanceUntilIdle()
+        Thread.sleep(10)
         history.recordVisit("http://www.mozilla.org/2", PageVisit(VisitType.DOWNLOAD))
-        advanceUntilIdle()
+        Thread.sleep(10)
         history.recordVisit("http://www.mozilla.org/3", PageVisit(VisitType.BOOKMARK))
-
+        Thread.sleep(10)
```
MatthewTighe, line 553: I think we'll definitely want to avoid adding these sleeps to the tests. Though short, the less we impact our CI pipeline's run time the better. I submitted a [patch](https://github.com/mozilla-mobile/firefox-android/pull/4005) that I think will get these tests back on track, could you rebase on it once it lands? That way, we can revert my patch if it doesn't work and go back to ignoring these until we found a solution

android-components/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
```diff
@@ -225,6 +223,9 @@ enum class VisitType(val type: Int) {
     DOWNLOAD(7),
     FRAMED_LINK(8),
     RELOAD(9),
+
+    // Internal visit type used for meta data updates. Doesn't represent an actual page visit.
+    UPDATE_PLACE(10),
```
MatthewTighe, line 228: Couple questions for my own edification:
I see a kdoc for this type that says these match Firefox desktop. Is that still true with this change?

More broadly, do you happen to know if there's reasoning behind redeclaring these enums on our side? Seems a little pointless at first glance to me to not just use the UniFFI types everywhere

MatthewTighe (commented):
This LGTM, just want to make sure I get another change to glance over once it's been rebased so holding off on approval for the moment.

android-components/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
skhamis: Great Question! So `update_place` is the only departure from desktop: https://searchfox.org/mozilla-central/source/toolkit/components/places/nsINavHistoryService.idl#947 and is specifically used only in android, so I'd rather not shake the boat too much without understanding the historical reasons here.

I'd ideally like to use only the Uniffi'd version and remove all this extra code but it would break some practices in android-components.

1. In the "abstract" History storage we have a VisitType defined here: https://github.com/mozilla-mobile/firefox-android/blob/471ef6ed0f7371996e72a3c432284d5a0bc12f99/android-components/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt#L211-L228
 - This is what is used across the app, however this defined type is in the `interface HistoryStorage` https://github.com/mozilla-mobile/firefox-android/blob/471ef6ed0f7371996e72a3c432284d5a0bc12f99/android-components/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt#L10

which means I'd have to remove this interface type and explicitly use the Syncs storage which would break the current "swappable storage layer" paradigm that currently exists. The closest thing I could do is just simply keeping the conversion types. Totally open to ideas to get rid of those! I think overall it still shaves a couple of abstraction layers tho.

android-components/components/browser/storage-sync/src/test/java/mozilla/components/browser/storage/sync/PlacesHistoryStorageTest.kt
skhamis: Thanks, yeah I figured it was probably bad but was curious if it sparked any ideas on simply ensuring we can get some sequential things going, since I _think_ it's now happening too fast. I don't mind holding off until that patch and will rebase after it lands.

skhamis: This has been rebased on your patch (btw, thanks for fixing that!)

android-components/components/concept/storage/src/main/java/mozilla/components/concept/storage/HistoryStorage.kt
MatthewTighe: Ah, the swappable implementation thing makes sense in theory. Let's not rock the boat to hard in that case and just keep the patch as is

MatthewTighe:
FYI, it looks like the [patch](https://github.com/mozilla-mobile/firefox-android/pull/4005) I landed to help with the tests didn't quite solve the issue and needed to be reverted. For now, let's just keep ignoring the tests and our team will try to sort that out in a follow-up

skhamis:
> FYI, it looks like the [patch](https://github.com/mozilla-mobile/firefox-android/pull/4005) I landed to help with the tests didn't quite solve the issue and needed to be reverted. For now, let's just keep ignoring the tests and our team will try to sort that out in a follow-up

That sounds good -- figured I'd give it a whirl but definitely needs more looking into as it's own patch. However rebasing got rid of it all that test changes so it's all good now!

MatthewTighe (approved):
Cool, I think we can land this now. What's the strategy for bringing this kind of breaking change in? Will there be a manual A-S release or will be just land this as soon as the usual nightly build is ingested?

skhamis:
Thanks! Yes since a-s has the PR approved -> https://github.com/mozilla/application-services/pull/5846 once I land that, I'll wait for the next nightly (which will fail I believe), then rebase this and land once the CI successfully runs! 

The only thing is I'm waiting for the iOS side of things to be approved: https://github.com/mozilla-mobile/firefox-ios/pull/16707  before I land the breaking change in a-s

PR #3888, opened 2023-10-02
Bug 1856261 - reset isTablet when onStop

jackyzy823:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1856261


jackyzy823:
[First version](https://github.com/mozilla-mobile/firefox-android/pull/3888/commits/05c1bb7aa8f79293c2a25f4b8323969609cda90b) couldn't pass test `WHEN fragment configuration change enables tablet size twice THEN tablet action items are only added once` .

I'm not sure what will happens if addTabletActions invoked by onUpdateToolbarForConfigurationChange again. (like 6 buttons?)

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -122,7 +122,10 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
 
         browserToolbarView.view.addNavigationAction(leadingAction)
 
-        updateToolbarActions(isTablet = resources.getBoolean(R.bool.tablet))
+        this.isTablet = resources.getBoolean(R.bool.tablet)
+        if (this.isTablet) {
+            addTabletActions(requireContext())
+        }
```
gabrielluong, line 128: Sorry drive-by: Do we ever care about the else branch here to remove the tablet actions if it's no longer a tablet? Or is it because this is only ever called on `onViewCreated` that we should expect the toolbar to not have any of the tablet actions?

jackyzy823: The second. We should expect the toolbar to not have any of the tablet actions.

gabrielluong:
bors try

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
mavduevskiy: That's actually a good question, I couldn't make the app reach `removeTabletActions()` block. I don't have a real foldable device, that might be the issue. @zmckenney Could you help us understand that block of could, in which case do we expect it to get executed?

mavduevskiy (commented):
Hey, jackyzy823, thanks for the PR! I remember seeing the bug, it's really easy to hit it. Thanks for taking a look into it.

The solution you provided works just fine, but there are a small things that make me a bit concerned to approve it right away. 

It removes part of the code flow, the change prevents `removeTabletActions` from ever to be called during that lifecycle call. Now we are not exactly sure if it was ever called, but technically it is possible for fragment parameter to be true while the function argument to be false. 

While it doesn't mean that we need to keep the `removeTabletActions` call, if we were to remove it – we should prove that it's a safe thing to do.

It is also related to lifecycle and tablet/foldable devices, so we would like to be extra careful as testing all the cases is challenging. The previous system was aiming to support a cycle of events, triggering decision making on two brunches every time, which is a good approach.

The actual culprit here is the flag that fragment is holding. The flag is meant to keep track of the fragment state (is it a tablet or not), but the Fragment instance can outlive the view, and in that case [this](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt#315) code stops working correctly. It's expected to be called at least once, but it doesn't happen if the fragment retains true value from the previous view. 

It looks like the proper fix would be just resetting the flag when the view gets destroyed, while your solution would kinda "brute force" that case and through it under the carpet. What do you think?

gabrielluong:
> It looks like the proper fix would be just resetting the flag when the view gets destroyed, while your solution would kinda "brute force" that case and through it under the carpet. What do you think?

I do like the idea here to reset the `this.isTablet` flag (possibly during `onStop`?). I had a similar thought when I looked at this PRs awhile back that I was just a bit nervous about avoiding the else branch in `updateToolbarActions` simply because I did not understand the code well and why it was structured the way it is. 

@mavduevskiy Thank you for taking a deeper look into this and considering the different edge cases even if we don't know yet why it was structured the way it is. This is a great instinct to keep developing.

mavduevskiy (changes requested):
Thanks, @jackyzy823!

Just a small change to cover for potential bug and we can land it!

mavduevskiy (approved):
thanks, @jackyzy823 
lgtm! shipping

willdurand:
@mergifyio backport release_v120

willdurand:
https://github.com/Mergifyio backport releases_v120

PR #3916, opened 2023-10-03
Bug 1856708 - Fix ReaderView to use `storage.session` correctly

willdurand:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1856708

### DIFF
--- a/android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview-background.js
+++ b/android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview-background.js
@@ -17,11 +17,11 @@ browser.runtime.onMessage.addListener(async (message, sender, sendResponse) => {
        });
        break;
      case 'addSerializedDoc':
-        storage.session.set(sender.contextId.toString(), message.doc);
+        browser.storage.session.set(sender.contextId.toString(), message.doc);
         break;
      case 'getSerializedDoc':
-       let doc = await storage.session.get(message.id);
-       storage.session.delete(message.id);
+       let doc = await browser.storage.session.get(message.id);
+       browser.storage.session.remove(message.id);
        sendResponse(doc);
        break;
      default:


android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview-background.js
```diff
@@ -17,11 +17,11 @@ browser.runtime.onMessage.addListener(async (message, sender, sendResponse) => {
        });
        break;
      case 'addSerializedDoc':
-        storage.session.set(sender.contextId.toString(), message.doc);
+        browser.storage.session.set(sender.contextId.toString(), message.doc);
```
zmckenney, line 20: The arguments for `browser.storage.session.set` are incorrect and I think we should be waiting (?) so suggesting an alternative. 

Also, in debugging I found that `addSerializedDoc` is called twice and `getSerializedDoc` is called twice when opening a readerview. After each closing of readerview when the user goes to open the same page in readerview the `contextId` has changed and it again calls `addSerializedDoc` and `getSerializedDoc` twice.

Unless I'm understanding incorrectly, shouldn't the correct behavior be the first opening of the readerview we add the serialized doc with a persistent contextId (or another type of UID) and then subsequent openings of this page call we just call `getSerializedDoc` and return the cached doc?
```suggestion
        let sessionContextKey = sender.contextId.toString();
        await browser.storage.session.set({sessionContextKey: message.doc});
```

boek:
@willdurand Possible I wasn't doing it right but it appears that `.set` requires an object now not a key and value parameter. I also took a whack at fixing this with https://github.com/mozilla-mobile/firefox-android/pull/3922

zmckenney (changes requested):

willdurand:
> @willdurand Possible I wasn't doing it right but it appears that `.set` requires an object now not a key and value parameter. I also took a whack at fixing this with https://github.com/mozilla-mobile/firefox-android/pull/3922

sigh, right.. can that patch be uplifted to beta?

android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview-background.js
willdurand: > The arguments for `browser.storage.session.set` are incorrect and I think we should be waiting (?) so suggesting an alternative.

right...

> the `contextId` has changed

looks like a pre-existing issue, but yeah...

willdurand:
Let's close this in favor of https://github.com/mozilla-mobile/firefox-android/pull/3922

PR #3917, opened 2023-10-03
Bug 1810047 - Implement UI for webextensions optional permissions.

willdurand:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1810047


willdurand:
image

fenix/app/src/main/java/org/mozilla/fenix/extension/WebExtensionPromptFeature.kt
```diff
@@ -108,10 +112,32 @@ class WebExtensionPromptFeature(
...
+        promptRequest: WebExtensionPromptRequest.AfterInstallation.OptionalPermissions,
+    ) {
+        val shouldGrantWithoutPrompt = Addon.localizePermissions(promptRequest.permissions, context).isEmpty()
+
+        // If we don't have any promptable permissions, just proceed.
+        if (shouldGrantWithoutPrompt) {
+            promptRequest.onConfirm(true)
+            consumePromptRequest()
+            return
+        }
```
Amejia481, line 133: Could you provide me more context around when this case could happen? As it seen a bit odd going the granting path instead of deny one, maybe I don't have enough context around it.

fenix/app/src/main/java/org/mozilla/fenix/extension/WebExtensionPromptFeature.kt
```diff
@@ -108,10 +112,32 @@ class WebExtensionPromptFeature(
...
+
+        showPermissionDialog(
+            // This is a bit of a hack so that the permission prompt only lists
+            // the optional permissions that are requested.
+            addon = addon.copy(permissions = promptRequest.permissions),
```
Amejia481, line 138: I think to avoid future issues, we could just add a new property on the prompt for optional permissions, this way we will explicit about each uses of the permissions  :)

fenix/app/src/main/java/org/mozilla/fenix/extension/WebExtensionPromptFeature.kt
willdurand: Yep, that's what we are doing on Desktop: https://searchfox.org/mozilla-central/rev/29bdf6ff9965a647c6f64d63fed2b5bd094532c7/browser/modules/ExtensionsUI.sys.mjs#283-287 - Also, we're doing that in `AddonUpdater`: https://github.com/mozilla-mobile/firefox-android/blob/b0d480a03901c249f779bcc1c2f6cd0d38f645ae/android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/update/AddonUpdater.kt#L214

fenix/app/src/main/java/org/mozilla/fenix/extension/WebExtensionPromptFeature.kt
willdurand: I considered this but... an add-on can have optional permissions listed in the manifest, and that's what the property on an `Addon` should probably contain.

Now, when the extension invokes `permissions.request()`, it can pass any of these optional permissions, either one, two or all, and this is what `promptRequest.permissions` contains here. That's why I am not sure a new property on `Addon` would be a better approach.

Thinking about it more, maybe the fragment shouldn't receive an `Addon` instance but the name, icon, permissions, etc.?

fenix/app/src/test/java/org/mozilla/fenix/extension/WebExtensionPromptFeatureTest.kt
```diff
@@ -216,4 +218,77 @@ class WebExtensionPromptFeatureTest {
...
+            extension = mockk(),
+            // The "scripting" API permission doesn't have a description so we should not show a dialog for it.
+            permissions = listOf("scripting"),
+            onConfirm = onConfirm,
+        )
+
+        webExtensionPromptFeature.handleOptionalPermissionsRequest(addon = mockk(relaxed = true), promptRequest = promptRequest)
+
+        verify(exactly = 0) { webExtensionPromptFeature.showPermissionDialog(any(), any(), any()) }
+        verify(exactly = 1) { onConfirm(true) }
```
Amejia481, line 276: If this is part of the core functionality, we should have this to AC level, where we auto-grant or reject without the UI to be informed. Wdyth?

fenix/app/src/main/java/org/mozilla/fenix/extension/WebExtensionPromptFeature.kt
Amejia481: > Thinking about it more, maybe the fragment shouldn't receive an Addon instance but the name, icon, permissions, etc.?

That sounds like a better approach!

Amejia481: Just bringing this related comment https://github.com/mozilla-mobile/firefox-android/pull/3917#discussion_r1345938566

fenix/app/src/test/java/org/mozilla/fenix/extension/WebExtensionPromptFeatureTest.kt
willdurand: Where would you see that happening? We need an `Addon` to determine whether we need to early return.

fenix/app/src/main/java/org/mozilla/fenix/extension/WebExtensionPromptFeature.kt
Amejia481: As this is an already known pattern, and we don't have nothing to do on the UI, wee could just add this logic to `GeckoEngine` and granting there, so the logic will be shared for all consumers.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -345,8 +345,22 @@ class GeckoEngine(
...
+                extension: org.mozilla.geckoview.WebExtension,
+                permissions: Array<out String>,
+                origins: Array<out String>,
+            ): GeckoResult<AllowOrDeny>? {
+                val result = GeckoResult<AllowOrDeny>()
+                webExtensionDelegate.onOptionalPermissionsRequest(
+                    GeckoWebExtension(extension, runtime),
+                    permissions.toList() + origins.toList(),
+                ) { allow ->
                     if (allow) result.complete(AllowOrDeny.ALLOW) else result.complete(AllowOrDeny.DENY)
```
Amejia481, line 364: @willdurand could we do the granting here if permissions are empty?

pmarks-net:
Thanks for [fixing my extension](https://discourse.mozilla.org/t/ipvfoo-doesnt-work/122979/5), but be aware that the permissions granted via this dialog are not visible (and thus cannot be revoked) from the extension's Permissions UI.

willdurand:
> and thus cannot be revoked

the extension can revoke them

willdurand:
@pmarks-net sorry for the quick reply the other day, to expand on that: yes, you're right. We don't have UI for optional permissions in the Fenix details view. There is a bug for that, see: https://bugzilla.mozilla.org/show_bug.cgi?id=1811563. We'll get that fixed eventually but I don't have an ETA yet. In the meantime, extensions can offer optional permission removal using `permissions.remove()`.

PR #3919, opened 2023-10-03
Bug 1856399 - Use distinct version code in Fenix AAB

gbrownmozilla:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1856399


gbrownmozilla:
Clear the 'g' bit in Fenix version codes used in AAB builds, to ensure that Fenix AAB version codes are always distinct from Fenix APK version codes.

I feel like there should be a more elegant way to detect AAB-ness without passing -Paab ... suggestions?

There are similar issues for Focus and Reference-Browser. Once this is sorted out, I can handle those cases in separate PRs.

jcristau:
I'm wondering if it'd be better to switch it around so the aab is the one with the higher version code?
Or does it not really matter because once we start shipping the bundle we'll stop uploading apks?

gbrownmozilla:
Right, I think we would never push APKs and AAB for the same build to Google Play...but who knows what surprises the future holds? 

I've flipped around the 'g' bit to be 1 for AAB, and also set the 'p' and 'x' bits for AAB: That should guarantee that the AAB version code is > any APK version code created at the same time.

android-components/plugins/config/src/main/java/ConfigPlugin.kt
```diff
@@ -144,17 +144,19 @@ object Config {
...
 
         // 'p' bit is 1 for 64-bit architectures.
-        if (abi == "arm64-v8a" || abi == "x86_64") {
+        if (aab || abi == "arm64-v8a" || abi == "x86_64") {
             version = version or (1 shl 1)
         }
 
-        // 'g' bit is currently always 1 (see comment above)
-        version = version or (1 shl 0)
+        // 'g' bit is 0 for APK, 1 for AAB
```
JohanLorenzo, line 156: 👍

JohanLorenzo (approved):
Changes make sense to me! I'm glad we're able to reemploy the smallest digit here. Thanks for updating the comments 👍

PR #3938, opened 2023-10-04
Bug 1857093 - Update Sentry to version 6.33.1

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1857093


gabrielluong:
@jonalmeida Could you figure out if you should be reviewing this change or potentially also see if there are also others that should learn how to test Sentry bumps as well.

jonalmeida:
What I usually do is:

1. Look through the change logs that Ryan links to (thank you!) in the bug and look for _behaviour changes_. Breaking API changes are easy to catch because the build will fail, but behaviours can change internally and result in confusion ~3 months later when it his release.
2. A sanity test to make sure the code paths that we implement still work as expected (here, can still create and send crash information up to Sentry).
3. Are there new features that we should be leveraging. We should file follow-up bugs for those to investigate further.

titooan (approved):
Seems good to me!

titooan (approved):
Looks good to me. 
I haven't seen any behaviour change in the changelogs for the new versions linked in the bug comments, neither any new feature that we should be leveraging.

PR #3975, opened 2023-10-06
Bug 1856957 - Translations UI Download Cycle Spinner.

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1856957


iorgamgabriel:
https://github.com/mozilla-mobile/firefox-android/assets/8118371/d8dcbc41-c6ec-4f62-b04f-a3d966979587

fenix/app/src/main/res/drawable/ic_download_circle_spinner.xml
```diff
@@ -0,0 +1,16 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
```
gabrielluong, line 1: When it comes to adding icons, we should pull the resource from https://github.com/FirefoxUX/acorn-icons/tree/main/icons/mobile. You will see the icons that are available in our iconography in this [figma](https://www.figma.com/file/Kv7wWr8vH6AQrGhu9vmjdE/Mobile-Icons?node-id=221%3A1223&mode=dev). These icons will have their paths optimized and formatted in acorn-icons, which we should treat as our source of truth.

In this case, we didn't need to import this because it should be the equivalent of [mozac_ic_sync_24.xml](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/ui/icons/src/main/res/drawable/mozac_ic_sync_24.xml). I am in the middle of updating our existing icons to use the latest paths in acorn-icons as part of [Bug 1842389](https://bugzilla.mozilla.org/show_bug.cgi?id=1842389). For the time being, I think we can remove this newly added icon and just use `mozac_ic_sync_24.xml`

fenix/app/src/main/java/org/mozilla/fenix/translations/LoadingIndicator.kt
```diff
@@ -0,0 +1,153 @@
...
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.compose.button.DEFAULT_MAX_LINES
+import org.mozilla.fenix.theme.FirefoxTheme
+
+@Composable
+@LightDarkPreview
+private fun DownloadIndicatorPreview() {
```
gabrielluong, line 46: Could we move the Preview to the bottom of the file?

fenix/app/src/main/java/org/mozilla/fenix/translations/LoadingIndicator.kt
```diff
@@ -0,0 +1,153 @@
...
+    modifier: Modifier = Modifier,
+    icon: Painter? = null,
+    backgroundColor: Color,
+    iconModifier: Modifier = Modifier,
+    tint: Color,
+) {
+    // Required to detect if font increased due to accessibility.
+    val fontScale: Float = LocalConfiguration.current.fontScale
+
+    Row(
```
gabrielluong, line 105: Since our use case for the download indicator is inside of a `PrimaryButton`, I think we should be looking to use a `Button` here instead of a `Row`. Would it make sense to replace this `Row` with the `Button` definition in https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt#54-61 for the time being?

gabrielluong: Actually, I am not fully correct here, we also have an use case with the loading indicator as a standalone list item. See https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7341%3A209640&mode=dev 

Screen Shot 2023-10-06 at 11 53 09 AM

gabrielluong: Here's some changes I am looking for:
- We should have a `LoadingIndicator` that displays the icon only (without the text) and handles the rotation animation as its own composable. This will be used in a `ListItem` as part of the `afterListAction` as seen in https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt#223. This will handle the use case in "Download languages" screen above and can also be reused in the Button use case.
- We should change your existing composable with the icon and text to be the Button use case. This means changing the `Row` to a `Button` as described earlier. I am specifically looking for you to split up the indicator to its own Composable that could be reuse for both use cases. I think we should make the button standalone for now instead of changing our existing [Button](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt) to accept other composable `content` for the time being and think about it as a follow up after we land our first iteration of this button.

iorgamgabriel: Hi, I don't think it should be a button because it is not clickable and at accessibility it shouldn't say "button" . I will modify all to support the use case for download language screen .

iorgamgabriel: From figma it should be an indicator not a button

gabrielluong: Wouldn't the use case inside of the bottom sheet be a button? https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7202%3A99073&mode=dev

Screen Shot 2023-10-06 at 12 20 12 PM

gabrielluong: We're also expecting this to be a button in the bottom as described in the accessibility notes in https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7224%3A198420&mode=dev

Screen Shot 2023-10-06 at 12 22 11 PM

gabrielluong: To be clear, I am asking for 2 Composables:
- Button
- Indicator

iorgamgabriel: https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7221:186934&mode=design#571221606

iorgamgabriel: I understand two elements , but I'm not sure it should be a button .

fenix/app/src/main/res/drawable/ic_download_circle_spinner.xml
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/LoadingIndicator.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/LoadingIndicator.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * Icon for Download indicator.
+ *
+ * @param modifier [Modifier] to be applied to the icon layout.
+ * @param icon [Painter] used to be displayed.
+ * @param tint to be applied to the icon.
+ * @param iconContentDescription Optional contentDescription for the icon.
+ * */
+@Composable
+fun DownloadIconIndicator(
```
ohall-m, line 95: I tested this component in TalkBack and I think it may need some accessibility refinement. Currently, TalkBack only says "Translating" on the moving spinner icon.

I looked around in our code base some more and an indeterminate progress bar has recently been added via [shopping](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProgressIndicator.kt#31,41,51).

To test the shopping loading icon feature:
1. Turn on TalkBack
2. Navigate to Walmart.com and select a more uncommon item (I picked things like grocery items)
3. Go to "Open review checker" in toolbar
4. "Check review quality" or something similar for an already processed item
5. Spinner will show up
6. Navigate to spinner
7. Talkback announces:  "In Progress", "Progress Bar". I think "In Progress" is the `<value>` and "Progress Bar" is the `<role>`. I looked in the [layout inspector semantics](https://developer.android.com/jetpack/compose/semantics) and it looks like it inherited from "ProgressBarRangeInfo". (Aside, I think it probably should have became a live region to hear that announcement without needing to manually navigate to it, but it could be my TalkBack settings?)

We used an [indeterminate progress circle indicator](https://developer.android.com/jetpack/compose/components/progress) here to make a [progress indicator](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProgressIndicator.kt#31,41,51).

A quick look at [CircularProgressIndicator](https://developer.android.com/reference/com/google/android/material/progressindicator/CircularProgressIndicator) doesn't seem to have a way to override the icon/animation (could be missing something here, please double check).

I'm thinking we will need to update this component to use [custom semantics](https://developer.android.com/jetpack/compose/accessibility) to make this have the `ProgressBarRangeInfo` [role](https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/SemanticsProperties#ProgressBarRangeInfo()).

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+    backgroundColor: Color,
+    iconModifier: Modifier = Modifier,
+    iconTint: Color,
+    iconContentDescription: String? = null,
+) {
+    // Required to detect if font increased due to accessibility.
+    val fontScale: Float = LocalConfiguration.current.fontScale
+
+    Button(
+        enabled = false,
```
ohall-m, line 132: Just leaving a note for reference - I think `enabled = false,` is the correct state  based on this [conversation](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7221:186934&mode=design#571221606) -> "not sure it should stay a button, but if so, it should be programmatically `disabled`". 

The TalkBack response for the button seemed as expected for `DownloadIndicatorPreview` (besides the progress bar issue mentioned elsewhere).

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
ohall-m: @anna-yeddi, does this seem to be the right direction?

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+            animation = tween(ANIMATION_DURATION_MILLIES, easing = LinearEasing),
+            repeatMode = RepeatMode.Restart,
+        ),
+    )
+    return angle
+}
+
+@Composable
+@LightDarkPreview
+private fun DownloadIconIndicatorPreview() {
```
ohall-m, line 200: Thanks for the previews for example usage!

ohall-m (changes requested):
Thanks for this patch, I tested it and it looks good! 

I think if you update the role to be `ProgressBarRangeInfo` on the spinner element, then it will provide the correct TalkBack responses.

fenix/app/src/main/java/org/mozilla/fenix/translations/LoadingIndicator.kt
ohall-m: Thanks for the patch! I think the `Indicator` should have the role `ProgressBarRangeInfo`. I left some more info below.

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+            color = textColor,
+            modifier = Modifier.clearAndSetSemantics { },
+            style = FirefoxTheme.typography.button,
+            maxLines = if (fontScale > 1.0f) Int.MAX_VALUE else DEFAULT_MAX_LINES,
+        )
+    }
+}
+
+@Composable
+private fun BaseLoadingIndicator(
```
gabrielluong, line 171: Let's remove these Base composables. Unlike our design components, I don't think we need such wrapper classes over the Material `Icon`.

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+                iconContentDescription = iconContentDescription,
+            )
+            Spacer(modifier = Modifier.width(8.dp))
+        }
+
+        Text(
+            text = text,
+            textAlign = TextAlign.Center,
+            color = textColor,
+            modifier = Modifier.clearAndSetSemantics { },
```
gabrielluong, line 163:
```suggestion
            modifier = Modifier.clearAndSetSemantics {},
```
nit: Prefer `{}` over `{ }`. Same in other places.

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+) {
+    BaseDownloadIndicator(
+        text = text,
+        textColor = textColor,
+        modifier = modifier,
+        buttonContentDescription = contentDescription,
+        icon = icon,
+        backgroundColor = backgroundColor,
+        iconModifier = iconModifier.then(
+            Modifier.wrapContentSize(),
```
gabrielluong, line 79: A bit of a naive question - is this `wrapContentSize()` doing anything for else? I have tried this with and without this and it appears the same in the preview. So, I am wondering what this is doing.

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
gabrielluong, line 163: Why is `clearAndSetSemantics` needed here? This doesn't exist in our current [Button](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt#44) implementation.

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+    modifier: Modifier = Modifier,
+    contentDescription: String? = null,
+    backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
+    textColor: Color = FirefoxTheme.colors.textActionPrimary,
+    icon: Painter? = null,
+    iconContentDescription: String? = null,
+    iconModifier: Modifier = Modifier,
+    iconTint: Color = FirefoxTheme.colors.iconActionPrimary,
+) {
+    BaseDownloadIndicator(
```
gabrielluong, line 71: Now that I have another chance to re-examine everything and see what parameters we need for the composable, I think I was wrong about needing to copy over over the base [Button](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt#44) in order to support our button with an animated icon and label. All we needed to add in `PrimaryButton` is an additional parameter to support `iconModifier`. 

I would propose either (1) making the changes to use PrimaryButton in this PR or (2) follow up with extending and reusing PrimaryButton.

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.compose.button.DEFAULT_MAX_LINES
+import org.mozilla.fenix.theme.FirefoxTheme
+
+private const val ANIMATION_DURATION_MILLIES = 2000
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,235 @@
...
+
+/**
+ * Download indicator.
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
iorgamgabriel: Done I refactor it using Primary Button.

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
iorgamgabriel: Done

ohall-m: Thank you for all the work on this. I think this solution will meet accessibility criteria with just a few more adjustments. 

So, this patch as it stands is manually overriding everything to provide an accurate screen reader description. However, we are losing the semantic value of the components. I think this might be okay, but needs more discussion.

So originally, I hoped we could do this:
```
        modifier = Modifier.semantics {
            role = Role.ProgressBarRangeInfo
        }
```

([Example](https://bryanherbst.com/2021/02/19/compose-role-semantics/). However, it seems compose does not support this (wasn't in the list of IDE choices when I imported role.)). If this was possible, then we would have gotten a nice inherited `role` and `value` from the Progress Bar.

I also didn't see a way to customize a [`CircularProgressIndicator()`](https://developer.android.com/reference/com/google/android/material/progressindicator/CircularProgressIndicator) for our needs e.g, changing the drawable to our sync icon. 


**I have a question:**
1. I think we should remove `clearAndSetSemantics` or could you provide further justification? That is overriding the semantic meaning of the elements and generally we need special reasoning to do so. I also do not see its use in other places in our code base. 

I'm bringing it up because setting the contentDescription this way via `clearAndSetSemantics` to "Translating in Progress" is losing the `role` of button and that it is not-enabled.

**My recommendation:**
* Please remove `clearAndSetSemantics` - it is overriding too much semantic content, unless there is further justification
* Add a parameter for us to set the `contentDescription` on the elements as a freestanding option 
* Set the default `contentDescription` of the icon to a string of `In progress`.
* Keep the default `contentDescription` of the translating button to a string of `Translating in progress`
*  (We will also need a `Downloading in progress` string in the future for use on the download screen)
* I'll add a note on the Figma about this to see if we get further recommendations and notify Emily of the string need here. (We can return and refine this later as well. I think these recommendations will push it fairly close to where it needs to be.)

ohall-m: (Aside note, I think when we add this to the sheet proper, we will want the containing area to be a polite live-region to notify the user of the button change.)

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -47,15 +50,18 @@ private fun Button(
     backgroundColor: Color,
     modifier: Modifier = Modifier,
     icon: Painter? = null,
+    iconModifier: Modifier = Modifier,
     tint: Color,
     onClick: () -> Unit,
+    enabled: Boolean = true,
```
gabrielluong, line 56: Let's reorder the `enabled` to be after the `modifier` parameter. We want to keep the lambda/content parameters as the last parameter so that we can `Button { onButtonClick() }` or `Button { Icon() }` if it was a `content` parameter.

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -99,6 +109,8 @@ fun PrimaryButton(
     textColor: Color = FirefoxTheme.colors.textActionPrimary,
     backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
     icon: Painter? = null,
+    iconModifier: Modifier = Modifier,
+    buttonEnabled: Boolean = true,
```
gabrielluong, line 113:
```suggestion
    enabled: Boolean = true,
```
I think this should be understood since it's the same naming used for the Material `Button` composable.

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -107,8 +119,10 @@ fun PrimaryButton(
         backgroundColor = backgroundColor,
         modifier = modifier,
         icon = icon,
+        iconModifier = iconModifier,
         tint = FirefoxTheme.colors.iconActionPrimary,
         onClick = onClick,
+        enabled = buttonEnabled,
```
gabrielluong, line 125: We want to reorder below `modifier` after making the changes above.

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
gabrielluong, line 113: Can you also add these to the other Buttons (eg Secondary, Destructive, etc)?

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,160 @@
...
+ * If it is set to a low number, the speed of the rotation will be higher.
+ */
+private const val ANIMATION_DURATION_MS = 2000
+
+/**
+ * Icon for Download indicator.
+ *
+ * @param modifier [Modifier] to be applied to the icon layout.
+ * @param icon [Painter] used to be displayed.
+ * @param tint to be applied to the icon.
```
gabrielluong, line 46:
```suggestion
 * @param tint Tint [Color] to be applied to the icon.
```

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,160 @@
...
+ * @param iconContentDescription Optional contentDescription for the icon.
```
gabrielluong, line 47:
```suggestion
 * @param iconContentDescription Optional content description for the icon.
```
Prefer plain english for the description. Code and variables should be wrapped in `[]`.

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,160 @@
...
+    )
+}
+
+/**
+ * Download indicator for translations screens.
+ * It indicates that the download of the language file is in progress.
+ * @param modifier [Modifier] to be applied to the layout.
+ * @param text The download indicator text to be displayed.
+ * @param contentDescription To be applied to the layout.
+ * @param icon Optional [Painter] used to display an [Icon] before the download indicator text.
```
gabrielluong, line 78:
```suggestion
 * Download indicator for translations screens.
 * It indicates that the download of the language file is in progress.
 *
 * @param modifier [Modifier] to be applied to the layout.
 * @param text The download indicator text to be displayed.
 * @param contentDescription Content description to be applied to the button.
 * @param icon Optional [Painter] used to display an [Icon] before the download indicator text.
```

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,160 @@
...
+            modifier = Modifier
+                .background(FirefoxTheme.colors.layer1)
+                .padding(16.dp),
+            verticalArrangement = Arrangement.spacedBy(16.dp),
+        ) {
+            DownloadIconIndicator(
+                icon = painterResource(id = R.drawable.mozac_ic_sync_24),
+                modifier = Modifier.wrapContentSize(),
+                iconContentDescription =
+                stringResource(id = R.string.translations_bottom_sheet_translating_in_progress),
```
gabrielluong, line 133: nit should this be indented?

gabrielluong (commented):
This has my conditional approval assuming my remaining feedback are approval and will not need another review from me. I will leave it to @ohall-m to provide the final approval to land. Everything is looking pretty good to me. @iorgamgabriel  Thank you for your patience in all of this!

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
ohall-m: Thanks for discussing this issue in Slack with me! 

@iorgamgabriel found a great solution. We are just waiting to get a little more information on the preferred labeling to know which route to take and this patch will be good to go with a quick update.  🎉

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * @param modifier [Modifier] to be applied to the layout.
+ * @param text The button text to be displayed.
+ * @param contentDescription Content description to be applied to the button.
+ * @param icon Optional [Painter] used to display an [Icon] before the button text.
+ */
+@Composable
+fun DownloadIndicator(
+    modifier: Modifier = Modifier,
+    text: String,
```
gabrielluong, line 81:
```suggestion
    text: String,
    modifier: Modifier = Modifier,
```

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
gabrielluong: Still looking for this change

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
ohall-m: Have some more info on the labling:

TalkBack should report: "Translating in progress, button, disabled"

via what you shared earlier: 

```
    modifier = modifier.then(
        Modifier
            .clearAndSetSemantics {
                disabled()
                role = Role.Button
                contentDescription?.let { this.contentDescription = contentDescription }
            },
    ),
```

And for the indicator, "Download in progress" as the content description. (More context in the Figma, if needed too.)

With that change and what @gabrielluong mentioned, and we should be good to go.

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,157 @@
...
+ *
+ * @param modifier [Modifier] to be applied to the icon layout.
+ * @param icon [Painter] used to be displayed.
+ * @param tint Tint [Color] to be applied to the icon.
+ * @param iconContentDescription Optional content description for the icon.
+ */
+@Composable
+fun DownloadIconIndicator(
+    modifier: Modifier = Modifier,
+    icon: Painter,
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/compose/button/Button.kt
```diff
@@ -87,26 +94,33 @@ private fun Button(
...
  */
 @Composable
 fun PrimaryButton(
     text: String,
     modifier: Modifier = Modifier.fillMaxWidth(),
+    enabled: Boolean = true,
     textColor: Color = FirefoxTheme.colors.textActionPrimary,
     backgroundColor: Color = FirefoxTheme.colors.actionPrimary,
     icon: Painter? = null,
+    iconModifier: Modifier = Modifier,
```
gabrielluong, line 113: I am also expecting the `iconModifier` param for the other buttons.

iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/DownloadIndicator.kt
```diff
@@ -0,0 +1,157 @@
...
+ * @param modifier [Modifier] to be applied to the icon layout.
+ * @param tint Tint [Color] to be applied to the icon.
+ * @param iconContentDescription Optional content description for the icon.
+ */
+@Composable
+fun DownloadIconIndicator(
+    icon: Painter,
+    modifier: Modifier = Modifier,
+    tint: Color = FirefoxTheme.colors.iconPrimary,
+    iconContentDescription: String? = null,
```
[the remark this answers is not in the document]
iorgamgabriel: Done

gabrielluong (approved):

PR #4018, opened 2023-10-10
Bug 1855291 - Handle FxA exceptions in getManageAccountURL

bendk:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1855291

### DIFF
--- a/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
+++ b/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt
@@ -162,7 +162,7 @@ interface OAuthAccount : AutoCloseable {
      * @param entryPoint A string which will be included as a query param in the URL for metrics.
      * @return The URL which should be opened in a browser tab.
      */
-    suspend fun getManageAccountURL(entryPoint: FxAEntryPoint): String
+    suspend fun getManageAccountURL(entryPoint: FxAEntryPoint): String?
 
     /**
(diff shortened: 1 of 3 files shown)


mhammond:
Is there a little more context available? I don't see the callers of this function and how they handle null, and in general it seems wrong that on exception callers now need to know to handle a null string when they will have no context about why it's null.

IOW, I'm wondering why the callers of this function don't handle the exception rather than handling the null value.

bendk:
> IOW, I'm wondering why the callers of this function don't handle the exception rather than handling the null value.

That's a good question.  I mainly went this way because that's how the other [FirefoxAccount](https://github.com/mozilla-mobile/firefox-android/blob/ac8183515fbb299e46472aeeb44f0d7fba4a11b6/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/FirefoxAccount.kt) functions handle exceptions.

The only caller is the [AccountSettingsFragment](https://github.com/mozilla-mobile/firefox-android/blob/9dc2d889b41d76da463be92285d41b7ba866bab4/fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt#L382), so we could also easily make that catch the exception.

mhammond (approved):
As discussed with Ben, this doesn't seem ideal in terms of error handling, but will fix an issue on released versions. Hopefully we can fix this as we fix fxa errors in general.

bendk:
@Mergifyio backport releases_v119

DonalMe:
@mergify backport releases_v119

PR #3958, opened 2023-10-05
Bug 1857092 - Add metrics for clicks on Firefox Suggestions in Fenix

tiftran:
### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1857092

PR #4050, opened 2023-10-11
Bug 1858542 - Add Glean metrics for Firefox Suggest impressions.

linabutler:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1858542

### DIFF
--- a/android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
+++ b/android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
@@ -102,6 +102,8 @@ interface AwesomeBar {
      * @property onChipClicked A callback to be executed when a [Chip] was clicked by the user.
      * @property score A score used to rank suggestions of this provider against each other. A suggestion with a higher
      * score will be shown on top of suggestions with a lower score.
+     * @property metadata Opaque metadata associated with this [Suggestion]. A [SuggestionProvider] can use this field
+     * to pass additional information about this suggestion.
      */
     data class Suggestion(
         val provider: SuggestionProvider,
@@ -116,6 +118,7 @@ interface AwesomeBar {
         val onSuggestionClicked: (() -> Unit)? = null,
         val onChipClicked: ((Chip) -> Unit)? = null,
         val score: Int = 0,
+        val metadata: Map<String, Any>? = null,
     ) {
         /**
          * Chips are compact actions that are shown as part of a suggestion. For example a [Suggestion] from a search
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
@@ -5,6 +5,7 @@
 package org.mozilla.fenix.components.appstate
 
 import androidx.annotation.VisibleForTesting
+import mozilla.components.concept.awesomebar.AwesomeBar
 import mozilla.components.service.pocket.PocketStory.PocketRecommendedStory
 import mozilla.components.service.pocket.PocketStory.PocketSponsoredStory
 import mozilla.components.service.pocket.ext.recordNewImpression
@@ -234,6 +235,16 @@ internal object AppStoreReducer {
         )
 
         is AppAction.ShoppingAction -> ShoppingStateReducer.reduce(state, action)
+
+        is AppAction.AwesomeBarAction.VisibilityStateUpdated -> {
+            state.copy(awesomeBarVisibilityState = action.visibilityState)
+        }
+        is AppAction.AwesomeBarAction.SuggestionClicked -> {
+            state.copy(clickedSuggestion = action.suggestion)
+        }
+        is AppAction.AwesomeBarAction.EngagementFinished -> {
+            state.copy(awesomeBarVisibilityState = AwesomeBar.VisibilityState(), clickedSuggestion = null)
+        }
     }
 }
 
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsMiddleware.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsMiddleware.kt
@@ -4,6 +4,7 @@
 
 package org.mozilla.fenix.components.metrics
 
+import mozilla.components.feature.fxsuggest.FxSuggestImpressionInfo
 import mozilla.components.lib.state.Middleware
 import mozilla.components.lib.state.MiddlewareContext
 import org.mozilla.fenix.components.appstate.AppAction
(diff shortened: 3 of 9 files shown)


android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/Suggestions.kt
```diff
@@ -57,6 +66,38 @@ internal fun Suggestions(
             }
         }
     }
+
+    val currentSuggestions by rememberUpdatedState(suggestions)
+
+    LaunchedEffect(Unit) {
```
linabutler, line 72: GitHub munches comments on individual commits after a force-push (which makes sense, but I didn't think about that—should've opened the draft PR a lot sooner! 😅), but here are breadcrumbs to previous conversations:

* @MatthewTighe's comment about testing, and my explanation of why I went with this approach: https://github.com/linabutler/firefox-android/commit/6055b58d7cc5c0b87112923133a073679767ef3f#r129642181
* Our conversations from an even earlier iteration, exploring different alternatives and figuring out how they work with Compose: https://github.com/mozilla-mobile/firefox-android/commit/e9d0ea629a5083f3b0bb1c3e3ca38921feeb46d2#r129168885, https://github.com/linabutler/firefox-android/commit/69a30b3641e3b2e340ee936fb45ada76aaa313ac#r129142449

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -105,6 +105,8 @@ interface AwesomeBar {
         val onSuggestionClicked: (() -> Unit)? = null,
         val onChipClicked: ((Chip) -> Unit)? = null,
         val score: Int = 0,
+        val onSuggestionShownForEngagement: ((Position) -> Unit)? = null,
+        val onSuggestionHiddenForEngagement: (() -> Unit)? = null,
```
MatthewTighe, line 109: I think it would make more sense to remove these and replace this workflow with a (optional, to avoid breaking changes) callback to `AwesomeBar` Composable. Something like:

```kotlin
@Composable
fun AwesomeBar(
...
    onAutoComplete: (AwesomeBar.SuggestionProviderGroup, AwesomeBar.Suggestion) -> Unit,
    onVisibleSuggestionsUpdated: (List<Pair<AwesomeBar.Suggestion, AwesomeBar.Suggestion.Position>>) -> Unit = {},
    onScroll: () -> Unit = {},
)
```

Then, in combination with my below comment, the `Suggestions` Composable would call that directly. Then, in `AwesomeBarWrapper` we would define the lambda:

```kotlin
onVisibleSuggestionsUpdated = { 
    context.components.appStore.dispatch(AppAction.UpdatePendingFxSuggestImpressions(it))
},
```

There's a bit of detail left out here, so please let me know if anything isn't clear!

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestImpression.kt
```diff
@@ -0,0 +1,33 @@
...
+            advertiser = advertiser,
+            iabCategory = iabCategory,
+        )
+        else -> null
+    }
+
+interface FxSuggestImpressionRecorder {
+    fun recordPendingImpression(impression: FxSuggestImpression, position: AwesomeBar.Suggestion.Position)
+    fun forgetPendingImpression(impression: FxSuggestImpression)
+}
```
MatthewTighe, line 33: I think we can ditch this interface in favor of the callback strategy I discussed above to hopefully remove a layer of abstraction here.

MatthewTighe (commented):
Ran out of time today to leave good comments about testing, but in general I would consider how the logic could be shifted around such that it can either be tested in:

1. the reducer
2. the telemetry middleware
3. standalone classes. for example, I think directly testing the `visibleSuggestions` function will be easier than trying to actually manipulate the Compose tree in an integration test

Happy to chat more about it on or offline tomorrow though!

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
```diff
@@ -233,6 +233,13 @@ internal object AppStoreReducer {
         )
 
         is AppAction.ShoppingSheetStateUpdated -> state.copy(shoppingSheetExpanded = action.expanded)
+
+        is AppAction.RecordPendingFxSuggestImpression -> state.copy(pendingFxSuggestImpressions = state.pendingFxSuggestImpressions + action.impression)
+        is AppAction.ForgetPendingFxSuggestImpression -> state.copy(pendingFxSuggestImpressions = state.pendingFxSuggestImpressions - action.impression)
+        is AppAction.ForgetAllPendingFxSuggestImpressions -> state.copy(pendingFxSuggestImpressions = emptySet())
+
+        is AppAction.SearchEngagementFinished -> state
+        is AppAction.SearchEngagementAbandoned -> state
```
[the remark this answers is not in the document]
linabutler: Oh, yes! When I wrote this part, I think I missed that the middleware can run _before_ the state is updated, because it controls when `next()` is called—so when it sees a `SearchEngagementFinished` action, it can capture the pending impressions, and send the Glean ping, before the reducer sees that action and clears them out. Thanks!

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
MatthewTighe: > a reporting URL, advertiser name, block ID, and category name

It looked like this was already implemented as the `Suggestion.toImpression()` extension function. The `MetricsMiddleware` should be able to just map over the visible suggestions and call that, right? Or am I missing something still?

linabutler: Ah, `Suggestion.toImpression` is an extension function on `mozilla.appservices.suggest.Suggestion`, but the provider returns `AwesomeBar.Suggestion`s.

fenix/app/src/main/java/org/mozilla/fenix/components/metrics/MetricsMiddleware.kt
```diff
@@ -20,18 +21,26 @@ class MetricsMiddleware(
...
             metrics.track(Event.GrowthData.FirstAppOpenForDay)
             metrics.track(Event.GrowthData.FirstWeekSeriesActivity)
             metrics.track(Event.GrowthData.UsageThreshold)
             metrics.track(Event.GrowthData.UserActivated(fromSearch = false))
         }
+        is AppAction.AwesomeBarAction.EngagementFinished -> {
+            for (info in context.state.awesomeBarVisibilityState.suggestionsInfo()) {
+                val impressionInfo = info.suggestion.metadata?.get("impressionInfo") as? FxSuggestImpressionInfo.Amp
+                    ?: continue
+                // TODO: Record Glean impression ping here.
```
linabutler, line 40: @tiftran is adding the new `fx-suggest` ping over in #3958. This logic will submit that same ping, but with its `ping_type` set to `fxsuggest-impression` instead of `fxsuggest-click`. We'll also add some new metrics to that ping, for the suggestion position.

So there are three PRs in all (whew!) for Suggest telemetry:

1. #4110, which adds the `AwesomeBar.VisibilityState` instrumentation.
2. #3958, which adds the `fx-suggest` ping (and telemetry for clicks).
3. This PR, which will send an `fx-suggest` ping for impressions.

PR #4059, opened 2023-10-12
Bug 1858018 - Persist review checker reanalysis state

rahulsainani:
### What
Restore reanalysis state if the product was in the process of reanalyzing when the sheet was closed, i.e. hold state of `productsAnalysed` for a session (no need for persistence).

### How
– Create `ShoppingState` to be held in `AppState`. Migrate existing shopping related state inside so it's all in one place.
– Create `ShoppingReducer` that actions on `ShoppingAction`s, this is a similar pattern to `BrowserStateReducer`.
– Add tests for the above, note that the tests don't just test the reducer but the logic from action dispatch, again similar to `BrowserAction`s, making refactoring of reducer state easier without updating tests.
– Use `AppStore` in `ReviewQualityCheckNetworkMiddleware` to dispatch `AddToProductAnalysed` when reanalysis has started and `RemoveFromProductAnalysed` once reanalysis is finished.
– On `FetchProductAnalysis` action, `ShoppingState` is used to restore the reanalysis state based on the productAnalysis response when `needsAnalysis` is true, that means that analysis wasn't completed, that starts polling again. Note that polling is stopped when sheet is closed, we merely display the progress that's happening on the backend.
– To use `ProductAnalysis.needsAnalysis`, `ReviewQualityCheckService.fetchProductReview` now returns `GeckoProductAnalysis`, moving the object check and tests from the mapper. Ideally, `ProductAnalysis` should have all the fields, something to update in AC as a future improvement.

### Note
Logs in middleware and `ShoppingStateReducer` are left to make reviews easier, will be removed before landing.

### Preview
https://github.com/mozilla-mobile/firefox-android/assets/38040960/85d6dc25-4d59-487d-a120-665cb6b1a6c7

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1858018

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/shopping/ShoppingState.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/shopping/ShoppingState.kt
@@ -1,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components.appstate.shopping
+
+/**
+ * State for shopping feature that's required to live the lifetime of a session.
+ *
+ * @property productsInAnalysis Set of product ids that are currently being analysed or were being
+ * analysed when the sheet was closed.
+ * @property shoppingSheetExpanded Boolean indicating if the shopping sheet is expanded and visible.
+ */
+data class ShoppingState(
+    val productsInAnalysis: Set<String> = emptySet(),
+    val shoppingSheetExpanded: Boolean? = null,
+)
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
@@ -6,8 +6,12 @@ package org.mozilla.fenix.shopping.middleware
 
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
+import mozilla.components.browser.engine.gecko.shopping.GeckoProductAnalysis
 import mozilla.components.lib.state.MiddlewareContext
 import mozilla.components.lib.state.Store
+import mozilla.components.support.base.log.logger.Logger
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.appstate.AppAction.ShoppingAction
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckAction
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckAction.FetchProductAnalysis
 import org.mozilla.fenix.shopping.store.ReviewQualityCheckAction.RetryProductAnalysis
@@ -21,13 +25,17 @@ import org.mozilla.fenix.shopping.store.ReviewQualityCheckState.OptedIn.ProductR
  *
  * @property reviewQualityCheckService The service that handles the network requests.
  * @property networkChecker The [NetworkChecker] instance to check the network status.
+ * @property appStore The [AppStore] instance to access state and dispatch [ShoppingAction]s.
  * @property scope The [CoroutineScope] that will be used to launch coroutines.
  */
 class ReviewQualityCheckNetworkMiddleware(
     private val reviewQualityCheckService: ReviewQualityCheckService,
     private val networkChecker: NetworkChecker,
+    private val appStore: AppStore,
     private val scope: CoroutineScope,
 ) : ReviewQualityCheckMiddleware {
+
+    private val logger = Logger("ReviewQualityCheckNetworkMiddleware")
 
     override fun invoke(
         context: MiddlewareContext<ReviewQualityCheckState, ReviewQualityCheckAction>,
@@ -55,8 +63,18 @@ class ReviewQualityCheckNetworkMiddleware(
         scope.launch {
             when (action) {
                 FetchProductAnalysis, RetryProductAnalysis -> {
-                    val productReviewState = fetchAnalysis()
+                    val productPageUrl = reviewQualityCheckService.selectedTabUrl()
+                    val productAnalysis = reviewQualityCheckService.fetchProductReview()
+                    val productReviewState = productAnalysis.toProductReviewState()
                     store.updateProductReviewState(productReviewState)
+
+                    productPageUrl?.let {
+                        store.restoreAnalysingStateIfRequired(
+                            productPageUrl = productPageUrl,
+                            productReviewState = productReviewState,
+                            productAnalysis = productAnalysis,
+                        )
+                    }
                 }
 
                 ReviewQualityCheckAction.ReanalyzeProduct, ReviewQualityCheckAction.AnalyzeProduct -> {
@@ -65,6 +83,12 @@ class ReviewQualityCheckNetworkMiddleware(
                     if (reanalysis == null) {
                         store.updateProductReviewState(ProductReviewState.Error.GenericError)
                         return@launch
+                    }
+
+                    // add product to the set of products that are being analysed
+                    val productPageUrl = reviewQualityCheckService.selectedTabUrl()
+                    productPageUrl?.let {
+                        appStore.dispatch(ShoppingAction.AddToProductAnalysed(it))
                     }
 
                     val status = pollForAnalysisStatus()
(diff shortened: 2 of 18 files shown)


fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
```diff
@@ -55,8 +63,16 @@ class ReviewQualityCheckNetworkMiddleware(
         scope.launch {
             when (action) {
                 FetchProductAnalysis, RetryProductAnalysis -> {
-                    val productReviewState = fetchAnalysis()
+                    val productPageUrl = reviewQualityCheckService.selectedTabUrl()
+                    val productAnalysis = reviewQualityCheckService.fetchProductReview()
+                    val productReviewState = productAnalysis.toProductReviewState()
                     store.updateProductReviewState(productReviewState)
+
+                    store.restoreAnalysingStateIfRequired(
```
MozillaNoah, line 71: Just to make sure I'm on the same page here, we don't need an `if` or any conditional here since `reviewQualityCheckService.fetchProductReview()` is a running suspend function, and `store.updateProductReviewState(productReviewState)` won't be called until that finishes? 

So essentially, we're firing off the fetch/re-fetch, and while that's running (since it's non-blocking) we call `restoreAnalysingStateIfRequired` to put the UI back in the reanalysis state (if applicable).

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
```diff
@@ -116,4 +145,26 @@ class ReviewQualityCheckNetworkMiddleware(
...
+    ) {
+        if (productPageUrl == null) {
+            return
+        }
+
+        val productsAnalysed = appStore.state.shoppingState.productsAnalysed
+        if (productReviewState.isAnalysisPresentOrNoAnalysisPresent() &&
+            productAnalysis?.needsAnalysis == true &&
+            productsAnalysed.contains(productPageUrl)
+        ) {
```
MozillaNoah, line 162: nit: you may be able to consolidate the ifs

```suggestion
        if (productPageUrl == null) {
            return
        }

        val productsAnalysed = appStore.state.shoppingState.productsAnalysed
        if (productPageUrl != null && 
            productReviewState.isAnalysisPresentOrNoAnalysisPresent() &&
            productAnalysis?.needsAnalysis == true &&
            productsAnalysed.contains(productPageUrl)
        ) {
```
OR keep the original and `if` and now do:
```suggestion
        productPageUrl?.let {
        ...
        }
```

fenix/app/src/test/java/org/mozilla/fenix/components/appstate/ShoppingActionTest.kt
```diff
@@ -0,0 +1,72 @@
...
+import mozilla.components.support.test.ext.joinBlocking
+import org.junit.Assert.assertEquals
+import org.junit.Test
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.appstate.shopping.ShoppingState
+
+class ShoppingActionTest {
+
+    @Test
+    fun `WHEN shopping sheet is expanded THEN state should reflect that`() {
```
MozillaNoah, line 16: the test names are flipped. this one should be collapsed the next one should be expanded.

fenix/app/src/test/java/org/mozilla/fenix/components/appstate/ShoppingActionTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        val store = AppStore()
```
MozillaNoah, line 17: should the store's shopping state have its default `shoppingSheetExpanded` be true in this test?

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
```diff
@@ -363,11 +371,145 @@ class ReviewQualityCheckStoreTest {
...
+                productReviewState = ProductAnalysisTestData.analysisPresent(
+                    analysisStatus = AnalysisStatus.REANALYZING,
+                ),
+                productRecommendationsPreference = false,
+                productVendor = ProductVendor.BEST_BUY,
+            )
+
+            // Since reanalyzing is an intermediate state and the tests completes to get to the final
+            // state, this checks if the intermediate state is present in the observed state.
+            assertTrue(observedState.contains(expected))
```
MozillaNoah, line 417: alternatively, we could do these checks with a `verify` call on a mocked or spied AppStore dispatch call.

MozillaNoah (approved):
some nits and questions, but no blockers. Going to approve, but let's sync up on a short call before landing 👍

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
rahulsainani: Mocks are always slowing down the tests, here we use the real object which is favorable, followed by fakes, followed by mocks.

fenix/app/src/test/java/org/mozilla/fenix/components/appstate/ShoppingActionTest.kt
rahulsainani: Very true!

fenix/app/src/test/java/org/mozilla/fenix/components/appstate/ShoppingActionTest.kt
rahulsainani: it defaults to null but it's not good for readibility, will update to true 👍

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
rahulsainani: > Just to make sure I'm on the same page here, we don't need an if or any conditional here since reviewQualityCheckService.fetchProductReview() is a running suspend function, and store.updateProductReviewState(productReviewState) won't be called until that finishes?

Yes, suspend functions run [sequentially](https://kotlinlang.org/docs/composing-suspending-functions.html#sequential-by-default).

> So essentially, we're firing off the fetch/re-fetch, and while that's running (since it's non-blocking) we call restoreAnalysingStateIfRequired to put the UI back in the reanalysis state (if applicable).
No, when the fetch is completed and the state is updated to `ProductReviewState` from the fetch, then we restore to reanalysis state if required.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/shopping/ShoppingState.kt
```diff
@@ -0,0 +1,16 @@
...
+package org.mozilla.fenix.components.appstate.shopping
+
+/**
+ * State for shopping feature that's required to live the lifetime of a session.
+ *
+ * @property productsAnalysed Set of product ids that have been analysed.
+ * @property shoppingSheetExpanded Boolean indicating if the shopping sheet is expanded and visible.
+ */
+data class ShoppingState(
+    val productsAnalysed: Set<String> = emptySet(),
```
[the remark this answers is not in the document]
rahulsainani: That's fair, `productsInAnalysis` reads well  👍

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
```diff
@@ -23,7 +24,7 @@ interface ReviewQualityCheckService {
      *
      * @return [ProductAnalysis] if the request succeeds, null otherwise.
      */
-    suspend fun fetchProductReview(): ProductAnalysis?
+    suspend fun fetchProductReview(): GeckoProductAnalysis?
```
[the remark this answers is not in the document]
rahulsainani: We always need a `ProductAnalysis` with all fields, I think here it was a mistake to have an interface that doesn't have all the fields. One idea is also to get rid of `GeckoProductAnalysis` and only have `ProductAnalysis` object that has all the fields exposed by GV.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
```diff
@@ -116,4 +147,21 @@ class ReviewQualityCheckNetworkMiddleware(
...
+    private fun Store<ReviewQualityCheckState, ReviewQualityCheckAction>.restoreAnalysingStateIfRequired(
+        productPageUrl: String,
+        productReviewState: ProductReviewState,
+        productAnalysis: GeckoProductAnalysis?,
+    ) {
+        if (productReviewState.isAnalysisPresentOrNoAnalysisPresent() &&
+            productAnalysis?.needsAnalysis == true &&
+            appStore.state.shoppingState.productsInAnalysis.contains(productPageUrl)
+        ) {
+            logger.debug("Found product in the set: $productPageUrl")
```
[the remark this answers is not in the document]
rahulsainani: Good catch! I actually meant to remove the logs before landing (as mentioned in the PR description) and forgot to remove, we should remove this log, being removed [here](https://github.com/mozilla-mobile/firefox-android/pull/4103). 

Good to know about `String.redactPartialUri` 👍

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/shopping/ShoppingState.kt
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * State for shopping feature that's required to live the lifetime of a session.
+ *
+ * @property productsInAnalysis Set of product ids that are currently being analysed or were being
+ * analysed when the sheet was closed.
+ * @property shoppingSheetExpanded Boolean indicating if the shopping sheet is expanded and visible.
+ */
+data class ShoppingState(
+    val productsInAnalysis: Set<String> = emptySet(),
+    val shoppingSheetExpanded: Boolean? = null,
```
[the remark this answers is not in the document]
rahulsainani: Good question, I think we can set false as default, let me double check and create a bug if necessary to update state or update doc and tests.

PR #4063, opened 2023-10-12
Bug 1807545 - Add new "Open in regular tab" feature.

mhammond:
This feature allows a private tab to be reopened as a regular tab. The feature is only enabled for users with the "Open links in a private tab" option enabled - such users who follow links from external applications will always have these tabs opened as private, so may wish to "opt-out" of individual tabs being private.
Users without this option set will not have private tabs opened by default, so have already opted-in to each private tab - thus, those users are unlikely to benefit from this feature.

The feature is added to the main menu - however, as above, it will only be visible when a private tab is open *and* "Open links in a private tab" is enabled, so will only be seen by a minority of users.

The feature is also enabled only for Nightly while it is polished, but the intention is that it eventually ride to release.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1807545

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
@@ -94,6 +94,7 @@ class DefaultBrowserToolbarMenuController(
     override fun handleToolbarItemInteraction(item: ToolbarMenu.Item) {
         val sessionUseCases = activity.components.useCases.sessionUseCases
         val customTabUseCases = activity.components.useCases.customTabsUseCases
+        val tabsUseCases = activity.components.useCases.tabsUseCases
         trackToolbarItemInteraction(item)
 
         when (item) {
@@ -252,6 +253,13 @@ class DefaultBrowserToolbarMenuController(
                         item.isChecked,
                         it.id,
                     )
+                }
+            }
+            is ToolbarMenu.Item.OpenInRegularTab -> {
+                currentSession?.let { session ->
+                    getProperUrl(session)?.let { url ->
+                        tabsUseCases.migratePrivateTabUseCase.invoke(session.id, url)
+                    }
                 }
             }
             is ToolbarMenu.Item.AddToTopSites -> {
@@ -443,6 +451,8 @@ class DefaultBrowserToolbarMenuController(
                 } else {
                     Events.browserMenuAction.record(Events.BrowserMenuActionExtra("desktop_view_off"))
                 }
+            is ToolbarMenu.Item.OpenInRegularTab ->
+                Events.browserMenuAction.record(Events.BrowserMenuActionExtra("open_in_regular_tab"))
             is ToolbarMenu.Item.FindInPage ->
                 Events.browserMenuAction.record(Events.BrowserMenuActionExtra("find_in_page"))
             is ToolbarMenu.Item.SaveToCollection ->
--- a/fenix/app/src/main/res/drawable/ic_open_in_regular_tab.xml
+++ b/fenix/app/src/main/res/drawable/ic_open_in_regular_tab.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+
+<!--
+
+NOTE: This is a copy of ic_open_in_new.xml with the following changes:
+* All widths and heights on the <vector> element are 24 instead of 18.
+* A containing group with "translate" values so the image is centered vertically
+  and horizontally.
+
+-->
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:width="24dp"
+    android:height="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <group
+        android:translateX="3"
+        android:translateY="3"
+    >
+        <path
+            android:fillColor="?attr/textPrimary"
+            android:fillType="evenOdd"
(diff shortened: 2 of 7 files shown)


fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
```diff
@@ -254,6 +256,18 @@ class DefaultBrowserToolbarMenuController(
...
+                // Even if we could "move" the tab (ie, allow it to keep the same tab ID and some
+                // session data) we'd choose to not do that - the possibility that other existing
+                // private tabs have this tab ID as the parent is one example of some of the risks
+                // we'd need to mitigate. So instead we dispatch "remove" and "add" actions.
+                currentSession?.let { session ->
+                    getProperUrl(session)?.let { url ->
+                        store.dispatch(TabListAction.RemoveTabAction(session.id, false))
+                        store.dispatch(TabListAction.AddTabAction(createTab(url), true))
+                    }
+                }
```
jonalmeida, line 269: Nice work on figuring out a solution for those edge cases! Can we move this to the `TabsUseCases` so that it can be reusable code if we need to move this around or use it in multiple places?

fenix/app/src/main/res/drawable/ic_open_in_regular_tab.xml
jonalmeida, line 1: This icon looks misaligned, but I'm not sure why. Maybe @gabrielluong as our icon connoisseur would know? Should this be in our `ui-icons` as well?

You can see with the Android dev tools option "Show Layound bounds" that the icon is a bit smaller and moved to the top-right corner of it's designated area.

Screenshot 2023-10-16 at 9 31 11 AM

jonalmeida (changes requested):
This looks good and with once the two comments below are addressed, we should land it!

It's behind a feature flag for Nightly, as well as only being visible with the "Open links in a private tab", and seems safe to land and get UX feedback on it.

I mentioned this offline already that this menu may not be the best location for this feature, however we both agreed that this is a good first iteration to get the ball rolling.

Excellent work!

fenix/app/src/main/res/drawable/ic_open_in_regular_tab.xml
gabrielluong: I bet we already have an icon for this. Looking.

mhammond: > So, we have this https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/drawable/ic_open_in_new.xml,

It is that ic_open_in_new.xml with the dimensions changed from 18x18 to 24x24 (which is why it is offset - every other 24x24 icon centres the image whereas this is at the top-left.) In the meantime I'm trying to abuse `translateX/Y` but welcome any advice on that too!

gabrielluong: Okay, I will follow up with @cwzilla so that we can get a proper 24x24 icon. In the meantime, we can go with a temporary solution here if we need to land this before the cut. Let's file a follow up issue about fixing this tmp solution with a proper icon from the design system.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt
jonalmeida: A UseCase would probably be valid here for a few reasons, I think: we have UseCases that only one line of code to make sharing of repeatable code easier amongst components, we can separate the testable parts of the code (these two actions) from the UI, and this is probably not the final destination of this change.

> It looks like the use-case would require as params the store, the tab ID and the tab URL (the latter because getProperUrl() is private) - while that would work, it seems like an odd signature (you'd expect the ID would be enough to give you the URL, but as above, getProperSession() can't be called by the use-case).

Yeah, that's fair! The `getProperSession` is an artifact from our reader view web extension integration so making that a nullable parameter would work for today and then when the new ReaderView GV implementation lands, it will take out that weird method and this param (we can file a bug with a dependant on BZ-1770985).

I was curious and took an opportunity to write out what this use case would look like with a few extra checks. You don't have to use this, but I think it was easier to write it out as code instead of words. 😄 

```kotlin
    class MigratePrivateTabUseCase(
        private val store: BrowserStore,
    ) {
        /**
         * @param alternativeUrl url to load if one from the tab does not exist.
         * @return the ID of the tab that was re-created as part of the move.
         */
        operator fun invoke(
            tabId: String,
            alternativeUrl: String? = null
        ): String {
            val tab = store.state.findTab(tabId) ?: throw IllegalStateException("Tab does not exist.")

            if (!tab.content.private) {
                throw IllegalArgumentException("The tab we are trying to move is not private.")
            }

            val url = alternativeUrl ?: tab.content.url
            val newTab = createTab(url)

            store.dispatch(TabListAction.RemoveTabAction(tabId, false))
            store.dispatch(TabListAction.AddTabAction(newTab, true))

            return newTab.id
        }
    }
```

fenix/detekt-baseline.xml
```diff
@@ -568,6 +569,7 @@
     <ID>UndocumentedPublicFunction:DefaultSyncInteractor.kt$SyncInteractor$fun onCameraPermissionsNeeded()</ID>
     <ID>UndocumentedPublicFunction:DefaultToolbarMenu.kt$DefaultToolbarMenu$@VisibleForTesting(otherwise = PRIVATE) fun canAddToHomescreen(): Boolean</ID>
     <ID>UndocumentedPublicFunction:DefaultToolbarMenu.kt$DefaultToolbarMenu$@VisibleForTesting(otherwise = PRIVATE) fun canInstall(): Boolean</ID>
+    <ID>UndocumentedPublicFunction:DefaultToolbarMenu.kt$DefaultToolbarMenu$@VisibleForTesting(otherwise = PRIVATE) fun shouldShowOpenInRegularTab(): Boolean</ID>
```
gabrielluong, line 572: Let's just add a kDoc for `shouldShowOpenInRegularTab` to avoid this

mhammond: I thought it would be quite odd for that being the only function in that file with a kDoc, but OK!

mhammond: done!

jonalmeida (approved):
Great! This looks good to me with some minor nits.

I'll start a conversation with UX to review this once we land it.

fenix/app/src/test/java/org/mozilla/fenix/toolbar/DefaultToolbarMenuTest.kt
```diff
@@ -57,6 +59,7 @@ class DefaultToolbarMenuTest {
                 tabs = listOf(
                     createTab(url = "https://firefox.com", id = "1"),
                     createTab(url = "https://getpocket.com", id = "2"),
+                    createTab(url = "https://private-browsing.com", id = "3", private = true),
```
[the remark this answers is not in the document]
mhammond: Done!

PR #4077, opened 2023-10-13
Bug 1858789 - vastly more efficient counting of bookmarks for the hom…

mhammond:
…e activity.

Relies on https://github.com/mozilla/application-services/pull/5853 which was very recently merged - hopefully the Nightlies job will see it 😅 .

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1858789


fenix/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkItemMenu.kt
```diff
@@ -36,7 +36,7 @@ class BookmarkItemMenu(
     @VisibleForTesting
     @SuppressWarnings("LongMethod")
     internal suspend fun menuItems(itemType: BookmarkNodeType, itemId: String): List<TextMenuCandidate> {
-        val hasAtLeastOneChild = !context.bookmarkStorage.getTree(itemId)?.children.isNullOrEmpty()
+        val hasAtLeastOneChild = !context.bookmarkStorage.getTree(itemId, false)?.children.isNullOrEmpty()
```
mhammond, line 39: This isn't technically part of the home activity changes, but while looking for other suspect uses of `getTree` I discovered this, which doesn't actually need the full content of each child but will work just fine with just the IDs of the children.

fenix/app/src/main/java/org/mozilla/fenix/library/bookmarks/DesktopFolders.kt
```diff
@@ -49,6 +49,10 @@ class DesktopFolders(
         }
     }
 
+    suspend fun count(): Int {
```
mhammond, line 52: This doesn't seem like it's *quite* the correct abstraction, but I can't see a better option and it seems close enough :)

jonalmeida:
@rahulsainani would you be interested in reviewing this patch? I see that you had written an extension function in this that does some clean-up so figured you might be curious to try this out.

As a further follow-up to BZ-1836447, we can reduce the `HomeActivity` by moving this out entirely and doing it some kind of telemetry observer that listens to lifecycle events. I'm not sure _why_ we are doing this in `onPause` to begin with, but let's not change that part for now.

rahulsainani:
Apologies, i haven't been able to look at this till now. Just got pinged by Jeff so I've added him as reviewer.

mhammond:
Thanks! Added a changelog entry and added a "needs landing" label, so I hope the rest takes care of itself :)

jonalmeida:
> Thanks! Added a changelog entry and added a "needs landing" label, so I hope the rest takes care of itself :)

Removing the label, sorry! I see some legitimate failures in the linter task (ignore the intermittent failures from the other tasks).

jonalmeida:
Spoke to mark offline and recommended two changes:
* Let's remove the previous `count()` method so that we don't keep using it.
* There is a failure in the unit test [`countBookmarks`](https://github.com/mozilla-mobile/firefox-android/pull/4077/files#diff-0a5316ef159859ea2dc35b7afb263c2b0a138566fbc43e080cd71a77cf011fd9R76-R81) that is failing because of a [long standing bug in mockk](https://github.com/mockk/mockk/issues/544). Let's remove this test in that case, because it doesn't add enough value compared to trying to land it with a workaround for this test framework bug.
```
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class kotlin.UInt (java.lang.Integer is in module java.base of loader 'bootstrap'; kotlin.UInt is in unnamed module of loader 'app')
```

cc: @boek

mhammond:
Best I can tell the only failures here are the ones main has.

fenix/app/src/test/java/org/mozilla/fenix/library/bookmarks/DesktopFoldersTest.kt
```diff
@@ -42,6 +43,7 @@ class DesktopFoldersTest {
     fun setup() {
         context = spyk(testContext)
         every { context.components.core.bookmarksStorage } returns mockk()
+        coEvery { context.components.core.bookmarksStorage.countBookmarksInTrees(any()) } returns 0U
```
jonalmeida, line 46: This line will throw errors if someone tries to make an assertion that executes this line in a future test. It might be best to remove it to avoid a footgun in the future.

jonalmeida (approved):
Cheers!

I left one comment below which I'll quickly fix myself so that we can land this now.

PR #4081, opened 2023-10-13
Bug 1855988 - Translation Options UI Dialog Screen

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1855988


fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,220 @@
...
+                        .clearAndSetSemantics {
+                            contentDescription = item.contentDescription
+                        },
+                )
+                if (item.hasDivider) {
+                    Divider(Modifier.padding(top = 4.dp, bottom = 4.dp))
+                }
+            }
+        }
+        Text(
```
gabrielluong, line 71: As a rule of thumb, could you add a new line in between each of your composable as we have done in other composables? This will help make things easier to read.
```suggestion
        TranslationOptionsDialogHeader(onBackClicked)

        LazyColumn {
            items(translationOptionsList) { item: TranslationSwitchItem ->
                SwitchWithLabel(
                    checked = item.isEnabled,
                    onCheckedChange = item.stateChange,
                    label = item.textLabel,
                    modifier = Modifier
                        .padding(start = 72.dp, end = 16.dp)
                        .clearAndSetSemantics {
                            contentDescription = item.contentDescription
                        },
                )

                if (item.hasDivider) {
                    Divider(Modifier.padding(top = 4.dp, bottom = 4.dp))
                }
            }
        }

        Text(
```
Snippet of some of the suggested changes, but please self-review your other changes and help improve readability of your code.

fenix/app/src/main/res/values/static_strings.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8" standalone="yes"?>
-<!-- This Source Code Form is subject to the terms of the Mozilla Public
+<?xml version="1.0" encoding="utf-8" standalone="yes"?><!-- This Source Code Form is subject to the terms of the Mozilla Public
```
gabrielluong, line 1: nit undo this change

fenix/app/src/main/res/values/static_strings.xml
```diff
@@ -124,4 +123,24 @@
...
+    <string name="translation_option_bottom_sheet_title">Translation Options</string>
+    <string name="translation_option_bottom_sheet_always_translate">Always offer to translate</string>
+    <string name="translation_option_bottom_sheet_always_translate_in_language">Always translate %1$s</string>
+    <string name="translation_option_bottom_sheet_spanish_language">Spanish</string>
+    <string name="translation_option_bottom_sheet_never_translate_in_language">Never translate %1$s</string>
+    <string name="translation_option_bottom_sheet_never_translate_site">Never translate this site</string>
+    <string name="translation_option_bottom_sheet_translation_settings">Translation settings</string>
+    <string name="translation_option_bottom_sheet_about_translations">About translations in %1$s</string>
+    <string name="accessibility_off_switch">Switch OFF</string>
+    <string name="accessibility_on_switch">Switch ON</string>
```
gabrielluong, line 136: Let's move this out of static strings since content descriptions aren't blocked on translation.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,220 @@
...
+    }
+}
+
+/**
+ * Returns a content description for a switch list item based on the enabled value.
+ */
+@Composable
+fun getSwitchContentDescription(textSwitch: String, isEnabled: Boolean) =
+    if (isEnabled) {
+        "$textSwitch " + stringResource(
```
gabrielluong, line 200: We shouldn't be appending strings for localization. Please consider using parameterize strings.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettingsFragmentView.kt
```diff
@@ -0,0 +1,148 @@
...
+import org.mozilla.fenix.compose.Divider
+import org.mozilla.fenix.compose.SwitchWithLabel
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Firefox Translation settings fragment compose view.
+ */
+@Composable
+fun TranslationSettingsFragmentView(
```
gabrielluong, line 31:
```suggestion
fun TranslationSettings(
```
Let's move away with the fragment style naming. Compose doesn't necessary need to be render in a fragment.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
```diff
@@ -0,0 +1,16 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.translations
+
+/**
+ * Translation switch list item that will appear on Translation screens.
+ */
+data class TranslationSwitchItem(
+    val isEnabled: Boolean,
+    val stateChange: (Boolean) -> Unit,
```
gabrielluong, line 12:
```suggestion
    val onStateChange: (Boolean) -> Unit,
```
I assume this param is a callback.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettingsFragmentView.kt
```diff
@@ -0,0 +1,148 @@
...
+        )
+        Text(
+            text = stringResource(id = R.string.translation_settings_fragment_automatic_translation),
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(start = 72.dp, end = 16.dp, bottom = 16.dp, top = 8.dp)
+                .clickable { onAutomaticTranslationClicked() },
+            color = FirefoxTheme.colors.textPrimary,
+            style = FirefoxTheme.typography.subtitle1,
+        )
```
gabrielluong, line 78: This should be using our `TextListItem` component.

gabrielluong: Don't worry about using `TextListItem` for the `Text` above with `headline8`. I think we should followup after seeing talking with UX to incorporate that use case into our design components.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.translations
+
+/**
+ * Translation switch list item that will appear on Translation screens.
```
gabrielluong, line 8: Let's also add the @param here.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
iorgamgabriel: Done

fenix/app/src/main/res/values/static_strings.xml
iorgamgabriel: Done

fenix/app/src/main/res/values/static_strings.xml
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettingsFragmentView.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettingsFragmentView.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
```diff
@@ -0,0 +1,148 @@
...
+            color = FirefoxTheme.colors.textAccent,
+            style = FirefoxTheme.typography.headline8,
+        )
+
+        TextListItem(
+            label = stringResource(id = R.string.translation_settings_fragment_automatic_translation),
+            modifier = Modifier
+                .fillMaxWidth()
+                .padding(start = 56.dp)
+                .clickable { onAutomaticTranslationClicked() },
```
gabrielluong, line 79: For the `clickable`, could we move this into using the API's `onClick` https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt#55

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
```diff
@@ -0,0 +1,148 @@
...
+        add(
+            TranslationSwitchItem(
+                isEnabled = true,
+                onStateChange = {},
+                hasDivider = false,
+                contentDescription = getSwitchContentDescription(
+                    textSwitchId = R.string.translation_settings_fragment_offer_to_translate_content_description,
+                    isEnabled = true,
+                ),
+                textLabel = stringResource(R.string.translation_settings_fragment_offer_to_translate),
```
gabrielluong, line 119: Could we reorder the params here to align with how we have defined the param order in `TranslationSwitchItem `

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * TranslationSwitchItem that will appear on Translation screens.
+ *
+ * @property textLabel the text that will appear on the switch.
+ * @property contentDescription for the switch item.
+ * It will be a single contentDescription for the switch and for the text next to the switch.
+ * @property isEnabled if the switch is checked or not.
+ * @property hasDivider if a divider should appear under the switch.
+ * @property onStateChange is Invoked when Switch is being clicked,
+ * therefore the change of checked state is requested.
```
gabrielluong, line 16:
```suggestion
 * @property textLabel The text that will appear on the switch item.
 * @property contentDescription The content description for the switch item.
 * It will be a single content description for the switch and for the text next to the switch.
 * @property isChecked Whether the switch is checked or not.
 * @property hasDivider Whether a divider should appear under the switch item.
 * @property onStateChange Invoked when the switch item is clicked,
 * the new checked state is passed into the callback.
```
We should avoid using the param name as the start of the sentence and just write complete sentences. The rendered view of these kDocs as is would look like: `onStateChange - is Invoked ...` when we want to `onStateChange - Invoked`.

Example rendered view:
Screen Shot 2023-10-17 at 11 31 39 AM

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2306,4 +2306,10 @@
     <string name="a11y_action_label_read_article">read the article</string>
     <!-- Action label for links to the Firefox Pocket website. Talkback will append this to say "Double tap to open link to learn more". -->
     <string name="a11y_action_label_pocket_learn_more">open link to learn more</string>
+    <!-- Action label for the switch element when it is off. Talkback will say, for example "Automatic translation switch off".-->
+    <!-- "Automatic translation" will be the text from the switch element. -->
+    <string name="accessibility_off_switch">Switch OFF</string>
```
gabrielluong, line 2311:
```suggestion
    <string name="accessibility_switch_off">Switch OFF</string>
```

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2306,4 +2306,10 @@
...
+    <!-- Action label for the switch element when it is on. Talkback will say, for example "Automatic translation switch on".-->
+    <!-- "Automatic translation" will be the text from the switch element. -->
+    <string name="accessibility_on_switch">Switch ON</string>
```
gabrielluong, line 2314:
```suggestion
    <string name="accessibility_switch_on">Switch ON</string>
```

ohall-m (commented):
I only managed to get through a partial review focusing on the accessibility of the component - I'll take a look more in depth later too.

fenix/app/src/main/res/values/strings.xml
gabrielluong, line 2314: I was expecting the strings to be `%1$s Switch ON`, but since we added the parameter to the content description strings instead which are in static_string. Let's move these back into static_strings for the time being. What worries me here is that we are now exposing these strings to translations, but they don't see the other side of the strings which are currently in static_strings. I know we included "Automatic translation" here, but without a localization note about what and how "Automatic translation" is used, I don't think there is enough context here for our translators at this moment.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
```diff
@@ -0,0 +1,148 @@
...
+                .clickable { onDownloadLanguageClicked() },
+        )
+    }
+}
+
+/**
+ * Return a list of Translation option switch list item.
+ */
+@Composable
+fun getTranslationSettingsSwitchList(): List<TranslationSwitchItem> {
```
gabrielluong, line 108: Maybe this should be `internal fun` since we only want this available in the package?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,234 @@
...
+import org.mozilla.fenix.compose.SwitchWithLabel
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.compose.list.TextListItem
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Firefox Translation options bottom sheet dialog.
+ */
+@Composable
+fun TranslationOptionsDialogBottomSheet(
```
gabrielluong, line 42: Please add @param for all our public composable functions.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,234 @@
...
+
+        LazyColumn {
+            items(translationOptionsList) { item: TranslationSwitchItem ->
+                SwitchWithLabel(
+                    checked = item.isEnabled,
+                    onCheckedChange = item.onStateChange,
+                    label = item.textLabel,
+                    modifier = Modifier
+                        .padding(start = 72.dp, end = 16.dp)
+                        .clearAndSetSemantics {
```
ohall-m, line 65: The labels for the toggles aren't associated with TalkBack - I think this might be causing it, not 100% sure and didn't look in-depth. 

When testing, and you go to the toggle using TalkBack, it should:
* Have a label that is programatically connected to it - `name`
* Know the state of the toggle - `value` - inherited from `Switch` (please don't override the content description)
* Know it is a toggle - `role` - inherited from `Switch`

For an example, check out the `Menu` -> `Accessibility` -> and try the toggle for `Automatic font sizing`.

Notice, you can also double tap the label and it will toggle the switch. It shows they were programmatically associated.

When testing this component, it should behave similarly.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,234 @@
...
+ * Returns a content description for a switch list item based on the enabled value.
+ */
+@Composable
+fun getSwitchContentDescription(
+    @StringRes textSwitchId: Int,
+    additionalTextSwitchArgs: MutableList<String> = mutableListOf(),
+    isEnabled: Boolean,
+): String {
+    if (isEnabled) {
+        additionalTextSwitchArgs.add(stringResource(R.string.accessibility_on_switch))
```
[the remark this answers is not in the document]
iorgamgabriel: https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7221:187721&mode=design#572340901
The designer said "These are a bit tricky: the focus could look like it is, with both the label and the control focused together but announced as one element" . This is the reason that I put contentDescription. I can leave the default implementation it is way more easy for me , but design must know it .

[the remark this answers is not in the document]
iorgamgabriel: https://github.com/mozilla-mobile/firefox-android/assets/8118371/96449d5d-f259-4328-9771-e175002a8670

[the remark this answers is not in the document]
iorgamgabriel: This is how the switch is read with the default implementation without contentDescription . Two elements .

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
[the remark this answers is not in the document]
ohall-m: Thanks for the video and the comment thread! In the video, the toggle element is missing a `name` or label (in the `name`, `role`, `value` [trifecta](https://www.afb.org/blog/entry/name-role-value)).

This can happen when the label and toggle are not programmatically linked or something is overriding the `name`. What is happening now is what happened to me with my earlier edited patch [here](https://github.com/mozilla-mobile/firefox-android/pull/4081#discussion_r1362376220), that advice now applies (but maybe not the reasoning).

I think that Figma thread is referring to the fact that they are okay if it is one focus-able element labeled (e.g., Focus on **row**: "Never translate, toggle on" or hiding the label from the focus order (e.g., Focus on **toggle**: "Never translate, toggle on"). We can figure out the grouping and focusable elements pretty quickly using the default implementation, lets get the base component and `name` situated first.

The inherited implementation is preferred because it will always use the TalkBack standard names and terminology for controls. (e.g., As a user, I have a visual expectation of how a toggle will present itself. As a user, I have an aural expectation of how a toggle will present itself. If we use non-standard content descriptions on standard components, we can break user expectations. Also worth mentioning, some screen reader users have the speech rate settings at 3x or much more the default speed, so a non-standard announcement sticks out even further.) 

When the label and element are programmatically linked, I expect they will show up as one element. (You can test this without TalkBack. If you tap a label and the component toggles/clicks, then the label and the component are programmatically tied. If they are programmatically tied and still have the wrong TalkBack announcement, then something is overriding the announcement. (e.g., something like `clearAndSetSemantics`, grouping incorrectly, or inheriting something strange.) If they are programmatically tied and showing as two elements, we can group the row as the focus-able element to ensure one element and will inherit the toggle semantic behavior.

Quick summary:
I suspect: 
1. `clearAndSetSemantics` is interfering with the `name` of the toggle 
2. OR else the label and toggle aren't connected programmatically
3. OR else something is overriding the label

We can solve focus grouping as a secondary problem, if fixing the `name` or label doesn't automatically connect the elements.

I'd suggest checking out how `Menu` -> `Accessibility` -> toggle for `Automatic font sizing` is setup too.

fenix/app/src/main/res/navigation/nav_graph.xml
```diff
@@ -426,6 +431,14 @@
         android:id="@+id/translationsDialogFragment"
         android:name="org.mozilla.fenix.translations.TranslationsDialogFragment"/>
 
+    <dialog
+        android:id="@+id/translationOptionsDialogFragment"
+        android:name="org.mozilla.fenix.translations.TranslationOptionsDialogFragment" />
+
+    <fragment
```
jonalmeida, line 438: Drive-by 🚗 : Why are is fragment, and two dialogs (the one in this change and the one above) global in our navigation graph? Shouldn't they only be accessible from their current entry points (browser and settings)?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,237 @@
...
+import org.mozilla.fenix.compose.list.TextListItem
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Firefox Translation options bottom sheet dialog.
+ *
+ * @param translationOptionsList A list with TranslationSwitchItems that will appear on the fragment.
+ * @param onBackClicked Callback when the user presses the back button.
+ * @param onTranslationSettingsClicked Callback when the user presses Translation Settings button.
+ * @param aboutTranslationClicked Callback when the user presses About Translation button.
```
gabrielluong, line 43:
```suggestion
 * @param translationOptionsList A list of [TranslationSwitchItem]s to display.
 * @param onBackClicked Invoked when the user clicks on the back button.
 * @param onTranslationSettingsClicked Invoked when the user clicks on the "Translation Settings" button.
 * @param aboutTranslationClicked Invoked when the user clicks on the "About Translation" button.
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -39,7 +40,14 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
     ): View = ComposeView(requireContext()).apply {
         setContent {
             FirefoxTheme {
-                TranslationsDialogBottomSheet()
+                TranslationsDialogBottomSheet(
+                    onSettingClicked = {
+                        findNavController().popBackStack()
+                        findNavController().navigate(
+                            TranslationsDialogFragmentDirections.actionGlobalToTranslationOptionsDialogFragment(),
```
gabrielluong, line 47: I am kinda wondering if we should be navigating from bottom sheet to another bottom sheet. Visually, this will close the first bottom sheet, and then opening the next bottom sheet. I know we do this elsewhere, but there's also a flow when we are viewing the site details where we do not have this opening and closing of the bottom sheet for navigation.

Screen Shot 2023-10-19 at 2 13 45 AM

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
```diff
@@ -0,0 +1,152 @@
...
+) {
+    Column(
+        modifier = Modifier
+            .background(
+                color = FirefoxTheme.colors.layer1,
+            ),
+    ) {
+        LazyColumn {
+            items(translationSwitchList) { item: TranslationSwitchItem ->
+                SwitchWithLabel(
```
gabrielluong, line 50: I think we need a follow up bug to change `SwitchWithLabel`. Right now, if we go to our Settings and look for an existing setting with a switch, clicking on the row will toggle the switch. I think the current implementation behaves differently from our actual desired use case.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
gabrielluong: We can address any changes in a follow up, but we should start asking for some UX feedback.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
iorgamgabriel: I think I fixed it . There were some bugs in the SwitchWithLabel component.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
iorgamgabriel: I fixed it . It works like in Settings now.

ohall-m: Please mention the global change to QA, just so they can be aware for testing other toggles. 

I do sense the `SwitchWithLabel` portion maybe should be a separate patch for this reason too. (Either works for me, but I think we should be especially in case we break other toggles.)

gabrielluong: Let's actually move the changes with `SwitchWithLabel` into a separate PR. I would be more comfortable pinging others about how this switch should work - whether or not we should even expose the ability to move the toggle or if it  should just be a visual toggle more than anything.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
ohall-m: (Please see other conversation about moving `SwitchWithLabel` to a different PR for more feedback. This implementation works for our use-case, but we should be careful about global changes.)

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
```diff
@@ -65,6 +75,7 @@ fun SwitchWithLabel(
         Switch(
             checked = checked,
             onCheckedChange = onCheckedChange,
+            modifier = Modifier.clearAndSetSemantics {},
```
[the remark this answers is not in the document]
ohall-m: We should also confirm that we want the row to be the toggable element. 

I recommending testing some with more with `mergeDescendants = false` to see if we can find a local solution.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: +1 on asking UX feedback early and often

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -82,7 +82,7 @@ private fun TranslationsDialogHeader() {
         Spacer(modifier = Modifier.width(4.dp))
 
         IconButton(
-            onClick = {},
+            onClick = { onSettingClicked() },
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,185 @@
...
+    }
+}
+
+@Composable
+private fun TranslationOptionsDialogHeader(
+    onBackClicked: () -> Unit,
+) {
+    Row(
+        verticalAlignment = Alignment.CenterVertically,
+        modifier = Modifier.padding(top = 16.dp, end = 16.dp, start = 16.dp, bottom = 0.dp),
```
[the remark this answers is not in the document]
iorgamgabriel: I added .defaultMinSize(minHeight = 56.dp) like TextListItem

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,185 @@
...
+                isChecked = false,
+                hasDivider = true,
+                onStateChange = {},
+            ),
+        )
+        add(
+            TranslationSwitchItem(
+                textLabel = stringResource(
+                    id = R.string.translation_option_bottom_sheet_always_translate_in_language,
+                    formatArgs = arrayOf(stringResource(R.string.translation_option_bottom_sheet_spanish_language)),
```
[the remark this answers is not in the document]
iorgamgabriel: done

fenix/app/src/main/res/values/static_strings.xml
```diff
@@ -124,4 +124,22 @@
     <string name="translations_bottom_sheet_translate_to">Translate to</string>
     <string name="translations_bottom_sheet_negative_button">Not now</string>
     <string name="translations_bottom_sheet_positive_button">Translate</string>
+
+    <string name="translation_option_bottom_sheet_title">Translation Options</string>
+    <string name="translation_option_bottom_sheet_always_translate">Always offer to translate</string>
+    <string name="translation_option_bottom_sheet_always_translate_in_language">Always translate %1$s</string>
+    <string name="translation_option_bottom_sheet_spanish_language">Spanish</string>
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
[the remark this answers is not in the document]
iorgamgabriel: It doesn't work with the mergeDescendants = false/true; the switch is read separated, I don't want that , I want to be read like a single component.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
iorgamgabriel: https://github.com/mozilla-mobile/firefox-android/pull/4181

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
iorgamgabriel: I asked on Slack .

fenix/app/src/main/res/values/static_strings.xml
```diff
@@ -124,4 +124,21 @@
...
+    <string name="translation_option_bottom_sheet_always_translate_in_language">Always translate %1$s</string>
+    <string name="translation_option_bottom_sheet_never_translate_in_language">Never translate %1$s</string>
+    <string name="translation_option_bottom_sheet_never_translate_site">Never translate this site</string>
+    <string name="translation_option_bottom_sheet_translation_settings">Translation settings</string>
+    <string name="translation_option_bottom_sheet_about_translations">About translations in %1$s</string>
+
+    <string name="translation_settings_fragment_toolbar_title">Translations</string>
+    <string name="translation_settings_fragment_translation_preference">Translation preferences</string>
+    <string name="translation_settings_fragment_offer_to_translate">Offer to translate when possible</string>
+    <string name="translation_settings_fragment_always_download">Always download languages  in data saving mode</string>
```
ohall-m, line 139: Could you check that there aren't two spaces here: ` languages  in`?

fenix/app/src/main/res/values/static_strings.xml
```diff
@@ -124,4 +124,21 @@
...
+    <string name="translation_settings_fragment_automatic_translation">Automatic translation</string>
+    <string name="translation_settings_fragment_automatic_never_translate_site">Never translate this site</string>
```
ohall-m, line 141: I think we need separate strings for the "Translations" page linked as "Translation settings". 

In the second page, it should be the plural of "Never translate these sites" via [this](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7202%3A113050&mode=dev). (This copy is [correct](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7202%3A99806&mode=dev) for the main "Translation Options" toggle.)

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogFragment.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.google.android.material.bottomsheet.BottomSheetBehavior
+import com.google.android.material.bottomsheet.BottomSheetDialogFragment
+import org.mozilla.fenix.R
+import org.mozilla.fenix.browser.BrowserFragmentDirections
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * A bottom sheet fragment displaying the Firefox Translation Options dialog.
+ */
+class TranslationOptionsDialogFragment : BottomSheetDialogFragment() {
```
ohall-m, line 23: One more question on layout, so in the Figma, in landscape, the panel is planned to span the full [width of the page](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7212%3A177225&mode=dev). However, this [design may change](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7335%3A209194&mode=dev).

I think if it is an easy fix, could you go ahead and make the panel full width? (If it isn't an easy fix, let's wait until we have the final design to make the change or follow-up on it.) What do you think?

fenix/app/src/main/res/values/static_strings.xml
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogFragment.kt
iorgamgabriel: Done

fenix/app/src/main/res/values/static_strings.xml
iorgamgabriel: Done

fenix/app/src/main/res/values/static_strings.xml
```diff
@@ -127,4 +127,21 @@
...
+    <string name="translation_option_bottom_sheet_never_translate_site">Never translate this site</string>
+    <string name="translation_option_bottom_sheet_translation_settings">Translation settings</string>
+    <string name="translation_option_bottom_sheet_about_translations">About translations in %1$s</string>
+
+    <string name="translation_settings_fragment_toolbar_title">Translations</string>
+    <string name="translation_settings_fragment_translation_preference">Translation preferences</string>
+    <string name="translation_settings_fragment_offer_to_translate">Offer to translate when possible</string>
+    <string name="translation_settings_fragment_always_download">Always download languages in data saving mode</string>
+    <string name="translation_settings_fragment_automatic_translation">Automatic translation</string>
+    <string name="translation_settings_fragment_automatic_never_translate_site">Never translate this sites</string>
```
ohall-m, line 144: Thanks for the change, very slight adjustment:
```suggestion
    <string name="translation_settings_fragment_automatic_never_translate_sites">Never translate these sites</string>
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: Thanks for getting the feedback and all the work on the patch! 

I think the patch will be good to go after this is resolved or, if it looks like a bigger change, a bug is filed to smooth the transition between sheets. 🐞

gabrielluong: Where is the Slack conversation happening? Could you ask over in #firefox-translations?

fenix/app/src/main/res/values/static_strings.xml
```diff
@@ -127,4 +127,21 @@
...
+
+    <string name="translation_option_bottom_sheet_title">Translation Options</string>
+    <string name="translation_option_bottom_sheet_always_translate">Always offer to translate</string>
+    <string name="translation_option_bottom_sheet_always_translate_in_language">Always translate %1$s</string>
+    <string name="translation_option_bottom_sheet_never_translate_in_language">Never translate %1$s</string>
+    <string name="translation_option_bottom_sheet_never_translate_site">Never translate this site</string>
+    <string name="translation_option_bottom_sheet_translation_settings">Translation settings</string>
+    <string name="translation_option_bottom_sheet_about_translations">About translations in %1$s</string>
+
+    <string name="translation_settings_fragment_toolbar_title">Translations</string>
```
gabrielluong, line 139:
```suggestion
    <string name="translation_settings_toolbar_title">Translations</string>
```
Let's not attach the word "fragment" into our string IDs since it's used in a Composable.

iorgamgabriel: I think I need to delete and bottom_sheet because it's the same.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
gabrielluong: @iorgamgabriel @ohall-m I think it would be okay to make the experience smooth as a followup issue since we already implemented 2 different screens here and prefer an iterative approach. I think we should try to limit only implementing a single screen per issue since the amount of review comments are also growing as we try to do more.

If you try out the site permission details for `cnn.com`, you will see for the tracking protection blocked / allow navigation is already smooth. This was implemented all in a single fragment/XML - think about it as 'pages' or views that are toggled visible manually. I think we can easily do this with Compose. However, I would be curious if we could do this with using [navigation in compose](https://developer.android.com/guide/navigation/use-graph/navigate#compose) while still be in the same bottom sheet.

fenix/app/src/main/res/values/static_strings.xml
iorgamgabriel: Done

fenix/app/src/main/res/values/static_strings.xml
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
iorgamgabriel: https://bugzilla.mozilla.org/show_bug.cgi?id=1861058

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
```diff
@@ -0,0 +1,147 @@
...
+                    Divider(Modifier.padding(top = 8.dp, bottom = 8.dp))
+                }
+            }
+
+            item {
+                Text(
+                    text = stringResource(
+                        id = R.string.translation_settings_translation_preference,
+                    ),
+                    modifier = Modifier
```
ohall-m, line 66: Could you add: 
`modifier = Modifier.semantics { heading() }`

(Apologizes for catching this late in the review, the [purple breaker text](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=8151%3A15473&mode=dev) and the [page title](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=8151%3A15378&mode=dev) should be semantic [headings](https://developer.android.com/jetpack/compose/accessibility#define-headings).)

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
```diff
@@ -0,0 +1,196 @@
...
+                painter = painterResource(id = R.drawable.mozac_ic_back_24),
+                contentDescription = stringResource(R.string.etp_back_button_content_description),
+                tint = FirefoxTheme.colors.iconPrimary,
+            )
+        }
+
+        Spacer(modifier = Modifier.width(32.dp))
+
+        Text(
+            text = stringResource(id = R.string.translation_option_bottom_sheet_title),
```
ohall-m, line 129: Could you add: 
`modifier = Modifier.semantics { heading() }`

(More info below.)

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettingsFragment.kt
```diff
@@ -0,0 +1,52 @@
...
+import org.mozilla.fenix.ext.showToolbar
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * A fragment displaying the Firefox Translation settings screen.
+ */
+class TranslationSettingsFragment : Fragment(), UserInteractionHandler {
+    override fun onResume() {
+        super.onResume()
+        showToolbar(getString(R.string.translation_settings_toolbar_title))
```
ohall-m, line 25: We may need to change this to a semantic heading. However, it seems like it might be a global issue? (Only did a spot check.)

ohall-m (commented):
Thanks, I tested the patch and it looks good! I mentioned a few places for semantic headings, otherwise LGTM. 

Could you also take a look at the `nav_graph? It sounds like we may need a change.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialogBottomSheet.kt
iorgamgabriel: Done

iorgamgabriel: Done

iorgamgabriel:
@ohall-m I think the navigation I should do it in a different pr because I should use nav compose and only one BottomSheetDialogFragment . If you are ok with this .

ohall-m (approved):
🤩 LGTM!

fenix/app/src/main/res/navigation/nav_graph.xml
jonalmeida: Ping @iorgamgabriel @ohall-m 🙂

iorgamgabriel: Hi, with this task https://bugzilla.mozilla.org/show_bug.cgi?id=1861058 I'm going to redo all the navigation.

PR #4090, opened 2023-10-14
Bug 1803465 - extend isURLLenient to match IPv6 literals

pmarks-net:
This pull request lets Firefox for Android accept IPv6 literals typed into the address bar.

For example: <a href="https://[2606:4700:4700::1111]/">https://[2606:4700:4700::1111]/</a>

The bug was originally reported as https://github.com/mozilla-mobile/fenix/issues/4343, but @YoRyan's https://github.com/mozilla-mobile/android-components/pull/5546 was rejected because it grew [a big slow regex](https://github.com/mozilla-mobile/android-components/issues/5594).

The big regex [no longer exists](https://github.com/mozilla-mobile/android-components/commit/f9282686cd8243dc3faead1cdeedbf2a73dd1215) (yay!) leaving this 1-line `isURLLenient` regex as the final roadblock.

---

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1803465
https://bugzilla.mozilla.org/show_bug.cgi?id=1807752


pmarks-net:
@csadilek and @mcomella originally wrote this regex.

pmarks-net:
I tested this using https://firefox-source-docs.mozilla.org/mobile/android/fenix.html and `./gradlew support-utils:test`.

I'm not sure if the 3 workflow failures are my fault, nor how to rerun them.

Djfe:
great, thank you for working on this 👍
Would it make sense to also introduce some assertFalse?
like for misspelled ipv6 style urls?

pmarks-net:
@Djfe this regex has no concept of a "misspelled" IP address.  It's not looking at the address itself, just very broad character classes to distinguish possible URLs from garbage.

pmarks-net:
Added more test cases.  When I type these into the UI, it either shows "Invalid Address" or a completely blank tab, depending on the mood of the downstream URL parser.

Note that both of these states are already reachable in Firefox 118:

- `http://x:x` -> "Invalid Address"
- `x://x` -> blank tab

pmarks-net:
I'm still waiting for a reviewer.  @csadilek perhaps?

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -88,6 +88,40 @@ class URLStringUtilsTest {
...
+        assertTrue(isURLLike("fe80::"))
+        assertTrue(isURLLike("x:["))
+        assertTrue(isURLLike("http://[::"))
+        assertTrue(isURLLike("http://[::/path"))
+        assertTrue(isURLLike("http://[::?query"))
+        assertTrue(isURLLike("[[http://banana]]"))
+        assertTrue(isURLLike("http://[[["))
+        assertTrue(isURLLike("[[[::"))
+        assertTrue(isURLLike("[[[::/"))
+        assertTrue(isURLLike("http://[1.2.3]"))
```
titooan, line 123: I think this block includes tests that assert that `isURLLike` works as expected (returning false for malformed IPv6 literals), as well as tests showing the limits of the current regex (returning the wrong result for wellformed and malformed IPv6 literals). 

It would be a bit clearer to separate the two types of tests in two groups described with a comment.

titooan: Also, we noticed that we could add tests with an URL with the wrong number of bytes : 
```kotlin 
        assertFalse(isURLLike("https://[1:2:3:4:5:6:7]/"))
        assertFalse(isURLLike("https://[1:2:3:4:5:6:7:8:9]/"))
```

Those test are currently failing. Would it be possible to adapt the regex so that we take this case into account? 

If not, we can put those tests in the block of tests showing the limits of the current regex.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -88,6 +88,40 @@ class URLStringUtilsTest {
...
+        assertTrue(isURLLike("[1::]"))
+        assertTrue(isURLLike("[::]/"))
+        assertTrue(isURLLike("[1:2:3:4:5:6:7:8]"))
+        assertTrue(isURLLike("https:[::]"))
+        assertTrue(isURLLike("https://[::]"))
+        assertTrue(isURLLike("https://[::1]"))
+        assertTrue(isURLLike("https://[1::]"))
+        assertTrue(isURLLike("https://[2001:db8::1.2.3.4]/"))
+        assertTrue(isURLLike("https://[1:2:3:4:5:6:7:8]/"))
+        assertTrue(isURLLike("[2001:db8::1.2.3.4]/"))
```
titooan, line 103: Here is a suggestion that I think makes it easier to test several variations of each "base IPv6 literal", including `http:`, `https:`, trailing slash... 
```suggestion
        val validBaseUrls = listOf(
            "[::]",
            "[::1]",
            "[1::]",
            "[::]/",
            "[1:2:3:4:5:6:7:8]",
            "[2001:db8::1.2.3.4]",
        )

        validBaseUrls.forEach { url ->
            assertTrue(isURLLike(url))
            assertTrue(isURLLike("$url/"))
            assertTrue(isURLLike("https://$url"))
            assertTrue(isURLLike("https://$url/"))
            assertTrue(isURLLike("http://$url"))
            assertTrue(isURLLike("http://$url/"))
        }
```

What do you think?

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/WebURLFinderTest.kt
```diff
@@ -59,6 +59,11 @@ class WebURLFinderTest {
         )
         assertEquals("http://ß.de/", find("http://ß.de/ çnn.çơḿ"))
         assertEquals("htt-p://ß.de/", find("çnn.çơḿ htt-p://ß.de/"))
+        assertEquals(
+            "http://[2001:db8::1.2.3.4]:8080/inner#anchor&arg=1",
+            find("test.com http://[2001:db8::1.2.3.4]:8080/inner#anchor&arg=1"),
+        )
+        assertEquals("http://[::]", find("test.com http://[::]"))
```
titooan, line 66: Can we also add some more similar test coverage from UrlStringUtils into WebUrlFinderTest?

titooan (changes requested):
Hi pmarks-net,
Thanks a lot for this PR and sorry for the late review. 

While we were reviewing this, we tried to think of all the cases and here are some suggestions we came up with, that you can find in this [patch file](https://github.com/mozilla-mobile/firefox-android/files/13378157/PR-4090.patch).


Please feel free to ask for clarifications is anything is not clear enough.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
pmarks-net: This regex makes no attempt to parse IP addresses, so cases like `[1:2:3:4:5:6:7]` belong in the "Invalid IPv6 literals incorrectly recognized as valid" bucket.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
pmarks-net: Done.  I removed "[::]/" because the / is tested later, added "[::1]:8080" for coverage, and added tests for "https:[::]"-style URLs.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/WebURLFinderTest.kt
pmarks-net: Added `isUrlLikeEmulated()`, copied all the test cases, and marked cases that differ with `INVERT()`.

pmarks-net (commented):
These changes were incorporated into  commit 674be27070f77e1d6e55b6bd67f0f456f994bb35

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
titooan: Good catch!

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/WebURLFinderTest.kt
```diff
@@ -108,4 +116,97 @@ class WebURLFinderTest {
...
+        assertTrue(INVERT(isURLLike("[[http://banana]]")))
+        assertTrue(INVERT(isURLLike("http://[[[")))
+        assertTrue(INVERT(isURLLike("[[[::")))
+        assertTrue(INVERT(isURLLike("[[[::/")))
+        assertTrue(INVERT(isURLLike("http://[1.2.3]")))
+        assertTrue(INVERT(isURLLike("https://[1:2:3:4:5:6:7]/")))
+        assertTrue(INVERT(isURLLike("https://[1:2:3:4:5:6:7:8:9]/")))
+
+        assertTrue(isURLLike("https://abc--cba.com/"))
+    }
```
titooan, line 211: I'm sorry, I think my comment about adding tests for WebUrlFinderTest was not clear. 

Here is what I had in mind: 
```kotlin 
    @Test
    fun testIPv6LiteralsInsideSentence() {
        val validIPv6Literals = listOf(
            "[::]",
            "[::1]",
            "[1::]",
            "[1:2:3:4:5:6:7:8]",
            "[2001:db8::1.2.3.4]",
            "[::1]:8080",
        ).flatMap { url ->
            listOf(
                "https://$url",
                "https://$url/",
                "https:$url",
                "http://$url",
                "http://$url/",
                "http:$url",
            )
        }

        validIPv6Literals.forEach { url ->
            assertEquals(url, find("random_text $url other_random_text"))
        }
    }
```

The idea would be to make sure the regex is able to find the IPv6 literals inside a sentance containing them. 
What do you think?

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -88,6 +88,54 @@ class URLStringUtilsTest {
...
+
+        validIPv6Literals.forEach { url ->
+            assertTrue(isURLLike(url))
+            assertTrue(isURLLike("$url/"))
+            assertTrue(isURLLike("https://$url"))
+            assertTrue(isURLLike("https://$url/"))
+            assertTrue(isURLLike("https:$url"))
+            assertTrue(isURLLike("http://$url"))
+            assertTrue(isURLLike("http://$url/"))
+            assertTrue(isURLLike("http:$url"))
```
titooan, line 109: I realize we could also add: 
```kotlin
            assertTrue(isURLLike("http:$url/"))
            assertTrue(isURLLike("https:$url/"))
```

titooan (approved):

titooan (changes requested):
Thanks a lot for the corrections you made! 
I added a few suggestions, please feel free to ask if anything is unclear.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
```diff
@@ -88,6 +88,54 @@ class URLStringUtilsTest {
...
+        assertFalse(isURLLike("[[[:/"))
+        assertFalse(isURLLike("http://]]]"))
+
+        // Invalid IPv6 literals correctly recognized as something else.
+        assertTrue(isURLLike("fe80::"))
+        assertTrue(isURLLike("x:["))
+
+        // Invalid IPv6 literals incorrectly recognized as valid.
+        // We allow these for now, until bug 1685152 is fixed.
+        assertTrue(isURLLike("[:::"))
```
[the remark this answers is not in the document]
pmarks-net: My `isUrlLikeEmulated()` tests indicate that "fe80::" and "x:[" belong in a different category, because the URI class thinks they are valid.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/WebURLFinderTest.kt
pmarks-net: I've updated isUrlLikeEmulated() to use `find("random_text $it other_random_text")`, so it covers the cases you listed and more.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/URLStringUtilsTest.kt
pmarks-net: done

pmarks-net:
GitHub says "Merging is blocked - Merging can be performed automatically once the requested changes are addressed", so I'm going to try the "Re-request review" button.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/WebURLFinderTest.kt
```diff
@@ -118,4 +126,101 @@ class WebURLFinderTest {
...
+    @Test
+    fun isUrlLikeEmulated() {
+        // autolinkWebUrlPattern uses a copy of the regex from URLStringUtils,
+        // so here we emulate isURLLike() and copy its tests.
+        val isURLLike: (String) -> Boolean = {
+            find("random_text $it other_random_text") == it.trim()
+        }
+
+        // All cases that behave differently are annotated with INVERT().
+        val INVERT: (Boolean) -> Boolean = { !it }
```
titooan, line 139: Sorry, I forgot to specify that in my previous comment: we are adding complexity with this annotation-like function, that I think we could replace with clear comments around the tests.

We could replace the use of `assertTrue(INVERT(isURLLike(xxx)))` by `assertFalse(isURLLike(xxx))`. Those asserts could be gathered with a comment above explaining that the ideal regex should have the same result as the one in `UrlStringUtils.kt`, but the current one gives a different result. Also explaining that we         allow it for now, until bug 1685152 is fixed.

What do you think?

titooan (changes requested):
Thanks a lot for those fixes! 

Let's just fix one last comment, and it's good to ship.

android-components/components/support/utils/src/test/java/mozilla/components/support/utils/WebURLFinderTest.kt
pmarks-net: > replace the use of assertTrue(INVERT(isURLLike(xxx))) by assertFalse(isURLLike(xxx)). Those asserts could be gathered with a comment

I intentionally kept assertTrue/assertFalse the same as the original tests, and replaced the comments with a programmatic annotation, in order to reduce the cognitive effort required to maintain a second copy.  Your suggestion would *increase* complexity when both functions are considered.

titooan: I do understand the idea and the value of trying to keep both tests as similar as possible.
Nevertheless, this is starting a new pattern in the codebase, and maybe we should try to follow the existing conventions instead?

Maybe we could get a second opinion from someone else? cc: @jonalmeida

jonalmeida: Hey! I agree with @titooan, I too would prefer avoiding new patterns. Adding these cases that should pass/fail can be in a separate section with a well-worded comment to explain that these should be have differently when the linked bug is fixed.

> I intentionally kept assertTrue/assertFalse the same as the original tests, and replaced the comments with a programmatic annotation, in order to reduce the cognitive effort required to maintain a second copy. Your suggestion would increase complexity when both functions are considered.

The felt complexity of both solutions are indeed subjective. In cases like this, I would recommend the advice of the code maintainers to decide what is the more preferred way, especially in terms of conventions the codebase follows. 🙂

pmarks-net: done.

titooan:
bors try

jonalmeida:
> bors try

It looks like the CI we depend on for our contributor PRs is down, so I've pushed [a separate branch](https://github.com/mozilla-mobile/firefox-android/pull/4566) to run all our CI tests.

jonalmeida:
> > bors try
> 
> It looks like the CI we depend on for our contributor PRs is down, so I've pushed [a separate branch](https://github.com/mozilla-mobile/firefox-android/pull/4566) to run all our CI tests.

CI has passed! I've pushed a small update to correct the changelog entry since we've cut the previous version already. Let's land this!

titooan (approved):
Thanks a lot for your contribution @pmarks-net, it looks pefect to me! 
Let's land it!

pmarks-net:
Firefox 122 (with the IPv6 literals fix) is now live on the Play store.

I'm mentioning https://github.com/mozilla-mobile/fenix/issues/4343 to notify the original reporters.

PR #4101, opened 2023-10-16
Bug 1857558 - [a11y] Dictate Correct CFR Popup Title

DreVla:
Previously when using Talkback and a CFR was displayed, the dictated caption was wrong. This patch aims to make the Popup focusable so once it appears while Talkback is on, it will dictate "Pop-Up window" followed by the text of the CFR Popup.

https://github.com/mozilla-mobile/firefox-android/assets/32488956/0e37456c-7aa5-476f-a909-e2cb7755dd7f

Extra video when TB is already turned on and user navigates to a view where a CFR opens

https://github.com/mozilla-mobile/firefox-android/assets/32488956/e8e3097d-f552-45e0-abd3-1fc3217c5df3

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1857558


android-components/components/compose/cfr/build.gradle
```diff
@@ -32,6 +32,10 @@ android {
         freeCompilerArgs += "-Xjvm-default=all"
     }
 
+    lint {
+        disable 'MissingTranslation'
+    }
```
MozillaNoah, line 37: what is this for?

DreVla: That is for ignoring lint MissingTranslation error for the `mozac_cfr_a11y_title` string added. It applies only for the compose cfr component.

MozillaNoah: meaning that it will ignore the lint warning when a string is used in this component/module, but the string will still be flagged for localization?

Do you know why this lint disable wasn't needed before?

DreVla: > meaning that it will ignore the lint warning when a string is used in this component/module, but the string will still be flagged for localization?

Exactly.

> Do you know why this lint disable wasn't needed before?

From what I have seen, it was already used in [AC here](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/android-lint.gradle), and it is applied to multiple components like this:
```
apply from: '../../../android-lint.gradle'
```
I could do the same here, but I am not sure if we want all the other warnings disabled. Let me know what you think.

MozillaNoah: Let's keep the lint changes light for now

android-components/components/compose/cfr/src/main/res/values/strings.xml
```diff
@@ -5,4 +5,7 @@
 <resources>
     <!-- Content description (not visible, for screen readers etc.): Description for the close button of a Contextual Feature Recommendation Popup. -->
     <string name="mozac_cfr_dismiss_button_content_description">Dismiss</string>
+
+    <!-- Content description: Text for a11y to be dictated when a CFR is displayed -->
+    <string name="mozac_cfr_a11y_title">Firefox notification</string>
```
MozillaNoah, line 10: Just caught this. Since this is AC, we shouldn't have any reference to Firefox here. Let's figure out a string to have in AC as the default and then pass-in a string through fenix with a parameter for the browser name.

[example usage](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/home/sessioncontrol/viewholders/FeltPrivacyInfoCard.kt#60)

android-components/components/compose/cfr/src/main/java/mozilla/components/compose/cfr/CFRPopupFullscreenLayout.kt
```diff
@@ -510,8 +510,22 @@ internal class CFRPopupFullscreenLayout(
...
+     */
+    private fun getApplicationName(): String {
+        val applicationInfo = anchor.context.applicationInfo
+        val stringId = applicationInfo.labelRes
+        return if (stringId == 0) {
+            applicationInfo.nonLocalizedLabel.toString()
+        } else {
+            anchor.context.getString(stringId)
+        }
+    }
```
MozillaNoah, line 527: so turns out we apparently already have an extension function to help us with this

[`val Context.appName: String`](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/content/Context.kt#56-57)

android-components/components/compose/cfr/src/main/res/values/strings.xml
```diff
@@ -5,4 +5,7 @@
 <resources>
     <!-- Content description (not visible, for screen readers etc.): Description for the close button of a Contextual Feature Recommendation Popup. -->
     <string name="mozac_cfr_dismiss_button_content_description">Dismiss</string>
+
+    <!-- Content description: Text for a11y to be dictated when a CFR is displayed -->
+    <string name="mozac_cfr_a11y_notification_title">notification</string>
```
MozillaNoah, line 10: let's make this a parameterized string, and the parameter can be the app name you obtain with `Context.appName`

MozillaNoah (approved):

PR #4110, opened 2023-10-16
Bug 1859408 - Instrument the awesomebar to report visible suggestions.

linabutler:
This commit adds an `onVisibilityStateUpdated` callback to the `AwesomeBar` composable. The new callback is invoked with the current `AwesomeBar.VisibilityState` whenever the list of visible suggestions changes.

I split this out of #4050 to make it a little easier to review, and so that we could land the instrumentation earlier, in case we need to uplift telemetry, or if there are any performance pitfalls.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1859408

### DIFF
--- a/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/Suggestions.kt
+++ b/android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/Suggestions.kt
@@ -4,15 +4,23 @@
 
 package mozilla.components.compose.browser.awesomebar.internal
 
+import android.os.Parcelable
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyListState
 import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.RememberObserver
+import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberUpdatedState
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.testTag
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.distinctUntilChangedBy
+import kotlinx.parcelize.Parcelize
 import mozilla.components.compose.browser.awesomebar.AwesomeBarColors
 import mozilla.components.compose.browser.awesomebar.AwesomeBarOrientation
 import mozilla.components.concept.awesomebar.AwesomeBar
@@ -25,6 +33,7 @@ internal fun Suggestions(
     orientation: AwesomeBarOrientation,
     onSuggestionClicked: (AwesomeBar.SuggestionProviderGroup, AwesomeBar.Suggestion) -> Unit,
     onAutoComplete: (AwesomeBar.SuggestionProviderGroup, AwesomeBar.Suggestion) -> Unit,
+    onVisibilityStateUpdated: (AwesomeBar.VisibilityState) -> Unit,
     onScroll: () -> Unit,
 ) {
     val state = rememberLazyListState()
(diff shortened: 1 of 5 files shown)


android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/Suggestions.kt
```diff
@@ -57,6 +66,26 @@ internal fun Suggestions(
             }
         }
     }
+
+    // The latest list of suggestions with which this composable was recomposed.
+    val currentSuggestions by rememberUpdatedState(suggestions)
+
+    // The latest `onVisibilityStateUpdated` callback with which this composable was composed.
```
MatthewTighe, line 73: nit: I do appreciate these comments because `rememberUpdatedState` is new to me but I think as Compose familiarity grows the behavior of these properties should be self-evident enough that we don't need the comments.

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
```diff
@@ -72,6 +72,17 @@ interface AwesomeBar {
...
 
+    /**
+     * Information about the [Suggestion]s that are currently displayed by the [AwesomeBar].
+     */
+    data class VisibilityState(
+        /**
+         * An ordered map of the currently visible [SuggestionProviderGroup]s, and the visible [Suggestion]s in each
+         * group. The groups are ordered top-to-bottom.
+         */
+        val groups: Map<SuggestionProviderGroup, List<Suggestion>> = emptyMap(),
```
MatthewTighe, line 83: nit:
```suggestion
        val suggestionGroups: Map<SuggestionProviderGroup, List<Suggestion>> = emptyMap(),
```

or even

```suggestion
        val visibleSuggestionGroups: Map<SuggestionProviderGroup, List<Suggestion>> = emptyMap(),
```

Super pedantic, but `VisibilityState` implies to my brain that it could also account for `Invsibile` items in the future because old Android View's had `INVISIBLE`, `GONE`, and `VISIBLE` options for their visibility property so this language is kinda in the general lexicon.

android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/Suggestions.kt
```diff
@@ -57,6 +66,26 @@ internal fun Suggestions(
...
+    val currentOnVisibilityStateUpdated by rememberUpdatedState(onVisibilityStateUpdated)
+
+    LaunchedEffect(Unit) {
+        // This effect is launched on the initial composition. The flow below emits new values when either the latest
+        // list of suggestions changes, or the visibility or position of any suggestion in that list changes.
```
MatthewTighe, line 78: Conversely I am actually very happy to see a functional composition block summarized in a comment this way.

android-components/components/compose/awesomebar/src/test/java/mozilla/components/compose/browser/awesomebar/internal/SuggestionsTest.kt
```diff
@@ -0,0 +1,126 @@
...
+import mozilla.components.concept.awesomebar.AwesomeBar.Suggestion
+import mozilla.components.concept.awesomebar.AwesomeBar.SuggestionProvider
+import mozilla.components.concept.awesomebar.AwesomeBar.SuggestionProviderGroup
+import mozilla.components.support.test.mock
+import mozilla.components.support.test.whenever
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class SuggestionsTest {
```
MatthewTighe, line 16: Awesome tests 🎉

MatthewTighe (approved):
Thanks for splitting these up, definitely made it easier! And great work overall, this was a tougher problem space than I anticipated and it's cool to see you work through these Compose intricacies! I learned quite a bit during these review cycles and only wish I'd known better at the start to be able to help guide more

android-components/components/concept/awesomebar/src/main/java/mozilla/components/concept/awesomebar/AwesomeBar.kt
linabutler: That makes sense; thanks for the additional context! 😊

android-components/components/compose/awesomebar/src/main/java/mozilla/components/compose/browser/awesomebar/internal/Suggestions.kt
linabutler: Agreed; removed both comments!

I was inspired to add them after reading [this helpful explanation](https://jetc.dev/slack/2021-08-28-why-need-rememberUpdateState.html) of why we need to use `rememberUpdatedState` with `LaunchedEffect`—and wanted to share it in case you find it helpful, too!

PR #4154, opened 2023-10-18
Bug 1859909 - Update AndroidX Compose to version 1.5.4

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1859909


mcarare (approved):
The only changes are related to the syncing of invalidation ( semantics and measurements) for Text, when changed or detached, in order to avoid some edge case crashes. So I am not expecting this upgrade to have any negative impact on our app.

PR #4170, opened 2023-10-19
Bug 1860130 - Fix getSerializedDoc message reply in readerview-background.js

rpl:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1860130

### DIFF
--- a/android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview-background.js
+++ b/android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview-background.js
@@ -5,16 +5,17 @@
 // This background script is needed to update the current tab
 // and activate reader view.
 
-browser.runtime.onMessage.addListener(async (message, sender, sendResponse) => {
+browser.runtime.onMessage.addListener(message => {
   switch (message.action) {
      case 'addSerializedDoc':
         browser.storage.session.set({ [message.id]: message.doc });
-        break;
+        return Promise.resolve();
      case 'getSerializedDoc':
-       let doc = await browser.storage.session.get(message.id);
-       browser.storage.session.remove(message.id);
-       sendResponse(doc);
-       break;
+        return (async () => {
+          let doc = await browser.storage.session.get(message.id);
+          browser.storage.session.remove(message.id);
+          return doc[message.id];
+        })();
      default:
        console.error(`Received unsupported action ${message.action}`);
    }
--- a/android-components/components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
+++ b/android-components/components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
@@ -30,6 +30,7 @@ import mozilla.components.support.ktx.kotlinx.coroutines.flow.filterChanged
 import mozilla.components.support.webextensions.WebExtensionController
 import org.json.JSONObject
 import java.lang.ref.WeakReference
+import java.net.URLEncoder
 import java.util.Locale
 import java.util.UUID
 
@@ -294,7 +295,12 @@ class ReaderViewFeature(
 
     private fun WebExtensionController.createReaderUrl(url: String, id: String): String? {
         val colorScheme = config.colorScheme.name.lowercase(Locale.ROOT)
-        return readerBaseUrl?.let { it + "readerview.html?url=$url&id=$id&colorScheme=$colorScheme" }
+        // Encode the original page url, otherwise when the readerview page will try to
+        // parse the url and retrieve the readerview url params (ir and colorScheme)
+        // the parser may get confused because the original webpage url being interpolated
+        // may also include its own search params non-escaped (See Bug 1860490).
+        val encodedUrl = URLEncoder.encode(url, "UTF-8")
+        return readerBaseUrl?.let { it + "readerview.html?url=$encodedUrl&id=$id&colorScheme=$colorScheme" }
     }
 
     @VisibleForTesting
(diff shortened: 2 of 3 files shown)


rpl:
@boek I included the STR to test this out manually in [Bug 1860130 comment 0](https://bugzilla.mozilla.org/show_bug.cgi?id=1860130#c0), but I'd really like to add a test case to cover this scenario explicitly, I notice that we have [fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ReaderViewTest.kt](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ReaderViewTest.kt) which seems to be testing the feature with actually switching a test page into readerview mode and I was wondering if it may be doable to add an additional test case to cover entering readerview mode on a doc that we can only retrieve using getSerializedDoc, but I haven't yet been able to give that a try, in the meantime I thought to ask you in this PR:
- if based on your knowledge that seems feasible or if the test is still mocking too much to be doable
- which gradlew command line I may use to run that test file only while trying it out some tweaks to it

android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview-background.js
```diff
@@ -5,16 +5,17 @@
...
+browser.runtime.onMessage.addListener(message => {
   switch (message.action) {
      case 'addSerializedDoc':
         browser.storage.session.set({ [message.id]: message.doc });
-        break;
+        return Promise.resolve();
      case 'getSerializedDoc':
-       let doc = await browser.storage.session.get(message.id);
-       browser.storage.session.remove(message.id);
-       sendResponse(doc);
```
rpl, line 16: @boek I wanted to mention this explicitly at least here: when a runtime.onMessage listener is an async function then when it is called its return value is always going to a promise and runtime.onMessage will in that case use the value resolved by the returned promise as the value to return to the sender of the message being handled (and sendResponse is ignored, see [here](https://searchfox.org/mozilla-central/rev/3fe52b82b0a00829d0a8224b4bd767cb20b7a27d/toolkit/components/extensions/ExtensionChild.sys.mjs#210-216)).

android-components/components/feature/readerview/src/main/assets/extensions/readerview/readerview-background.js
```diff
@@ -5,16 +5,17 @@
...
-       break;
+        return (async () => {
+          let doc = await browser.storage.session.get(message.id);
+          browser.storage.session.remove(message.id);
+          return doc[message.id];
```
rpl, line 17: @boek this part of the change was necessary to retrieve the actual value we are looking for in the session storage, the API call will be returning an object with the value we are looking for set as the value of the `message.id` key set on that object.

android-components/components/feature/readerview/src/main/java/mozilla/components/feature/readerview/ReaderViewFeature.kt
```diff
@@ -294,7 +295,12 @@ class ReaderViewFeature(
 
     private fun WebExtensionController.createReaderUrl(url: String, id: String): String? {
         val colorScheme = config.colorScheme.name.lowercase(Locale.ROOT)
-        return readerBaseUrl?.let { it + "readerview.html?url=$url&id=$id&colorScheme=$colorScheme" }
+        // Encode the original page url, otherwise when the readerview page will try to
+        // parse the url and retrieve the readerview url params (ir and colorScheme)
+        // the parser may get confused because the original webpage url being interpolated
+        // may also include its own search params non-escaped (See Bug 1860490).
+        val encodedUrl = URLEncoder.encode(url, "UTF-8");
+        return readerBaseUrl?.let { it + "readerview.html?url=$encodedUrl&id=$id&colorScheme=$colorScheme" }
```
rpl, line 303: @boek I added this additional change after having investigated [Bug 1860490](https://bugzilla.mozilla.org/show_bug.cgi?id=1860490) deeply enough to finally see how we were hitting that readerview mode issue.

I've currently rolled it into this pull request to allow @dholbert to be able to also give it a try and confirm this additional change along with the original changes part of this pull request are going to fix [Bug 1860490](https://bugzilla.mozilla.org/show_bug.cgi?id=1860490).

I verified on my side that with the changes from this pull request applied [Bug 1860490](https://bugzilla.mozilla.org/show_bug.cgi?id=1860490) was not hit anymore on my Fenix instance, but I had to go through a VPN to be able to reach the content from the url in [Bug 1860490 comment 0](https://bugzilla.mozilla.org/show_bug.cgi?id=1860490#c0) and so I'd like to confirm that the apk built for this PR does fix it also for @dholbert 

(btw, a huge thank you also goes to @Amejia481 :heart:  for giving me access to a VPN with a US exit note I could use to reproduce it and investigate it).

Having said that, it may be as well reasonable to split it out into its own separate patch and pull request built on top of this one, if we prefer that then I'll be more than happy to do that, let me know wdyt.

boek:
@rpl looks good! We will have to [fix the test](https://firefoxci.taskcluster-artifacts.net/Bv9ljYL1SG6hRzXn22Jrcg/0/public/reports/tests/testReleaseUnitTest/classes/mozilla.components.feature.readerview.ReaderViewFeatureTest.html#show%20reader%20view%20dispatches%20LoadUrlAction%20and%20UpdateReaderActiveAction) that is failing before we can land

boek (approved):

boek:
@Mergifyio backport releases_v120

PR #4181, opened 2023-10-20
Bug 1860274  - Modify the SwitchWithLabel behaviour for clicking and accessibility.

iorgamgabriel:
When the user clicks on the row will toggle the switch ,like in Settings.
Accessibility should read it like a single component, not like now two components (the text and the switch).

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1860274


fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
```diff
@@ -47,9 +50,16 @@ fun SwitchWithLabel(
     enabled: Boolean = true,
 ) {
     Row(
-        modifier = modifier,
         horizontalArrangement = Arrangement.spacedBy(16.dp),
         verticalAlignment = Alignment.CenterVertically,
+        modifier = modifier.then(
```
rahulsainani, line 55: If it's only for talkback to read it as one component and not have the whole Row/component toggleable, we should use `.semantics(mergeDescendants = true) { }` modifier.

iorgamgabriel: I think all the element should be toggeable just like in Settings Screen.

iorgamgabriel: This request comes from this pr https://github.com/mozilla-mobile/firefox-android/pull/4081#discussion_r1364976300

rahulsainani: Thanks for the context, since this is an existing ui component that's ported to compose I'm assuming UX is on the same page.

```suggestion
        modifier = modifier
            .toggleable(
                value = checked,
                enabled = enabled,
                role = Role.Switch,
                onValueChange = onCheckedChange,
            )
            .semantics(mergeDescendants = true) {},
```
1. Here we don't need to use .then but can simply add the toggleable modifier. 
2. Let's add the enabled param since we want the whole time to be disabled based on the passed in param. 
3. `onValueChange = onCheckedChange` same pattern in the file.
4. I think `mergeDescedants` would be better than clearing semantics of the switch. If we add another composable in the row, this would handle it.

iorgamgabriel:
https://github.com/mozilla-mobile/firefox-android/assets/8118371/495c6fb5-0c6c-437e-b14d-9436ea100132

iorgamgabriel:
Without clearing semantics the switch is still read .

rahulsainani:
@iorgamgabriel gotcha, let's keep that modifier on `Switch` in that case.

fenix/app/src/main/java/org/mozilla/fenix/compose/SwitchWithLabel.kt
iorgamgabriel: I made the changes without  .semantics(mergeDescendants = true) {},

rahulsainani (approved):

PR #4194, opened 2023-10-23
Bug 1857215 - Part 1:Integrate review checker recommendations

rahulsainani:
### What
First step to integrate the product recommendations API

### How
– Integrate the exposed AC API
– Use the first item of the list if present or return null
– Extract `EnumMapper` since it's used in multiple places and will be used for product recommendations mapping as well.

### What's next
Add actions, mapping and middleware  logic.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1857215


fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/EnumMapper.kt
```diff
@@ -0,0 +1,12 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.shopping.middleware
+
+/**
+ * Converts a string to an enum value, ignoring case. If the string does not match any of the
+ * enum values, the default value is returned.
+ */
+inline fun <reified T : Enum<T>> String.asEnumOrDefault(defaultValue: T? = null): T? =
+    enumValues<T>().firstOrNull { it.name.equals(this, ignoreCase = true) } ?: defaultValue
```
calumozilla, line 12: are these enumMapper changes related to recommendations? should they be in this PR

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
```diff
@@ -43,6 +44,13 @@ interface ReviewQualityCheckService {
      * Returns the selected tab url.
      */
     fun selectedTabUrl(): String?
+
+    /**
+     * Fetches product recommendations related to the product user is browsing in the current tab.
+     *
+     * @return [ProductRecommendation] if request succeeds, null otherwise.
+     */
+    suspend fun productRecommendations(): ProductRecommendation?
```
MozillaNoah, line 53: the plurality/singularity verbiage here is mismatched. The API returns a list of recommendations, but based on the default implementation, we only ever want the first one that's returned?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
```diff
@@ -110,8 +118,23 @@ class DefaultReviewQualityCheckService(
...
-    private inline fun <reified T : Enum<T>> String.asEnumOrDefault(defaultValue: T? = null): T? =
-        enumValues<T>().firstOrNull { it.name.equals(this, ignoreCase = true) } ?: defaultValue
+    override suspend fun productRecommendations(): ProductRecommendation? =
+        withContext(Dispatchers.Main) {
+            suspendCoroutine { continuation ->
+                browserStore.state.selectedTab?.let { tab ->
+                    tab.engineState.engineSession?.requestProductRecommendations(
+                        url = tab.content.url,
+                        onResult = {
+                            continuation.resume(it.firstOrNull())
```
MozillaNoah, line 128: let's include a comment why we only care about the first recommendation and ignoring whatever else the API returns

MozillaNoah (approved):

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
rahulsainani: Indeedio, it was returning a list earlier and then middleware was responsible for the logic of getting the first one but makes sense to do it here at the edge since we only care about the first recommendation.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/EnumMapper.kt
rahulsainani: Yes, since this will be used in the recommendations mapping logic (next PR) along with the 3 places it's used currently, the idea was to split this bug into 2 PRs(commits) in effort to make it easier for reviews, so the second part won't have these changes of extracting the enum mapper and tests.

PR #2370, opened 2023-06-08
Bug 1834747 - Fix talkback linktext access url issue

rahulsainani:
### How
– Set modifier clickable when screen reader is enabled, along with setting the Role as `Role.Button` which tells the accessibility services that double tap is available.
– For non screen reader enabled cases, nothing has changed.

### Why
– This was chosen over `UrlAnnotation`, which has better talkback support, as that has a [bug](https://issuetracker.google.com/issues/253292081), which prevents talkback from reading the url, it's fixed in compose-ui:1.5.0-beta.
– Trying the beta, in talkback mode, the link is shown in a separate dialog and the on clicking the link, it takes the user to the default browser (which could be other browsers). Since the component calls the lambda passed onClick, this approach seems better than uses `UrlAnnotation`.

### Video
https://github.com/mozilla-mobile/firefox-android/assets/38040960/5b895766-8e76-4585-ac70-10dacf6bbd90

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation

Used by GitHub Actions.

PR #4200, opened 2023-10-23
Bug 1855286 - Fix Talk Back with link texts with multiple embedded links

MozillaNoah:
Talkback was only discovering one of the two links in Shopping onboarding. After doing some adjacent pairing with Jeff, I discovered we could add `addUrlAnnotation` when constructing the annotated string to allow TalkBack to properly get two links inside its context menu but also do the link chime sound effect when reading off the blurb. The proper links also had to be plugged-in for `LinkText` to function properly.

This should also fix [Bug 1857850](https://bugzilla.mozilla.org/show_bug.cgi?id=1857850).

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1855286

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
@@ -11,11 +11,10 @@ import androidx.compose.foundation.text.ClickableText
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
-import androidx.compose.ui.semantics.onClick
-import androidx.compose.ui.semantics.semantics
-import androidx.compose.ui.text.AnnotatedString
+import androidx.compose.ui.text.ExperimentalTextApi
 import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.UrlAnnotation
 import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextDecoration
@@ -49,6 +48,7 @@ data class LinkTextState(
  * @param linkTextColor [Color] applied to the clickable part of the text.
  * @param linkTextDecoration [TextDecoration] applied to the clickable part of the text.
  */
+@OptIn(ExperimentalTextApi::class)
 @Composable
 fun LinkText(
     text: String,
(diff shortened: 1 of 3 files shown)


fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
```diff
@@ -129,8 +104,14 @@ internal fun buildUrlAnnotatedString(
...
 
+        addUrlAnnotation(
+            urlAnnotation = UrlAnnotation(it.url),
+            start = startIndex,
+            end = endIndex,
+        )
+
         addStringAnnotation(
-            tag = URL_TAG,
+            tag = URL_TAG + "_${it.text}",
```
MozillaNoah, line 114: this is to allow the string annotations to all have unique tags

fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
```diff
@@ -67,46 +67,21 @@ fun LinkText(
         linkTextDecoration,
     )
 
-    // When using UrlAnnotation, talkback shows links in a separate dialog and
-    // opens them in the default browser. Since this component allows the caller to define the
-    // onClick behaviour - e.g. to open the link in in-app custom tab, here StringAnnotation is used
-    // and modifier is enabled with Role.Button when screen reader is enabled.
```
rahulsainani, line 73: This was the reason why `StringAnnotation` was used. [PR](https://github.com/mozilla-mobile/firefox-android/pull/2370) for reference. As mentioned, with this change, talkback will open the links in a default browser. 

We have to make a choice:
1. Either we are fine with users being navigated out of firefox into their default browser and use `UrlAnnotation`.
2. Create a custom dialog that shows all the links so we can handle the `onClick` callbacks for those and open them in firefox.
3. Allow `LinkText` to support 2 behaviors (number 1 that's mentioned above and the changes in the PR) and let the callers decide, defaulting to 1, since Onboarding also uses it -> that means new users, who don't have firefox set as default. This will give us time to implement 2 properly. 

wdyt?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ReviewQualityCheckContextualOnboarding.kt
```diff
@@ -26,13 +27,13 @@ import org.mozilla.fenix.compose.LinkText
 import org.mozilla.fenix.compose.LinkTextState
 import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.compose.button.PrimaryButton
+import org.mozilla.fenix.shopping.middleware.GetReviewQualityCheckSumoUrl
+import org.mozilla.fenix.shopping.middleware.ReviewQualityCheckNavigationMiddleware
```
rahulsainani, line 31: ui should not depend on middleware.
💭 discussion: maybe we can make the links part of the state or move the constants to store package since it's part of the business logic. This also depends on our approach based on this [comment](https://github.com/mozilla-mobile/firefox-android/pull/4200/files#r1369778333).

fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
MozillaNoah: I think at the very least we need to get Talkback working in general. Having only one link work with a11y services obviously isn't ideal, so as a first step, I'd like us to lean towards option 1 (for now).

Jeff and I talked about this a bit, and this is definitely a side-symptom of applying a desktop design convention to mobile of having 2+ clickable links in one body of text. iOS doesn't even comply with this. They are specced-out to have each of the privacy policy and terms of use links on their own lines.

image

So realistically, while we are trying to solve, in front of us, an engineering problem, we could definitely argue that it's instead a design + a11y problem. Having two links in one body of text makes more sense in a world where the user has the precision of a mouse pointer. What we have now is an a11y issue we're trying to code around, but is also an experience issue for everyone because the tap targets are way too small to have multi-embedded links. We kinda brush it under the rug here, likely because they are links to support information and not core parts of an experience.

rahulsainani: If I'm understanding that right, wouldn't it then be better to update the texts in both features (shopping, TPBM) to have one link and to keep using `StringAnnotation` and control the callback, since one link is already working on talkback. If we update to `UrlAnnotation` as this patch is doing, we potentially navigate users away firefox, definitely for onboarding, which would be a major bug IMO but I'm interested in opinions from UX/a11y on this.

boek: @MozillaNoah In these cases it is probably okay to push back (like the iOS team has) on the designs where they are deviating from the happy path. I agree with @rahulsainani that we should not be navigating users away from the app and we also should not be breaking Talkback.

MozillaNoah: Forgot to loop back here last week, but I posted in the ENG-UX slack channel and the approach is being discussed amongst Design, Copy, and Localization because of various conflicts of those three channels/needs. The iOS design needs to be updated too as a result of that decision.

MozillaNoah:
Closing this as it was resolved in a different PR. See ticket for details

PR #4205, opened 2023-10-24
Bug 1860587 - Set allow_checks_interruption to false

dsmithpadilla:
By setting allow_checks_interruption false, Mergify will not interrupt the checks. Instead, a pull request with a higher priority will be inserted just after the pull requests that currently have checks running.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1860587


jcristau (commented):
Doesn't this kind of defeat the purpose of the "high priority" rule?

dsmithpadilla:
@jcristau My understanding is that this only stops it from cutting the 1st in line that is being processed (when set to false). So in a queue of 10 it will be No. 2 but wont interrupt the 1st one that probably is already almost ready to be merged.

jcristau:
> @jcristau My understanding is that this only stops it from cutting the 1st in line that is being processed (when set to false). So in a queue of 10 it will be No. 2 but wont interrupt the 1st one that probably is already almost ready to be merged.

Ah, I see, I guess that's because speculative_checks is 1 and so no checks are running for any other queued PRs?

dsmithpadilla:
> > @jcristau My understanding is that this only stops it from cutting the 1st in line that is being processed (when set to false). So in a queue of 10 it will be No. 2 but wont interrupt the 1st one that probably is already almost ready to be merged.
> 
> Ah, I see, I guess that's because speculative_checks is 1 and so no checks are running for any other queued PRs?

Right. So the checks are only happening on the tip of the queue. The rest of the items in queue are simply in line. Setting this to false moves the urgent PR to the top of the line of those who have no checks currently running. Currently, it interrupts the tip.

jcristau (approved):

PR #4226, opened 2023-10-25
Bug 1855988 - Translation Options UI Dialog Screen

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslationOptions.kt
```diff
@@ -0,0 +1,113 @@
...
+    subText: String,
+    selected: Boolean = false,
+    onClick: () -> Unit,
+) {
+    Row(
+        modifier = Modifier
+            .padding(start = 16.dp, top = 6.dp, bottom = 6.dp, end = 16.dp)
+            .clickable { onClick() },
+    ) {
+        RadioButton(
```
ohall-m, line 84: We should take a look at the accessibility on this component a little closer. 

I was correctly able to make a selection by activating the text in the list  TalkBack. (The other areas worked great too!) However, I was also able to navigate focus to the RadioButton itself as an unlabled component. 

We should follow the same pattern as we did on the switch, where the list item is selectable, but the radio button alone is not focusable or else knows it is connected.

fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslation.kt
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+}
+
+@Composable
+internal fun getAutomaticTranslationList(): List<AutomaticTranslationItem> {
+    return mutableListOf<AutomaticTranslationItem>().apply {
+        add(
+            AutomaticTranslationItem(
+                languageLocale = Locale.CANADA,
```
ohall-m, line 83: Thanks for adding the placeholder data!

This area will change when we do integration into AC, but I'd prefer if our placeholder data was a bit more similar. 

**Background:**
Our GV API calls here will likely be [`getLanguageSettings()`](https://searchfox.org/mozilla-central/rev/42acdc9cd5ae89222bdceeeaed7bacac755be48f/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#239) and [`listSupportedLanguages()`](https://searchfox.org/mozilla-central/rev/42acdc9cd5ae89222bdceeeaed7bacac755be48f/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#133). I'm still thinking through this in AC, but we will likely call something that makes these calls more convenient, something like:

`List<AutomaticTranslationPreference> getAutomaticTranslationPrefs` (Will be a new AC call that coordinates the two GV calls.):

`AutomaticTranslationPreference`:
* [`Language`](https://searchfox.org/mozilla-central/rev/42acdc9cd5ae89222bdceeeaed7bacac755be48f/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#1043):
  * `code` - String BCP 47 code of the language
  * `localizedDislayName` - String language name
* [`LanguageSetting`](https://searchfox.org/mozilla-central/rev/42acdc9cd5ae89222bdceeeaed7bacac755be48f/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#549-567) - enum 

**Requested Change:**
* Change `languageLocale` to `displayName` (For the placeholder data, just make the `Locale.displayLanguage` call here)
  * (This placeholder will ultimately probably become `Language`, which is about to be introduced as soon as this [patch](https://github.com/mozilla-mobile/firefox-android/pull/4521/files) finishes review.)
* (More of a thought or item to look at): Not sure best how to orcestrate it, but when we are ready to commit the change the user made (e.g., offer, always, never), we will need to call [`setLanguageSettings`](https://searchfox.org/mozilla-central/rev/42acdc9cd5ae89222bdceeeaed7bacac755be48f/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#278). So, at the point of commiting the change to GV, we will need to know the BCP 47 `code` and the radio `LanguageSetting` enum to use. (What are your thoughts?)

fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslationItem.kt
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * AutomaticTranslationItem that will appear on Automatic Translation screen.
+ *
+ * @property languageLocale The text that will appear in the list.
+ * @property automaticTranslationOption The option that the user selected.
+ */
+@Parcelize
+data class AutomaticTranslationItem(
+    val languageLocale: Locale,
```
ohall-m, line 20: Please adjust to a simple string for now, will ultimately probably become `Language`.

fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslationItem.kt
```diff
@@ -0,0 +1,68 @@
...
+@Parcelize
+sealed class AutomaticTranslationOption(
+    open val titleId: Int,
+    open val summaryId: List<Int>,
+) : Parcelable {
+
+    /**
+     * The app will offer to translate sites in the selected language.
+     */
+    data class OfferToTranslate(
```
ohall-m, line 39: Nice descriptions on this class! I think this class works just like we need it to for integration. 

Only one thing to keep in mind is we will probably will need to add  [`LanguageSetting`](https://searchfox.org/mozilla-central/rev/a9cb8b07d4cf4290e693bc31ba23412d0f042cb8/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#552) down the road for the final call to commit the change. Thought I'd mention it in-case you see it being an issue/something to think about.

fenix/app/src/main/res/navigation/nav_graph.xml
```diff
@@ -1443,6 +1443,23 @@
                 app:destination="@id/translationsDialogFragment"
                 app:popUpTo="@id/translationSettingsFragment"
                 app:popUpToInclusive="true" />
+            <action
+                android:id="@+id/action_translationSettingsFragment_to_automaticTranslationFragment"
```
ohall-m, line 1447: nit: I sense we might want to do some kind of `pref` or `preference` prefixing here to keep everything organized.

ohall-m (commented):
Thanks, this is looking good! Only had a few comments, mostly related to organization and what we will need to expect for integrating with AC.

fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslationOptions.kt
ohall-m: Current behavior:
* Good -> Can focus TalkBack on list and correctly change item
* Needs adjustment -> Can focus TalkBack on the radio button (label is not identified)

Requested change:
* Similar to how we made the switch component before.
* Should be able to focus on the text/list element and change selection, radio button alone should not be focusable alone aka the list pair should be only focusable as one element.

fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslation.kt
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Locale
+
+/**
+ * Firefox Automatic Translate fragment compose view.
+ *
+ * @param automaticTranslationList List of [AutomaticTranslationItem]s to display.
+ * @param onItemClick Invoked when the user clicks on the a item from the list.
+ */
+@Composable
+fun AutomaticTranslation(
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslation.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslationOptions.kt
iorgamgabriel: Done

fenix/app/src/main/res/navigation/nav_graph.xml
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslationItem.kt
```diff
@@ -0,0 +1,67 @@
...
+) : Parcelable {
+
+    /**
+     * The app will offer to translate sites in the selected language.
+     */
+    data class OfferToTranslate(
+        override val titleId: Int = R.string.pref_automatic_translation_option_offer_to_translate_title,
+        override val summaryId: List<Int> = listOf(
+            R.string.pref_automatic_translation_option_offer_to_translate_summary,
+            R.string.app_name,
```
ohall-m, line 42: I would adjust this string and the others using `R.string.app_name` because the copy is:

`Firefox will offer to translate sites in this language.`

and using `app_name` on my build produces:

`Firefox Fenix will offer to translate sites in this language.`

I did a check through the strings.xml and it seems "Firefox" is used in strings. (I didn't see a freestanding "Firefox" for us to substitute in, but maybe there is somewhere?)

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslationOptions.kt
```diff
@@ -0,0 +1,120 @@
...
+import androidx.compose.ui.semantics.clearAndSetSemantics
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.compose.button.RadioButton
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Firefox Automatic Translation Options fragment compose view.
+ *
+ * @param selectedOption Selected option that will come from Gecko.
```
ohall-m, line 30: nit: I keep using the name "Gecko" too, but we should try to avoid using platform specific language!

```suggestion
 * @param selectedOption Selected option that will come from the translations engine.
```

ohall-m (commented):
I think we need to make a slight copy change and this looks good to go for me!

fenix/app/src/main/java/org/mozilla/fenix/translations/automatic/AutomaticTranslationOptions.kt
ohall-m: Tested in TalkBack, works great, thanks!

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslationItem.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslationOptions.kt
iorgamgabriel: Done

ohall-m (approved):
Thanks, LGTM!

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslation.kt
```diff
@@ -0,0 +1,117 @@
...
+    prefAutomaticTranslationList: List<PrefAutomaticTranslationItem>,
+    onItemClick: (PrefAutomaticTranslationItem) -> Unit,
+) {
+    Column(
+        modifier = Modifier
+            .background(
+                color = FirefoxTheme.colors.layer1,
+            ),
+    ) {
+        Text(
```
gabrielluong, line 43: Did a `TextListItem` not work here?

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslation.kt
```diff
@@ -0,0 +1,117 @@
...
+import java.util.Locale
+
+/**
+ * Firefox Preferences Automatic Translate fragment compose view.
+ *
+ * @param prefAutomaticTranslationList List of [PrefAutomaticTranslationItem]s to display.
+ * @param onItemClick Invoked when the user clicks on the a item from the list.
+ */
+@Composable
+fun PrefAutomaticTranslation(
```
gabrielluong, line 33:
```suggestion
fun AutomaticTranslationPreference(
```
Gonna suggest a different way of naming all of these preferences. My thinking is that we would rather know upfront what the title of the screen says. wdyt?

fenix/app/src/main/res/values/static_strings.xml
```diff
@@ -144,4 +144,12 @@
     <string name="translations_bottom_sheet_translating_in_progress">In progress</string>
     <string name="translations_bottom_sheet_translating_in_progress_content_description">Translating in Progress</string>
 
+    <string name="pref_automatic_translation_toolbar_title">Automatic translation</string>
```
gabrielluong, line 147: @ohall-m This is for a followup. Wondering if we are at the point of landing strings if we know the strings have been finalized?

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslation.kt
ohall-m: I made one of these suggestions [originally](https://github.com/mozilla-mobile/firefox-android/pull/4226#discussion_r1399554662). The other patch will need to be updated too, if we change directions on conventions.

My thoughts with keeping the prefix are that we are going to end up with a lot of preferences on this feature and it would be nice to have a way to categorize them as distinct from other elements. Open to whatever you think!

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslationOptions.kt
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Top level radio button for the automatic translation options.
+ *
+ * @param text The main text to be displayed.
+ * @param subText The subtext to be displayed.
+ * @param selected [Boolean] that indicates whether the radio button is currently selected.
+ * @param onClick Invoked when the radio button is clicked.
+ */
+@Composable
+fun AutomaticTranslationOptionRadioButton(
```
gabrielluong, line 75: I am gonna ask for a follow up bug here where we extend our existing [ListItem](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt#216) to add the Radio Button use case. We should try to avoid implementing one-off components when we know the source of these components are coming from the design system. In this case, this is an extension of our existing ListItem.

To visualize this, you can click on "Go to main component" when inspecting a component in figma. 
Screen Shot 2023-11-27 at 11 31 14 AM

This will take you to https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?node-id=3016%3A12658&mode=dev. We are trying to build out the design components composable that we are using.

fenix/app/src/main/res/values/static_strings.xml
ohall-m: They should be finalized at this point except when otherwise noted! L10N has also been consulted too. `Copy: COMPLETE` and `L10N: COMPLETE` on the Figma and I also have that impression from other conversations.

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslation.kt
ohall-m: #4356 is the other bug we will need to comment on, if we move to a different convention.

gabrielluong: @ohall-m I like the idea of grouping these things into folders as you suggested with `translations/preferences/automatic` in your original suggestion. I think I am also gravitating towards having `Pref` or `Preference` as the suffix instead of prefix.

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslation.kt
```diff
@@ -0,0 +1,117 @@
...
+import androidx.compose.ui.semantics.semantics
+import androidx.compose.ui.unit.dp
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.compose.list.TextListItem
+import org.mozilla.fenix.theme.FirefoxTheme
+import java.util.Locale
+
+/**
+ * Firefox Preferences Automatic Translate fragment compose view.
```
gabrielluong, line 27:
```suggestion
 * Automatic Translate preference screen.
```

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslation.kt
ohall-m: Sounds good to me!

fenix/app/src/main/res/values/static_strings.xml
gabrielluong: Yes, I agree that a big string PR is indeed very difficult to review and there is good reason to worry. However, if we continue to allow our static strings to grow bigger and land later. It will only fuel an even bigger string PR that would be even more difficult to review.

We can start small and open smaller PRs for strings in individual screens that have their strings finalized.

ohall-m: Ahh, I see what you are saying, sounds good to me! I'll file some follow ups. I'm thinking`Core Translation Flow`, `Preferences`, `Errors`, and `Dialogs`. Can we land directly to `strings.xml` on patches like this or is that a bad practice?

ohall-m: Perfect, thank you for the info!

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/PrefAutomaticTranslationOptions.kt
iorgamgabriel: https://bugzilla.mozilla.org/show_bug.cgi?id=1868105

gabrielluong (approved):

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/AutomaticTranslationOptionsFragmentPreference.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.ui.platform.ComposeView
+import androidx.fragment.app.Fragment
+import androidx.navigation.fragment.navArgs
+import org.mozilla.fenix.ext.showToolbar
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * A fragment displaying the Firefox Preference Automatic Translation Options screen.
+ */
+class AutomaticTranslationOptionsFragmentPreference : Fragment() {
```
gabrielluong, line 20:
```suggestion
class AutomaticTranslationOptionsPreferenceFragment : Fragment() {
```

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/AutomaticTranslationFragmentPreference.kt
```diff
@@ -0,0 +1,48 @@
...
+import androidx.fragment.app.Fragment
+import androidx.navigation.fragment.findNavController
+import org.mozilla.fenix.R
+import org.mozilla.fenix.ext.showToolbar
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * A fragment displaying the Firefox Automatic Translation list screen.
+ */
+class AutomaticTranslationFragmentPreference : Fragment() {
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/preferences/automatic/AutomaticTranslationOptionsFragmentPreference.kt
iorgamgabriel: Done

PR #4254, opened 2023-10-27
Bug 1861475 - Add logic for Shopping recommendation impression

MozillaNoah:
This was done over two commits:
1. Refactor `Modifier.onShown` out of Pocket Composables to be reusable. 
2. Plug in the Action to hook up the onshown trigger to the Store

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1861475


fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -56,6 +57,8 @@ import java.util.SortedMap
 private val combinedParentHorizontalPadding = 32.dp
 private val productRecommendationImageSize = 60.dp
 
+private const val PRODUCT_RECOMMENDATION_SETTLE_TIME = 1500
```
rahulsainani, line 60: Let's suffix this with `_MS`

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -425,8 +449,16 @@ private enum class Highlight(
 private fun ProductRecommendation(
     product: RecommendedProductState.Product,
     onClick: () -> Unit,
+    onImpression: (RecommendedProductState.Product) -> Unit,
```
rahulsainani, line 452: Could be (String) -> Unit since we only need to pass the aid, same for all the functions calling into this

MozillaNoah: True true. I was initially leaning towards passing the entire data model, so the UI isn't necessarily aware of the API requirements, especially if they were to change down the road. 

Another option for consideration could be passing back nothing and having the middleware read the state directly and obtain the `aid` that way.

rahulsainani: hmm I think just passing the id would be make sense, in case we start showing two recommendations here, middleware woulnd't be able to know without extra info which recommendations had an impression and click.

I think we should also add make `onClick` of type `(String) -> Unit`, since we need to know which recommendation was clicked as well.

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
```diff
@@ -100,3 +115,110 @@ fun Modifier.thenConditional(
...
+ * Add a callback for when this Composable is "shown" on the screen.
+ * This checks whether the composable has at least [threshold] ratio of it's total area drawn inside
+ * the screen bounds.
+ * Does not account for other Views / Windows covering it.
+ *
+ * @param threshold The ratio of the total area to be within the screen bounds to trigger [onVisible].
+ * @param settleTime The amount of time to wait before calling [onVisible].
+ * @param onVisible Invoked when the UI is visible to the user.
+ */
+fun Modifier.onShown(
```
MozillaNoah, line 136: Doing some quick research, this Modifier extension cannot be covered by a unit test and will need a UI test if we choose to add coverage here.

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
rahulsainani: Think this was missed

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -425,8 +449,16 @@ private enum class Highlight(
 private fun ProductRecommendation(
     product: RecommendedProductState.Product,
     onClick: () -> Unit,
+    onImpression: (String) -> Unit,
 ) {
-    Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
+    Column(
+        modifier = Modifier.onShown(
+            threshold = 0.5f,
```
rahulsainani, line 456: Let's extract this as a constant, Is 0.5 confirmed?

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
rahulsainani: Gotcha, do you think we can add it in this PR?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -71,10 +74,11 @@ private val productRecommendationImageSize = 60.dp
  * @param onShowMoreRecentReviewsClicked Invoked when the user clicks to show more recent reviews.
  * @param onExpandSettings Invoked when the user expands the settings card.
  * @param onRecommendedProductClick Invoked when the user clicks on the product recommendation.
+ * @param onProductRecommendationImpression Invoked when the user has seen the product recommendation.
  * @param modifier The modifier to be applied to the Composable.
  */
 @Composable
-@Suppress("LongParameterList")
+@Suppress("LongParameterList", "LongMethod")
```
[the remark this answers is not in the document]
MozillaNoah: I'll double check 👍

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
MozillaNoah: Looks like I only did the one settle time. thanks for catching this

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
MozillaNoah: It's what's been used for Pocket, but I can double check for here

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
MozillaNoah: We certainly could if you feel this is a requirement for landing the change. I don't feel strongly either way

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
rahulsainani: Thanks!

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
rahulsainani: It would increase our confidence in this logic and allow for easy refactoring later on when we move away from `composed {}`

MozillaNoah:
Pushed a WIP of what I'm working on. I have three tests but one is failing and currently trying to figure out how to best construct it.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ModifierTest.kt
```diff
@@ -0,0 +1,129 @@
...
+            )
+        }
+
+        composeTestRule.scrollToOnShownIndex()
+
+        assertFalse(onShown)
+
+        composeTestRule.mainClock.advanceTimeUntil(ON_SHOWN_SETTLE_TIME + 5000L) {onShown}
+
+//        composeTestRule.waitUntil(ON_SHOWN_SETTLE_TIME + 5000L) { onShown }
```
MozillaNoah, line 68: This is the test I'm currently struggling to find a good way to implement. There isn't a good way to do a true "wait" (nor is it advised), so I'm currently stuck trying to find a good way to have a test cover the settle time of `Modifier.onShown`

fenix/app/src/debug/AndroidManifest.xml
```diff
@@ -26,6 +26,8 @@
     <application
             tools:replace="android:name"
             android:name="org.mozilla.fenix.DebugFenixApplication">
+
+        <activity android:name="androidx.activity.ComponentActivity" />
```
MozillaNoah, line 30: for some reason, and only in the Fenix manifest, I needed to add this activity tag in order to get `createComposeRule()` to work in a UI test in Fenix. (An empty `ComponentActivity` is stood up via `createComposeRule()` that we can then set the content on. This way we don't have to do something as costly as working around `HomeActivity` to just test a Modifier.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ModifierTest.kt
rahulsainani: You can use java.time.Clock interface that's a param for the modifier. The production code uses `SystemClock` and the test can use `FixedClock` so we can control the time for tests.

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
```diff
@@ -100,3 +115,110 @@ fun Modifier.thenConditional(
...
+    var lastVisibleCoordinates: LayoutCoordinates? = null
+
+    return composed {
+        if (inComposePreview) {
+            Modifier
+        } else {
+            val context = LocalContext.current
+            var wasEventReported by remember { mutableStateOf(false) }
+
+            val toolbarHeight = context.resources.getDimensionPixelSize(R.dimen.browser_toolbar_height)
```
rahulsainani, line 151: Apologies that I missed this previously, Looking into the tests made me realize about this issue. I'm a bit worried that this modifier code has a lot of dependencies to things it shouldn't directly need. For instance, the common `onShown` modifier shouldn't worry about the rest of the UI hierarchy, in the case of RC bottom sheet, the toolbar placement doesn't even matter since the composable is in a bottom sheet and not a list on home. Wondering if we should use this common logic or think of a different way?

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ModifierTest.kt
MozillaNoah: I shall check this out!

fenix/app/src/main/java/org/mozilla/fenix/compose/ext/Modifier.kt
MozillaNoah: I agree that this Modifier has a lot of dependencies and is a good candidate for further refactoring. 

Good catch on the toolbar stuff. If this Modifier is to be reusable, its API will need to be changed to factor something like this in when necessary and not assume it's relevant for 100% of use cases (like review checker). I'll need to think of a good way to accommodate both of these paths, so this will further delay this landing and I may need to hand this off, if it comes to that. I've already spent a good amount of time on this just standing up the requirements of the ticket but also trying to get the UI tests stood up, so I'm hoping this won't take too much longer.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/ModifierTest.kt
MozillaNoah: To close this loop, I was able to get the test to pass once I removed the toolbar logic from the Modifier!

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
MozillaNoah: Finally got confirmation that 50% is good for this impression

rahulsainani (approved):
Looks good! Thanks for adding the tests and refactoring!

As you suggested offline, we can add `screenBounds: Rect? = null` param to improve the API for  Modifier.`onShown()`.

And we can be consistent with passing the data for click/impression for `RecommendedProduct`, either in `ProductAnalysis` or in `RecommendedProduct`. Both are fine.

PR #4284, opened 2023-10-30
Bug 1861415 - Filter sponsored suggestion URLs from other providers when the "show sponsored suggestions" setting is enabled

linabutler:
Hi @jonalmeida and @MatthewTighe! 👋🏼 This turned into a bit of a refactor, so I wanted to put this up as a draft for your initial thoughts, in case you'd like to approach it differently!

The big idea is that, if the "show sponsored suggestions" setting is switched on, we want to hide suggestions whose URLs have `mfadid=adm` in their query params.

I saw we already do that for Top Sites via `frecencyFilterQuery`, and set about extending that to suggestions. I also noticed that we already have logic for filtering suggestions to only show URLs for the current search engine when a shortcut engine is selected, which got me thinking about ways to combine these filters.

So I did an exploratory refactor that:

1. Changed the providers to take a predicate function, instead of a URL or a host, for filtering results. The synced tabs provider is the only interesting case here: all our other providers give us an `android.net.Uri` to work with, but the synced tabs provider works with `String`s. I kept that behavior, instead of trying to turn it into a `Uri`, in case it's intentional.
2. Moved the filters into a new `SearchResultFilter` type, to capture the two types of filtering that we want to do.
3. Changed `AwesomeBarView.getProvidersToAdd` to build a list of `SearchResultFilter`s.
4. Changed the `AwesomeBarView.get*Providers` methods to take that list of filters, and turn them into a predicate to pass down to the providers.

Please let me know what you think about this direction, and if you like it, or if you have a better / easier way to approach it! I updated the tests in the first commit, but held off on updating the tests in the second commit for now.

### Pull Request checklist

### After merge

8. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
9. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
10. Click on `View task in Taskcluster` in the new `DETAILS` section.
11. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1861415

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
@@ -32,6 +32,8 @@ import mozilla.components.feature.syncedtabs.DeviceIndicators
 import mozilla.components.feature.syncedtabs.SyncedTabsStorageSuggestionProvider
 import mozilla.components.feature.tabs.TabsUseCases
 import mozilla.components.support.ktx.android.content.getColorFromAttr
+import mozilla.components.support.ktx.android.net.sameHostWithoutMobileSubdomainAs
+import mozilla.components.support.ktx.kotlin.tryGetHostFromUrl
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
@@ -363,7 +365,7 @@ class AwesomeBarView(
                 maxNumberOfSuggestions = METADATA_SUGGESTION_LIMIT,
                 showEditSuggestion = false,
                 suggestionsHeader = activity.getString(R.string.firefox_suggest_header),
-                resultsUriFilter = searchEngineUriFilter,
+                resultsUriFilter = { it.sameHostWithoutMobileSubdomainAs(searchEngineUriFilter) },
             )
         } else {
             HistoryStorageSuggestionProvider(
@@ -374,7 +376,7 @@ class AwesomeBarView(
                 maxNumberOfSuggestions = METADATA_SUGGESTION_LIMIT,
                 showEditSuggestion = false,
                 suggestionsHeader = activity.getString(R.string.firefox_suggest_header),
-                resultsUriFilter = searchEngineUriFilter,
+                resultsUriFilter = { it.sameHostWithoutMobileSubdomainAs(searchEngineUriFilter) },
             )
         }
     }
@@ -483,7 +485,9 @@ class AwesomeBarView(
                 getDrawable(activity, R.drawable.ic_search_results_device_tablet),
             ),
             suggestionsHeader = activity.getString(R.string.firefox_suggest_header),
-            resultsHostFilter = searchEngineHostFilter,
+            resultsUrlFilter = searchEngineHostFilter?.let {
+                { url -> url.tryGetHostFromUrl() == searchEngineHostFilter }
+            },
         )
     }
 
@@ -514,7 +518,9 @@ class AwesomeBarView(
             getDrawable(activity, R.drawable.ic_search_results_tab),
             excludeSelectedSession = !fromHomeFragment,
             suggestionsHeader = activity.getString(R.string.firefox_suggest_header),
-            resultsUriFilter = searchEngineUriFilter,
+            resultsUriFilter = searchEngineUriFilter?.let {
+                { uri -> uri.sameHostWithoutMobileSubdomainAs(it) }
+            },
         )
     }
 
@@ -545,7 +551,9 @@ class AwesomeBarView(
             engine = engineForSpeculativeConnects,
             showEditSuggestion = false,
             suggestionsHeader = activity.getString(R.string.firefox_suggest_header),
-            resultsUriFilter = searchEngineHostFilter,
+            resultsUriFilter = searchEngineHostFilter?.let {
+                { uri -> uri.sameHostWithoutMobileSubdomainAs(it) }
+            },
         )
     }
 
--- a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
@@ -6,6 +6,7 @@ package org.mozilla.fenix.search.awesomebar
 
 import android.content.Context
 import android.graphics.drawable.Drawable
+import android.net.Uri
 import androidx.annotation.VisibleForTesting
 import androidx.appcompat.content.res.AppCompatResources
 import androidx.core.graphics.BlendModeColorFilterCompat.createBlendModeColorFilterCompat
@@ -34,6 +35,7 @@ import mozilla.components.feature.tabs.TabsUseCases
 import mozilla.components.support.ktx.android.content.getColorFromAttr
 import mozilla.components.support.ktx.android.net.sameHostWithoutMobileSubdomainAs
 import mozilla.components.support.ktx.kotlin.tryGetHostFromUrl
+import mozilla.components.support.ktx.kotlin.urlContainsQueryParameters
 import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
@@ -41,6 +43,7 @@ import org.mozilla.fenix.components.Components
 import org.mozilla.fenix.components.Core.Companion.METADATA_HISTORY_SUGGESTION_LIMIT
 import org.mozilla.fenix.components.Core.Companion.METADATA_SHORTCUT_SUGGESTION_LIMIT
 import org.mozilla.fenix.ext.components
+import org.mozilla.fenix.ext.containsQueryParameters
 import org.mozilla.fenix.ext.settings
 import org.mozilla.fenix.search.SearchEngineSource
 import org.mozilla.fenix.search.SearchFragmentState
@@ -255,7 +258,7 @@ class AwesomeBarView(
         }
     }
 
-    @Suppress("ComplexMethod")
+    @Suppress("ComplexMethod", "LongMethod")
     @VisibleForTesting
     internal fun getProvidersToAdd(
         state: SearchProviderState,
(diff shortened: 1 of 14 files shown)


linabutler:
I was thinking about this some more, and wanted to make a few changes. Clearing review for now 😊

fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -278,45 +283,77 @@ class AwesomeBarView(
...
         }
 
         if (state.showHistorySuggestionsForCurrentEngine) {
-            getHistoryProvidersForSearchEngine(state.searchEngineSource)?.let {
-                providersToAdd.add(it)
-            }
+            providersToAdd.add(
+                getHistoryProvider(
+                    filters = listOfNotNull(
+                        getFilterToExcludeSponsoredSuggestions(state),
```
linabutler, line 297: This will always be `null` here, because we never show sponsored suggestions when a shortcut engine is selected—IOW, there's never a case where `state.show*SuggestionsForCurrentEngine == true && state.showSponsoredSuggestions == true` today—but I kept it for symmetry with the `showAll*Suggestions` cases.

fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -278,45 +283,77 @@ class AwesomeBarView(
...
+                        getFilterForCurrentEngine(state),
+                    ),
+                ),
+            )
         }
 
         if (activity.browsingModeManager.mode == BrowsingMode.Normal && state.showAllSessionSuggestions) {
-            providersToAdd.add(getLocalTabsProvider(state.searchEngineSource))
+            // Unlike other providers, we don't exclude sponsored results for open tabs.
```
linabutler, line 347: This matches the Desktop behavior that I observed: Desktop hides sponsored suggestions from history, bookmarks, and synced tabs, but not open local tabs.

MatthewTighe (commented):
First pass. Overall I think the direction of this is good! It does feel like the overall complexity of this area is a little high, but I don't think that's doing of this patch.

MatthewTighe:
Oh also, wanted to double-check whether this was intentionally a branch from the upstream? Typically, we prefer [merging from forks](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/CONTRIBUTING_code.md#creating-a-pull-request)

fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -566,6 +621,28 @@ class AwesomeBarView(
...
+         * A filter that only includes results whose URLs have the same host as [uri].
+         */
+        data class IncludeHost(val uri: Uri) : SearchResultFilter
+
+        /**
+         * A filter that excludes results whose URLs contain the given [query] parameters.
+         */
+        data class ExcludeQuery(val query: String) : SearchResultFilter
+
+        fun matches(uri: Uri): Boolean = when (this) {
```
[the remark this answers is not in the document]
linabutler: 🦀 🦀 🦀

fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
linabutler: Sounds good, that makes sense!

linabutler:
Alrighty, I think this is officially ready for review!

fenix/app/src/test/java/org/mozilla/fenix/search/awesomebar/AwesomeBarViewTest.kt
```diff
@@ -180,16 +267,45 @@ class AwesomeBarViewTest {
...
+                },
+            ),
+            showSponsoredSuggestions = false,
+        )
+
+        val result = awesomeBarView.getProvidersToAdd(state)
+
+        val historyProvider = result.firstOrNull { it is CombinedHistorySuggestionProvider }
+        assertNotNull(historyProvider)
+        assertNotNull((historyProvider as CombinedHistorySuggestionProvider).resultsUriFilter)
```
MatthewTighe, line 302: I'm wondering if there's an easy way to test which filter has been applied just so that we can verify that we're not missing any existing functionality.

fenix/app/src/test/java/org/mozilla/fenix/search/awesomebar/AwesomeBarViewTest.kt
```diff
@@ -623,17 +765,42 @@ class AwesomeBarViewTest {
...
+
+        val result = awesomeBarView.getProvidersToAdd(state)
+
+        val localSessionsProviders = result.filterIsInstance<SessionSuggestionProvider>()
+        assertEquals(1, localSessionsProviders.size)
+        assertNull(localSessionsProviders[0].resultsUriFilter)
+    }
+
+    @Test
+    fun `GIVEN normal browsing mode and needing to show all local tabs suggestions and sponsored suggestions are disabled WHEN configuring providers THEN add the tabs provider`() {
```
MatthewTighe, line 791: should we be expecting no filters here? Are we relying on a different area of the code to filter the sponsored suggestions while they're disabled?

MatthewTighe (approved):
This LGTM, thanks for working through all these corner cases and helping me understand it a bit better offline! I do think because of some of the inherent complexity here that we should wait for a solid QA pass before we consider uplifting, but I don't see anything that looks particularly risky in the code itself

fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt
```diff
@@ -278,45 +283,68 @@ class AwesomeBarView(
...
         if (state.showAllHistorySuggestions) {
-            if (activity.settings().historyMetadataUIFeature) {
-                providersToAdd.add(defaultCombinedHistoryProvider)
-            } else {
-                providersToAdd.add(defaultHistoryStorageProvider)
-            }
+            providersToAdd.add(
+                getHistoryProvider(
+                    filter = getFilterToExcludeSponsoredResults(state),
+                ),
```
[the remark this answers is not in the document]
linabutler: We talked about this offline, but I wanted to capture that discussion here, so that we can leave some breadcrumbs for our future selves!

Previously—on tip-of-tree Fenix today; before my PR—we have a few cases:

* If `state.showAllHistorySuggestions == true`, we use `defaultCombinedHistoryProvider` or `defaultHistoryProvider`. These actually [don't do any filtering at all](https://github.com/mozilla-mobile/firefox-android/blob/ff06d1ef8f05887291213d0ea0d8c25eb999e15a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt#L120-L140), and omit the `resultsUriFilter` param.
* If `state.showHistorySuggestionsForCurrentEngine == true`, we call `getHistoryProvidersForSearchEngine()`. That method [returns `null`](https://github.com/mozilla-mobile/firefox-android/blob/ff06d1ef8f05887291213d0ea0d8c25eb999e15a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt#L354) if `state.searchEngineSource.searchEngine == null`, or a new provider (combined or not) with a `resultUriFilter` param, and the [same other options](https://github.com/mozilla-mobile/firefox-android/blob/ff06d1ef8f05887291213d0ea0d8c25eb999e15a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt#L358-L365) as the default providers.

To make things even more complicated, the way we handle "`state.searchEngineSource.searchEngine == null`" for history is consistent with how we handle [search term suggestions](https://github.com/mozilla-mobile/firefox-android/blob/ff06d1ef8f05887291213d0ea0d8c25eb999e15a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt#L402), but inconsistent with the way we handle [bookmarks](https://github.com/mozilla-mobile/firefox-android/blob/ff06d1ef8f05887291213d0ea0d8c25eb999e15a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt#L535-L538), [synced tabs](https://github.com/mozilla-mobile/firefox-android/blob/ff06d1ef8f05887291213d0ea0d8c25eb999e15a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt#L471-L474), and [local tabs](https://github.com/mozilla-mobile/firefox-android/blob/ff06d1ef8f05887291213d0ea0d8c25eb999e15a/fenix/app/src/main/java/org/mozilla/fenix/search/awesomebar/AwesomeBarView.kt#L504-L507). Bookmarks, synced tabs, and local tabs will return a provider _without_ a filter if we're only showing suggestions for the current filter, but `state.searchEngineSource.searchEngine == null`.

And to get even further into the weeds, I think the only case where `searchEngineSource.searchEngine == null` is for [`SearchEngineSource.None`](https://github.com/mozilla-mobile/firefox-android/blob/f2afc4e57598b31fdab25600b575de354a53cc2a/fenix/app/src/main/java/org/mozilla/fenix/search/SearchFragmentStore.kt#L38-L43). IIUC, that's only used (1) in test code, or (2) when the user hasn't selected a search engine, and we don't have a default for their region.

With my PR, things are a little different:

* `getHistoryProvidersForSearchEngine()` is now `getHistoryProvider()`, and behaves like the other `get*Provider()` methods.
*  If `state.showAllHistorySuggestions == true` and `state.showSponsoredSuggestions == false`, `getHistoryProvider()` returns `defaultCombinedHistoryProvider` or `defaultHistoryProvider`.
* If `state.showAllHistorySuggestions == true` and `state.showSponsoredSuggestions == true`, `getHistoryProvider()`returns a (combined or not) history provider, with a filter to exclude sponsored suggestions.
* If `state.showHistorySuggestionsForCurrentEngine == true`, we call `getFilterForCurrentEngineResults()`. If that returns `null`—in the case where `searchEngineSource.searchEngine == null`—we don't call `getHistoryProvider()` at all. Otherwise, we call `getHistoryProvider()` with that filter, which will only include suggestions with the same host as `searchEngineSource.searchEngine.resultsUrl`.

I also changed bookmarks, synced tabs, and local tabs to work like history in the case where `searchEngineSource.searchEngine == null`. I don't think that this is a user-visible change—because I don't think there's any way to select a `None` engine—but it was more for consistency later, so that we aren't wondering, "why do we not show any history for `None`, but all bookmarks and tabs?"

fenix/app/src/test/java/org/mozilla/fenix/search/awesomebar/AwesomeBarViewTest.kt
linabutler: Local tabs are the one exception: we don’t want to filter out any open sponsored results, to match Desktop’s behavior. I added a comment to `AwesomeBarView` about that, but I’ll add one to this test, too; thanks!

fenix/app/src/test/java/org/mozilla/fenix/search/awesomebar/AwesomeBarViewTest.kt
linabutler: I wonder if `::` gives us a stable lambda, so that we can check if it’s equal here. I‘ll check back later this evening, or tomorrow 😊

PR #4126, opened 2023-10-18
Bug 1845409 - File RFC 0010 proposing handling navigation through middleware

MatthewTighe:
This patch includes an RFC for a new architectural goal of handling navigation through middleware moving forward. Currently, there are two commits. 

1. Old proposal for observing navigation state and reacting accordingly.
2. New proposal for handling navigation through middleware.

The first proposal seemed suboptimal after offline conversations with the team, so I've reworked it accordingly. It's included here for context.

Tentative deadline is Feb 2nd.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1845409

PR #4323, opened 2023-11-01
Nav proto

MatthewTighe:
This is a prototype patch to show case the two main proposals from RFC 0010: https://github.com/mozilla-mobile/firefox-android/pull/4126

I've separated them into two commits:
1. Adding a `Screen` property to the `AppState` which will be observed from the `HomeActivity` in order to control app-wide navigation.
2. A small test for creating a middleware to handle "transient" effects in response to `Action`s. This commit is less polished and is meant mostly to try and demonstrate that piece of the proposal.

Due to time constraints, only the 1st item is actually functional. To see it in action, check out the first commit and try clicking pocket stories on the home page or the home button from the browser to see it in action.

I've left comments in-line in the patch to try and explain some outstanding concerns/questions of my own and to address areas that would need more investigation and refinement if we were to adopt them for real.

Much of the code here is copied from where it currently lives. I think there are further improvements that could be made if this work were actually done to fit better into the unidirectional, single-source of truth patterns defined by `lib-state`. 

Used by GitHub Actions.

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/AppNavigationBinding.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/AppNavigationBinding.kt
@@ -1,0 +1,122 @@
+package org.mozilla.fenix
+
+import androidx.annotation.IdRes
+import androidx.navigation.NavController
+import androidx.navigation.NavDirections
+import androidx.navigation.NavHost
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.distinctUntilChangedBy
+import mozilla.components.lib.state.helpers.AbstractBinding
+import org.mozilla.fenix.addons.AddonDetailsFragmentDirections
+import org.mozilla.fenix.addons.AddonPermissionsDetailsFragmentDirections
+import org.mozilla.fenix.addons.AddonsManagementFragmentDirections
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.appstate.AppState
+import org.mozilla.fenix.components.appstate.Screen
+import org.mozilla.fenix.exceptions.trackingprotection.TrackingProtectionExceptionsFragmentDirections
+import org.mozilla.fenix.ext.alreadyOnDestination
+import org.mozilla.fenix.ext.nav
+import org.mozilla.fenix.home.HomeFragment
+import org.mozilla.fenix.home.HomeFragmentDirections
+import org.mozilla.fenix.library.bookmarks.BookmarkFragmentDirections
+import org.mozilla.fenix.library.history.HistoryFragmentDirections
+import org.mozilla.fenix.library.historymetadata.HistoryMetadataGroupFragmentDirections
+import org.mozilla.fenix.library.recentlyclosed.RecentlyClosedFragmentDirections
+import org.mozilla.fenix.search.SearchDialogFragmentDirections
+import org.mozilla.fenix.settings.CookieBannersFragmentDirections
+import org.mozilla.fenix.settings.HttpsOnlyFragmentDirections
+import org.mozilla.fenix.settings.SettingsFragmentDirections
+import org.mozilla.fenix.settings.TrackingProtectionFragmentDirections
+import org.mozilla.fenix.settings.about.AboutFragmentDirections
+import org.mozilla.fenix.settings.logins.fragment.LoginDetailFragmentDirections
+import org.mozilla.fenix.settings.logins.fragment.SavedLoginsAuthFragmentDirections
+import org.mozilla.fenix.settings.search.SaveSearchEngineFragmentDirections
+import org.mozilla.fenix.settings.search.SearchEngineFragmentDirections
+import org.mozilla.fenix.settings.studies.StudiesFragmentDirections
+import org.mozilla.fenix.settings.wallpaper.WallpaperSettingsFragmentDirections
+import org.mozilla.fenix.share.AddNewDeviceFragmentDirections
+import org.mozilla.fenix.shopping.ReviewQualityCheckFragmentDirections
+import org.mozilla.fenix.tabstray.TabsTrayFragmentDirections
+import org.mozilla.fenix.trackingprotection.TrackingProtectionPanelDialogFragmentDirections
+
+class AppNavigationBinding(
+    appStore: AppStore,
+    private val navController: NavController,
+) : AbstractBinding<AppState>(appStore) {
+    override suspend fun onState(flow: Flow<AppState>) = flow.distinctUntilChangedBy { it.screen }
+        .collect {
+            when (val screen = it.screen) {
+                is Screen.Home -> {
+                    navController.navigate(R.id.homeFragment)
+                }
+                is Screen.Browser -> {
+                    val from = screen.from
+                    if (navController.alreadyOnDestination(R.id.browserFragment)) return@collect
+                    @IdRes val fragmentId = if (from.fragmentId != 0) from.fragmentId else null
+                    val directions = getNavDirections(from, screen.customTabSessionId)
+                    navController.nav(fragmentId, directions)
+                }
+            }
+        }
+
+    private fun getNavDirections(
+        from: BrowserDirection,
+        customTabSessionId: String?,
+    ): NavDirections = when (from) {
+        BrowserDirection.FromGlobal ->
+            NavGraphDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromHome ->
+            HomeFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromWallpaper ->
+            WallpaperSettingsFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromSearchDialog ->
+            SearchDialogFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromSettings ->
+            SettingsFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromBookmarks ->
+            BookmarkFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromHistory ->
+            HistoryFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromHistoryMetadataGroup ->
+            HistoryMetadataGroupFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromTrackingProtectionExceptions ->
+            TrackingProtectionExceptionsFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromCookieBanner ->
+            CookieBannersFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromHttpsOnlyMode ->
+            HttpsOnlyFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromAbout ->
+            AboutFragmentDirections.actionGlobalBrowser(customTabSessionId)
+        BrowserDirection.FromTrackingProtection ->
+            TrackingProtectionFragmentDirections.actionGlobalBrowser(customTabSessionId)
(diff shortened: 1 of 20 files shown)


fenix/app/src/main/java/org/mozilla/fenix/AppNavigationBinding.kt
```diff
@@ -0,0 +1,122 @@
...
+                    val from = screen.from
+                    if (navController.alreadyOnDestination(R.id.browserFragment)) return@collect
+                    @IdRes val fragmentId = if (from.fragmentId != 0) from.fragmentId else null
+                    val directions = getNavDirections(from, screen.customTabSessionId)
+                    navController.nav(fragmentId, directions)
+                }
+            }
+        }
+
+    private fun getNavDirections(
```
MatthewTighe, line 62: I'm not actually sure of the benefit of these `NavDirections`. All the ones that I checked seemed to reference the area they were called from, which would mean we could just push new Destinations onto the stack and rely on normal popping behavior.

All that to say, I think there's a possibility that all of this could be removed to reduce the code in this file.

fenix/app/src/main/java/org/mozilla/fenix/AppTabMiddleware.kt
```diff
@@ -0,0 +1,119 @@
...
+    private fun loadBrowserTab(screen: Screen.Browser) {
+        val startTime = profiler?.getProfilerTime()
+        val mode = BrowsingMode.Normal
+//        val mode = browsingModeManager.mode
+
+
+//        val private = when (mode) {
+//            BrowsingMode.Private -> true
+//            BrowsingMode.Normal -> false
+//        }
```
MatthewTighe, line 52: This isn't available because the `BrowsingModeManager` is currently owned by the `HomeActivity` instead of `Components`, where the middleware is initialized.

I already have an in-progress patch to remove `BrowsingModeManager` (see [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1861459)) and move this state into the `AppStore` which would allow this middleware to read the state directly.

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -1026,7 +1026,7 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
     @OptIn(DelicateCoroutinesApi::class)
     open fun downloadWallpapers() {
         GlobalScope.launch {
-            components.useCases.wallpaperUseCases.initialize()
+            components.wallpaperUseCases.initialize()
```
MatthewTighe, line 1029: There was a circular dependency issue adding usecases to the new AppStore middleware, so I had to move the wallpapers usecase up a level.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -563,6 +568,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
                 "finishing" to isFinishing.toString(),
             ),
         )
+
+        navigationBinding.stop()
```
MatthewTighe: Whoops, this isn't actually necessary 😅 handled automatically by adding the binding to the list of lifecycle observers above.

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -260,12 +260,6 @@ abstract class BaseBrowserFragment :
         val originalContext = ActivityContextWrapper.getOriginalContext(activity)
         binding.engineView.setActivityContext(originalContext)
 
-        browserFragmentStore = StoreProvider.get(this) {
-            BrowserFragmentStore(
-                BrowserFragmentState(),
-            )
-        }
-
```
MatthewTighe, line 268: There are some awkward lifecycle constraints here where the `BrowserToolbarView` is recreated as the fragment is, whereas this store would survive. Since I copied the existing CFR handling directly, I did not think through the correct way to handle lifecycle management if this Store actually survived configuration changes.

fenix/app/src/main/java/org/mozilla/fenix/components/Components.kt
```diff
@@ -85,12 +91,20 @@ class Components(private val context: Context) {
...
+        // Required to even access context.filesDir property and to retrieve current locale
+        val (rootStorageDirectory, currentLocale) = strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
+            val rootStorageDirectory = context.filesDir
+            val currentLocale = LocaleManager.getCurrentLocale(context)?.toLanguageTag()
+                ?: LocaleManager.getSystemDefault().toLanguageTag()
+            rootStorageDirectory to currentLocale
+        }
+        WallpapersUseCases(context, appStore, core.client, rootStorageDirectory, currentLocale)
+    }
+
```
MatthewTighe, line 107: See comment above about circular dependency issue

fenix/app/src/main/java/org/mozilla/fenix/components/UseCases.kt
```diff
@@ -107,15 +104,4 @@ class UseCases(
...
-    val wallpaperUseCases by lazyMonitored {
-        // Required to even access context.filesDir property and to retrieve current locale
-        val (rootStorageDirectory, currentLocale) = strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
-            val rootStorageDirectory = context.filesDir
-            val currentLocale = LocaleManager.getCurrentLocale(context)?.toLanguageTag()
-                ?: LocaleManager.getSystemDefault().toLanguageTag()
-            rootStorageDirectory to currentLocale
-        }
-        WallpapersUseCases(context, appStore, client, rootStorageDirectory, currentLocale)
-    }
```
MatthewTighe, line 120: Circular dependency thing

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
```diff
@@ -79,4 +83,22 @@ data class AppState(
...
+        val newTab: Boolean,
+        val from: BrowserDirection,
+        val customTabSessionId: String? = null,
+        val engine: SearchEngine? = null,
+        val forceSearch: Boolean = false,
+        val flags: EngineSession.LoadUrlFlags = EngineSession.LoadUrlFlags.none(),
+        val requestDesktopMode: Boolean = false,
+        val historyMetadata: HistoryMetadataKey? = null,
+        val additionalHeaders: Map<String, String>? = null,
+    ) : Screen()
```
MatthewTighe, line 103: I expect that with more thought that some of this could be simplified or moved into a Store.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/ToolbarCfrMiddleware.kt
```diff
@@ -0,0 +1,152 @@
...
+        when (action) {
+            is BrowserFragmentAction.Init -> {
+                val isPrivate = customTabId?.let {
+                    browserStore.state.findCustomTabOrSelectedTab(it)?.content?.private
+                } ?: browserStore.state.selectedTab?.content?.private ?: false
+                val cfr = getCFRToShow(isPrivate)
+                handleCfr(cfr)
+            }
+        }
+    }
```
MatthewTighe, line 56: The important piece here is that basically - the view will be created and the middleware will handle checking which CFR should be shown.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarView.kt
```diff
@@ -181,6 +183,20 @@ class BrowserToolbarView(
...
+                initialState = BrowserFragmentState(),
+                middleware = listOf(ToolbarCfrMiddleware(
+                    components.core.store,
+                    view,
+                    components.settings,
+                    DefaultShoppingExperienceFeature(),
+                    customTabSession?.id,
+                    {},
+                    {},
+                    {},
```
MatthewTighe, line 197: I think there is opportunity to improvise a bit in cases like this - but the easiest case is defining the CFR callbacks inline in the view unless there is a strong desire to unit test them.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
gabrielluong: I think this is a good idea. We should start with something ugly and then make it better iteratively.

fenix/app/src/main/java/org/mozilla/fenix/AppNavigationBinding.kt
```diff
@@ -0,0 +1,122 @@
...
+import org.mozilla.fenix.settings.search.SaveSearchEngineFragmentDirections
+import org.mozilla.fenix.settings.search.SearchEngineFragmentDirections
+import org.mozilla.fenix.settings.studies.StudiesFragmentDirections
+import org.mozilla.fenix.settings.wallpaper.WallpaperSettingsFragmentDirections
+import org.mozilla.fenix.share.AddNewDeviceFragmentDirections
+import org.mozilla.fenix.shopping.ReviewQualityCheckFragmentDirections
+import org.mozilla.fenix.tabstray.TabsTrayFragmentDirections
+import org.mozilla.fenix.trackingprotection.TrackingProtectionPanelDialogFragmentDirections
+
+class AppNavigationBinding(
```
gabrielluong, line 42: I am kinda struggling with this binding in terms of wrapping my brain around everything, but this is more of a reaction without much further thought on why and an understanding of everything - I feel like this could be a middleware.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserFragmentStore.kt
```diff
@@ -7,21 +7,27 @@
...
+class BrowserFragmentStore(
+    initialState: BrowserFragmentState,
+    middleware: List<Middleware<BrowserFragmentState, BrowserFragmentAction>>
+) :
     Store<BrowserFragmentState, BrowserFragmentAction>(initialState, ::browserStateReducer)
 
 /**
  * The state for the Browser Screen
  */
 class BrowserFragmentState : State
```
gabrielluong, line 26: This whole file is kinda wild that it still exists. At the top, it says it is unused and it doesn't even hold any state. This might be better renamed `BrowserToolbarStore` to better serve your purpose. 

Alternatively, I am also wondering if we should react to the `InitAction` from `BrowserAction` for your toolbar CFR middleware.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
```diff
@@ -79,4 +83,22 @@ data class AppState(
     val wallpaperState: WallpaperState = WallpaperState.default,
     val standardSnackbarError: StandardSnackbarError? = null,
     val shoppingState: ShoppingState = ShoppingState(),
+    val screen: Screen = Screen.Home,
```
MozillaNoah, line 86: one thought for future consideration here: how will the default value here be updated based on the user's "Opening screen" preference?

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -30,6 +30,7 @@ import org.mozilla.fenix.wallpapers.Wallpaper
  * [Action] implementation related to [AppStore].
  */
 sealed class AppAction : Action {
+    data class ChangeScreen(val screen: Screen) : AppAction()
```
MozillaNoah, line 33: I might be getting hung up on semantics, but what about:

```
data class Navigate(val destination: NavigationDestination) : AppAction()
```

There will be instances, like Tabs Tray or the review checker, where we navigate to a non-fullscreen feature flow.

Tangentially, but I imagine at some point our CFRs getting an overhaul and having an engine that's centered around a similar Redux pattern.

fenix/app/src/main/java/org/mozilla/fenix/AppNavigationBinding.kt
MozillaNoah: +1 to a navigation middleware

PR #4389, opened 2023-11-07
Bug 1856336 - Integrate rc impression and click APIs in network middleware

rahulsainani:
Described in the bug.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1856336


fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
```diff
@@ -137,6 +147,39 @@ class DefaultReviewQualityCheckService(
                 }
             }
         }
+
+    override suspend fun recordRecommendedProductClick(productAid: String) =
+        withContext(Dispatchers.Main) {
+            suspendCoroutine { continuation ->
+                browserStore.state.selectedTab?.engineState?.engineSession?.sendClickAttributionEvent(
```
MozillaNoah, line 154: I can't imagine doing an optional call chain like this in the Java days 😅

fenix/app/src/test/java/org/mozilla/fenix/shopping/fake/FakeReviewQualityCheckService.kt
```diff
@@ -30,5 +32,14 @@ class FakeReviewQualityCheckService(
     override suspend fun analysisStatus(): AnalysisStatusDto? = status
 
     override fun selectedTabUrl(): String? = selectedTabUrl
+
     override suspend fun productRecommendation(): ProductRecommendation? = productRecommendation
+
+    override suspend fun recordRecommendedProductClick(productAid: String) {
+        recordClick(productAid)
+    }
```
MozillaNoah, line 40: super curious if this will work
```suggestion
    override suspend fun recordRecommendedProductClick(productAid: String) = recordClick
```

MozillaNoah (approved):

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
rahulsainani: _nullpointerexceptions intensifies_ 😅 

to be fair, we could theoretically assert these as non null since these calls can only happen from UI actions when we have a selected tab, but it's safer to not assume that since this class doesn't know about where the triggers can come from.

fenix/app/src/test/java/org/mozilla/fenix/shopping/fake/FakeReviewQualityCheckService.kt
rahulsainani: 😄 doesn't work, `recordClick` is of type  `(String) -> Unit`, and this function returns a `Unit`

PR #4396, opened 2023-11-07
Bug 1834820 - Update RemoteSettings downloader class

sarah541:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1834820


android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -164,7 +150,7 @@ sealed class RemoteSettingsResult {
     /**
      * Represents a successful outcome of a remote settings operation.
      */
-    data class Success(val records: List<RemoteSettingsRecord>) : RemoteSettingsResult()
+    data class Success(val records: RemoteSettingsResponse) : RemoteSettingsResult()
```
MatthewTighe, line 153:
```suggestion
    data class Success(val response: RemoteSettingsResponse) : RemoteSettingsResult()
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -102,29 +103,14 @@ class RemoteSettingsClient(
...
-                    )
-                }
+                val records: Any? = jsonObject.get("records")
+
                 fileReader.close()
-                RemoteSettingsResult.Success(records)
+                if (records is RemoteSettingsResponse) {
+                    RemoteSettingsResult.Success(records)
+                } else {
+                    RemoteSettingsResult.DiskFailure(IOException("Invalid data format"))
```
MatthewTighe, line 112: nit:
```suggestion
                    RemoteSettingsResult.DiskFailure(IOException("Invalid or corrupted saved data"))
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -102,29 +103,14 @@ class RemoteSettingsClient(
...
-                        ),
-                    )
-                }
+                val records: Any? = jsonObject.get("records")
+
                 fileReader.close()
-                RemoteSettingsResult.Success(records)
+                if (records is RemoteSettingsResponse) {
+                    RemoteSettingsResult.Success(records)
+                } else {
```
MatthewTighe, line 111: Bit of oddness here - we're `get`ting the "records" field but the expected type is `RemoteSettingsResponse`? In general, let's try to use `response` as a variable name where appropriate. If possible, I would also prefer if we could try to avoid the `Any?` type and instead directly cast it. Something like
```
val response = jsonObject.get("<name>") as? RemoteSettingsResponse
```

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -93,7 +75,26 @@ class RemoteSettingsClient(
     }
 
     /**
-     * Reads all records for a collection  found in the local storage.
+     * Updates the local storage with [records] data only if records changed from last time.
+     */
+    suspend fun writeLastModified(records: RemoteSettingsResponse, lastModified: ULong): RemoteSettingsResult =
```
MatthewTighe, line 80: nit:
```suggestion
    suspend fun writeIfModifiedSince(records: RemoteSettingsResponse, lastModified: ULong): RemoteSettingsResult =
```
Let's make sure we have `@param` descriptions in the kdoc as well, since this is a public API

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -66,25 +63,10 @@ class RemoteSettingsClient(
...
-                    put("deleted", it.deleted)
-                    put("attachment", it.attachment)
-                    put("fields", it.fields)
-                }
-                jsonArray.put(recordJson)
-            }
-
             val jsonObject = JSONObject()
-            jsonObject.put("records", jsonArray)
+            jsonObject.put("records", records)
```
MatthewTighe, line 69: Here as well we are writing a `RemoteSettingsResponse` using a "record" name. I think we should try to align those things.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -93,7 +75,26 @@ class RemoteSettingsClient(
...
+                    jsonObject.put("records", records)
+                    AtomicFile(file).writeString { jsonObject.toString() }
+                    RemoteSettingsResult.Success(records)
+                } else {
+                    RemoteSettingsResult.NoChanges // Return a specific result for the case where no changes are made
+                }
+            } catch (e: IOException) {
+                RemoteSettingsResult.DiskFailure(e)
+            }
+        }
```
MatthewTighe, line 94: Couple things here:

1. I think this could call the existing `write` function.
2. I was under the impression that we were looking to add functionality to this client to fetch and then write if the response was newer than what was stored. The way this function is reading, it still seems like it would be the responsibility of the consumer to handle that logic.

For 2, I think we have a few options:
1. Expose the `lastModified` property from the client and let the consumer handle writing updates to storage based on whether the response was newer.
2. Keep this function as is. 
3. Change this function to be a new extension function with a name like `fetchAndWriteIfResponseUpdated` that does a fetch _and_ and update if the response is newer than the previous `lastModified`.

Personally, I would opt for the 3rd option. I am just not sure we're adding a ton of value by adding another function where the only real difference is the `if (records.lastModified != lastModified)` condition. A consumer could probably handle that themselves.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -176,4 +162,9 @@ sealed class RemoteSettingsResult {
      * (e.g., network not available, server error).
      */
     data class NetworkFailure(val error: Exception) : RemoteSettingsResult()
+
+    /**
+     * Indicates that no changes were made
+     */
+    object NoChanges : RemoteSettingsResult()
```
MatthewTighe, line 169: Is there an existing scenario where a consumer would care to know there were no changes? I'm just wondering if there would ever be an effect or action taken as a result of this. IMO, it seems likely that we could re-use `Success` here, even if the resulting state does not change.

MatthewTighe (commented):
Had a few questions/comments, but this is overall looking good! I think the idea of returning the `RemoteSettingsResponse` as part of the `Success` result is definitely the overall direction we want to take things.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
gabrielluong: I am kinda surprised detekt isn't flagging this 🤔

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
sarah541: I added etchAndWriteIfResponseUpdated extension and writeIfLastModified functions as they both can be utilized individually.

MatthewTighe (approved):
Nice, glad to see how simple these functions become with the switch to `response`. This LGTM, mostly had some nits around comment wording. I did have a question about the comparison in `writeIfModifiedSince`, I'll approve now and leave it up to you whether that needs to change or not

PR #4426, opened 2023-11-09
Bug 1821005 - Restore previous Gradle 8 upgrade reverted due to startup crashes

mcarare:
This reverts commit 592e88b3547d2e96ae04d8b5bd69cbab2ac3e97e.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1821005


focus-android/gradle.properties
```diff
@@ -25,3 +25,5 @@ libUrl=https://github.com/mozilla-mobile/firefox-android/tree/main/focus-android
 libVcsUrl=https://github.com/mozilla-mobile/firefox-android.git
 libLicense=MPL-2.0
 libLicenseUrl=https://www.mozilla.org/en-US/MPL/2.0/
+android.nonTransitiveRClass=false
+android.nonFinalResIds=false
```
rahulsainani, line 29: Are we enabling r8 full mode in focus?

mcarare: No, I tested and minifying with R8 full mode also crashes Focus. 
As for the sample apps they do not have minify enabled.

rahulsainani: Yes, that's what my understanding was. I see we've now added the property below 👍

mcarare: Yes, thanks for pointing that out!

jonalmeida (approved):
Thanks for keeping the new changes in a separate commit for easier reviewing! Let's land this now, so that the next build at [4am UTC](https://github.com/mozilla-mobile/firefox-android/blob/main/.cron.yml#L13) will have this landed so we can monitor this in nightly during working hours.

Could you please land similar changes into Reference Browser as well (enabling minification too in our release types) so that we can catch future bugs as well? Thanks!

PR #4436, opened 2023-11-09
Bug 1863754 - Disable sponsored and non-sponsored suggestions when Firefox Suggest is disabled.

linabutler:
This commit fixes an issue where it was possible for Firefox to show sponsored and non-sponsored suggestions if the user explicitly disabled and re-enabled the settings, and then unenrolled from the experiment. In that case, the `show{Non}SponsoredSuggestions` settings would still be `true` because they were user-set, even though `enableFxSuggest` was now `false`.

As of this commit, the `show{Non}SponsoredSuggestions` settings are only used if `enableFxSuggest` is `true`.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1863754

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/search/SearchFragmentStore.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/search/SearchFragmentStore.kt
@@ -185,9 +185,9 @@ fun createInitialSearchFragmentState(
         showSessionSuggestionsForCurrentEngine = false,
         showAllSessionSuggestions = true,
         showSponsoredSuggestions = activity.browsingModeManager.mode == BrowsingMode.Normal &&
-            settings.showSponsoredSuggestions,
+            settings.enableFxSuggest && settings.showSponsoredSuggestions,
         showNonSponsoredSuggestions = activity.browsingModeManager.mode == BrowsingMode.Normal &&
-            settings.showNonSponsoredSuggestions,
+            settings.enableFxSuggest && settings.showNonSponsoredSuggestions,
         tabId = tabId,
         pastedText = pastedText,
         searchAccessPoint = searchAccessPoint,
@@ -281,9 +281,9 @@ private fun searchStateReducer(state: SearchFragmentState, action: SearchFragmen
                 showAllSyncedTabsSuggestions = action.settings.shouldShowSyncedTabsSuggestions,
                 showSessionSuggestionsForCurrentEngine = false, // we'll show all local tabs
                 showSponsoredSuggestions = action.browsingMode == BrowsingMode.Normal &&
-                    action.settings.showSponsoredSuggestions,
+                    action.settings.enableFxSuggest && action.settings.showSponsoredSuggestions,
                 showNonSponsoredSuggestions = action.browsingMode == BrowsingMode.Normal &&
-                    action.settings.showNonSponsoredSuggestions,
+                    action.settings.enableFxSuggest && action.settings.showNonSponsoredSuggestions,
                 showAllSessionSuggestions = true,
             )
         is SearchFragmentAction.SearchShortcutEngineSelected ->
--- a/fenix/app/src/test/java/org/mozilla/fenix/search/SearchFragmentStoreTest.kt
+++ b/fenix/app/src/test/java/org/mozilla/fenix/search/SearchFragmentStoreTest.kt
@@ -64,6 +64,7 @@ class SearchFragmentStoreTest {
         every { settings.shouldShowHistorySuggestions } returns true
         every { settings.shouldShowSearchSuggestions } returns true
         every { settings.shouldShowSearchSuggestionsInPrivate } returns false
+        every { settings.enableFxSuggest } returns true
         every { settings.showSponsoredSuggestions } returns true
         every { settings.showNonSponsoredSuggestions } returns true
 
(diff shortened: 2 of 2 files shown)


fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1896,30 +1896,31 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+     * Defaults to `true` if Firefox Suggest is enabled and the user has not explicitly changed the
+     * setting. Always `false` and can't be changed if Firefox Suggest is disabled.
      */
-    var showSponsoredSuggestions by lazyFeatureFlagPreference(
+    var showSponsoredSuggestions by featureFlagPreference(
         key = appContext.getPreferenceKey(R.string.pref_key_show_sponsored_suggestions),
-        default = { enableFxSuggest },
-        featureFlag = FeatureFlags.fxSuggest,
+        default = true,
+        featureFlag = enableFxSuggest,
```
linabutler, line 1913: This is a "creative" use of `featureFlagPreference`, and I wanted to drop a comment with some more details why—and maybe see if there's a better way of doing that?

I think the behavior we want is:

* If `enableFxSuggest` is `true`, then `showSponsoredSuggestions` (and `showNonSponsoredSuggestions`) should be the user-set value. If there's no user-set value, `show{Non}SponsoredSuggestions` should default to `true`.
* If `enableFxSuggest` is `false`, then `show{Non}SponsoredSuggestions` should always be `false`—regardless of whether they've previously had a user-set value or not.

`enableFxSuggest` can be `true` in three ways:

* You're on a Nightly or Debug build, or...
* You're enrolled in the Nimbus experiment, or...
* You turn on the secret setting.

In other cases, it's `false`.

linabutler: Another approach would be to change all places that check `settings.show{Non}SponsoredSuggestions` to check `settings.enableFxSuggest && settings.show{Non}SponsoredSuggestions`—but this felt slightly nicer to me 😊

MatthewTighe: I imagine this logic is pretty subject to change as rollouts continue, so I'm happy landing this as is. That said, a couple alternative suggestions if you are interested in them:

- since `enableFxSuggest` seems outside of user control entirely, we could remove it as a cached preference and read from nimbus / feature flag directly. this could be abstracted into a function instead, like with `isAddressFeatureEnabled()`
- all of the logic could be converted to a function, something like `amoCollectionOverrideConfigured`

MatthewTighe (approved):
Had a couple alternative suggestions, but I also think this can work as is!

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
linabutler: > seems outside of user control entirely

There is one case where it's user-controllable (😅): they can enable or disable it via the Secret Settings menu.

We added that to make testing easier—before we had the preview experiment—but it seems like it could still be useful after that. If the experiment goes well and we roll Suggest out for everyone, I think we'll want to have `enableFxSuggest` check the current region for the default value—like `isAddressFeatureEnabled()`!—so that it's only visible and enabled for `en-US` users, but still have the secret setting as an escape hatch.

MatthewTighe: > There is one case where it's user-controllable (😅): they can enable or disable it via the Secret Settings menu.

I am kinda of the opinion that those don't count and that we should revisit our persistence workflows there, since we don't even persist access to the menu _but_ point taken! Again since I think this work is fairly in-flight still as long as it's working as intended then it's good by me 🙂

linabutler: Gotcha, thanks! I put the (😅) because I was bemused at how complicated this is, and definitely didn't mean to make a point or anything!

"A flag that's controllable by a Nimbus feature variable now, and by a region check in the future, but can be overridden, and also affects the visibility of suggestions and Settings switches, and also also is still in-flight" is the shortest summary I can think of! 😆

Thanks for the review, and suggesting those other approaches!

linabutler:
I did some testing on my device, and realized there's a snag with using `featureFlagPreference(..., featureFlag = enableFxSuggest)`: the value of `enableFxSuggest` is evaluated once at startup. As a result, if you flip the secret setting, you now have to restart Fenix after, or things will behave oddly.

If `enableFxSuggest` was `true` on launch, and you flip the secret setting off, you'll still see suggestions (but not the new settings) until you restart. And if `enableFxSuggest` was `false` on launch, and you flip the secret setting on, you'll see the new settings, but they won't work—and you won't see suggestions—until you restart.

It's a small thing, and I know we recommend / require restarting for some of our other secret settings—but we check `settings.show{Non}SponsoredSuggestions` in so few places that doing this:

> Another approach would be to change all places that check `settings.show{Non}SponsoredSuggestions` to check `settings.enableFxSuggest && settings.show{Non}SponsoredSuggestions`

...Actually doesn't seem too bad!

MatthewTighe (approved):

PR #4457, opened 2023-11-13
Bug 1859671 - Remember review checker card states for the session

rahulsainani:
### What
Described in the bug

### How
– Use a map for product url to `CardState` to store the expanded states for each card in `AppStore`.
– Adds reducer logic and tests.
– Use existing actions for expanding/collapsing settings, info and highlights cards to dispatch actions on `AppStore`.
– Restore `CardState` for the selected product url if present.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1859671


fenix/app/src/main/java/org/mozilla/fenix/components/appstate/shopping/ShoppingState.kt
```diff
@@ -7,11 +7,28 @@ package org.mozilla.fenix.components.appstate.shopping
...
  * analysed when the sheet was closed.
  * @property shoppingSheetExpanded Boolean indicating if the shopping sheet is expanded and visible.
+ * @property productCardState Map of product url to [CardState] that contains the state of different
+ * cards in the shopping sheet.
  */
 data class ShoppingState(
     val productsInAnalysis: Set<String> = emptySet(),
     val shoppingSheetExpanded: Boolean? = null,
-)
+    val productCardState: Map<String, CardState> = emptyMap(),
```
MozillaNoah, line 19: Do we foresee any potential fragility with using the URL as the key here? In other words, is it guaranteed that the URL we use from the Fakespot service is the same between API calls?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckPreferencesMiddleware.kt
```diff
@@ -103,6 +126,53 @@ class ReviewQualityCheckPreferencesMiddleware(
...
+                        )
+                    },
+                )
+            }
+        }
+    }
+
+    private fun Store<AppState, AppAction>.dispatchShoppingAction(
+        reviewQualityCheckState: ReviewQualityCheckState,
+        action: (productPageUrl: String, optedIn: OptedIn) -> ShoppingAction,
```
MozillaNoah, line 170: why the lambda instead of just passing a `ShoppingAction` as the parameter?

MozillaNoah (commented):
Tests look good. Just one outstanding question about URLs as keys

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckPreferencesMiddleware.kt
rahulsainani: The `ShoppingAction` needs `productPageUrl` and the right expanded boolean to be constructed. For instance,

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/shopping/ShoppingState.kt
rahulsainani: Good question. For the case you mentioned, this works fine. The url we use is the `selectedTabUrl`.

I am concerned about the cases when the url changes because of some query params but the product id is the same, aka same pdp. This is the same for persisting reanalysis states, which also use product urls in a set. So far, we haven't encountered any issues with it. But if we do, we could use an object with vendor domain and product id instead. While working on persisting renalaysis, that was complicating it before we even encountered an issue with the url approach. Maybe I'm overthinking this, however, since we don't know all the different product pages, it's a bit hard to optimize this solution based on the current information. wdyt?

MozillaNoah: I also had a concern about pages changing the URL whenever the user clicks on site elements.

Yeah, this may be part of a larger conversation as this feature continues to iterate. Incorporating the Product ID could be a potential candidate. Tab ID could be an option as well; it just wouldn't cover the scenario that a user has the product open in more than tab. 

I'm cool with landing as-is, since we have this on our radar and have a couple hypothetical solutions in case issues do arise once this feature is truly in the wild.

MozillaNoah (approved):

PR #4231, opened 2023-10-25
Bug 1844523 - Translate and Restore to Original Engine Support

ohall-m:
This patch adds engine support for translate and restoring the page after a translation. `requestTranslate` and `requestTranslationRestore` are the Gecko engine session functions.

This corresponding actions for these functions are `TranslateAction`, and `TranslateRestoreAction`. The success and failure correspond to the actions `TranslateSuccessAction` and `TranslateExceptionAction`.

Sample usage is provided on reference browser. The sample is currently hardcoded to support Spanish to English page translations until the next API is introduced.

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1844523

PR #4465, opened 2023-11-13
Bug 1860844 - Add GeckoTranslateSessionDelegate to Android Components

ohall-m:
This bug adds a translation session delegate for receiving information from GeckoView on the state of the toolkit translation engine and also when to expect and offer translations on a given page.

It also updates reference browser to use auto-detected defaults for translating.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1860844


android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslateSessionDelegate.kt
```diff
@@ -0,0 +1,79 @@
...
+        var detectedLanguages = DetectedLanguages(
+            geckoTranslationsState?.detectedLanguages?.docLangTag,
+            geckoTranslationsState?.detectedLanguages?.isDocLangTagSupported,
+            geckoTranslationsState?.detectedLanguages?.userLangTag,
+        )
+        var pair = TranslationPair(
+            geckoTranslationsState?.requestedTranslationPair?.fromLanguage,
+            geckoTranslationsState?.requestedTranslationPair?.toLanguage,
+        )
+        var translationsState = GeckoTranslationState(
```
Amejia481, line 68: Nit: This  can be `val`

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/GeckoTranslationState.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+* The representation of the Gecko translations engine toolkit state.
+*
+* @property detectedLanguages Detected information about preferences and page information.
+* @property error If an error state occurred or an error was reported by toolkit.
+* @property isEngineReady If the toolkit translation engine is primed for use or will need to be loaded.
+* @property requestedTranslationPair The language pair to translate. Usually populated after first request.
+*/
+
+data class GeckoTranslationState(
```
Amejia481, line 16: nit: I think we have to omit the "Gecko" prefix, as concept data classes are engine agnostic, they are used to represent any data coming from any `Engine` implementation.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslateSessionDelegate.kt
```diff
@@ -0,0 +1,79 @@
...
+    ) {
+        var detectedLanguages = DetectedLanguages(
+            geckoTranslationsState?.detectedLanguages?.docLangTag,
+            geckoTranslationsState?.detectedLanguages?.isDocLangTagSupported,
+            geckoTranslationsState?.detectedLanguages?.userLangTag,
+        )
+        var pair = TranslationPair(
+            geckoTranslationsState?.requestedTranslationPair?.fromLanguage,
+            geckoTranslationsState?.requestedTranslationPair?.toLanguage,
+        )
```
Amejia481, line 67: Could `docLangTag`, `isDocLangTagSupported`, `userLangTag`, and `toLanguage` just be part of `TranslationState` class?  Could you help me to clarify why they need to be part of separate classes.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -843,6 +844,46 @@ sealed class ContentAction : BrowserAction() {
  * [BrowserAction] implementations related to translating a web content page.
  */
 sealed class TranslationsAction : BrowserAction() {
+    /**
+     * Indicates that the translations toolkit expects the user may want to translate the page on
```
Amejia481, line 848: nit: Maybe we should omit the uses of `toolkit` as this refers back to the Gecko specific implementation, browser state is engine agnostic.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -843,6 +844,46 @@ sealed class ContentAction : BrowserAction() {
...
+     * translations are an option.
+     *
+     * @property tabId The ID of the tab the [EngineSession] should be linked to.
+     */
+    data class TranslateOfferAction(
+        override val tabId: String,
+    ) : TranslationsAction(), ActionWithTab
+
+    /**
+     * Indicates the translation state of the Gecko translations toolkit on the given [tabId].
```
Amejia481, line 873: Same as https://github.com/mozilla-mobile/firefox-android/pull/4465/files#r1399758138 but with Gecko.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/GeckoTranslationState.kt
ohall-m: Thanks! We have a bit of a namespace collision here with `TranslationState` coming from [GeckoView](https://searchfox.org/mozilla-central/rev/a9cb8b07d4cf4290e693bc31ba23412d0f042cb8/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#942) versus the browser's [translation state](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt). 

 I'm open to any naming, but we need to decide based on that information. Some of the [original discussion is here too](https://github.com/mozilla-mobile/firefox-android/pull/4231#discussion_r1380306894).

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/GeckoTranslationState.kt
```diff
@@ -0,0 +1,21 @@
...
+    val detectedLanguages: DetectedLanguages? = null,
+    val error: String? = null,
+    val isEngineReady: Boolean? = false,
+    val requestedTranslationPair: TranslationPair? = null,
```
Amejia481, line 20: Could properties `DetectedLanguages` and `TranslationPair` could be part of the `GeckoTranslationState` class?

ohall-m: They could, but I think we may lose some of the conceptual lifecycle if we go that route.

 Main reasoning for current implementation is that this state change on those items only happens through the `onTranslationStateChange`.  I would like to make it conceptually clear that those items all share the same lifecycle. Open to this change though, just want to make sure it is the direction we'd like to go!

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslateSessionDelegate.kt
ohall-m: Possible, but would need a little more discussion we'd like to go that route! (More info in this [thread](https://github.com/mozilla-mobile/firefox-android/pull/4465#discussion_r1399765495 ).) Mostly just a decision if we should hide that those components have the same lifecycle or not.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/GeckoTranslationState.kt
Amejia481: Sorry, I'm not that familiar with the translation APIs do we have some technical docs where I can get more context about the flow of translations. From what you mentioned, it looks like we are could have two sources of request for translation, I would like to get more context about it 😅 ?

ohall-m: Thanks for taking a look! I really appreciate the review, esp. where there is so much context!

A translation can occur either:
1. If Fenix requests a translation or 
2. If the pref is set to automatically translate 

When the translation engine in toolkit does any operation it will call `onTranslationStateChange` to show a translation lifecycle event occurred. It gets called fairly regularly, maybe a couple of times per page load.

So, on an automatically translate call page load:
* Before Page load - onTranslationStateChange - no translation pair, most items will be null
* Page Load - onTranslationStateChange - will have a docLangTag of the detected language
* It translated -  onTranslationStateChange - will have the pair state it auto translated to

When Fenix calls translate, will see the same events coming from `onTranslationStateChange`; however, we will also have our progress [state set](https://searchfox.org/mozilla-mobile/rev/1eb55a6c369e7f208ce7c24934104b0fa24ac0ce/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt#16-17). (Test with some [examples](https://searchfox.org/mozilla-central/rev/a9cb8b07d4cf4290e693bc31ba23412d0f042cb8/mobile/android/geckoview/src/androidTest/java/org/mozilla/geckoview/test/TranslationsTest.kt#150) of the lifecycle too.)

Also, this made me thing of something that might influence how we organize this, `onTranslationStateChange` should probably also flip [`isTranslated`](https://searchfox.org/mozilla-mobile/rev/1eb55a6c369e7f208ce7c24934104b0fa24ac0ce/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt#15) when we see that an auto translate event occurred (criteria would be that we have a translation pair set, therefore it was translated). Which, makes the lifecycle a less separate. 

Let me know what state organization you'd prefer!

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslateSessionDelegate.kt
ohall-m: Ty, updated!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
ohall-m: Thanks, updated language from `toolkit` to `translations engine` and avoided Gecko specific language!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
ohall-m: Ty, updated!

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/GeckoTranslationState.kt
ohall-m: >I guess then we have two type of translations, one Automatically and one Requested by the user, right?

Yes, that is right!

Amejia481: We have a similar case for Prompt quests, where we have different type of prompts that shared common behaviour but each one has its particular properties. We modeled it creating the [PromptRequest.kt](https://github.com/mozilla-mobile/firefox-android/blob/1942ff65ff06690d8ed892ca3d514b3a6a6562d0/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/prompt/PromptRequest.kt#L28).

It has a `sealed class PromptRequest`  which is the base class for all the requests, and all other requests are children `data class`s, which has its own properties and meaning,  for example [Alert](https://github.com/mozilla-mobile/firefox-android/blob/1942ff65ff06690d8ed892ca3d514b3a6a6562d0/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/prompt/PromptRequest.kt#L76C5-L76C21).

Maybe having `sealed class TranslationRequest`, `data class AppTranslationRequest` and `data class EngineSuggestionTranslationRequest` 😅  could be an alternative? Also tagging @gabrielluong to hear his opinion.

ohall-m: Just want to add one point of clarification, [GeckoTranslationState](https://searchfox.org/mozilla-central/rev/a9cb8b07d4cf4290e693bc31ba23412d0f042cb8/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#942) isn't a translations request. It is more a packet of information on the Gecko toolkit translations engine state. So, for example, on page load, we will receive information that the page can be translated or not. `GeckoTranslationState` changes independently of a translations requests, but will update in response to a translation request.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/DetectedLanguages.kt
```diff
@@ -0,0 +1,20 @@
...
+* The representation of a translations detected document and user language.
+*
+* @property docLangTag The user's preferred language tag. Usually used for determining the best
+ * guess for translating "to".
+* @property isDocLangTagSupported If the translation engine supports the document language.
+* @property userLangTag The auto-detected language tag of page. Usually used for determining the
+ * best guess for translating "from".
+*/
+data class DetectedLanguages(
+    val docLangTag: String? = null,
```
Amejia481, line 17: nit: Maybe `userPreferredLanguageTag`? 😅

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/DetectedLanguages.kt
```diff
@@ -0,0 +1,20 @@
...
+    val isDocLangTagSupported: Boolean? = false,
```
Amejia481, line 18: nit: Maybe `supportDocumentLange`? 😅

Amejia481 (approved):
LGTM!

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/GeckoTranslationState.kt
ohall-m: Filed [1865919](https://bugzilla.mozilla.org/show_bug.cgi?id=1865919) as a follow up to refactor into a sealed class!

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/DetectedLanguages.kt
ohall-m: Thanks, updated, much clearer! (Also realized I had documentation switched 😅)

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/DetectedLanguages.kt
ohall-m: Updated!

PR #4489, opened 2023-11-15
Bug 1864759 - Add secret setting and nimbus flag for toolbar redesign

HarrisonOg:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864759

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
@@ -78,4 +78,14 @@ object FeatureFlags {
      * Enable Meta attribution.
      */
     val metaAttributionEnabled = Config.channel.isNightlyOrDebug
+
+    /**
+     * Enable Toolbar Redesign components and behaviors ready for Nightly.
+     */
+    val completeToolbarRedesignEnabled = Config.channel.isNightlyOrDebug
+
+    /**
+     * Enable Toolbar Redesign partial components and behaviors.
+     */
+    val incompleteToolbarRedesignEnabled = Config.channel.isDebug
 }
--- a/fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
@@ -1942,4 +1942,24 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = { enableFxSuggest },
         featureFlag = FeatureFlags.fxSuggest,
     )
+
+    /**
+     * Indicates if the user is shown new redesigned Toolbar UI.
+     */
+    var enableRedesignToolbar by lazyFeatureFlagPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_toolbar_use_redesign),
+        default = { false },
+        featureFlag = FeatureFlags.completeToolbarRedesignEnabled,
+    )
+
+    /**
+     * Indicates if the user is shown incomplete new redesigned Toolbar UI components and behaviors.
+     *
+     * DEV ONLY
+     */
+    var enableIncompleteToolbarRedesign by lazyFeatureFlagPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_toolbar_use_redesign_incomplete),
+        default = { false },
+        featureFlag = FeatureFlags.incompleteToolbarRedesignEnabled,
+    )
 }
--- a/fenix/app/src/main/res/values/preference_keys.xml
+++ b/fenix/app/src/main/res/values/preference_keys.xml
@@ -138,6 +138,10 @@
     <string name="pref_key_toolbar_top" translatable="false">pref_key_toolbar_top</string>
     <string name="pref_key_toolbar_bottom" translatable="false">pref_key_toolbar_bottom</string>
 
+    <!-- Toolbar Redesign -->
+    <string name="pref_key_toolbar_use_redesign" translatable="false">pref_key_toolbar_use_redesign</string>
+    <string name="pref_key_toolbar_use_redesign_incomplete" translatable="false">pref_key_toolbar_use_redesign_incomplete</string>
+
     <!-- Privacy Pop Window -->
     <string name="pref_key_privacy_pop_window" translatable="false">pref_key_privacy_pop_window</string>
 
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/RedesignToolbarFeature.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/RedesignToolbarFeature.kt
@@ -1,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components.toolbar
+
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * An abstraction for the Toolbar Redesign feature.
+ */
+interface RedesignToolbarFeature {
+
+    /**
+     * Returns true if the toolbar redesign feature is enabled.
+     */
+    val isEnabled: Boolean
+}
+
+/**
+ * The complete portions of the redesigned Toolbar ready for Nightly.
+ *
+ */
+class CompleteRedesignToolbarFeature(
+    private val settings: Settings,
+) : RedesignToolbarFeature {
+
+    override val isEnabled: Boolean
+        get() = settings.enableRedesignToolbar
+}
+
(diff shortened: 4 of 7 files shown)


fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1928,4 +1928,12 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = { enableFxSuggest },
         featureFlag = FeatureFlags.fxSuggest,
     )
+
+    /**
+     * Indicates if the user is shown new redesigned Toolbar UI.
+     */
+    var enableRedesignedToolbar by booleanPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_toolbar_use_redesign),
+        default = FxNimbus.features.toolbar.value().toolbarRedesignEnabled,
```
gabrielluong, line 1937: Should this be using `lazyFeatureFlagPreference` with a `featureFlag`? I am not fully sure if I am understanding shared preference myself, but from reading `lazyFeatureFlagPreference`, I believe while this is feature flagged off, then we will not be storing the `default` value into a user's shared preference and instead be using a dummy object that just returns false. 

If we used `booleanPreference` preference here, we will basically be setting `pref_key_toolbar_use_redesign` to be `false` by default for all users that has Fenix installed and receives this patch. Consequently, I think this will mean when we do enable this, we will have user with `pref_key_toolbar_use_redesign` already set to `false` since `toolbarRedesignEnabled` is false. I believe this is how it works, but I think I need someone to help me verify if what I am saying is actually true.

rahulsainani: GL is highlighting good points here. 

My 2 cents here would be to simply use nimbus flag directly via `RedesignToolbarFeature` and use the experiment settings to self enroll for users/testers who want to see this in action.

gabrielluong: > My 2 cents here would be to simply use nimbus flag directly via RedesignToolbarFeature and use the experiment settings to self enroll for users/testers who want to see this in action.

I would generally say we should have a secret setting pref that can enable the feature manually. Otherwise, you will also have to make a change to the nimbus flag during local development to enable the feature. We ran into the issue where we couldn't have UX review a design because they couldn't enable the feature themselves unless the nimbus experiment was running.

HarrisonOg: hello!
great points all around, I'm actually going to rework this to remove the nimbus flag for now and check out `lazyFeatureFlagPreference`.
The way this feature will roll out will be to Nightly users for a bit and let it soak while we measure performance metrics and then eventually roll up to Beta if behavior changes are not negative for search revenue.

rahulsainani: @gabrielluong We could default nimbus flag to true only for development in the config though.

For my own curiosity, if we use `lazyFeatureFlagPreference`, the nimbus flag still needs to be true right? So either the nimbus flag has to be set to true for nightly config, or there should be a nightly experiment from the beginning that testers/UX can turn on in the experiment settings?

From the experiment testing perspective, I think the latter one seems better since it tests the experiment config in nightly as well, which would be the same as the beta experiment config, wdyt?

gabrielluong: > @gabrielluong We could default nimbus flag to true only for development in the config though.

I think it depends on the state of the feature development. Only when the feature is ready to ship in Nightly should the feature be enabled by default in Debug and Nightly. Enabling it by default for debug builds in Nimbus mean we are no longer testing the state that is shipped to our users, which will lead to other issues such as we are no longer catching problems early ourselves during day-to-day development, and we will also have to programmatically turn the nimbus flag on/off if we wanted to see the old/new state. We may run to issues where we even accidentally commit this temporary code change (this has already happened before with Compose Top Sites).

I see 2 hard requirements:
- We need a way to toggle on/off a feature via secret settings so that we don't have to programmatically make a change and rebuild every time we are doing trying to test the different feature states.
- We need a preference that can be used for Nimbus experimentation that could be override by the secret settings. 
To obtain both of these requirements, we do no have to solely depend on using the same preference key. It shouldn't also be that we can only do or the another. We need to be able to do both so that we aren't programmatically toggling the preference which will affect our developer experience.

> For my own curiosity, if we use lazyFeatureFlagPreference, the nimbus flag still needs to be true right? So either the nimbus flag has to be set to true for nightly config, or there should be a nightly experiment from the beginning that testers/UX can turn on in the experiment settings?

The Nimbus and feature flag would need to be true. If we set the preference too early without `lazyFeatureFlagPreference`, we are still running to the original problem where `booleanPreference` sets the value on the first read. I think there is possibly a deeper rooted problem with how we are using preferences to store the Nimbus default value. From the experiment testing perspective, we should only ever read from the Nimbus value getter directly instead of storing that value in a preference because once we create the shared preference with the Nimbus default value, we are kinda stuck with that default value that has been determined by the treatment group placed upon by experimenter. When we do let a feature ride to all the channels that has been using Nimbus and these preferences, we need to make sure their existing preferences are no longer used or removed completely. Otherwise, we might be run into a scenario where an user was in the treatment group where the feature was turned off, but the preference used here will never be reset unless we also remove the preference.

mavduevskiy:
We had discussed the rollout of the feature with the UX team, and I summed the results [here](https://www.figma.com/file/8fsCvtnweBkWGZ9NfL907l?node-id=5503:276236&mode=design#582046598). What it means for us, probably is this. We want to:
1.  be able to enable stuff for us, developers, without interrupting work flow for others (dev build + secret settings);
2. be able to enable parts of the feature on nightly, without ability to disable new UI for the users;
3. be able to run the changes as an experiment, at some point in the future.

gabrielluong (approved):

fenix/app/src/main/res/values/preference_keys.xml
```diff
@@ -138,6 +138,11 @@
     <string name="pref_key_toolbar_top" translatable="false">pref_key_toolbar_top</string>
     <string name="pref_key_toolbar_bottom" translatable="false">pref_key_toolbar_bottom</string>
 
+    <!-- Toolbar Redesign -->
+    <string name="pref_key_toolbar_use_redesign" translatable="false">pref_key_toolbar_use_redesign</string>
+    <string name="pref_key_toolbar_use_redesign_incomplete" translatable="false">pref_key_toolbar_use_redesign_incomplete</string>
+
+
```
mavduevskiy, line 145: extra line

fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
```diff
@@ -78,4 +78,14 @@ object FeatureFlags {
...
+
+    /**
+     * Enable Toolbar Redesign components and behaviors ready for Nightly.
+     */
+    val completeToolbarRedesignEnabled = Config.channel.isNightlyOrDebug
+
+    /**
+     * Enable Toolbar Redesign partial components and behaviors.
+     */
+    const val incompleteToolbarRedesignEnabled = false
```
mavduevskiy, line 90: shouldn't we also enable it for dev build? I don't think it's going to work in combination with `var enableIncompleteToolbarRedesign` well. Though I might be mistaken

mavduevskiy (commented):
Looks good! a question about `incompleteToolbarRedesignEnabled`

fenix/app/src/main/java/org/mozilla/fenix/FeatureFlags.kt
HarrisonOg: good catch! switched to `channel.isDebug` as we want devs to turn it off and on.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -1942,4 +1942,24 @@ class Settings(private val appContext: Context) : PreferencesHolder {
...
+     * Indicates if the user is shown new redesigned Toolbar UI.
+     */
+    var enableRedesignToolbar by lazyFeatureFlagPreference(
+        key = appContext.getPreferenceKey(R.string.pref_key_toolbar_use_redesign),
+        default = { false },
+        featureFlag = FeatureFlags.completeToolbarRedesignEnabled,
+    )
+
+    /**
+     * Indicates if the user is shown incomplete new redesigned Toolbar UI components and behaviors.
```
gabrielluong, line 1956: @HarrisonOg @mavduevskiy I am wondering why we would want to have a separate settings for this? For developers, we would be able to see the WIP at any time by going to the secret settings with a single pref. At the moment, I would imagine the `enableRedesignToolbar` would also just as well represent the WIP toolbar redesign.

PR #4490, opened 2023-11-15
Bug 1857021 – Support delete webchannel message

mavduevskiy:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -67,21 +68,31 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
...
+            forceCloseAndClearData()
+        }
+    }
 
-                // Remove the device name when we log out.
-                context?.let {
-                    val deviceNameKey = it.getPreferenceKey(R.string.pref_key_sync_device_name)
-                    preferenceManager.sharedPreferences?.edit()?.remove(deviceNameKey)?.apply()
-                }
+    private fun forceCloseAndClearData() {
```
jonalmeida, line 75: nit: naming. Force close on Android has a different meaning (kill the app).

```suggestion
    private fun closeAndClearData() {
```

fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
```diff
@@ -67,21 +68,31 @@ class AccountSettingsFragment : PreferenceFragmentCompat() {
...
         }
     }
 
     override fun onResume() {
         super.onResume()
         showToolbar(getString(R.string.preferences_account_settings))
+
+        // The user might have deleted the account through the "manage account" tab, in which case
+        // the account settings fragment should not be shown.
+        if (accountManager.accountProfile() == null) {
```
jonalmeida, line 93: This is race-y. Please use an account observer or observe the `SyncStore`.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -539,6 +548,10 @@ open class FxaAccountManager(
             resetAccount()
             Event.Progress.LoggedOut
         }
+        ProgressState.DeletingAccount -> {
+            resetAccount()
```
jonalmeida, line 552: You may need to use/make a separate method. `resetAccount` calls `account.disconnect()` which wouldn't make sense to call if we are receiving the web channel response that the account is deleted.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/store/SyncStoreSupport.kt
```diff
@@ -113,6 +113,11 @@ internal class FxaAccountObserver(
         store.dispatch(SyncAction.UpdateSyncStatus(SyncStatus.LoggedOut))
         store.dispatch(SyncAction.UpdateAccount(null))
     }
+
+    override fun onAccountDeleted() {
+        store.dispatch(SyncAction.UpdateSyncStatus(SyncStatus.LoggedOut))
```
jonalmeida, line 118: If you introduce a new `SyncStatus` state, then you would know when an account is deleted and you can observe this `SyncStore` to know when to do the rest of the UI changes you have in Fenix. See the above comments for context as well.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -362,6 +362,8 @@ class Settings(private val appContext: Context) : PreferencesHolder {
     val shouldShowSecurityPinWarning: Boolean
         get() = secureWarningCount.underMaxCount()
 
+    var shouldShowAccountDeletedSnackbar: Boolean = false
```
jonalmeida, line 365: Same comment as the others, we may not need this if we are observing state changes.

jonalmeida (changes requested):
Left some comments on how we should react to state changes, but this direction looks good. :+1:

android-components/components/feature/accounts/src/main/java/mozilla/components/feature/accounts/FxaWebChannelFeature.kt
```diff
@@ -363,11 +370,33 @@ class FxaWebChannelFeature(
...
+                val uid = data.getString("uid")
+                if (uid.isNotEmpty() && uid == accountManager.accountProfile()?.uid) {
+                    CoroutineScope(Dispatchers.Main).launch {
+                        accountManager.onAccountDeleted()
+                    }
+                }
+
+                null
+            } catch (e: JSONException) {
+                // TODO ideally, this should log to Sentry.
```
jonalmeida, line 388: Let's leave a TODO with link to a filed bug if we're not going to do it in this patch, other it never gets done. :)

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -418,6 +418,12 @@ open class FxaAccountManager(
      */
     suspend fun logout() = withContext(coroutineContext) { processQueue(Event.Account.Logout) }
 
+    /**
+     * Should be called once the user deletes the account.
+     * It clears the account data and notifies the observers that account has been deleted.
+     */
+    suspend fun onAccountDeleted() = withContext(coroutineContext) { processQueue(Event.Account.Delete) }
```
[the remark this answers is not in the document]
mavduevskiy: Was debating on that myself, I don't like the current name because of the callback association, but I find `deleteAccount` to be a poor choice as well. `FxaAccountManager` doesn't delete the account (and can't), the deletion happened outside, and the manager reacts to the external event. Could we find a better naming? `handleAccountDeletion`, `processAccountDeletion`, `respondToAccountDeletion`?

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -481,9 +485,34 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         }
     }
 
+    private fun subscribeToAccountManager() {
+        requireComponents.backgroundServices.accountManagerAvailableQueue.runIfReadyOrQueue {
+            // By the time this code runs, we may not be attached to a context or have a view lifecycle owner.
+            if ((this@BrowserFragment).view?.context == null) {
```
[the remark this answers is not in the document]
mavduevskiy: I am not sure I understand the question, but since we are waiting for a queue, the fragment might be closed by the time it finishes. Is the approach wrong?

fenix/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
```diff
@@ -262,4 +262,9 @@ internal class TelemetryAccountObserver(
         SyncAuth.signOut.record(NoExtras())
         context.settings().signedInFxaAccount = false
     }
+
+    override fun onAccountDeleted() {
+        context.settings().signedInFxaAccount = false
+        context.settings().shouldShowAccountDeletedSnackbar = true
```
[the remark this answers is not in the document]
mavduevskiy: Could we rely on callbacks though in this case? Fragments where we need to display snackbar doesn't exist/are not active by the time accountManager changes it's state. The callback triggers navigation from the custom tab activity back to the app, and somehow we need to pass the knowledge that it's about time we show the message

fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
mavduevskiy: I don't think it is, the CustomTabActivity is closed only when the accountManager moves into loggedOut state, so there is no change the profile exist at this point.

Also, account observer in this fragment wouldn't catch the delete event, because the callback is called when the user is on CustomTabActivity

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
mavduevskiy: From the [doc](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/concept/sync/src/main/java/mozilla/components/concept/sync/OAuthAccount.kt#190):

     * Reset internal account state and destroy current device record.
     * Use this when device record is no longer relevant, e.g. while logging out. On success, other
     * devices will no longer see the current device in their device lists.
     
     Looks pretty relevant #3825

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/store/SyncStoreSupport.kt
mavduevskiy: I am still hesitant to use that approach, because of the life-cycle of fragments. events sent while the observer isn't subscribed are lost, right? we need somehow to register an account deleted status in a custom tab activity and then show snackbar in a different fragment, once the navigation happens.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
mavduevskiy: same concern as above

mavduevskiy:
Thanks for the review @jonalmeida! I addressed the more straightforward changes, and shared my concern with the proposed observer approach. I would be happy to hop on a call/pair up to settle the details if discussion ...intensifies.

I realized that tests are missing, but given that we are debating listener/flag approach I will wait for decision on that first, since it might affect the tests greatly.

mavduevskiy:
As you can see in diagram, when `onAccountDeleted` is triggered, we are in a different Activity, and we need a way to somehow pass down the information to `HomeFragment`/`BrowserFragment` to display the snackbar.

```mermaid
graph TD
    A[CustomTabActivity] -->|User deletes account| B[Account Deletion]
    B --> C[AuthorizationManager]
    C -->|Set flag| D[Settings]
    C -->|Finish Activity| E[CustomTabActivity Ends]
    E --> F[HomeActivity Starts]
    F --> G[HomeFragment]
    G -->|Check flag| H[Display Snackbar]
    H -->|Reset flag| I[Settings Updated]

    style A fill:#f96,stroke:#333,stroke-width:2px
    style B fill:#fc6,stroke:#333,stroke-width:2px
    style C fill:#9cf,stroke:#333,stroke-width:2px
    style D fill:#c6f,stroke:#333,stroke-width:2px
    style E fill:#f96,stroke:#333,stroke-width:2px
    style F fill:#f96,stroke:#333,stroke-width:2px
    style G fill:#9cf,stroke:#333,stroke-width:2px
    style H fill:#9f9,stroke:#333,stroke-width:2px
    style I fill:#c6f,stroke:#333,stroke-width:2px

**Comment by mavduevskiy** [comment]
If we were to create a middle man that would handle Snackbars in orderly manner, I see it as something holding a queue of Snackbar messages and handles display validation priority and timing, that Fragments could subscribe to and handle the side effects. In our case, an Activity would push a message into the queue, and fragments on Resume would receive a message to handle. Something looking like this
```mermaid
sequenceDiagram
    participant CustomTabActivity
    participant AuthorizationManager
    participant Settings
    participant HomeActivity
    participant HomeFragment
    participant Snackbar

    rect rgb(224, 224, 224)
    note right of CustomTabActivity: User deletes account via WebView in CustomTabActivity
    CustomTabActivity->>AuthorizationManager: Intercept "account_deleted" message
    AuthorizationManager->>Settings: Set "shouldShowAccountDeletedSnackbar" flag
    AuthorizationManager->>CustomTabActivity: Trigger "onAccountDeleted" callback
    CustomTabActivity->>HomeActivity: Finish and return to HomeActivity
    end

    rect rgb(224, 224, 224)
    note right of HomeActivity: HomeActivity becomes active
    HomeActivity->>HomeFragment: Initialize or resume fragment
    HomeFragment->>Settings: Check "shouldShowAccountDeletedSnackbar" flag
    Settings-->>HomeFragment: Return true
    HomeFragment->>Snackbar: Display "Account Deleted" Snackbar
    Snackbar-->>Settings: Reset "shouldShowAccountDeletedSnackbar" flag
    end

But I am not sure if it's worth it. Do we have any other places in the app where snackbars shouldn't be shown in a different place from where the user was when the even that needs UI feedback happened? Is so, such a system could be a follow up.

**Comment by mavduevskiy** [comment]
But I am not sure if it's worth it. Do we have any other places in the app where snackbars shouldn't be shown in a different place from where the user was when the even that needs UI feedback happened? Is so, such a system could be a follow up.

File: android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/store/SyncStoreSupport.kt (hunk 4 above)
Comment (jonalmeida) [reply android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/store/SyncStoreSupport.kt:118]: Since we're putting `Event.Account.Delete` in the same place as `Event.Account.Logout`, that would probably mean we should make it symmetrical on the `SyncStore`.

File: fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt [hunk 10]
```diff
@@ -481,9 +485,34 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
         }
     }
 
+    private fun subscribeToAccountManager() {
```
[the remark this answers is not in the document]
jonalmeida: See the comment below for an example.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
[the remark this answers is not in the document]
jonalmeida: In fact, if we use the `SyncStore`, then you don't even need the `runIfReadyOrQueue` because you would be observing the store and not eagerly initializing the `AccountManager`, so this way, you will always have the `Context` available (because that's when you start observing, and you will be able to see the state when the account manager is eventually initialized.

```kotlin
class AccountDeletionSnackbarObserver(
    private val containerView: View,
    private val anchorView: View,
    private val store: SyncStore,
) : AbstractBinding<SyncState>(store) {
    override suspend fun onState(flow: Flow<SyncState>) {
        flow.map { it.status }
            .distinctUntilChanged()
            .collectLatest { status ->
                if (status == SyncStatus.Deleted) {
                    FenixSnackbar.make(
                        view = containerView,
                        duration = FenixSnackbar.LENGTH_SHORT,
                        isDisplayedWithBrowserToolbar = true,
                    )
                        .setText(containerView.context.getString(R.string.snackbar_account_deleted))
                        .setAnchorView(anchorView)
                        .show()

                    // Consume the state so we don't keep showing the snackbar.
                    // Maybe this should be done somewhere else since changing the sync status should not be specific
                    // to the snackbar observer.
                    store.dispatch(SyncAction.UpdateSyncStatus(SyncStatus.NotInitialized))
                }
            }
    }
}
```

fenix/app/src/main/java/org/mozilla/fenix/components/BackgroundServices.kt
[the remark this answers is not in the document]
jonalmeida: The code example in [the comment above](https://github.com/mozilla-mobile/firefox-android/pull/4490/files#r1394720727) satisfies these concerns. Let me know what you think?

jonalmeida:
> But I am not sure if it's worth it. Do we have any other places in the app where snackbars shouldn't be shown in a different place from where the user was when the even that needs UI feedback happened? Is so, such a system could be a follow up.

I'm not aware of any that would work in this case. I left a code snippet above that I believe simplifies the flow.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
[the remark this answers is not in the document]
mavduevskiy: If we stick to this approach, the resetting of `SyncStatus` should happen after after the `Snackbar` is displayed, which doesn't sound right to me, because change of the state has nothing to do with the snackbar display. Adding a new `SyncStatus` might make sense for different cases, but in this particular one we still have a problem of knowing if we should show a snackbar or not when `status == SyncStatus.Deleted`

What I think is happening, we will be introducing a `SyncStatus.Deleted` status as a channel to pass down the account deleted event, which isn't the cleanest approach either. The cleaner approach would be to intoduce a `SnackbarManager`, that would queue the events and pass it down to subscribers to handle, the system that would receive an even from `AccountObserver.onAccountDeleted`, but wouldn't meddle with the `SyncStore` state. But since we are not going for it, why not just stick to a global flag for this case? Setting it to true and resetting after snackbar display seems like a cleaner approach than utilizing a new `SyncState` for that.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
mavduevskiy: When the account has been deleted, the device record is not longer relevent, so we should clear the device record.

fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
mavduevskiy: My thinking was, we [shouldn't](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt#62) be showing the fragment if the user is logged out or there is an Auth problem since it's closed automatically. But the fragment isn't active when the "accountDeleted" is called, so I inroduced a check during fragment creation. Isn't profile always available is the user in logged in state and there is no auth problem? We probable could do some other check to determine if the user is logged it.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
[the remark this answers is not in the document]
jonalmeida: > If we stick to this approach, the resetting of SyncStatus should happen after after the Snackbar is displayed, which doesn't sound right to me, because change of the state has nothing to do with the snackbar display.

Yes, this is the problem that I mentioned in my comment in the code snippet. We can still solve this though.

> What I think is happening, we will be introducing a SyncStatus.Deleted status as a channel to pass down the account deleted event, which isn't the cleanest approach either. The cleaner approach would be to intoduce a SnackbarManager, that would queue the events and pass it down to subscribers to handle, the system that would receive an even from AccountObserver.onAccountDeleted, but wouldn't meddle with the SyncStore state.

We need to update the SyncStore with the new deleted state regardless, because we want the Store to be the new way to observe the store and not use the AccountManager directly. So since we have to do that, we could then use it to for the snackbar case as well.

> But since we are not going for it, why not just stick to a global flag for this case? Setting it to true and resetting after snackbar display seems like a cleaner approach than utilizing a new SyncState for that.

We have established patterns of how we hold and observe state that we know works, so if we're going to create a new pattern, we should have strong reasons for it.

If you prefer using a global flag, then please put it in the AppStore (established pattern) and observe that. As mentioned above, we do still need to keep the SyncStore up-to-date.

fenix/app/src/main/java/org/mozilla/fenix/settings/account/AccountSettingsFragment.kt
jonalmeida: > Isn't profile always available is the user in logged in state and there is no auth problem?

No, it isn't, which is why you need to always observe for when it's available.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
jonalmeida: Correct, we do that with `accountStorage.clear()` locally. Why do we need to tell the server to delete the device record as well (after we received the account deletion)?

mhammond: ah, sorry, missed this ping - IIUC, the account deleted notification means the account *has been* deleted. I expect any attempts to do anything, including revoking the device record, would fail.

mhammond: (I haven't really looked at this patch - we *should* delete the device record on local disconnect.)

gabrielluong:
@mavduevskiy @jonalmeida @mhammond In preparation of the monorepo migration and cleaning up our pull request backlog, are we looking to continue making progress here or could this PR be closed and reopen in the new repo?

gabrielluong:
We have moved our firefox-android repo over to [mozilla-central](https://hg.mozilla.org/mozilla-central). Please feel free to reopen a new PR against the new repo. You will have to clone the new repository and reapply your changes. Thank you again, and sorry for the inconvenience.

For more information, please see https://github.com/mozilla-mobile/firefox-android/wiki#upcoming-migration-to-mozilla-central.

PR #4491, opened 2023-11-15
Bug 1864926 - Update AndroidX Browser to version 1.7.0

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864926


mcarare (approved):
The update only adds some functionality for custom tabs, so it should not have any effect on current code base.

PR #4493, opened 2023-11-15
Bug 1864692 - Create new startup-test Robo Test Task for Fenix/Focus w/opt-enabled Nightly

AaronMT:
https://bugzilla.mozilla.org/show_bug.cgi?id=1864692

See also discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1864043

This pull-request adds two new additional `Tasks` to  `startup-arm`. Added are a new Fenix and Focus Nightly specific`Task` that fetch the `signing-apk-fenix-nightly` and `signing-apk-focus-nightly` build artifact and initiates Robo Test for the purposes of Nightly sanity testing. These artifact builds are the ones with optimization (ProGuard/Minify) enabled (opt). These differ from existing `Task` in that the current fetch the non-optimized builds (`disable-optimization` is set in `build-apk`) [[here](https://github.com/mozilla-mobile/firefox-android/blob/5aac35da96f041c8f78358fd44e8af87cd35a46d/taskcluster/ci/build-apk/kind.yml#L254)]. 

Historically, we require having no optimization on the Instrumented builds as it leads to instances of ProGuard stripping required Classes and other stability issues (e.g, https://github.com/android/android-test/issues/1712) specifically with android-test and Instrumentation. A different method of Instrumentation is used in Robo Test so that testing with opt builds will work in this case (confirmed through Test Lab community and testing with our artifacts).

I've tested publishing the artifacts from `signing-apk-fenix-nightly` and `signing-apk-focus-nightly` to Test Lab to see how they perform, and they are good: [Fenix (Nightly)](https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.80a1a071cfb2770e/matrices/8902475991130342812) and [Focus Nightly](https://console.firebase.google.com/project/moz-focus-android/testlab/histories/bh.99d37f16502bd2e2/matrices/5686412163623215945).

We can see how this runs for a couple days and decide to either keep the existing non-opt tasks or not.

This PR  
* Copies the `robo-test.sh` script to Focus that we have in Fenix
* Defines two new `Tasks`
* Creates Flank configurations for both projects, Robo (5 minute timeout)

I'm using the `MediumPhone.arm` as it's the new recommendation from Firebase (in the new year they will deprecate `Pixel2.arm`).
https://bugzilla.mozilla.org/show_bug.cgi?id=1864692
https://bugzilla.mozilla.org/show_bug.cgi?id=1864692
https://bugzilla.mozilla.org/show_bug.cgi?id=1864692
https://bugzilla.mozilla.org/show_bug.cgi?id=1864692
https://bugzilla.mozilla.org/show_bug.cgi?id=1864692
https://bugzilla.mozilla.org/show_bug.cgi?id=1864692
https://bugzilla.mozilla.org/show_bug.cgi?id=1864692


AaronMT:
Haven't forgotten about this, (priority swap), but I'll rebase in Jan.

taskcluster/ci/startup-test/kind.yml
```diff
@@ -30,6 +30,29 @@ task-defaults:
     run-on-tasks-for: []
 
 tasks:
+    fenix-arm-nightly-robo-opt:
+        attributes:
+            build-type: fenix-nightly
+            shipping-product: fenix
+        description: Fenix Nightly Opt Robo Test (ARM)
+        dependencies:
+            signing: signing-apk-fenix-nightly
```
jonalmeida, line 39: Aaron and I have been talking about this offline: I'm not sure if this `signing-apk-fenix-nightly` task is something that is generated during runtime when a `signing-apk` task for the build type `fenix-nightly` is executed.

If so, then should be fine to depend on.

taskcluster/ci/startup-test/kind.yml
```diff
@@ -30,6 +30,29 @@ task-defaults:
...
+        run:
+            secrets:
+                - name: project/mobile/firefox-android/fenix/firebase
```
jonalmeida, line 42: Is it fine to use this firebase secret with a non-firebase task?

jonalmeida (approved):
Based on the thread in signing-apk, it might be fine to land this. So approving in advance in case we want to land this in nightly and see the results there.

taskcluster/ci/startup-test/kind.yml
AaronMT: These new tasks still require the same setup as others in that they require the same authentication to run on Firebase Test Lab, so yes.

taskcluster/ci/startup-test/kind.yml
AaronMT: I'll double check this week here (been swamped).

jonalmeida: Thanks for following up, this is great to hear!

PR #4494, opened 2023-11-15
Bug 1862536 - Add a Nimbus variable for Firefox Suggest providers.

linabutler:
This commit adds the `suggestion-provider.available-suggestion-types` Nimbus variable, which holds a list of all available suggestion types that the `FxSuggestSuggestionProvider` can return. This lets us toggle suggestion types remotely as part of our experiments.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1862536

### DIFF
--- a/android-components/components/feature/fxsuggest/fxsuggest.fml.yaml
+++ b/android-components/components/feature/fxsuggest/fxsuggest.fml.yaml
@@ -1,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+about:
+  description: Nimbus Feature Manifest for the Firefox Suggest feature.
+  android:
+    package: mozilla.components.feature.fxsuggest
+    class: .FxSuggestNimbus
+channels:
+  - debug
+  - release
+features:
+  awesomebar-suggestion-provider:
+    description: Configuration for the Firefox Suggest awesomebar suggestion provider.
+    variables:
+      available-suggestion-types:
+        description: >
+          A map of suggestion types to booleans that indicate whether or not the provider should
+          return suggestions of those types.
+        type: Map<SuggestionType, Boolean>
+        default: {
+          "amp": false,
+          "wikipedia": true,
+        }
+enums:
+  SuggestionType:
+    description: The type of a Firefox Suggest search suggestion.
+    variants:
+      amp:
+        description: A Firefox Suggestion from adMarketplace.
+      wikipedia:
+        description: A Firefox Suggestion for a Wikipedia page.
--- a/android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
+++ b/android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
@@ -48,10 +48,14 @@ class FxSuggestSuggestionProvider(
             emptyList()
         } else {
             val providers = buildList() {
-                if (includeSponsoredSuggestions) {
+                val availableSuggestionTypes = FxSuggestNimbus.features
+                    .awesomebarSuggestionProvider
+                    .value()
+                    .availableSuggestionTypes
+                if (includeSponsoredSuggestions && availableSuggestionTypes[SuggestionType.AMP] == true) {
                     add(SuggestionProvider.AMP)
                 }
-                if (includeNonSponsoredSuggestions) {
+                if (includeNonSponsoredSuggestions && availableSuggestionTypes[SuggestionType.WIKIPEDIA] == true) {
                     add(SuggestionProvider.WIKIPEDIA)
                 }
             }
--- a/android-components/.buildconfig.yml
+++ b/android-components/.buildconfig.yml
@@ -818,8 +818,10 @@ projects:
     - feature-session
     - lib-publicsuffixlist
     - lib-state
+    - service-nimbus
     - support-base
     - support-ktx
+    - support-locale
     - support-test
     - support-utils
     - tooling-lint
--- a/android-components/components/feature/fxsuggest/build.gradle
+++ b/android-components/components/feature/fxsuggest/build.gradle
@@ -4,6 +4,22 @@
 
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
+buildscript {
+    repositories {
+        maven {
+            url "https://maven.mozilla.org/maven2"
+        }
+    }
+
+    dependencies {
+        classpath "${ApplicationServicesConfig.groupId}:tooling-nimbus-gradle:${ApplicationServicesConfig.version}"
+    }
+}
+
+plugins {
+    id "com.jetbrains.python.envs" version "$python_envs_plugin"
+}
+
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 
@@ -36,6 +52,7 @@ dependencies {
     implementation project(':concept-awesomebar')
     implementation project(':concept-engine')
     implementation project(':feature-session')
+    implementation project(':service-nimbus')
     implementation project(':support-base')
     implementation project(':support-ktx')
 
@@ -54,4 +71,17 @@ dependencies {
 
 apply from: '../../../android-lint.gradle'
 apply from: '../../../publish.gradle'
+apply plugin: "org.mozilla.appservices.nimbus-gradle-plugin"
+nimbus {
+    // The path to the Nimbus feature manifest file
+    manifestFile = "fxsuggest.fml.yaml"
+
+    channels = [
+        debug: "debug",
+        release: "release",
+    ]
+
+    applicationServicesDir = gradle.hasProperty('localProperties.autoPublish.application-services.dir')
+        ? gradle.getProperty('localProperties.autoPublish.application-services.dir') : null
+}
 ext.configurePublish(config.componentsGroupId, archivesBaseName, project.ext.description)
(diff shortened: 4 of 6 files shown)


android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
```diff
@@ -18,16 +18,17 @@ import java.util.UUID
...
  */
 class FxSuggestSuggestionProvider(
     private val resources: Resources,
     private val loadUrlUseCase: SessionUseCases.LoadUrlUseCase,
-    private val includeSponsoredSuggestions: Boolean,
-    private val includeNonSponsoredSuggestions: Boolean,
+    private val providers: List<SuggestionProvider> = listOf(
+        SuggestionProvider.AMP,
+        SuggestionProvider.WIKIPEDIA,
+    ),
```
MatthewTighe, line 31: Do we intentionally want to include a sponsored provider as a default here?

MatthewTighe (approved):
Had one question, but this otherwise LGTM!

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
linabutler: Oooh, no, good point! Apart from a few tests, we're always passing a list of `providers`, so we can drop the default entirely.

linabutler:
I didn't think this through yesterday 😅 The logic I wrote will pass the contents of the Nimbus variable through, but we actually want to pass the _intersection_ of the current `SearchProviderState` and the value of the Nimbus variable.

IOW, the Nimbus variable functions as an allowlist, of all the Firefox Suggest suggestion types that we _can_ show; but the list of `providers` that we pass to `FxSuggestSuggestionProvider` is the intersection of the user's settings with that allowlist. Here are a couple of truth tables to illustrate:

|                                             | `FxSuggestProvider.AMP` in `FxNimbus.features.fxSuggest.value().providers` | `FxSuggestProvider.AMP` not in `FxNimbus.features.fxSuggest.value().providers` | 
|---------------------------------------------|----------------------------------------------------------------------------|--------------------------------------------------------------------------------|
| `state.showSponsoredSuggestions` is `true`  | add `SuggestionProvider.AMP` to the list of providers to query             |                                                                                |
| `state.showSponsoredSuggestions` is `false` |                                                                            |                                                                                |

|                                                | `FxSuggestProvider.WIKIPEDIA` in `FxNimbus.features.fxSuggest.value().providers` | `FxSuggestProvider.WIKIPEDIA` not in `FxNimbus.features.fxSuggest.value().providers` | 
|------------------------------------------------|----------------------------------------------------------------------------------|--------------------------------------------------------------------------------------|
| `state.showNonSponsoredSuggestions` is `true`  | add `SuggestionProvider.WIKIPEDIA` to the list of providers to query             |                                                                                      |
| `state.showNonSponsoredSuggestions` is `false` |                                                                                  |                                                                                      |

(Side note, Android Studio has great Markdown table formatting! 🤩)

Right now, the "show sponsored suggestions" switch means just `SuggestionProvider.AMP`, and the "non-sponsored" suggestions switch means `SuggestionProvider.WIKIPEDIA`—but the exact set of suggestions controlled by those two switches might change in the future.

For example, we might add `SuggestionProvider.POCKET`, `SuggestionProvider.AMO`, and `SuggestionProvider.MDN` to the set of non-sponsored suggestions eventually, or even a hypothetical future where we have `SuggestionProvider.MOZILLA_ADS` as another sponsored provider.

I force-pushed a commit with a fix for the logic—but I'm wondering if there's actually a better way to implement it? My hangup with what I've pushed is that it feels like `AwesomeBarView` is the wrong level of abstraction for mapping `state.show{Non}SponsoredSuggestions` to the list of `mozilla.appservices.suggest.SuggestionProvider`s. But maybe it actually is the right level?

WDYT @MatthewTighe? (@jonalmeida, I'd love your thoughts, too; I know you've done lots with Nimbus as well!) Am I overthinking this? 😅

My other thought was that we could add an `fxsuggest.fml.yaml` manifest to the `feature-fxsuggest` component, and take advantage of [Nimbus imports](https://experimenter.info/fml-spec/#linking-files-from-other-components) to import it into Fenix's manifest—like we do for GeckoView [here](https://github.com/mozilla-mobile/firefox-android/blob/f8049f17f006fed2614f29a2c82af8d773b527c0/android-components/components/browser/engine-gecko/geckoview.fml.yaml) and [here](https://github.com/mozilla-mobile/firefox-android/blob/f8049f17f006fed2614f29a2c82af8d773b527c0/fenix/app/nimbus.fml.yaml#L135-L143)). Then the `providers` Nimbus variable could live in `fxsuggest.fml.yaml`, we could keep the `include{Non}SponsoredSuggestions` Boolean options for `FxSuggestSuggestionProvider`, and `FxSuggestSuggestionProvider` could figure out what providers those Booleans map to. But maybe that's too complicated, for no clear win... 🤔

linabutler:
(This isn't urgent for 121, and since we're in the soft freeze—and I'm heading out on holiday tomorrow! 🦃—there's no hurry from my side to land this 😊)

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
```diff
@@ -18,16 +19,14 @@ import java.util.UUID
...
+ * @param providers The [SuggestionProvider]s to query for search suggestions.
  * @param suggestionsHeader An optional header title for grouping the returned suggestions.
  * @param contextId The contextual services user identifier, used for telemetry.
  */
 class FxSuggestSuggestionProvider(
     private val resources: Resources,
     private val loadUrlUseCase: SessionUseCases.LoadUrlUseCase,
-    private val includeSponsoredSuggestions: Boolean,
-    private val includeNonSponsoredSuggestions: Boolean,
+    @get:VisibleForTesting val providers: List<SuggestionProvider>,
```
MatthewTighe, line 29: Honestly, I don't see a real reason to not just let this be public since the type is an immutable type and the property is immutable as well.

MatthewTighe (commented):
I agree that `AwesomeBarView` is probably not the ideal abstraction level. I like the idea of an fml specific to the component since it localizes the Nimbus configuration and logic to the feature. I'm not familiar with that setup really, but given that there is an existing GV and "messaging" example I don't see any reason not to pursue that.

That said, I'm also happy to stick with what's here for now if that's easiest. It would be good to file a follow-up on record if that ends up being the decision

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
linabutler: Thanks—I reverted this change as part of adding the component-specific FML, but I'll keep it in mind for the future!

linabutler:
A couple of copypastas and [build quirks](https://mozilla-hub.atlassian.net/browse/EXP-4097) later, I was able to add the component-specific manifest! 🎉 It ended up being a small change, and I think it's actually a bit easier to understand.

@jhugman Could you please double-check that I added the component-specific manifest correctly? I copied how `messaging.fml.yaml` and `geckoview.fml.yaml` did it, but would love an extra pair of eyes to make sure it's good! 😊 Thanks!

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestSuggestionProvider.kt
```diff
@@ -48,10 +48,11 @@ class FxSuggestSuggestionProvider(
             emptyList()
         } else {
             val providers = buildList() {
-                if (includeSponsoredSuggestions) {
+                val availableSuggestionTypes = FxSuggestNimbus.features.suggestionProvider.value().availableSuggestionTypes
+                if (includeSponsoredSuggestions && availableSuggestionTypes.contains(SuggestionType.AMP)) {
                     add(SuggestionProvider.AMP)
                 }
-                if (includeNonSponsoredSuggestions) {
+                if (includeNonSponsoredSuggestions && availableSuggestionTypes.contains(SuggestionType.WIKIPEDIA)) {
```
MatthewTighe, line 55: Yeah, this definitely looks like the right call to me. Thanks for putting it together!

MatthewTighe (approved):
Looks good from my perspective, but I think it makes sense to let James take a quick look over the Nimbus configuration before landing.

jhugman (approved):
This is lovely to see.

I had a couple of questions which we resolved in Slack. So: I'd suggest you use `available-suggestions` with type `Map<SuggestionType: Boolean>`. This will mean you can flip individual suggestions on or off independently.

Thank you!

android-components/components/feature/fxsuggest/fxsuggest.fml.yaml
```diff
@@ -0,0 +1,29 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+about:
+  description: Nimbus Feature Manifest for the Firefox Suggest feature.
```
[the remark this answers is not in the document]
linabutler: Will do! Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1867691 as a follow-up to add the metadata for both this variable, and the `fx-suggest` one in Fenix's main FML file.

PR #4503, opened 2023-11-16
Bug 1865086 - More FxA error reporting

bendk:
Include the exception message when reporting
 FxaDeviceConstellationException.EnsureCapabilitiesFailed
 (https://bugzilla.mozilla.org/show_bug.cgi?id=1865086)

 Report restoring an FxA account that's not connected
 (https://bugzilla.mozilla.org/show_bug.cgi?id=1864667)

 Report a breadcrumb if we queue multiple auth recovery attempts at once
 (https://bugzilla.mozilla.org/show_bug.cgi?id=1864994)

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1865086

### DIFF
--- a/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/AccountStorage.kt
+++ b/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/AccountStorage.kt
@@ -7,6 +7,7 @@ package mozilla.components.service.fxa
 import android.content.Context
 import android.content.SharedPreferences
 import androidx.annotation.VisibleForTesting
+import mozilla.appservices.fxaclient.FxaRustAuthState
 import mozilla.components.concept.base.crash.CrashReporting
 import mozilla.components.concept.sync.AccountEvent
 import mozilla.components.concept.sync.AccountEventsObserver
@@ -160,7 +161,16 @@ internal class SharedPrefAccountStorage(
             ?: return null
 
         // May throw a generic FxaException if it fails to process saved JSON.
-        return FirefoxAccount.fromJSONString(savedJSON, crashReporter)
+        val account = FirefoxAccount.fromJSONString(savedJSON, crashReporter)
+        val state = account.getAuthState()
+        if (state != FxaRustAuthState.CONNECTED && crashReporter != null) {
+            crashReporter.submitCaughtException(
+                AbnormalAccountStorageEvent.RestoringNonConnectedAccount(
+                    "Restoring account from an unexpected state: $state",
+                ),
+            )
+        }
+        return account
     }
 
     override fun write(accountState: String) {
@@ -185,11 +195,12 @@ internal class SharedPrefAccountStorage(
 /**
  * A base class for exceptions describing abnormal account storage behaviour.
  */
-internal abstract class AbnormalAccountStorageEvent : Exception() {
+internal abstract class AbnormalAccountStorageEvent(message: String? = null) : Exception(message) {
     /**
      * Account state was expected to be present, but it wasn't.
      */
-    internal class UnexpectedlyMissingAccountState : AbnormalAccountStorageEvent()
+    internal class UnexpectedlyMissingAccountState(message: String? = null) : AbnormalAccountStorageEvent(message)
+    internal class RestoringNonConnectedAccount(message: String? = null) : AbnormalAccountStorageEvent(message)
 }
 
 /**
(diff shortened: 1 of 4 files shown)


android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/AccountStorage.kt
```diff
@@ -190,6 +196,10 @@ internal abstract class AbnormalAccountStorageEvent : Exception() {
      * Account state was expected to be present, but it wasn't.
      */
     internal class UnexpectedlyMissingAccountState : AbnormalAccountStorageEvent()
+    internal class RestoringNonConnectedAccount(val state: FxaRustAuthState) : AbnormalAccountStorageEvent() {
+        override val message: String
+            get() = "AbnormalAccountStorageEvent.RestoringNonConnectedAccount($state)"
+    }
```
bendk, line 202: @mavduevskiy Is this the right way to get message strings included in the Sentry report?

mavduevskiy: It is certainly a way to get it there! Though one could argue that a more canonical approach for Java/Kotlin throwables would be passing a message into the constructor. We also don't really use `state` as far as I could tell, so 

```
internal class RestoringNonConnectedAccount(override val message: String) : AbnormalAccountStorageEvent()
```
used as
```
crashReporter?.submitCaughtException(
                    AbnormalAccountStorageEvent.RestoringNonConnectedAccount(
                        "AbnormalAccountStorageEvent.RestoringNonConnectedAccount($state)",
                    )
                )
```
could be a way to go.

bendk: I swear I saw a message from you about this, but I don't see it anymore.  In any case, I changed this to use pass a `message` parameter to the exception constructor instead.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -449,10 +455,16 @@ open class FxaAccountManager(
     internal suspend fun encounteredAuthError(
         operation: String,
         errorCountWithinTheTimeWindow: Int = 1,
-    ) = withContext(coroutineContext) {
-        processQueue(
-            Event.Account.AuthenticationError(operation, errorCountWithinTheTimeWindow),
-        )
+    ) {
+        if (authRecoveryScheduled.getAndSet(true)) {
+            crashReporter?.recordCrashBreadcrumb(Breadcrumb("multiple auth recoveries scheduled at once"))
```
mavduevskiy, line 460: How do we know that multiple auth recoveries were scheduled at this point?  The `encounteredAuthError` function is the only place where `authRecoveryScheduled` is set to true. Also, should we rename it `authRecoveryScheduled` to `multipleAuthRecoveryScheduled` if it's the intention of the flag?

mavduevskiy (changes requested):
Looking good! A few changes to make sure that we are passing down the message to the stack trace (should go all the way into the Exception constructor). And a question about `authRecoveryScheduled` usage.

bendk:
Thanks for the help with this one.  I implemented all the suggestions, except the auth check one which I tried to improve a different way.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -127,6 +129,19 @@ open class FxaAccountManager(
     @Volatile
     private var latestAuthState: String? = null
 
+    // Used to detect multiple auth recovery attempts at once
+    // (https://bugzilla.mozilla.org/show_bug.cgi?id=1864994)
+    private var authRecoveryScheduled = AtomicBoolean(false)
+    private fun willScheduleAuthRecovery() {
```
mavduevskiy, line 135: This is much cleaner way to organize the code!

mavduevskiy (approved):
This looks good! Approving.

One thing to consider is the naming of `willScheduleAuthRecovery` funciton, that sounds like it will schedule the recovery, but it actually checks for multiple scheduled recovers and reports to crash reporter if it's true. My two cents, something like `checkForMultipleRequiveryCalls` would be a better fit.

bendk:
Sounds good, I updated the name and added the `needs-landing` tag.

PR #4520, opened 2023-11-17
Bug 1864070 - Add secret setting to enable the debug drawer

MozillaNoah:
After some iteration of exploring the use of the legacy Preferences API + AppState, I've opted to use DataStore here instead. With this approach, we'll have an API that's easy to consume in Compose while also not introducing a huge hack into SecretSettingsFragment in order to handle our use case of an observable preference.

[DataStore is currently being leveraged for Pocket](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/datastore/DataStores.kt), so there is no library addition here.

Hopefully, this will serve as a potential migration path for the rest of our settings. 

In order to get the secret settings into the debug drawer, I'll likely need to convert all of them to Compose + DataStore.

### Pull Request checklist

### After merge

4. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864070

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
@@ -6,15 +6,21 @@ package org.mozilla.fenix.settings
 
 import android.os.Bundle
 import androidx.core.content.edit
+import androidx.lifecycle.lifecycleScope
 import androidx.navigation.fragment.findNavController
 import androidx.preference.EditTextPreference
 import androidx.preference.Preference
 import androidx.preference.PreferenceFragmentCompat
 import androidx.preference.SwitchPreference
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.launch
 import org.mozilla.fenix.BuildConfig
 import org.mozilla.fenix.Config
 import org.mozilla.fenix.FeatureFlags
 import org.mozilla.fenix.R
+import org.mozilla.fenix.debugsettings.data.debugDrawerEnabled
+import org.mozilla.fenix.debugsettings.data.debugSettings
+import org.mozilla.fenix.debugsettings.data.updateDebugDrawerEnabled
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.ext.settings
@@ -96,6 +102,23 @@ class SecretSettingsFragment : PreferenceFragmentCompat() {
             isVisible = true
             isChecked = context.settings().feltPrivateBrowsingEnabled
             onPreferenceChangeListener = SharedPreferenceUpdater()
+        }
+
+        lifecycleScope.launch {
+            val debugDataStore = requireContext().debugSettings
+
+            // During initial development, this will only be available in Nightly or Debug builds.
+            requirePreference<SwitchPreference>(R.string.pref_key_enable_debug_drawer).apply {
+                isVisible = Config.channel.isNightlyOrDebug
+                isChecked = debugDataStore.debugDrawerEnabled.first()
+                onPreferenceChangeListener =
+                    Preference.OnPreferenceChangeListener { _, newValue ->
+                        lifecycleScope.launch {
+                            debugDataStore.updateDebugDrawerEnabled(enabled = newValue as Boolean)
+                        }
+                        true
+                    }
+            }
         }
 
         // for performance reasons, this is only available in Nightly or Debug builds
--- a/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/data/DebugSettingsDataStore.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/data/DebugSettingsDataStore.kt
@@ -1,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.debugsettings.data
+
+import android.content.Context
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import androidx.datastore.preferences.core.booleanPreferencesKey
+import androidx.datastore.preferences.core.edit
+import androidx.datastore.preferences.preferencesDataStore
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+
+/**
+ * [DataStore] for accessing debugging settings.
+ */
+val Context.debugSettings: DataStore<Preferences> by preferencesDataStore(name = "debug_settings")
+
+private val DEBUG_DRAWER_ENABLED = booleanPreferencesKey("debug_drawer_enabled")
+
+/**
+ * Updates whether the debug drawer is enabled.
+ *
+ * @param enabled Whether the debug drawer is enabled.
+ */
+suspend fun DataStore<Preferences>.updateDebugDrawerEnabled(enabled: Boolean) {
+    edit { preferences ->
+        preferences[DEBUG_DRAWER_ENABLED] = enabled
+    }
+}
+
+/**
+ * [Flow] for checking whether the Debug Drawer is enabled.
+ */
+val DataStore<Preferences>.debugDrawerEnabled: Flow<Boolean>
+    get() = data.map { preferences ->
+        preferences[DEBUG_DRAWER_ENABLED] ?: false
+    }
--- a/fenix/app/src/main/res/values/preference_keys.xml
+++ b/fenix/app/src/main/res/values/preference_keys.xml
@@ -78,6 +78,7 @@
     <string name="pref_key_home_blocklist">pref_key_home_blocklist</string>
     <string name="pref_key_hidden_engines_restored" translatable="false">pref_key_hidden_engines_restored</string>
     <string name="pref_key_enable_fxsuggest" translatable="false">pref_key_fxsuggest_enabled</string>
+    <string name="pref_key_enable_debug_drawer" translatable="false">pref_key_enable_debug_drawer</string>
 
     <!-- Data Choices -->
     <string name="pref_key_telemetry" translatable="false">pref_key_telemetry</string>
--- a/fenix/app/src/main/res/values/static_strings.xml
+++ b/fenix/app/src/main/res/values/static_strings.xml
@@ -74,6 +74,8 @@
     <string name="preferences_debug_settings_toolbar_redesign" translatable="false">Enable Toolbar Redesign incomplete portions</string>
     <!-- Label for enabling Felt Privacy -->
     <string name="preferences_debug_felt_privacy" translatable="false">Enable Felt Privacy</string>
+    <!-- Label for enabling the Debug Drawer -->
+    <string name="preferences_debug_settings_debug_drawer" translatable="false">Enable Debug Drawer</string>
 
     <!-- A secret menu option in the tabs tray for making a tab inactive for testing. -->
     <string name="inactive_tabs_menu_item">Make inactive</string>
--- a/fenix/app/src/main/res/xml/secret_settings_preferences.xml
+++ b/fenix/app/src/main/res/xml/secret_settings_preferences.xml
@@ -45,6 +45,11 @@
         app:iconSpaceReserved="false"
         android:title="@string/preferences_debug_felt_privacy"
         />
+    <SwitchPreference
+        android:defaultValue="false"
+        android:key="@string/pref_key_enable_debug_drawer"
+        android:title="@string/preferences_debug_settings_debug_drawer"
+        app:iconSpaceReserved="false" />
     <EditTextPreference
         android:key="@string/pref_key_custom_glean_server_url"
         android:title="@string/preferences_debug_settings_custom_glean_server_url"


fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
```diff
@@ -86,6 +87,18 @@ class SecretSettingsFragment : PreferenceFragmentCompat() {
...
 
+        requirePreference<SwitchPreference>(R.string.pref_key_enable_debug_drawer).apply {
+            isVisible = Config.channel.isNightlyOrDebug
+            isChecked = context.settings().enableDebugDrawer
+            onPreferenceChangeListener =
+                Preference.OnPreferenceChangeListener { _, newValue ->
+                    context.settings().enableDebugDrawer = newValue as Boolean
+                    context.components.appStore.dispatch(AppAction.UpdateDebugDrawerEnabled(newValue))
+
+                    true
```
MatthewTighe, line 98: IMO if we are going to manage this state in a Store, then a middleware should be handling the `Settings` side-effects

MatthewTighe (commented):
> 1. Does it make sense to have debugging state inside of AppState? Do we already have precedence for this?

I'm not seeing that we do have precedence for it. At first glance, my intuition is telling me that a separate Store for the Debug Drawer would be more appropriate. I think this could potentially benefit from prototyping though. The open questions in my mind are things like: what kind of other state would live in it? How would that state be used? 

A few cases that seem like they could provide interesting learnings, which coincidentally include your existing plans:
1. the tabs management you're already planning. This will naturally need to have an effect on the browser store, and experimenting with the communication between the two components seems informative
2. Overwriting secret settings. This begs the question of where these preferences live - now they're accessed directly, but do we want to move towards a world where our Stores provide the interface to our persisted properties (my answer would be yes, but then we have more work to do)?
3. Nimbus features. This demonstrates a similar problem to `2`, in that we'll need to explore how we manipulate the existing source of truth with the debug drawer.

Tangentially, creating a separate Store may make it easier to disable when not in use. And if we ever move towards Scoped Stores, it may be easier to combine disparate stores than try to split out a monolith into the appropriate scopes (this is very hypothetical opinion).

> 2. If we do want to have a debugging state inside of AppState, what would be the best way to manage this? Right now, we might only need a boolean for the debug drawer, but we could easily foresee a future where more debugging state information is kept at the app level for whatever purpose. In that case, should we instead have a substate, a DebugState if you will, similar to ShoppingState and WallpaperState?

I think if we wanted to include it in AppStore for now, than it should absolutely be encapsulated into a distinct `DebugState`. In general I think the AppStore needs to broken apart a bit more, either into more distinct states (Pocket) or even more Stores (HomeFragment).

MozillaNoah:
After discussing with @gabrielluong, I've just pushed a new version of the draft which utilizes `DataStore` instead of shared preferences. An `AppAction` triggers the data store write. 

Depending how this iteration goes. I may need to port the rest of the secret settings to `DataStore` in order to create the Composified UI in the debug drawer.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
```diff
@@ -234,6 +234,8 @@ internal object AppStoreReducer {
         )
 
         is AppAction.ShoppingAction -> ShoppingStateReducer.reduce(state, action)
+
+        is AppAction.UpdateDebugDrawerEnabled -> state.copy() // consumed by preferences middleware
```
MatthewTighe, line 238:
```suggestion
        is AppAction.UpdateDebugDrawerEnabled -> state // consumed by preferences middleware
```

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
```diff
@@ -28,6 +34,11 @@ class SecretSettingsFragment : PreferenceFragmentCompat() {
     }
 
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
+        var debugDrawerEnabled = false
+        lifecycleScope.launch {
+            debugDrawerEnabled = context?.debugSettings?.debugDrawerEnabledFlow?.first() ?: false
```
MatthewTighe, line 39: Do we want to collect the flow rather than just take the value out of it so that we get reactivity?

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/PreferencesMiddleware.kt
```diff
@@ -0,0 +1,47 @@
...
+    private val debugSettingsDataStore: DataStore<Preferences>,
+    private val scope: CoroutineScope = CoroutineScope(Dispatchers.IO),
+) : Middleware<AppState, AppAction> {
+
+    override fun invoke(
+        context: MiddlewareContext<AppState, AppAction>,
+        next: (AppAction) -> Unit,
+        action: AppAction,
+    ) {
+        handleAction(action)
```
MatthewTighe, line 32: Since we can probably assume that the actions here are all disk writes, it might help convey some semantic value to do `scope.launch { handleAction(action) }` and make `handleAction` suspending. It would send a message to future readers that `handleAction` doesn't block `next`.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -262,4 +262,11 @@ sealed class AppAction : Action {
             val expanded: Boolean,
         ) : ShoppingAction()
     }
+
+    /**
+     * Action used to update the visibility of the Debug Drawer trigger.
+     *
+     * @property enabled Whether the debug drawer has been enabled as part of this Action.
+     */
+    data class UpdateDebugDrawerEnabled(val enabled: Boolean) : AppAction()
```
MatthewTighe, line 271: If we aren't keeping the enabled state in the Store, I'm not sure if we need to use an action/middleware to control the side-effects

MatthewTighe (commented):
Reflecting more on my earlier comment:

> At first glance, my intuition is telling me that a separate Store for the Debug Drawer would be more appropriate.

I think that I may have been wrong there. I'm not sure it makes sense to keep a 3rd global store around, since there seems to be a high likelihood that we will want to pass it into various components that we're already passing the `AppStore`.

So I think two options could be:
1. `DebugState` as part of `AppState`. I like that this is in keeping with the rest of our architecture, and it could even give us a bit of a playground to test future patterns.
2. Directly accessing debug state through the datastore helpers. I like that this separates code that governs "internal" behavior and user-facing behavior. 

I am maybe a little more inclined towards the first option at this point, but I'll leave it up to you what you want to try first. There could be other options too, I'm not sure that's exhaustive. I think it could be good to get a second opinion here, since this is having a bit more of an impact on discussions of architecture.

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
MozillaNoah: I only need the one-time read, since the UI will update correctly with the user's input

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
MozillaNoah: That's fair. Since this doesn't need any other side effects, the Action/Middleware is a little heavy handed.

MozillaNoah:
> So I think two options could be:
> 
>     1. `DebugState` as part of `AppState`. I like that this is in keeping with the rest of our architecture, and it could even give us a bit of a playground to test future patterns.
> 
>     2. Directly accessing debug state through the datastore helpers. I like that this separates code that governs "internal" behavior and user-facing behavior.
> 
> 
> I am maybe a little more inclined towards the first option at this point, but I'll leave it up to you what you want to try first. There could be other options too, I'm not sure that's exhaustive. I think it could be good to get a second opinion here, since this is having a bit more of an impact on discussions of architecture.

I think at this point it's a matter of trade offs

1. `DebugState` as part of `AppState`.
    -  Keeps the state global
    - Has the pattern we already know how to implement and consume
    - The state is really living in two places simultaneously (shared preferences & app state)
2. Using `DataStore` as the proof of concept observable persistent preferences.
    - This is a new pattern and preference architecture ([it is used for Pocket though](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/datastore/DataStores.kt#14))
    - Non-zero learning curve
    - Easily consumable in Compose

Either way, there will be some degree of clunkiness because we're dealing with the legacy XML prefs and wanting to live in a modern, observable world. As it stands right now, I'll likely need and want to convert Secret Settings as a whole into DataStore, since the preferences API so tightly coupled to XML and the shared preferences disk writing, preventing me to do a simple "migration" to Compose from a preferences XML file and PreferenceFragment.

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
MatthewTighe: Is the read guaranteed to complete before we finish setting up the UI initially? Since we're launching a coroutine to collect the read in, it seems like we could reach the end up setup before `debugDrawerEnabled` has been updated the result

MozillaNoah: I don't believe so, but we have a couple options here:
- Use `runBlocking` or our `runBlockingIncrement` for this singular read
- Pre-load the data into memory ([doc](https://developer.android.com/topic/libraries/architecture/datastore#synchronous))

For option 2, we just fire off pre-read of the store inside of onCreate, essentially.

fenix/app/src/main/java/org/mozilla/fenix/datastore/DebugDataStores.kt
t-p-white, line 1: What do you think about calling this`DebugDrawerDataStore`? `DebugDataStores` sounds a little generic and not sure if the pluralizing is needed?

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
t-p-white: Seems like option 2 would be preferable based on the doc? 

I haven't tested this but throwing out there, could you wrap the preference in a `lifecycleScope.launch` such as:
```
 lifecycleScope.launch {
            val debugDrawerEnabled = context?.debugSettings?.isDebugDrawerEnabled?.first() ?: false

            requirePreference<SwitchPreference>(R.string.pref_key_enable_debug_drawer).apply {
                isVisible = Config.channel.isNightlyOrDebug
                isChecked = debugDrawerEnabled
                onPreferenceChangeListener =
                    Preference.OnPreferenceChangeListener { _, newValue ->
                        lifecycleScope.launch {
                            context.debugSettings.updateDebugDrawerEnabled(enabled = newValue as Boolean)
                        }
                        true
                    }
            }
        }
``` 
or will this have an issue drawing async?

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
```diff
@@ -86,6 +97,18 @@ class SecretSettingsFragment : PreferenceFragmentCompat() {
             }
         }
```
t-p-white, line 99: Might be nice to include a small comment similar to the block below? e.g.  "We only want to enable the debug drawer for Nightly or Debug builds because..."

t-p-white (commented):
Looking good Noah! Just a few comments from me

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
t-p-white: nit: I also wonder whether the `?.first() ?: false` should be move directly to `debugDrawerEnabledFlow`

fenix/app/src/main/java/org/mozilla/fenix/datastore/DebugDataStores.kt
MozillaNoah: The pluralization was on accident, so thanks for catching that.

I initially made it just `debug` as an intentional future catch all, but having the specifically-scoped naming makes more sense looking at it.

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
MozillaNoah: I'll give the preference wrap a try.

I'll move it into the data store file and use `false` as a default static value

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
```diff
@@ -92,6 +98,23 @@ class SecretSettingsFragment : PreferenceFragmentCompat() {
             onPreferenceChangeListener = SharedPreferenceUpdater()
         }
 
+        lifecycleScope.launch {
+            val debugDataStore = requireContext().debugSettings
+
+            // During initial development, this will only be available in Nightly or Debug builds.
+            requirePreference<SwitchPreference>(R.string.pref_key_enable_debug_drawer).apply {
+                isVisible = Config.channel.isNightlyOrDebug
+                isChecked = debugDataStore.debugDrawerEnabled.first()
```
t-p-white, line 107: nit: IMO I think moving the `first()` call directly to `debugDrawerEnabled` would keep the responsibility in a single place instead of the caller having to make the call on a `flow` as we only want the boolean value and the flow won't update after the first call? e.g.
```kotlin
suspend fun DataStore<Preferences>.debugDrawerEnabled(): Boolean = data.map { preferences ->
        preferences[DEBUG_DRAWER_ENABLED] ?: false
    }.first()
```
Not a blocker though

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
```diff
@@ -92,6 +98,23 @@ class SecretSettingsFragment : PreferenceFragmentCompat() {
             onPreferenceChangeListener = SharedPreferenceUpdater()
         }
 
+        lifecycleScope.launch {
+            val debugDataStore = requireContext().debugSettings
```
t-p-white, line 102: nit: this could maybe be inlined if the suggestion below is used

t-p-white (commented):
>This is a new pattern and preference architecture ([it is used for Pocket though](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/datastore/DataStores.kt#14))

Do we know why it was used for Pocket?

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
MozillaNoah: We'll need the observe the flow as part of https://github.com/mozilla-mobile/firefox-android/pull/4612/commits/32fc424e1482913e97c21c478ae88062b9bed65b. 

Should we keep the `single` in that case, or maybe instead just expose a non-observable API?

t-p-white: Ah okay, so just checking my understanding:
* Prefs screen needs the `Boolean` value only as this retrieves the value only when created
* The overlay needs to observe the `Flow` to update its enabled state as this is always present
?

t-p-white:
@MozillaNoah after having a bit more of a read through of the `DataStore` docs, if we are going for:

>     2. Using `DataStore` as the proof of concept observable persistent preferences.
>        
>        * This is a new pattern and preference architecture ([it is used for Pocket though](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/datastore/DataStores.kt#14))
>        * Non-zero learning curve
>        * Easily consumable in Compose

Should we also establish a pattern where-by we use a top level 'grouping' DataStore such as a 'secret settings preferences' DataStore in `DataStores` e.g.:

```kotlin
/**
 * Class that handles saving and retrieving secret settings preferences
 */

private const val SECRET_SETTING_PREFERENCES_NAME = "secret_settings_preferences"

// Create a DataStore instance using the preferencesDataStore delegate, with the Context as
// receiver.
val Context.dataStore: DataStore<Preferences> by preferencesDataStore(
    name = SECRET_SETTING_PREFERENCES_NAME
)
```
For reference, I was using the guide and code from this [CodeLab](https://github.com/google-developer-training/android-basics-kotlin-words-app/blob/datastore/app/src/main/java/com/example/wordsapp/data/SettingsDataStore.kt).

MatthewTighe:
Just wanting share some thoughts about a couple things in the thread above, but I don't think we need to spend too long analyzing this as we iterate on it. Since this isn't user-facing or critical path at all, it should provide an easy area for more rapid adjustment than elsewhere in the app.

>1. DebugState as part of AppState.
> - Keeps the state global
> - Has the pattern we already know how to implement and consume
> - The state is really living in two places simultaneously (shared preferences & app state)
> - Keeps us stuck with the monolithic Preferences API

Specifically:

> The state is really living in two places simultaneously (shared preferences & app state)

This is true of any state whose real source of truth is behind an async API. It's also true for architectures that aren't ours, like MVVM. The issue is that we often want to update the local state before we can fetch the latest from async (whether disk or network), but this runs risk of divergence. 

The Redux pattern tends to encourage more eagerly updating local state (through the reducer), but I don't think it's _required_. We could instead:
1. Dispatch an action like `DebugDrawerPreferenceChanged` 
2. Write the update to disk from a middleware
3. Ignore the action in the reducer
4. Observe the datastore from some kind of binding which then updates the store with like `DebugDrawerPreferenceRead`
5. the reducer uses the result of the datastore through the second action

I think, generally speaking, we just trust that the async operations succeed and that the remote and local sources of truth stay in sync. There's probably lots to explore here in terms of out-of-sync error handling, caching mechanisms, etc.

I don't think we need to do any of this for this use case now (or ever if we don't want to), but I do think  in general that `DataStore` state would be best served by living in a Store.

MatthewTighe (approved):
This LGTM as a first iteration based on our current goals. As I mention in my comment above, this code should be easier to change down the line than other code in the app.

I do want to also mention that as we switch to DataStore we may want to create abstract super types for testing. Something like:

```
interface DebugSettingsCache {
    fun debugDrawerEnabledFlow(): Flow<Boolean>
    fun setDebugDrawerEnabled(isEnabled: Boolean)
}

class DebugSettingsDataStore(
    context: Context,
    private val writeScope: CoroutineScope = CoroutineScope(Dispatchers.IO)
) : DebugSettingsCache {
    private val debugCache = context.debugSettings
    override fun debugDrawerEnabledFlow(): Flow<Boolean> =
        debugCache.data.map { preferences ->
            preferences[DEBUG_DRAWER_ENABLED] ?: false
        }

    override fun setDebugDrawerEnabled(isEnabled: Boolean) {
        writeScope.launch {
            debugCache.edit { preferences ->
                preferences[DEBUG_DRAWER_ENABLED] = isEnabled
            }
        }
    }
}
```

This would just make swapping out a fake in tests much easier. Otherwise, I think we'd have to like mock the return value of the extension function you've defined on `Context`? 

Again, probably not necessary for this use case but we should be careful that this isn't used as a definitive precedent down the line.

fenix/app/src/main/java/org/mozilla/fenix/settings/SecretSettingsFragment.kt
MozillaNoah: That's correct!

MozillaNoah:
> @MozillaNoah after having a bit more of a read through of the `DataStore` docs, if we are going for:
> 
> > ```
> > 2. Using `DataStore` as the proof of concept observable persistent preferences.
> >    
> >    * This is a new pattern and preference architecture ([it is used for Pocket though](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/datastore/DataStores.kt#14))
> >    * Non-zero learning curve
> >    * Easily consumable in Compose
> > ```
> 
> Should we also establish a pattern where-by we use a top level 'grouping' DataStore such as a 'secret settings preferences' DataStore in `DataStores` e.g.:
>
> For reference, I was using the guide and code from this [CodeLab](https://github.com/google-developer-training/android-basics-kotlin-words-app/blob/datastore/app/src/main/java/com/example/wordsapp/data/SettingsDataStore.kt).

That's certainly an option! I'm going to investigate Secret Settings in general as part of [Bug 1864712](https://bugzilla.mozilla.org/show_bug.cgi?id=1864712), but I currently foresee two ways to group our preferences that will be maintained via `DataStore`.

1. Have a data store to mirror the UI of our settings flows (e.g. data store for tabs, data store for search, etc.)
2. Have a data store scoped to the use case/purpose (e.g. debugging, everything related to tabs, 

Basically, do we couple the data stores to their code adjacency or their product adjacency? We don't have to solve this now, but this work and [Bug 1864712](https://bugzilla.mozilla.org/show_bug.cgi?id=1864712) will undoubtedly inform us how to proceed with porting the rest of `Settings.kt` and creating new preferences going forward.

MozillaNoah:
> This LGTM as a first iteration based on our current goals. As I mention in my comment above, this code should be easier to change down the line than other code in the app.
> 
> I do want to also mention that as we switch to DataStore we may want to create abstract super types for testing. Something like:
> 
> ```
> interface DebugSettingsCache {
>        ...
> }
> 
> class DebugSettingsDataStore(
>     context: Context,
>     private val writeScope: CoroutineScope = CoroutineScope(Dispatchers.IO)
> ) : DebugSettingsCache {
>     ...
> }
> ```
> 
> This would just make swapping out a fake in tests much easier. Otherwise, I think we'd have to like mock the return value of the extension function you've defined on `Context`?
> 
> Again, probably not necessary for this use case but we should be careful that this isn't used as a definitive precedent down the line.

I love this idea! I'll write a ticket to add this abstraction. It will certainly make tests easier when the redux layer is set up as part of [Bug 1864073](https://bugzilla.mozilla.org/show_bug.cgi?id=1864073) and then we can test the data store in isolation!

MozillaNoah:
Hey @t-p-white, I'm going to land this today, but let's continue the discussion of DataStores and preferences as this feature evolves!

PR #4549, opened 2023-11-21
Bug 1865904 - Add correct default elevation to FAB

MozillaNoah:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1865904

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/compose/button/FloatingActionButton.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/compose/button/FloatingActionButton.kt
@@ -12,6 +12,8 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.material.FloatingActionButton
+import androidx.compose.material.FloatingActionButtonDefaults
+import androidx.compose.material.FloatingActionButtonElevation
 import androidx.compose.material.Icon
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


fenix/app/src/main/java/org/mozilla/fenix/compose/button/FloatingActionButton.kt
```diff
@@ -35,6 +37,7 @@ import org.mozilla.fenix.theme.FirefoxTheme
  * @param modifier [Modifier] to be applied to the action button.
  * @param contentDescription The content description to describe the icon.
  * @param label Text to be displayed next to the icon.
+ * @param elevation Elevation parameter set.
```
gabrielluong, line 40:
```suggestion
 * @param elevation  FloatingActionButtonElevation used to resolve the elevation for this FAB in different states.
```
This is from the `FloatingActionButton` kDoc from for this parameter.

fenix/app/src/main/java/org/mozilla/fenix/compose/button/FloatingActionButton.kt
```diff
@@ -43,13 +46,15 @@ fun FloatingActionButton(
     modifier: Modifier = Modifier,
     contentDescription: String? = null,
     label: String? = null,
+    elevation: FloatingActionButtonElevation = FloatingActionButtonDefaults.elevation(defaultElevation = 5.dp),
```
gabrielluong, line 49: I do see in the figma that we have 3 values 5, 18, 10 dp which probably represents the other states. I know we only care about the default state from the ticket, but is there any reason we are not adjusting the elevation for the hovered, pressed and focused state while we are already adjusting the elevation here?

I think the default values for the other state are fine, but unclear why we are excluding the elevation.

gabrielluong (approved):

fenix/app/src/main/java/org/mozilla/fenix/compose/button/FloatingActionButton.kt
MozillaNoah: I'll be honest, I have no idea how Figma is deriving the 18 and 10 values in the Compose code generator tool. I got the single value of 5 by checking the XML code.

As far as I can tell, the design system doesn't have a language for these other states, so I just wanted to get this change landed and defer that conversation for now, as we have no precedence for specifically setting the elevation (or colors) for pressed and hovered states.

gabrielluong: sgtm

PR #4568, opened 2023-11-22
Bug 1866293 - Add support for installing an add-on a from local file.

Amejia481:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1866293

### DIFF
--- a/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
+++ b/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
@@ -233,40 +233,53 @@ class GeckoEngine(
         onSuccess: ((WebExtension) -> Unit),
         onError: ((String, Throwable) -> Unit),
     ): CancellableOperation {
-        val onInstallSuccess: ((org.mozilla.geckoview.WebExtension) -> Unit) = {
-            val installedExtension = GeckoWebExtension(it, runtime)
-            webExtensionDelegate?.onInstalled(installedExtension)
-            installedExtension.registerActionHandler(webExtensionActionHandler)
-            installedExtension.registerTabHandler(webExtensionTabHandler, defaultSettings)
-            onSuccess(installedExtension)
-        }
-
-        val geckoResult = if (url.isResourceUrl()) {
-            runtime.webExtensionController.ensureBuiltIn(url, id).apply {
+        return if (url.isResourceUrl()) {
+            val geckoResult = runtime.webExtensionController.ensureBuiltIn(url, id).apply {
                 then(
                     {
-                        onInstallSuccess(it!!)
+                        onExtensionInstalled(it!!, onSuccess)
                         GeckoResult<Void>()
                     },
                     { throwable ->
-                        onError(id, GeckoWebExtensionException.createWebExtensionException(throwable))
+                        onError(
+                            id,
+                            GeckoWebExtensionException.createWebExtensionException(throwable),
+                        )
                         GeckoResult<Void>()
                     },
                 )
             }
+            geckoResult.asCancellableOperation()
         } else {
-            runtime.webExtensionController.install(url).apply {
-                then(
-                    {
-                        onInstallSuccess(it!!)
-                        GeckoResult<Void>()
-                    },
-                    { throwable ->
-                        onError(id, GeckoWebExtensionException.createWebExtensionException(throwable))
-                        GeckoResult<Void>()
-                    },
-                )
-            }
+            this.installWebExtension(
+                url,
+                onSuccess,
+                onError = { throwable ->
+                    onError(id, throwable)
+                },
+            )
+        }
+    }
+
+    /**
+     * See [Engine.installWebExtension].
+     */
+    override fun installWebExtension(
+        url: String,
+        onSuccess: ((WebExtension) -> Unit),
+        onError: ((Throwable) -> Unit),
+    ): CancellableOperation {
+        val geckoResult = runtime.webExtensionController.install(url).apply {
+            then(
+                {
+                    onExtensionInstalled(it!!, onSuccess)
+                    GeckoResult<Void>()
+                },
+                { throwable ->
+                    onError(GeckoWebExtensionException.createWebExtensionException(throwable))
+                    GeckoResult<Void>()
+                },
+            )
         }
         return geckoResult.asCancellableOperation()
     }
@@ -1232,6 +1245,17 @@ class GeckoEngine(
         } else {
             null
         }
+    }
+
+    private fun onExtensionInstalled(
+        ext: org.mozilla.geckoview.WebExtension,
+        onSuccess: ((WebExtension) -> Unit),
+    ) {
+        val installedExtension = GeckoWebExtension(ext, runtime)
+        webExtensionDelegate?.onInstalled(installedExtension)
+        installedExtension.registerActionHandler(webExtensionActionHandler)
+        installedExtension.registerTabHandler(webExtensionTabHandler, defaultSettings)
+        onSuccess(installedExtension)
     }
 }
 
--- a/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionRuntime.kt
+++ b/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionRuntime.kt
@@ -38,6 +38,29 @@ interface WebExtensionRuntime {
     }
 
     /**
+     * Installs a [WebExtension] from the provided [url] in this engine.
+     *
+     * @param url the url pointing to either a resources path for locating the extension
+     * within the APK file (e.g. resource://android/assets/extensions/my_web_ext) or to a
+     * local (e.g. resource://android/assets/extensions/my_web_ext.xpi) or remote
+     * (e.g. https://addons.mozilla.org/firefox/downloads/file/123/some_web_ext.xpi) XPI file.
+     * @param onSuccess (optional) callback invoked if the extension was installed successfully,
+     * providing access to the [WebExtension] object for bi-directional messaging.
+     * @param onError (optional) callback invoked if there was an error installing the extension.
+     * This callback is invoked with an [UnsupportedOperationException] in case the engine doesn't
+     * have web extension support.
+     */
+    @Suppress("LongParameterList")
+    fun installWebExtension(
+        url: String,
+        onSuccess: ((WebExtension) -> Unit) = { },
+        onError: ((Throwable) -> Unit) = { _ -> },
+    ): CancellableOperation {
+        onError(UnsupportedOperationException("Web extension support is not available in this engine"))
+        return CancellableOperation.Noop()
+    }
+
+    /**
      * Updates the provided [extension] if a new version is available.
      *
      * @param extension the extension to be updated.
--- a/android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
+++ b/android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
@@ -145,16 +145,40 @@ class AddonManager(
             id = addon.id,
             url = addon.downloadUrl,
             onSuccess = { ext ->
-                val installedState = toInstalledState(ext)
-                val installedAddon = Addon.newFromWebExtension(ext, installedState)
-                    .copy(iconUrl = addon.iconUrl)
-                addonUpdater.registerForFutureUpdates(installedAddon.id)
-                completePendingAddonAction(pendingAction)
-                onSuccess(installedAddon)
+                onAddonInstalled(ext, pendingAction, onSuccess)
             },
             onError = { id, throwable ->
                 completePendingAddonAction(pendingAction)
                 onError(id, throwable)
+            },
+        )
+    }
+
+    /**
+     * Installs an [Addon] from the provided [url].
+     *
+     *  @param url the url pointing to either a resources path for locating the extension
+     *  within the APK file (e.g. resource://android/assets/extensions/my_web_ext) or to a
+     *  local (e.g. resource://android/assets/extensions/my_web_ext.xpi) or remote
+     *  (e.g. https://addons.mozilla.org/firefox/downloads/file/123/some_web_ext.xpi) XPI file.
+     * @param onSuccess (optional) callback invoked if the addon was installed successfully,
+     * providing access to the [Addon] object.
+     * @param onError (optional) callback invoked if there was an error installing the addon.
+     */
+    fun installAddon(
+        url: String,
+        onSuccess: ((Addon) -> Unit) = { },
+        onError: ((Throwable) -> Unit) = { _ -> },
+    ): CancellableOperation {
+        val pendingAction = addPendingAddonAction()
+        return runtime.installWebExtension(
+            url = url,
+            onSuccess = { ext ->
+                onAddonInstalled(ext, pendingAction, onSuccess)
+            },
+            onError = { throwable ->
+                completePendingAddonAction(pendingAction)
+                onError(throwable)
             },
         )
     }
@@ -404,6 +428,19 @@ class AddonManager(
             it.start()
         }
         return Handler(iconThread.looper).asCoroutineDispatcher("WebExtensionIconDispatcher")
+    }
+
+    private fun onAddonInstalled(
+        ext: WebExtension,
+        pendingAction: CompletableDeferred<Unit>,
+        onSuccess: ((Addon) -> Unit),
+    ) {
+        val installedState = toInstalledState(ext)
+        val installedAddon = Addon.newFromWebExtension(ext, installedState)
+
+        addonUpdater.registerForFutureUpdates(installedAddon.id)
+        completePendingAddonAction(pendingAction)
+        onSuccess(installedAddon)
     }
 
     companion object {
--- a/android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
+++ b/android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
@@ -1,0 +1,81 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.addons.ui
+
+import android.content.ActivityNotFoundException
+import android.content.Context
+import android.content.Intent
+import android.net.Uri
+import androidx.activity.result.ActivityResultCaller
+import androidx.activity.result.ActivityResultLauncher
+import androidx.activity.result.contract.ActivityResultContracts
+import mozilla.components.feature.addons.Addon
+import mozilla.components.feature.addons.AddonManager
+import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.ktx.android.net.toFileUri
+
+/**
+ * Allows to launch a file picker to select an add-on file.
+ * @param context the application context
+ * @param onInstallationFailed (optional) callback invoked if there was an error installing the add-on.
+ */
+class AddonFilePicker(
+    val context: Context,
+    private val addonManager: AddonManager,
+    private val onInstallationFailed: (() -> Unit) = { },
+) {
+    internal lateinit var activityLauncher: ActivityResultLauncher<Array<String>>
+    private val logger = Logger("AddonFilePicker")
+
+    /**
+     * Launch an Android file picker where the user can select an add-on file.
+     * @returns a [Boolean] indicating if the file picker was launched successfully or not.
+     */
+    fun launch(): Boolean {
+        return try {
+            activityLauncher.launch(emptyArray())
+            true
+        } catch (e: ActivityNotFoundException) {
+            logger.error("Unable to find an app to select an XPI file", e)
+            false
+        }
+    }
+
+    /**
+     * Registers a listener for file picker results.
+     * @param resultCaller The [ActivityResultCaller] on which results will be observed.
+     */
+    fun registerForResults(resultCaller: ActivityResultCaller) {
+        activityLauncher =
+            resultCaller.registerForActivityResult(AddonOpenDocument(), ::handleUriSelected)
+    }
+
+    internal fun handleUriSelected(uri: Uri?) {
+        uri?.let {
+            val fileUri = convertToFileUri(uri)
+            val onSuccess: ((Addon) -> Unit) = {
+                logger.info("Add-on from $fileUri installed successfully")
+            }
+            val onError: ((Throwable) -> Unit) = { throwable ->
+                onInstallationFailed()
+                logger.error("Unable to install add-on from $fileUri", throwable)
+            }
+            addonManager.installAddon(
+                fileUri,
+                onSuccess,
+                onError,
+            )
+        }
+    }
+
+    internal fun convertToFileUri(uri: Uri): String = uri.toFileUri(context, "XPIs").toString()
+}
+
+internal open class AddonOpenDocument : ActivityResultContracts.OpenDocument() {
+    override fun createIntent(context: Context, input: Array<String>): Intent {
+        // We're restricting the mime type to XPI files, because that's expected by GeckoView currently.
+        return super.createIntent(context, arrayOf("application/x-xpinstall"))
+    }
+}
--- a/android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
+++ b/android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
@@ -1,0 +1,88 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.addons.mozilla.components.feature.addons.ui
+
+import android.content.ActivityNotFoundException
+import android.net.Uri
+import androidx.activity.result.ActivityResultCaller
+import androidx.activity.result.ActivityResultLauncher
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import mozilla.components.feature.addons.AddonManager
+import mozilla.components.feature.addons.ui.AddonFilePicker
+import mozilla.components.feature.addons.ui.AddonOpenDocument
+import mozilla.components.support.test.any
+import mozilla.components.support.test.mock
+import mozilla.components.support.test.robolectric.testContext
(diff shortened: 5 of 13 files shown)


android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,71 @@
...
+
+    internal fun handleUriSelected(uri: Uri?) {
+        uri?.let {
+            // We want to verify that we are not exposing any private data
+            val fileUri = uri.toFileUri(context,"XIPs").toString()
+            val onSuccess: ((Addon) -> Unit) = {
+                logger.info("Add installed successfully")
+            }
+            val onError: ((String, Throwable) -> Unit) = { _, _ ->
+                logger.info("Unable to install add")
```
willdurand, line 54:
```suggestion
                logger.info("Add-on installed successfully")
            }
            val onError: ((String, Throwable) -> Unit) = { _, _ ->
                logger.info("Unable to install add-on")
```

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,71 @@
...
+        activityLauncher =
+            resultCaller.registerForActivityResult(AddonOpenDocument()) { uri ->
+                handleUriSelected(uri)
+            }
+    }
+
+    internal fun handleUriSelected(uri: Uri?) {
+        uri?.let {
+            // We want to verify that we are not exposing any private data
+            val fileUri = uri.toFileUri(context,"XIPs").toString()
```
willdurand, line 49:
```suggestion
            val fileUri = uri.toFileUri(context,"XPIs").toString()
```

maybe?

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,71 @@
...
+            val onSuccess: ((Addon) -> Unit) = {
+                logger.info("Add installed successfully")
+            }
+            val onError: ((String, Throwable) -> Unit) = { _, _ ->
+                logger.info("Unable to install add")
+            }
+            addonManager.installAddon(
+                Addon(id = "myAddon", downloadUrl = fileUri),
```
willdurand, line 57: ha, so we have to fake the `id` here?

fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -386,6 +394,11 @@ class SettingsFragment : PreferenceFragmentCompat() {
             resources.getString(R.string.pref_key_nimbus_experiments) -> {
                 SettingsFragmentDirections.actionSettingsFragmentToNimbusExperimentsFragment()
             }
+            resources.getString(R.string.pref_key_install_local_addon) -> {
+                // TODO: Report telemetry
```
willdurand, line 398: why?

fenix/app/src/main/res/values/strings.xml
```diff
@@ -610,6 +610,8 @@
 
     <!-- Preference for add_ons -->
     <string name="preferences_addons">Add-ons</string>
+    <!-- Preference for installing a local add-on -->
+    <string name="preferences_install_local_addon">Install local Addon</string>
```
willdurand, line 614:
```suggestion
    <string name="preferences_install_local_addon">Install add-on from file</string>
```

to match what we have in `about:addons`:

Screenshot 2023-11-23 at 09 39 21

fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
Amejia481: Should we collect info about how offence users use the option?

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
willdurand: I see. That's interesting. Maybe we could sub-class `Addon` with a `LocalAddon` that would hardcode the `id` internally?

fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
willdurand: maybe yeah but that'd be a separate bug/ticket (that we need to file)

Amejia481: I'll file a separated ticket for it!

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,83 @@
...
+        uri?.let {
+            val fileUri = convertToFileUri(uri)
+            val onSuccess: ((Addon) -> Unit) = {
+                logger.info("Add-on installed successfully")
+            }
+            val onError: ((String, Throwable) -> Unit) = { _, _ ->
+                logger.info("Unable to install add-on")
+            }
+            addonManager.installAddon(
+                Addon(id = UUID.randomUUID().toString(), downloadUrl = fileUri),
```
Amejia481, line 65: @willdurand as the `id` is not used as part of the installation process, I added `randomUUID`, after the extension is install the add-on manger provide us correct id from the manifest.

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
Amejia481: See https://github.com/mozilla-mobile/firefox-android/pull/4568#discussion_r1403655894

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
willdurand: Any thought on changing `installAddon()` to accept a `url` param instead of a full (and yet partial) `Addon` instance?

Amejia481:
> So I think to move forward, I'd like to (1) discuss if we can change installAddon() to receive a url argument instead of a (possibly) empty Addon instance, and (2) investigate why it isn't possible to load a ZIP file with this patch.


That's sound good, I'm taking a look to the API to see how we can make it work!

For (2), we might want to defer that to a different bug if we want to, but that is going to be a valid use case for Nightly and maybe Beta users (well, extension developers mainly).

Perfect, I'll add a validations.

Amejia481:
@willdurand I pushed another commit with you suggestion. Before landing I'll squash changes.

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
Amejia481: I added a new commit with the suggested changes https://github.com/mozilla-mobile/firefox-android/pull/4568/commits/ade1f097bda2d282520bcd7506ee4e68dfa1b10a :)

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -159,6 +156,36 @@ class AddonManager(
...
+     *
+     *  @param url the url pointing to either a resources path for locating the extension
+     *  within the APK file (e.g. resource://android/assets/extensions/my_web_ext) or to a
+     *  local (e.g. resource://android/assets/extensions/my_web_ext.xpi) or remote
+     *  (e.g. https://addons.mozilla.org/firefox/downloads/file/123/some_web_ext.xpi) XPI file.
+     * @param onSuccess (optional) callback invoked if the addon was installed successfully,
+     * providing access to the [Addon] object.
+     * @param onError (optional) callback invoked if there was an error installing the addon.
+     */
+    fun installAddon(
```
willdurand, line 170: Same here.

And I think the logic with `BLOCKED_PERMISSIONS` is no longer needed. Gecko verifies the privileged permissions too, and this would only work for installs from the Fenix manager.

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,85 @@
...
+            activityLauncher.launch(emptyArray())
+            true
+        } catch (e: ActivityNotFoundException) {
+            logger.error("Unable to find an app to select an XPI file", e)
+            false
+        }
+    }
+
+    /**
+     * Registers a lister for file picker results.
```
willdurand, line 48:
```suggestion
     * Registers a listener for file picker results.
```

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    internal fun handleUriSelected(uri: Uri?) {
+        uri?.let {
+            val fileUri = convertToFileUri(uri)
+            val onSuccess: ((Addon) -> Unit) = {
+                logger.info("Add-on from $fileUri installed successfully")
+            }
+            val onError: ((Throwable) -> Unit) = { _ ->
+                onInstallationFailed()
+                logger.info("Unable to install add-on from $fileUri")
```
willdurand, line 64:
```suggestion
                logger.error("Unable to install add-on from $fileUri")
```

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,85 @@
...
+            )
+        }
+    }
+
+    internal fun convertToFileUri(uri: Uri): String = uri.toFileUri(context, "XPIs").toString()
+}
+
+internal open class AddonOpenDocument : ActivityResultContracts.OpenDocument() {
+    override fun createIntent(context: Context, input: Array<String>): Intent {
+        val intent = super.createIntent(context, input)
```
willdurand, line 79: Can we restrict to xpi files only for now? Mime type is `application/x-xpinstall`.

willdurand (changes requested):
This is looking great but I have questions on the overall design. I'd like to avoid duplicating methods since that will lead to confusion in 6 months..

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
```diff
@@ -1233,6 +1248,14 @@ class GeckoEngine(
             null
         }
     }
+
+    private fun postSuccessfulExtensionInstallation(ext: org.mozilla.geckoview.WebExtension): GeckoWebExtension {
```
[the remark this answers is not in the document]
Amejia481: Addressed!

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionRuntime.kt
```diff
@@ -37,6 +37,29 @@ interface WebExtensionRuntime {
...
+     * local (e.g. resource://android/assets/extensions/my_web_ext.xpi) or remote
+     * (e.g. https://addons.mozilla.org/firefox/downloads/file/123/some_web_ext.xpi) XPI file.
+     * @param onSuccess (optional) callback invoked if the extension was installed successfully,
+     * providing access to the [WebExtension] object for bi-directional messaging.
+     * @param onError (optional) callback invoked if there was an error installing the extension.
+     * This callback is invoked with an [UnsupportedOperationException] in case the engine doesn't
+     * have web extension support.
+     */
+    @Suppress("LongParameterList")
+    fun installWebExtension(
```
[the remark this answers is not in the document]
Amejia481: I think it will be a significant one, with have around 37 usages of this signature, and all their dependencies that will need changes, without counting which other issue we could find. I would rather prefer to file a follow up bug, land this functionality and start collecting feedback, while we explore the possible refactor. 

image

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: I think we should file a follow up bug to verify and confirm, the behaviour before removing the logic, as it looks like it's out side of the scope of the bug?

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
Amejia481: I'll give a try, as `application/x-xpinstall` is not a while spread mime type not sure if Android will recognized it.

Amejia481: It's working 😅 , looks like [Android recognize it](https://cs.android.com/search?q=application%2Fx-xpinstall&sq=)

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt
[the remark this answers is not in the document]
Amejia481: Addressed!

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
Amejia481: Addressed!

Amejia481:
@willdurand address some of the comments and added some clarification comments.

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
willdurand: Yes, file a follow-up to consider removing `BLOCKED_PERMISSIONS`.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionRuntime.kt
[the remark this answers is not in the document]
willdurand: aren't these occurrences mostly in test files?

fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -698,6 +717,16 @@ class SettingsFragment : PreferenceFragmentCompat() {
...
 
+    @VisibleForTesting
+    internal fun setupInstallAddonFromFilePreference(settings: Settings) {
+        with(requirePreference<Preference>(R.string.pref_key_install_local_addon)) {
+            isVisible = settings.showSecretDebugMenuThisSession &&
+                isNightlyOrDebugOrBeta()
+        }
+    }
+
+    internal fun isNightlyOrDebugOrBeta() = Config.channel.isNightlyOrDebug || Config.channel.isBeta
```
willdurand, line 728: Nit: not sure if introducing this variable is worthwhile

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,81 @@
...
+            )
+        }
+    }
+
+    internal fun convertToFileUri(uri: Uri): String = uri.toFileUri(context, "XPIs").toString()
+}
+
+internal open class AddonOpenDocument : ActivityResultContracts.OpenDocument() {
+    override fun createIntent(context: Context, input: Array<String>): Intent {
+        return super.createIntent(context, arrayOf("application/x-xpinstall"))
```
willdurand, line 79: Let's add a comment to indicate that we restrict the mime type to XPI files because that's expected by GeckoView currently.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionRuntime.kt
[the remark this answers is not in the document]
Amejia481: We have around 10 of them that aren't tests files but those occurrences have tests and dependencies, changing them will require updating their dependencies. As the change is a refactor that it's not related to the functionality, I think we could land the functionality start collecting feedback and discovering possible bugs, while simultaneously addressing the possible refactor and separated ticket.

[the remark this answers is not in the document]
willdurand: I am OK with deferring the refactoring to a follow-up bug but I don't think we should introduce a new method that we will remove shortly after. Maybe we should refactor the code first then, and rebase this PR on top of the refactored codebase?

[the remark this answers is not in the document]
Amejia481: The main issue that I see is that we won't know the extend of the refactor until we do it, as it's a **possible refactor**. Meanwhile, we're blocking a feature which is ready to land to give a try to a **possible refactor**,  which block us from start collecting feedback early.  Normally in Fenix we prioritize an agile process, where we try to have small incremental patches which speed up the feedback loop between in detecting issues.

Also keeping a PR opened have a maintenance cost, which is the burner of maintain it from being stale, as the source code is alive and keeps changing.

An example of keeping the PR from being stale and it's this merge conflict. 
image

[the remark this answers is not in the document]
willdurand: Ok, so I started the refactoring and it looks like the way built-ins are treated is going to annoy us... https://searchfox.org/mozilla-central/rev/9bc264fbc5d6e618d8f3b9677a8f5e8550b94dbc/mobile/android/modules/geckoview/GeckoViewWebExtension.sys.mjs#903,911

I guess that forces me to reconsider this patch 😛

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
```diff
@@ -406,6 +433,15 @@ class AddonManager(
         return Handler(iconThread.looper).asCoroutineDispatcher("WebExtensionIconDispatcher")
     }
 
+    private fun postSuccessfulAddonInstallation(ext: WebExtension, pendingAction: CompletableDeferred<Unit>): Addon {
```
willdurand, line 436: Maybe we could do this, and avoid returning an `addon` in addition to calling `registerXXX` and so on

```suggestion
    private fun onAddonInstalled(ext: WebExtension, pendingAction: CompletableDeferred<Unit>, onSuccess: ((Addon) -> Unit)) {
```

fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -698,6 +717,16 @@ class SettingsFragment : PreferenceFragmentCompat() {
         }
     }
 
+    @VisibleForTesting
+    internal fun setupInstallAddonFromFilePreference(settings: Settings) {
+        with(requirePreference<Preference>(R.string.pref_key_install_local_addon)) {
+            isVisible = settings.showSecretDebugMenuThisSession &&
+                isNightlyOrDebugOrBeta()
```
willdurand, line 724: I think we don't need this condition, thanks for mentioning it during the standup 😅 We do want this feature in Release.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionRuntime.kt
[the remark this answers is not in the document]
Amejia481: How should we proceed?  😛

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: 👍🏽

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/webextension/WebExtensionRuntime.kt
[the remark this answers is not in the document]
willdurand: Let's file a follow-up bug to refactor the `install` methods: I don't think the `id` is used for normal extensions, only for built-in ones.

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
willdurand: oh, hm. and why is this not a problem for the newly introduced install method then?

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: https://bugzilla.mozilla.org/show_bug.cgi?id=1867297

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: I think it's because we don't have the previous add-on to extract the `iconUrl`, I remembered we had a bug with the icon, I don't quite recall what was the issue about but we introduced the change [here](https://github.com/mozilla-mobile/firefox-android/commit/e73e3a7c4a9ea8de53e44618134000b3320a8721). My main worry is that we re introduce the issue 😅 .

Amejia481: I guess it should be safe as we are loading the icon when we are installing

Amejia481: I'll give a try without the `iconUrl` !

fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
Amejia481: Removed

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
Amejia481: Added

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/AddonManager.kt
Amejia481: I added the suggested changes!

fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
Amejia481: removed!

Amejia481:
@willdurand I updated the PR with the suggested changes!

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -0,0 +1,82 @@
...
+import mozilla.components.feature.addons.Addon
+import mozilla.components.feature.addons.AddonManager
+import mozilla.components.support.base.log.logger.Logger
+import mozilla.components.support.ktx.android.net.toFileUri
+
+/**
+ * Allows to launch a file picker to select an add-on file.
+ * @param context the application context
+ * @param onInstallationFailed (optional) callback invoked if there was an error installing the add-on.
+ *
```
willdurand, line 23:
```suggestion
```

fenix/app/src/test/java/org/mozilla/fenix/settings/SettingsFragmentTest.kt
```diff
@@ -95,6 +97,33 @@ class SettingsFragmentTest {
         assertTrue(preferenceAmoCollectionOverride.isVisible)
     }
 
+    @Test
+    fun `Install add-on from file pref is visible if debug menu active and feature is enabled and isNightlyOrDebug or isBeta`() = runTestOnMain {
```
willdurand, line 101:
```suggestion
    fun `Install add-on from file pref is visible if debug menu active and feature is enabled`() = runTestOnMain {
```

willdurand (approved):
Thanks, r+wc

fenix/app/src/test/java/org/mozilla/fenix/settings/SettingsFragmentTest.kt
Amejia481: Addressed!

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    @Before
+    fun setup() {
+        addonManager = mock()
+        filePicker = spy(
+            AddonFilePicker(testContext, addonManager),
+        )
+    }
+
+    @Test
+    fun `WHEN registerForResults is called THEN register for AddonOpenDocument`() {
```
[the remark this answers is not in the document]
Amejia481: Addressed!

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
Amejia481: Addressed!

jackyzy823:
Hello @Amejia481  , should we clean up the cache folder "/data/data/org.mozilla.firefox/cache/XPIS" after add-on installation success/failure ? (A similar issue  like this https://bugzilla.mozilla.org/show_bug.cgi?id=1860472)

willdurand:
> Hello @Amejia481 , should we clean up the cache folder "/data/data/org.mozilla.firefox/cache/XPIS" after add-on installation success/failure ? (A similar issue like this [bugzilla.mozilla.org/show_bug.cgi?id=1860472](https://bugzilla.mozilla.org/show_bug.cgi?id=1860472))

If the cache is never cleared, I think it would make sense yeah, please file a bug

PR #4580, opened 2023-11-23
Bug 1866318 - Retry intent referrer access once

MorMundHS-MA:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1866318

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
@@ -7,6 +7,7 @@ package org.mozilla.fenix
 import android.app.Activity
 import android.content.Intent
 import android.content.pm.PackageManager
+import android.net.Uri
 import android.os.Build
 import android.os.Bundle
 import android.os.StrictMode
@@ -126,7 +127,8 @@ class IntentReceiverActivity : Activity() {
             return
         }
         // NB: referrer can be spoofed by the calling application. Use with caution.
-        val r = referrer ?: return
+        val r = tryGettingReferrer() ?: return
+
         intent.putExtra(EXTRA_ACTIVITY_REFERRER_PACKAGE, r.host)
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
             // Category is supported for API>=26.
@@ -138,6 +140,25 @@ class IntentReceiverActivity : Activity() {
                     // At least we tried.
                 }
             }
+        }
+    }
+
+    @Suppress("TooGenericExceptionCaught")
+    private fun tryGettingReferrer(): Uri? {
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.LOLLIPOP_MR1) {
+            return null
+        }
+
+        return try {
+            referrer
+        } catch (e: RuntimeException) {
+            try {
+                referrer
+            } catch (e: Exception) {
+                null
+            }
+        } catch (e: Exception) {
+            null
         }
     }
 }


fenix/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt
```diff
@@ -140,6 +142,25 @@ class IntentReceiverActivity : Activity() {
...
+        } catch (e: RuntimeException) {
+            try {
+                referrer
+            } catch (e: Exception) {
+                null
+            }
+        } catch(e: Exception) {
+            null
+        }
+    }
```
MorMundHS-MA, line 162: For some reason, the second access does not cause a `RuntimeException` for the `referrer` access. It does not seem to be timing based, as waiting with a break point does not prevent the error from happening. Inspecting the variable does however throw the first time as well, but not the second time. 

However this whole behavior might be very specific to Spotify's implementation and I'm not sure how important the referrer information is. It seems to work fine without it, so a less complex fix is a simple try catch around the `referrer` access.

jonalmeida:
Hi @MorMundHS-MA , thanks for the patch! But as you've noticed, this patch might not be the best solution.

We have had situations like this before with all intents that can come from third-party apps. We've found it's safer to `SafeIntent` which catches these kind of exceptions. Could you add another accessor for the `referrer`.

It might be a bit tricky because I can see there are some extension functions like `sanitize` and `stripUnwantedFlags` on the [`IntentReceiverActivity`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt#47-48).

Hopefully, this is a helpful direction to getting this patch ready for landing. Please feel free to ping me if I can help more. 🙂

jonalmeida (dismissed):
Hi @MorMundHS-MA , thanks for the patch! But as you've noticed, this patch might not be the best solution.

We have had situations like this before with all intents that can come from third-party apps. We've found it's safer to `SafeIntent` which catches these kind of exceptions. Could you add another accessor for the `referrer`.

It might be a bit tricky because I can see there are some extension functions like `sanitize` and `stripUnwantedFlags` on the [`IntentReceiverActivity`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/IntentReceiverActivity.kt#47-48). (EDIT: looks like [focus converts to a SafeIntent after santizing](https://searchfox.org/mozilla-mobile/source/firefox-android/focus-android/app/src/main/java/org/mozilla/focus/activity/IntentReceiverActivity.kt#28))

Hopefully, this is a helpful direction to getting this patch ready for landing. Please feel free to ping me if I can help more. 🙂

MorMundHS-MA:
@jonalmeida Hi, thanks for the information! It gave me a direction to look further into the issue.

Unfortunately using `SafeIntent` won't really help imo. `getReferrer` is a method of the `Activity` itself not the intent. It *does* potentially get its information from the `Intent` ([AOSP code](https://github.com/aosp-mirror/platform_frameworks_base/blob/fe9c3c82cec4c0b7149fa40cd5d3934d172b1177/core/java/android/app/Activity.java#L6768)). But from my testing the actual value of `getReferrer` comes from the `mReferrer` variable in the `Activity`.

So I could use the `SafeIntent` to copy most of the `getReferrer` behavior, this is [what chromium does](https://github.com/chromium/chromium/blob/96df497cc1081929f79c544c5e89df3911da86b9/chrome/android/java/src/org/chromium/chrome/browser/IntentHandler.java#L480). This has the benefit of working on all SDK versions, where as `getReferrer` needs API>=22. But it does not get the referrer in most cases (which I'm still not sure how important the information really is).

Alternatively, I could add a `getSafeReferrer` extension method to `Activity` to mirror the design of `toSafeIntent` somewhat. That would still leave the question if I should fix the Spotify bug where it throws only the first time the `Intent` is accessed or if I should ignore it.

Lastly, there is always the option of reading `mReferrer` through reflection, which I don't think anyone would consider a good idea.

Overall I think my original solution is still the overall cleanest, unfortunately.

jonalmeida:
I don't have capacity to review this so leaving it up to the Fenix team to take over.

gabrielluong:
Hello @MorMundHS-MA , Thank you so much for your contribution. We have moved our firefox-android repo over to [mozilla-central](https://hg.mozilla.org/mozilla-central). Please feel free to reopen a new PR against the new repo. You will have to clone the new repository and reapply your changes. Thank you again, and sorry for the inconvenience.

For more information, please see https://github.com/mozilla-mobile/firefox-android/wiki#upcoming-migration-to-mozilla-central.

PR #4546, opened 2023-11-21
Bug 1866062 - New high priority Firefox suggest UI tests

AndiAJ:
Summary:
1.  Created a new test class and moved the previously created UI tests to it (`verifyFirefoxSuggestSponsoredSearchResultsTest` and `verifyFirefoxSuggestNonSponsoredSearchResultsTest`)

2.  Created new 5 UI tests (marked with "High" priority in TestRail)
     All 5 successfully passed 100x on Firebase ✅ 
     
[List with all keywords](https://drive.google.com/file/d/1o2nLj1L1IAHZCSUL08QHd7sJCEpS8ZZ0/view)

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
4. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
5. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
6. Click on `View task in Taskcluster` in the new `DETAILS` section.
7. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1866062

PR #4590, opened 2023-11-24
Bug 1868371 - Add more test data to Fx suggest related UI tests

AndiAJ:
Based on this [suggestion](https://github.com/mozilla-mobile/firefox-android/pull/4546#pullrequestreview-1744598867):

- Created a map for sponsored/non sponsored suggestions which will be randomly used in the UI tests
- Created `getSponsoredFxSuggestPlaceHolder` helper to help us properly check the sponsored suggestions links (some of the links contain a placeholder `%YYYYMMDDHH%`)

All 7 UI tests successfully passed 100x on Firebase ✅ 

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1868371

### DIFF
--- a/fenix/app/src/androidTest/java/org/mozilla/fenix/helpers/DataGenerationHelper.kt
+++ b/fenix/app/src/androidTest/java/org/mozilla/fenix/helpers/DataGenerationHelper.kt
@@ -22,6 +22,8 @@ import org.junit.Assert
 import org.mozilla.fenix.ext.components
 import org.mozilla.fenix.helpers.TestHelper.mDevice
 import org.mozilla.fenix.utils.IntentUtils
+import java.time.LocalDate
+import java.time.LocalTime
 
 object DataGenerationHelper {
     val appContext: Context = InstrumentationRegistry.getInstrumentation().targetContext
@@ -73,6 +75,33 @@ object DataGenerationHelper {
         val clipData = ClipData.newPlainText("label", message)
 
         clipBoard.setPrimaryClip(clipData)
+    }
+
+    /**
+     * Returns the date and time placeholder used in sponsored Fx suggest links.
+     */
+    fun getSponsoredFxSuggestPlaceHolder(): String {
+        val currentDate = LocalDate.now()
+        val currentTime = LocalTime.now()
+        val currentDay = currentDate.dayOfMonth
+        val currentMonth = currentDate.monthValue
+        val currentYear = currentDate.year
+        val currentHour = currentTime.hour
+        var placeholder: String
+
+        if (currentMonth < 10 && currentDay < 10 && currentHour < 10) {
+            placeholder = "$currentYear" + "0" + "$currentMonth" + "0" + "$currentDay" + "0" + "$currentHour"
+        } else if (currentMonth < 10 && currentDay < 10) {
+            placeholder = "$currentYear" + "0" + "$currentMonth" + "0" + "$currentDay$currentHour"
+        } else if (currentMonth < 10 && currentHour < 10) {
+            placeholder = "$currentYear" + "0" + "$currentMonth" + "$currentDay" + "0" + "$currentHour"
+        } else if (currentDay < 10 && currentHour < 10) {
+            placeholder = "$currentYear$currentMonth" + "0" + "$currentDay" + "0" + "$currentHour"
+        } else {
+            placeholder = "$currentYear$currentMonth$currentDay$currentHour"
+        }
+
+        return placeholder
     }
 
     /**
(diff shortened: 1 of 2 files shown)


sv-ohorvath (approved):

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/FirefoxSuggestTest.kt
```diff
@@ -33,22 +35,86 @@ class FirefoxSuggestTest {
...
+            ),
+            "Bloom" to listOf(
+                "Bloomingdales.com - Official Site",
+                "bloomingdales.com/?cm_mmc=Admarketplace-_-Privacy-_-Privacy-_-privacy%20instant%20suggest-_-319093353us1228${getSponsoredFxSuggestPlaceHolder()}-_-kclickid__kenshoo_clickid_&mfadid=adm",
+            ),
+            "Groupon" to listOf(
+                "groupon.com - Discover & Save!",
+                "groupon.com/?utm_source=google&utm_medium=cpc&utm_campaign=us_dt_sea_ggl_txt_smp_sr_cbp_ch1_nbr_k*{keyword}_m*{match-type}_d*ADMRKT_319093357us1279${getSponsoredFxSuggestPlaceHolder()}&mfadid=adm",
+            ),
+        )
```
AaronMT, line 66: Verify hard for me to review this because I don't know enough about these. These links seem highly structured with campaign details and such. Is there any way to pull these rather than hard-code? One single character change in the URLs and these break.

AndiAJ: Forgot to add in PR the details the [list](https://drive.google.com/file/d/1o2nLj1L1IAHZCSUL08QHd7sJCEpS8ZZ0/view) based on which, each keyword points to a sponsored website and the corresponding link.

Will check to see if these remain the same or could be changed at any point.

jjSDET: I love the approach of using these to cycle through your tests. It's the first step in a data-driven approach to dynamically create your tests through self-discoverability.

The next steps would be generating this data from a source of truth, like the text doc where all these sponsored links live.

Something to consider for doing that would be either starting with saving that doc as a fixture, parsing it, and then pulling in, say, the first 5 links from each category, or something like that; lots of different approaches work there.

If that source of truth is hosted somewhere, you could make a call to it and do that same parsing. That's typically a third step as it adds a hosting/network dependency you need to defend against.

All-in-all, I think it's a great entry-point into some more dynamic, eventually self-writing test automation that will get more coverage for less code and much less maintenance costs. 👍

fenix/app/src/androidTest/java/org/mozilla/fenix/helpers/DataGenerationHelper.kt
```diff
@@ -75,6 +77,33 @@ object DataGenerationHelper {
...
+    fun getSponsoredFxSuggestPlaceHolder(): String {
+        val currentDate = LocalDate.now()
+        val currentTime = LocalTime.now()
+        val currentDay = currentDate.dayOfMonth
+        val currentMonth = currentDate.monthValue
+        val currentYear = currentDate.year
+        val currentHour = currentTime.hour
+        var placeholder: String
+
+        if (currentMonth < 10 && currentDay < 10 && currentHour < 10) {
```
jjSDET, line 92:
```suggestion
        if (currentMonth < 10 && currentDay < 10 && currentHour < 10) {
```
While these if-else conditions may work perfectly and have no bugs, I've burned myself so many times over the years missing something in complex conditional checking.

Anytime I find myself writing more than one chained `if-else` I immediately take a step back and ask myself how I can simplify my logic and steps because so often bugs come up in weird ways when there are more than one conditional being checked.

It isn't necessary, just something I want to call out because more often than not an issue comes up with these and they are really tough to extend should another requirement come up.

I'm in the camp of `never use else` and use `guard-clause` type checking, but that doesn't always fit into what you're trying to do.

Other approaches could be use a `switch-case`, though stylistically some disapprove of that depending on the `coding sensibilities` and rules of the project.

An alternate approach that can be considered that is more robust is using a dictionary/hash-table/map-function where the key is your specific condition and the value is your string concatenation operation or something like that.

Using that approach also makes it much easier to extend later should you want to add more specific conditions later.

Again, nothing wrong with these, I just want to mention this as it might help avoid bugs and the pain of maintenance down the road.

My only 'heavy' recommendation is to add a comment explaining `why you have these conditions and what you want to return` so the intent of the code is very clear for anyone looking at it in the future or when maintenance needs to happen here. That will also help identify (hopefully more quickly) if there's an unintended issue.

jjSDET (approved):
LGTM

PR #4612, opened 2023-11-27
Bug 1864071 - Add debug drawer fab to an app-wide Composable overlay

MozillaNoah:
Note: this PR is based off of https://github.com/mozilla-mobile/firefox-android/pull/4520, so when reviewing this before the root PR lands, make sure to look at the commit just for 1864071 to keep to the actual changes of four files.

https://github.com/mozilla-mobile/firefox-android/assets/87384386/629334a3-7314-4a1a-9a96-7591b59ed686

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864071


fenix/app/src/main/java/org/mozilla/fenix/FenixOverlay.kt
```diff
@@ -0,0 +1,87 @@
...
+    ) {
+        DebugDrawerFAB(
+            visible = debugDrawerFabEnabled,
+            onClick = {
+                scope.launch {
+                    snackbarState.showSnackbar("Show debug drawer")
+                }
+            },
+            modifier = Modifier.align(Alignment.Center),
+        )
```
MozillaNoah, line 59: For now, I think it'll make sense to include just the debug fab in this wrapper composable until we have more debugging UI to overlay the app. In other words, when the drawer UI is added, this might move into an intermediary wrapper `DebugOverlay`, if it makes sense to do so.

fenix/app/src/main/java/org/mozilla/fenix/debugdrawer/ui/DebugDrawerFAB.kt
```diff
@@ -0,0 +1,78 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.debugdrawer.ui
```
MozillaNoah, line 5: with this package addition, the thought is to have something like:

`debugdrawer/ui` for all of the Compose code
`debugdrawer/redux` for the store/state/middlewares
`debugdrawer/feature_name/` for any of the functionality (tab manager, secret settings, crash viewer, etc.)

t-p-white: perhaps we should use debugsettings as the overall feature name? Then Debug Drawer would be part of the UI package, wdyt?

fenix/app/src/main/java/org/mozilla/fenix/debugdrawer/ui/DebugDrawerFAB.kt
```diff
@@ -0,0 +1,78 @@
...
+    visible: Boolean,
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    AnimatedVisibility(
+        visible = visible,
+        modifier = modifier,
+    ) {
+        FloatingActionButton(
+            icon = painterResource(R.drawable.ic_new),
```
t-p-white, line 45: we could use a "bug" type icon? 🙂

fenix/app/src/main/java/org/mozilla/fenix/FenixOverlay.kt
```diff
@@ -0,0 +1,87 @@
...
+private fun FenixOverlayContent(
+    debugDrawerFabEnabled: Boolean,
+) {
+    val snackbarState = remember { SnackbarHostState() }
+    val scope = rememberCoroutineScope()
+
+    Box(
+        modifier = Modifier.fillMaxSize(),
+    ) {
+        DebugDrawerFAB(
```
t-p-white, line 51: nit: Not sure about the positioning of the button, but appreciate this may change or is desirable in this use case

t-p-white (commented):
Nice job, I'll hold off the final review until https://github.com/mozilla-mobile/firefox-android/pull/4520#event-11074801083 is merged

fenix/app/src/main/java/org/mozilla/fenix/debugdrawer/ui/DebugDrawerFAB.kt
MozillaNoah: I'm actually working with Crystal to create an icon just for the drawer! [Ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1866128)

fenix/app/src/main/java/org/mozilla/fenix/FenixOverlay.kt
MozillaNoah: For 0.1, I'm thinking center of the screen. A follow-up once the drawer is stable is to make the FAB draggable, so it can be repositioned anywhere on the screen!

MatthewTighe: Maybe center-left to start? Think that would avoid the more common gesture-based back nav, and may leave more of the screen generally interact-able

fenix/app/src/main/java/org/mozilla/fenix/debugdrawer/ui/DebugDrawerFAB.kt
MatthewTighe: Both of these make sense to me

fenix/app/src/main/java/org/mozilla/fenix/debugdrawer/ui/DebugDrawerFAB.kt
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Floating Action Button used to reveal the debug drawer.
+ *
+ * @param visible Whether the fab is visible.
+ * @param onClick Invoked when the user clicks on the fab.
+ * @param modifier Modifier to apply to the content.
+ */
+@Composable
+fun DebugDrawerFAB(
```
MatthewTighe, line 35: nit: [typical java convention for acronyms of 3+ letters](https://stackoverflow.com/questions/2236807/java-naming-convention-with-acronyms) is for just the first letter to be uppercase. I see other fabs in the codebase follow that as well

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -278,6 +279,13 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         window.decorView.layoutDirection = TextUtils.getLayoutDirectionFromLocale(Locale.getDefault())
 
         binding = ActivityHomeBinding.inflate(layoutInflater)
+
+        binding.homeActivityComposeOverlay.setContent {
+            FirefoxTheme {
+                FenixOverlay()
+            }
+        }
```
MatthewTighe, line 287: It might run a lower risk of performance impact to observe the data store here and call this whole stack conditionally

MatthewTighe (commented):
Had a couple comments but this is looking pretty ready to me once the previous lands.

I do think we should [monitor for performance regressions](https://mozilla-hub.atlassian.net/wiki/spaces/Mobile/pages/222234014/Android+Performance+Regression+Monitoring) once this lands just so we can be aware of any impact. If there is any, we could consider shipping it only on debug builds first or something.

fenix/app/src/main/java/org/mozilla/fenix/datastore/DebugSettingsDataStore.kt
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.datastore
```
rahulsainani, line 5: 💡 This could be in the `debugdrawer.data` package since it's related to that feature and other packages don't need to access it.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
rahulsainani: +1 here

fenix/app/src/main/java/org/mozilla/fenix/datastore/DebugSettingsDataStore.kt
MozillaNoah: made this change over in https://github.com/mozilla-mobile/firefox-android/pull/4520

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
MozillaNoah: I'll look into this. I didn't necessarily want to do a debug settings data store observation at this level, but maybe there's another way to do this. Some initial thoughts could be:
- Have a separate observable boolean `fenixOverlayEnabled` that is driven by a series of other conditions (e.g. `debugDrawerEnabled`)
- Scrap the initial idea of having a universal, global Composable for the entire app and just scope it to the debugging. So make `FenixOverlay` into `DebugOverlay` and then the scoped observation would make more sense.

fenix/app/src/main/java/org/mozilla/fenix/FenixOverlay.kt
MozillaNoah: so like center along the left edge of the screen? or biased like 50 just left of center?

MatthewTighe: left edge would be my suggestion. i'm mostly just worried that repositioning it might be hard and so punted on, and center screen seems like important real-estate

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
MatthewTighe: I do think we can probably land something and iterate on it as needed, but the way I'm thinking about it is this:

What if we wanted to use a feature flag or build flag to control whether the overlay was present? For example only including it in debug builds to start. My assumption would be that that would be controlled before the feature package/fragment/whatever got entered, whereas here we are entering the feature first and then determining if we should do anything with it

MozillaNoah:
Recent push includes a rebase from main with the now-landed secret setting, the FAB lowercase fix, and the change to make the fab align to the left edge of the screen.

MozillaNoah:
recent push removes `FenixOverlay`, consolidates the UI to `DebugOverlay` and moves the observation to `HomeActivity`

MatthewTighe (approved):
LGTM! Let's just monitor perf for the rest of the nightly cycle

PR #4625, opened 2023-11-28
Bug 1826883 - Remove juno prefix from onboarding

rahulsainani:
### What
Remove juno prefixes and references from "all" places except string ids and nimbus flag as they shouldn't be changed for a live feature.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1826883


fenix/app/metrics.yaml
```diff
@@ -1271,7 +1271,7 @@ onboarding:
   turn_on_notifications:
     type: event
     description: |
-      User tapped on turn on notifications button in juno onboarding.
+      User tapped on turn on notifications button in  onboarding.
```
gabrielluong, line 1274:
```suggestion
      User tapped on turn on notifications button in onboarding.
```
This happens a few times here.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/OnboardingFragment.kt
```diff
@@ -224,8 +224,7 @@ class JunoOnboardingFragment : Fragment() {
         showNotificationPage: Boolean,
         showAddWidgetPage: Boolean,
     ): List<OnboardingPageUiData> {
-        val junoOnboardingFeature = FxNimbus.features.junoOnboarding.value()
-        val jexlConditions = junoOnboardingFeature.conditions
+        val jexlConditions = FxNimbus.features.junoOnboarding.value().conditions
```
gabrielluong, line 227: Should we follow up with renaming the Nimbus feature? I believe this will help if we do it sooner since a year from now, people will be wondering what is juno.

gabrielluong (approved):

fenix/app/metrics.yaml
rahulsainani: Good catch, fixed 👍

t-p-white (approved):
Changes lgtm 👍

fenix/app/src/main/java/org/mozilla/fenix/onboarding/OnboardingFragment.kt
t-p-white: Should we have a follow ticket for removing 'juno' from Nimbus, where we capture this & the need to update the [Experimenter](https://experimenter.info/onboarding/onboarding-feature-android/) docs?

t-p-white: Awesome 🙌 

>The experimenter docs only need updating if we do decide to update the string keys, right?

Yes that's right! 

>Can we simply update the string keys

Good question, I honestly don't know the answer to this, @Delphine would you be able to provide some guidance on this please? e.g. The actual strings text stays the same but the key/reference would be renamed

If the answer is yes, then it seems like it'd be a nice to have - if not, then I think we leave them as-is, wdyt?

Delphine: When string IDs change, that invalidates existing translations - so I would suggest keeping the string ID the same if the source string doesn't change.

PR #4646, opened 2023-11-30
Bug 1861459 - Move PBM state from BrowsingModeManager to AppStore

MatthewTighe:
This ended up being pretty large, so I tried to split into logical commits. I'm happy to split them out into separate PRs if that's useful, but I'm hoping to get a full smoke pass of private browsing from QA and am guessing it may be easier to do that in one go than over 3 patches. Definitely open to suggestions there though.

The 3 commits:
1. removes usages of `BrowsingModeManager` from all areas outside of the home screen. These usages were just accesses and updates to state. This is a large number of file changes but they should all be small diffs with mechanical changes.
2. removes usages of `BrowsingModeManager` from the home screen. This includes updates to read the initial app state from disk, and some of the logic here is definitely worth a closer look at from a reviewer.
3. just removes the (now unused) `BrowsingModeManager` and does a final bit of clean-up

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1861459


fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
```diff
@@ -317,10 +317,7 @@ class HistoryFragment : LibraryPageFragment<History>(), UserInteractionHandler,
                 (selectedItem as? History.Regular)?.url ?: (selectedItem as? History.Metadata)?.url
             }
 
-            (activity as HomeActivity).apply {
-                browsingModeManager.mode = BrowsingMode.Private
-                supportActionBar?.hide()
-            }
+            (activity as HomeActivity).supportActionBar?.hide()
```
MatthewTighe, line 320: Note that `openItemsInNewTab` which is called on the line before this dispatches the PBM state update, so I removed the redundancy here.

fenix/app/src/main/java/org/mozilla/fenix/library/historymetadata/HistoryMetadataGroupFragment.kt
```diff
@@ -188,10 +187,7 @@ class HistoryMetadataGroupFragment :
                     selectedItem.url
                 }
 
-                (activity as HomeActivity).apply {
-                    browsingModeManager.mode = BrowsingMode.Private
-                    supportActionBar?.hide()
-                }
+                (activity as HomeActivity).supportActionBar?.hide()
```
MatthewTighe, line 190: Note that `openItemsInNewTab` which is called on the line before this dispatches the PBM state update, so I removed the redundancy here.

fenix/app/src/main/java/org/mozilla/fenix/PrivateBrowsingBinding.kt
```diff
@@ -0,0 +1,56 @@
...
+class PrivateBrowsingBinding(
+    appStore: AppStore,
+    private val themeManager: ThemeManager,
+    private val retrieveWindow: () -> Window,
+    private val settings: Settings,
+    private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
+) : AbstractBinding<AppState>(appStore) {
+    override suspend fun onState(flow: Flow<AppState>) {
+        flow.distinctUntilChangedBy { it.mode }.collect {
+            themeManager.currentTheme = it.mode
```
MatthewTighe, line 39: https://bugzilla.mozilla.org/show_bug.cgi?id=1867689

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -251,12 +258,11 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
         maybeShowSplashScreen()
 
-        // There is disk read violations on some devices such as samsung and pixel for android 9/10
-        components.strictMode.resetAfter(StrictMode.allowThreadDiskReads()) {
```
MatthewTighe, line 255: I _think_ that this was triggered by reading from `settings` on the main thread to fetch the cached PBM state and that it's safe to remove, but it's worth keeping an eye on.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -868,14 +875,18 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
...
         intent?.toSafeIntent()?.let {
             if (it.hasExtra(PRIVATE_BROWSING_MODE)) {
                 val startPrivateMode = it.getBooleanExtra(PRIVATE_BROWSING_MODE, false)
-                return BrowsingMode.fromBoolean(isPrivate = startPrivateMode)
+                // Since the mode is initially set from settings during AppAction.Init, this action
+                // will overwrite the state once the action is processed in the queue if called from
+                // onCreate.
+                components.appStore.dispatch(
+                    AppAction.ModeChange(BrowsingMode.fromBoolean(isPrivate = startPrivateMode)),
+                )
```
MatthewTighe, line 887: There is a bit of a race condition with the UI here, but I don't anticipate that it will result in any (more) perceivable jank. Plus, now we aren't blocking the main thread while we read from settings so this may end up being an improvement in reactivity

fenix/app/src/main/java/org/mozilla/fenix/components/AppStore.kt
```diff
@@ -19,4 +19,8 @@ import org.mozilla.fenix.components.appstate.AppStoreReducer
 class AppStore(
     initialState: AppState = AppState(),
     middlewares: List<Middleware<AppState, AppAction>> = emptyList(),
-) : Store<AppState, AppAction>(initialState, AppStoreReducer::reduce, middlewares)
+) : Store<AppState, AppAction>(initialState, AppStoreReducer::reduce, middlewares) {
+    init {
+        dispatch(AppAction.Init)
+    }
```
MatthewTighe, line 25: This is a pattern copied from `BrowserStore` and something I expect will be commonplace as we introduce additional Stores

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -1391,8 +1392,9 @@ abstract class BaseBrowserFragment :
      */
     @VisibleForTesting
     internal fun updateThemeForSession(session: SessionState) {
-        val sessionMode = BrowsingMode.fromBoolean(session.content.private)
-        (activity as HomeActivity).browsingModeManager.mode = sessionMode
+        requireComponents.appStore.dispatch(
+            AppAction.ModeChange(BrowsingMode.fromBoolean(session.content.private)),
+        )
```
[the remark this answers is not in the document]
MatthewTighe: The selected tab is `BrowserState`, not `AppState`, so we'd need to create a middleware to dispatch an action from `BrowserStore` to `AppStore`. That doesn't seem like a huge win over what is basically an inline state binding, but I'm happy to file a follow-up if you'd like

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -30,6 +30,13 @@ import org.mozilla.fenix.wallpapers.Wallpaper
  * [Action] implementation related to [AppStore].
  */
 sealed class AppAction : Action {
+    /**
+     * [AppAction] dispatched to indicate that the store is initialized and
+     * ready to use. This action is dispatched automatically before any other
+     * action is processed. Its main purpose is to trigger initialization logic
+     * in middlewares. The action itself should have no effect on the [AppState].
+     */
+    object Init : AppAction()
```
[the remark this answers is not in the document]
MatthewTighe: Looking at `BrowserStore`, the `InitAction` seems to be more of an object lifecycle than Android lifecycle thing that indicates that the middlewares should do anything they need to fetch initial state.

Personally, I think it adds a little more clarity to see the `InitAction` dispatched inline from the Store constructor rather than as a response to `FenixApplication`. Plus, I'm guessing we get stronger guarantees about action ordering if we dispatch before the constructor completes than we would otherwise.

I'm still not opposed, but if we make changes I think we should keep the two Stores' workflows in sync.

fenix/app/src/main/java/org/mozilla/fenix/home/TabCounterView.kt
```diff
@@ -26,14 +27,14 @@ import org.mozilla.fenix.tabstray.Page
...
- * @param tabCounter The [TabCounter] that will be setup with event handlers.
+ * @property context An Android [Context].
+ * @property appStore Store for reading PBM state from and dispatching updates to.
+ * @property navController [NavController] used for navigation.
+ * @property tabCounter The [TabCounter] that will be setup with event handlers.
  */
 class TabCounterView(
     private val context: Context,
-    private val browsingModeManager: BrowsingModeManager,
+    private val appStore: AppStore,
```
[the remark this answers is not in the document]
MatthewTighe: Immediate reaction without investigating more deeply: do we have any concerns about the view outliving a change to the mode? We'd then need to either pass a callback or observe the store

fenix/app/src/main/java/org/mozilla/fenix/home/intent/HomeDeepLinkIntentProcessor.kt
```diff
@@ -76,7 +79,7 @@ class HomeDeepLinkIntentProcessor(
     private fun handleDeepLinkSideEffects(deepLink: Uri) {
         when (deepLink.host) {
             "enable_private_browsing" -> {
-                activity.browsingModeManager.mode = BrowsingMode.Private
+                appStore.dispatch(AppAction.ModeChange(BrowsingMode.Private))
```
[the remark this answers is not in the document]
MatthewTighe: Yeah, good point! I was mostly aiming for 1-1 mechanical replacements so didn't think it through from the perspective.

My only (potential) concern with this advice in some places is introducing an Action that then is only used to update the PBM state and isn't later updated to also refactor the rest of the local workflow to lib-state. 

For example, in the `ReEngagementIntentProcessor` update below:
would it be a `AppAction.OpenInBrowser` action? For now, that would do nothing but update the private browsing mode

MatthewTighe:
@boek I addressed the comments that were focused on injecting just the mode into the View classes instead of the whole Store. I would like to address your other two major comments in follow-ups as we talked about offline. My understanding of them is:

1. Create a binding to observe BrowserStore and use that to update the mode in AppStore as appropriate, instead of dispatching separate actions where new tabs are created
2. Rename any renaming actions from `ModeChange` to be more reflective of the actual event taking place where appropriate

Otherwise, I do think we should land these 3 commits together, as I discovered a bug in the first where the Home Screen doesn't react correctly to switching the mode by creating a tab of a different mode in the tabs tray. This is because the first commit changes the tabs tray to update the store, but the HomeFragment doesn't yet react to it.

Does that sound good? Still open to a different plan, but that would be my suggestion.

fenix/app/src/test/java/org/mozilla/fenix/PrivateBrowsingBindingTest.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @Test
+    fun `GIVEN screenshots not allowed in private mode WHEN mode changes to private THEN secure flag added to window`() = runTestOnMain {
+        whenever(window.addFlags(anyInt())).then { }
+        whenever(settings.allowScreenshotsInPrivateMode).thenReturn(false)
+
+        appStore.dispatch(AppAction.ModeChange(BrowsingMode.Private)).joinBlocking()
+//        this.runCurrent()
+//        this.advanceUntilIdle()
```
gabrielluong, line 82: Just a friendly reminder to clean up these commented lines up depending on what is needed at the end.

MatthewTighe: Whoops, thanks 😅

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarController.kt
```diff
@@ -174,13 +177,13 @@ class DefaultBrowserToolbarController(
                 }
             }
             is TabCounterMenu.Item.NewTab -> {
-                activity.browsingModeManager.mode = BrowsingMode.Normal
+                appStore.dispatch(AppAction.ModeChange(BrowsingMode.Normal))
```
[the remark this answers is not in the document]
MatthewTighe: Just a heads up that some form of this is still required, because this button doesn't actually create a new tab so the binding to the browser store won't cover it. This just navigates to the home screen in the expected mode.

fenix/app/src/main/java/org/mozilla/fenix/library/history/HistoryFragment.kt
MatthewTighe: Now, this opens a new tab which includes the browsing mode and triggers the `BrowserStoreBinding`

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -30,6 +31,28 @@ import org.mozilla.fenix.wallpapers.Wallpaper
...
+
+    /**
+     * NOTE: This action is not yet functional and will require https://bugzilla.mozilla.org/show_bug.cgi?id=1845409
+     * to be resolved. This is part of an ongoing lib-state refactor.
```
MatthewTighe, line 44: I'm open to other suggestions here, but this felt like the appropriate name for what was occurring at callsites. They tend to dispatch this action, and then call `HomeActivity::openToBrowserAndLoad`. Ideally, that function will be handled by middleware that react to this action in an upcoming patch.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppStoreReducer.kt
```diff
@@ -100,6 +106,9 @@ internal object AppStoreReducer {
                 else -> state.recentSyncedTabState
             },
         )
+        is AppAction.SelectedTabChanged -> state.copy(
```
MatthewTighe, line 109: I'm not sure the best way to reconcile this with `OpenTabInBrowser`. We need to be able to react to TabSessionState changes through `SelectedTabChanged` because there are things outside the Redux model that can cause updates, like `TabsUseCases`. 

However, eventually `OpenTabInBrowser` will also be kicking off those kind of state changes, which means we'd hit the reducer twice.

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -179,8 +177,9 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
     protected val homeActivityInitTimeStampNanoSeconds = SystemClock.elapsedRealtimeNanos()
 
     private lateinit var binding: ActivityHomeBinding
-    lateinit var themeManager: ThemeManager
-    lateinit var browsingModeManager: BrowsingModeManager
+    val themeManager by lazy {
+        createThemeManager()
```
gabrielluong, line 181: Could we simplify this by removing the `createThemeManager` function?

fenix/app/src/main/java/org/mozilla/fenix/BrowsingModeBinding.kt
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Binding to react to Private Browsing Mode changes in AppState.
+ *
+ * @param appStore AppStore to observe state changes from.
+ * @param themeManager Theme will be updated based on state changes.
+ * @param retrieveWindow Get window to update privacy flags for.
+ * @param settings Determine user settings for privacy features.
+ * @param ioDispatcher Dispatcher to launch disk reads. Exposed for test.
+ */
+class BrowsingModeBinding(
```
gabrielluong, line 30: I believe having this binding will allow us to remove the secure window flag handling in other fragments and just have a centralized place where we handle this. I am wondering if there is any plans to clean this up in other fragments afterrwards. Example can be seen in https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/settings/PrivateBrowsingFragment.kt#48-54

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -813,7 +808,7 @@ class HomeFragment : Fragment() {
 
     override fun onResume() {
         super.onResume()
-        if (browsingModeManager.mode == BrowsingMode.Private) {
+        if (requireComponents.appStore.state.mode == BrowsingMode.Private) {
```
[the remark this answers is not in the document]
MatthewTighe: :smart:

fenix/app/src/test/java/org/mozilla/fenix/home/DefaultSessionControlControllerTest.kt
```diff
@@ -1117,15 +1117,11 @@ class DefaultSessionControlControllerTest {
...
-            every { mode } returns mockk {
-                every { isPrivate } returns true
-            }
-        }
+        every { appStore.state.mode } returns BrowsingMode.Private
         val wallpaperState = WallpaperState.default.copy(
             availableWallpapers = makeFakeRemoteWallpapers(
                 THUMBNAILS_SELECTION_COUNT,
-                true,
+                false,
```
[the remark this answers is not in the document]
MatthewTighe: I don't remember, but it's possible I thought that the test would be more accurate that way. Looking at the implementation, it doesn't make an actual difference so I'll swap it back

fenix/app/src/main/java/org/mozilla/fenix/BrowsingModeBinding.kt
MatthewTighe: At a glance it does look like this is at least worth a slightly more thorough look, so I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1872945

MatthewTighe:
Addressed the latest round of feedback. I did just slightly more refactoring with the introduction of `WallpaperAction.OpenToHome` by creating a `WallpaperReducer` since there were a handful of other wallpapers actions.

fenix/app/src/main/java/org/mozilla/fenix/BrowsingModeBinding.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.mozilla.fenix.browser.browsingmode.BrowsingMode
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.appstate.AppState
+import org.mozilla.fenix.theme.ThemeManager
+import org.mozilla.fenix.utils.Settings
+
+/**
+ * Binding to react to Private Browsing Mode changes in AppState.
+ *
+ * @param appStore AppStore to observe state changes from.
```
gabrielluong, line 24:
```suggestion
 * Binding to react to Private Browsing Mode changes in [AppState].
 *
 * @param appStore [AppStore] to observe state changes from.
```
Let's make these references linkable(or able to jump to definition) where possible. This will also help us refactor things if we ever rename them.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
gabrielluong, line 44: Seeing TODO type comments usually gives me a bit of a pause, couple thoughts here are that we may have grown this PR a bit too big in terms of refactoring considering some of the changes are not yet functional and also seeing the additional wallpaper reducer refactoring. It would probably be better to not land non-functional changes because (1) they are only half-functional in this case and increases the maintenance burden if they were fully non-functional, and (2) there's no guarantee we will get to reference bug any time soon and can cause confusion.

Now I don't really know for sure if this is actually non-functional given the fact that I see this in the reducer, but I am assuming we actually want to open a tab instead of just changing the mode. 

I would encourage future work to try and avoid doing too much in a big PR.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
```diff
@@ -251,4 +284,44 @@ sealed class AppAction : Action {
...
+        /**
+         * App should be opened to the home screen in [BrowsingMode.Normal].
+         */
+        object NewTab : TabsTrayAction()
+
+        /**
+         * App should be opened to the home screen in [BrowsingMode.Private].
+         */
+        object NewPrivateTab : TabsTrayAction()
+    }
```
gabrielluong, line 316: Comments might be too generic here since there's no direct reference to what is happening from Toolbar or Tabs Tray.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
```diff
@@ -64,6 +65,7 @@ data class AppState(
     val collections: List<TabCollection> = emptyList(),
     val expandedCollections: Set<Long> = emptySet(),
     val mode: BrowsingMode = BrowsingMode.Normal,
+    val selectedTabId: String? = null,
```
gabrielluong, line 68: Should we really have this in the `AppState` if it's already known in the `BrowserStore`?

fenix/app/src/main/java/org/mozilla/fenix/home/TabCounterView.kt
```diff
@@ -27,30 +26,32 @@ import org.mozilla.fenix.tabstray.Page
  * Helper class for building the [FenixTabCounterMenu].
  *
  * @param context An Android [Context].
- * @param browsingModeManager [BrowsingModeManager] used for fetching the current browsing mode.
  * @param navController [NavController] used for navigation.
  * @param tabCounter The [TabCounter] that will be setup with event handlers.
+ * @param mode The current [BrowsingMode].
+ * @param itemTapped Callback to update the [BrowsingMode].
```
gabrielluong, line 32:
```suggestion
 * @param onBrowsingModeChange Callback to update the [BrowsingMode].
```
Maybe `onBrowsingModeChange` to keep with how we name our lambdas

fenix/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
```diff
@@ -195,8 +195,6 @@ class DefaultBookmarkController(
                         startLoading = load,
                     )
                 }
-                activity.browsingModeManager.mode =
-                    BrowsingMode.fromBoolean(mode == BrowsingMode.Private)
```
gabrielluong, line 199: Is the browsing mode change here handled anywhere? You had comments elsewhere where we have remove this redundancy, but it's not obvious to me in this file.

fenix/app/src/main/java/org/mozilla/fenix/home/TabsTrayReducer.kt
```diff
@@ -0,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.home
```
gabrielluong, line 5: Maybe these reducers should be in the same package as the appstate (maybe something like appstate/home)?

fenix/app/src/main/java/org/mozilla/fenix/home/intent/IntentReducer.kt
```diff
@@ -0,0 +1,22 @@
...
+package org.mozilla.fenix.home.intent
+
+import org.mozilla.fenix.browser.browsingmode.BrowsingMode
+import org.mozilla.fenix.components.appstate.AppAction.IntentAction
+import org.mozilla.fenix.components.appstate.AppState
+
+/**
+ * Reducer to handle updating [AppState] with the result of an [IntentAction].
+ */
+object IntentReducer {
```
gabrielluong, line 14: Same thing about the packaging

fenix/app/src/main/java/org/mozilla/fenix/library/LibraryPageFragment.kt
```diff
@@ -32,8 +30,6 @@ abstract class LibraryPageFragment<T> : Fragment() {
                     tabsUseCases.addTab.invoke(url, private = private)
                 }
         }
-
-        (activity as HomeActivity).browsingModeManager.mode = BrowsingMode.fromBoolean(private)
```
gabrielluong, line 36: Just a bit dizzy from reading all the code - I am just not seeing any use case file changes in the PR. Where do we handle the browsing mode updates when it goes through the use cases?

MatthewTighe: Through the `BrowserStoreBinding`. The tabs use cases will update the selected session id, which will propagate to a mode change in the `AppStore` through the binding.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
MatthewTighe: Good question! This is required for the line in the binding [here](https://github.com/mozilla-mobile/firefox-android/blob/cfb8dc0f934a06be117e8d5d15c2e599275c2eac/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/bindings/BrowserStoreBinding.kt#L31). Since `appState.mode` and `browserState.selectedTabId` are are not necessarily in sync at all times. For example, if you navigate to a tab, go home, and click the PBM button the selected tab will still be set to the tab, but the mode will be updated. 

This means that we need a way to track actual changes to the selected tab, so we can prevent lifecycle events from recreating the binding, triggering the initial observation, and overwriting the `appState.mode` with the mode from the selected tab. 

I'm open to other suggestions on how to avoid that, but this was what I came up.

I am saving this as an example for why we should move to a single store model eventually, as we could naturally avoid this problem under that paradigm.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppAction.kt
MatthewTighe: I do generally agree with all of the above of course. You are right that the action is not entirely non-functional, since it's used in the reducer. I can update the comment accordingly if you'd like. 

But yes, I would not consider this fully functional because it does not actually trigger the logic to open a tab and navigate to the browser (yet). The unfortunate truth is that much of this refactoring work is deeply inter-related, so I'm not sure there is an ordering of work that will let us avoid temporary hackiness like this. I actually started on the `BrowsingMode` stuff because it was used in `HomeActivity::openToBrowserAndLoad` and I thought refactoring that method would be easier if this state already lived in the Store.

Finally, on the size of the PR. I do completely agree, which is why this patch was initially split into 4 commits with the offer to split into multiple patches if reviewers desired. However, I do think that small patches are intended to accomplish the following:

1. reduce risk of unintended breakage
2. allow simple backouts
3. ease review burden

For the first, I think one benefit by landing something of this size is that it allows QA to do a thorough smokepass of the feature entirely, instead of expecting them to do so regularly as individual patches are landed.

For the second, I do think that lots of small patches would quickly become difficult to revert individually, because I think there is risk of small follow-up updates being included in later patches. That can be attributed to engineer error, but it's still a real risk. 

For the last, @boek and I had come to some agreements to just go ahead and do the thing while we were here.

fenix/app/src/main/java/org/mozilla/fenix/library/bookmarks/BookmarkController.kt
MatthewTighe: Same answer here as in the [below comment](https://github.com/mozilla-mobile/firefox-android/pull/4646#discussion_r1442251706)

fenix/app/src/main/java/org/mozilla/fenix/home/TabsTrayReducer.kt
MatthewTighe: I am happy to do this in the interim if you would like, but my thoughts for the long-term are that we would want state, reducers, and actions to live in individual feature modules and then be composed into the single store.  So this was intended to mirror that future, rather than group everything related to State into a single package. One issue here is that we already have a `TabsTrayReducer` for the existing `TabsTrayStore`, so we need some kind of distinction between the two for now.

What the future might look::

```
// in Tabs Tray module
data class TabsTrayState(...)

sealed class TabsTrayAction {
    object NewTab : TabsTrayAction
}

object TabsTrayReducer {
    fun reduce(...)
}

// in top level module
data class AppState(
    val tabsTrayState: TabsTrayState
)

// need a dedicated wrapping action so that we can still have a sealed class and compile-time exhaustiveness checking
sealed class AppAction {
    data class TabsTrayAppAction(val inner: TabsTrayAction)
}

object AppReducer {
    fun reduce(state: AppState, action: AppAction): AppState = when (action) {
        is TabsTrayAppAction -> TabsTrayReducer.reduce(action.inner)
    }
}
```

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
gabrielluong: Just looking at the [comment](https://github.com/mozilla-mobile/firefox-android/blob/cfb8dc0f934a06be117e8d5d15c2e599275c2eac/fenix/app/src/main/java/org/mozilla/fenix/components/appstate/bindings/BrowserStoreBinding.kt#L29-30) about re-emission, I am wondering what did you observe from re-emission.

fenix/app/src/main/java/org/mozilla/fenix/home/TabsTrayReducer.kt
gabrielluong: **Aside from the actual PR:** I am curious if this is something that you tried prototyping to validate that it will work. I must've tried to do something similar with these Stores many years ago right before pocmo left along with Christian to see if we could have a single Store, but couldn't get it to work. I don't really remember what it is anymore, but the general idea I am trying to convey here is to make sure said high-level architecture could work.

MatthewTighe: I did some prototyping earlier last year, but unfortunately lost it when my workstation was destroyed. I do firmly believe it's possible and is just a matter of improving our existing infrastructure. It's a high background priority for me, and would definitely expect a robust prototype and RFC stage before we made any commitments to move forward with something like that.

fenix/app/src/main/java/org/mozilla/fenix/components/appstate/AppState.kt
MozillaNoah: if `AppState.selectedTabId` is safe to consume, some of the `TabsTray` Composable APIs could be simplified as a result

MatthewTighe: @MozillaNoah Would be curious for more details there, but I would hesitate to suggest rushing for replacements just yet. Long-term, we may want to combine these Stores instead of replicate this state between them

gabrielluong: I agree let's not rush this. 

> Long-term, we may want to combine these Stores instead of replicate this state between them
I think I would like to see more details here because I simply don't see how we want consumer application state inside of `BrowserStore`, which could be consumed by any external app.

MozillaNoah: TLDR: `BrowserStore` is being passed around as a dependency for the tab lists _solely_ for obtaining the `selectedTabId`, making the root `TabsTray` composable dependent on three stores 😕

gabrielluong: How come we don't just pass down only the `selectedTabId`?

PR #4699, opened 2023-12-04
Bug 1867698 – add a basic nav bar composable

mavduevskiy:
### Screenshots
 Old | New
 --- | --- 
 |  

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1867698


fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -436,6 +438,13 @@ abstract class BaseBrowserFragment :
             lifecycleOwner = viewLifecycleOwner,
         )
 
+        if (IncompleteRedesignToolbarFeature(context.settings()).isEnabled) {
+            NavigationBarView(
+                context = context,
+                container = binding.browserLayout,
```
gabrielluong, line 444: Implementing this in Compose seems okay to me, but I want to look slightly ahead to make sure our bases are covered. I am wondering if we'll also have an appropriate binding for this when the awesomebar toolbar is also on the [bottom placement](https://www.figma.com/file/8fsCvtnweBkWGZ9NfL907l/Toolbar-Redesign?node-id=6150%3A377941&mode=dev).

Screen Shot 2023-12-05 at 12 12 52 PM

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
```diff
@@ -0,0 +1,165 @@
...
+                            Icon(
+                                painter = painterResource(it.iconId),
+                                contentDescription = it.label,
+                                tint = FirefoxTheme.colors.iconPrimary,
+                            )
+                        }
+                    }
+                    ItemType.TAB_COUNTER -> {
+                        IconButton(onClick = {}) {
+                            TabsIcon(it, 0)
```
gabrielluong, line 67: nit: add named parameters

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
```diff
@@ -0,0 +1,165 @@
...
+    STANDARD, TAB_COUNTER
+}
+
+/**
+ * Provides a collection of standard navigation items used in the application's navigation bar.
+ */
+object StandardNavigationItems {
+    val home = ActionItem(
+        iconId = R.drawable.mozac_ic_home_24,
+        label = "Home",
```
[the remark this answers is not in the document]
mavduevskiy: I went with a bit different name of `descriptionResourceId`, since we have to access string resources during compose phase and pass it not as a `string` but as an `id`

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
mavduevskiy: Is it okay to keep it for now? Right now, the placing of the compose element into xml layout is more of a proof of concept. The proper implementation will be done in this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1868260), where we will have to either integrate nav bar into toolbar view or introduce some form of a container to handle hiding/showing logic.

gabrielluong: > The proper implementation will be done in this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1868260), where we will have to either integrate nav bar into toolbar view or introduce some form of a container to handle hiding/showing logic.

This is more of where I want the conversation headed, which is how nicely or easily does this integrate with the toolbar view. 

> Is it okay to keep it for now?

I think whether or not we should land this - I think yes we could, but there are some considerations depending on the above conservation. If this compose implementation is only intended as a proof of concept and may not work well with the AC toolbar view, then I might suggest that instead of landing this into `main`, we could create a remote development branch of `main` to help prototype and have you and @HarrisonOg work off of that branch until things are ready.

This will help prevent having reviews until there is sufficient confidence that the implementation where we want it to be while allowing y'all to iterate quickly until you arrive at that point. We have done this in the past where we have to redesign the status bar where it was better to work off of a development branch until we know things are where it needs to be.

Ultimately, I think we are a bit nervous about how this will integrate with the existing toolbar view and what changes that will involve. So, I am expecting some sort of proposal about how we'll approach that, but it sounds like we need more time to prototype.

HarrisonOg: I like this idea of making a separate remote development branch we can rebase on `main` as needed.  This means we can get rid of that secondary "incomplete" feature flag.  It'll also ensure we don't push confusing partial behaviors to Nightly.

How would we handle moving whole components over to `main`?
@gabrielluong

gabrielluong: I think as soon as you reach a point where you know the proposed implementation in the development branch is gonna be the one that will ship. You should start re-landing the components that are ready for the redesign into `main`. As long as you're still opening PRs that are still reviewable, we will avoid reviewing a huge PR altogether when you are . So just remember to keep breaking things down, open individual PRs for the bugs that you have filed, help each other do spot reviews as you continue to land things in the development branch. You can even just open PRs and have each other review it going into the development branch. The review in this case could be a bit relaxed knowing this is to get towards a proof of concept.

If we could avoid landing prototype code into `main`, that would be preferred because there are just other ways around it.

mavduevskiy: It was a proof of concept that the Composable can be integrated into the xml. There are some areas that I am slightly concerned about (ex, integrating menu button and tab counter), but I decided to try to do that in compose since we generally add new elements in compose. But if it proves to be too much work, I might switch to the xml implementation.

I also see the benefits of having a remote branch. But the way I see it, there should be multiple dev branches then – one for nav bar (stuff I am working on), one for toolbar redesign. They are not really affecting each other in terms of that might be landed separately into main, but they shouldn't appear in nightly without each other. So, I would still suggest keeping the two flags that @HarrisonOg has introduced.

So, I am going to create a branch "navigation_bar" on main, and will be merging stuff there until we are ready to merge it into main. It's going to be a pretty bit patch in the end, right? I guess I need more guidance how to do this dev branch work @gabrielluong

gabrielluong: > It's going to be a pretty bit patch in the end, right? I guess I need more guidance how to do this dev branch work 

Sure, I can provide thoughts on how to make this manage-able.

mavduevskiy: Now that we discussed the risks of doing nav bar in compose, I wonder if the risk is worth it. It's a pretty simple element that holds a bunch of buttons that have already good implementation in android view style. I guess I had a feeling I have to follow the rule of doing new ui elements in compose, but it might be sensible to make an exception for this case? what do you think, @gabrielluong ?

I would still consider taking your approach with a dev brunch, to make smaller commits to for ease of reviewing. The only concern I have is merging, we will be merging the dev branch into main in one piece, right?

HarrisonOg: if we're moving to a dev branch for intermediate steps, I can remove the `IncompleteRedesignToolbarFeature` feature flag so we only have one feature flag going forward for the entire toolbar redesign.

HarrisonOg (approved):
looks good, ready to merge into the dev branch

mavduevskiy:
I will wait a bit before landing it, giving myself some time to come up with a draft that will determine how feasible going for compose implementation is. Tabs counter, menu and scrolling behaviour are done with android views, and, if integrating those proves itself to be too much work, we might just go with the xml nav bar for now.

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
mavduevskiy: @gabrielluong 
Here is a [draft](https://github.com/mozilla-mobile/firefox-android/pull/4887) addressing the compose integration concerns. Please let me know what you think. I will be moving forward with this PR if there is no hard nos.

gabrielluong: Since this PR is just landing in the dev_navigation_bar branch, please feel free to proceed with moving forward with landing this PR.

gabrielluong: To go a bit further, I think you should always expect to always be able to land on your dev branch. The purpose of it is for your team to make as much progress as possible to come up with the strategy that you can propose that will make the most effective sense for what we are trying to accomplish. We can leave the discussions to the other PRs and move forward with this in the immediate step.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
```diff
@@ -0,0 +1,166 @@
...
+ */
+@Composable
+fun NavigationBar(actionItems: List<ActionItem>) {
+    Box(
+        modifier = Modifier
+            .background(FirefoxTheme.colors.layer1)
+            .height(48.dp)
+            .fillMaxWidth(),
+    ) {
+        Divider(color = FirefoxTheme.colors.borderPrimary)
```
gabrielluong, line 46: We should prefer to use our design components https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/Divider.kt#28

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
```diff
@@ -0,0 +1,166 @@
...
+ * [TAB_COUNTER] - Represents a specialized item used to display a count, such as the number of open tabs in a browser.
+ */
+enum class ItemType {
+    STANDARD, TAB_COUNTER
+}
+
+/**
+ * Provides a collection of standard navigation items used in the application's navigation bar.
+ */
+object StandardNavigationItems {
```
gabrielluong, line 122:
```suggestion
object NavigationItems {
```
Perhaps a simpler name here unless there's a different use case where the navbar becomes non-standard.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
```diff
@@ -0,0 +1,166 @@
...
+            painter = painterResource(id = item.iconId),
+            contentDescription = stringResource(id = item.descriptionResourceId),
+            tint = FirefoxTheme.colors.iconPrimary,
+        )
+
+        Text(
+            text = tabsCount.toString(),
+            color = FirefoxTheme.colors.iconPrimary,
+            fontSize = 10.sp,
+            fontWeight = FontWeight.W700,
```
gabrielluong, line 90: Wondering if this patches with any of our typography?

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
mavduevskiy: Got it! I guess I got a bit confused by this [comment](https://github.com/mozilla-mobile/firefox-android/pull/4699#discussion_r1416239214) which was a direct reply to a question "if it's okay to land this as it is". I thought the answer was no and I need to come up with a working prototype to be able to proceed further. Thanks for clarification :)

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
mavduevskiy: I am not sure I understand the question. What is out typography? 

The [component](https://www.figma.com/file/Kv7wWr8vH6AQrGhu9vmjdE/Mobile-Icons?type=design&node-id=221-1234&mode=design&t=5RfQ7uSnVewM1hFy-4) has the `FontWeight` of 700, and `TabsTrayBanner` uses the [same](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayBanner.kt#369) font weight for the element!

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
mavduevskiy: that's a good piece of advice, will be checking out the `compose` folder!

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
gabrielluong: That was related to whether or not it should land on `main`. To sum that comment up, it really came down to "it depends".

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
gabrielluong: I meant if this lined up with any of the defined font styles in https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/theme/FenixTypography.kt#37

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
gabrielluong: @mavduevskiy We do have quite a good coverage of our design components that are also defined in this [figma](https://www.figma.com/file/ujdiReypSypu2GhRgzArea/Android-Components?type=design&node-id=1808-16937&mode=design&t=gQbzNWuKwh4Fi8IE-0). Background: We have a design system that introduced these common design components that are used throughout the application. By writing these reusable composable components that matches these design components, it provides us a single source of truth where if a component changes (text, padding, color, etc), we will only need to edit it in one place.

I would suggest familiarizing yourself with [Button](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/button) and [ListItem](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt) as probably the most common components that we will see in any screen. If you want a better overview, you can always ping @MozillaNoah or myself to dive deeper.

PR #4713, opened 2023-12-05
Bug 1868434 - Update Firebase Cloud Messaging to version 23.3.1

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1868434


jonalmeida (approved):
Thanks! None of the changes in the log seems useful to take immediate action on as well so bumping is good enough here.

PR #4723, opened 2023-12-06
Bug 1867246 - Add placeholder image for product recommendations

DreVla:
Implemented possibility to add custom placeholder/fallback when an image is loading or fails to be loaded. By default the placeholder is set to a solid color.

| Dark mode | Light mode |
|:----:|:--:|
| fakespot_ads_placeholder_imge_dark | fakespot_ads_placeholder_imge_light |

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1867246


fenix/app/src/main/java/org/mozilla/fenix/compose/ImagesPlaceholder.kt
```diff
@@ -21,61 +25,81 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
     contentDescription: String? = null,
 ) {
-    Image(ColorPainter(FirefoxTheme.colors.layer2), contentDescription, modifier)
+    Box(
+        modifier = Modifier
+            .background(
+                color = FirefoxTheme.colors.layer3,
+                shape = RoundedCornerShape(8.dp),
+            ),
+    ) {
```
rahulsainani, line 86: Not sure this should be in the default image placeholder, seems like the intention was to just have a color?

fenix/app/src/main/java/org/mozilla/fenix/compose/Image.kt
```diff
@@ -46,9 +49,13 @@ fun Image(
     contentDescription: String? = null,
     alignment: Alignment = Alignment.Center,
     contentScale: ContentScale = ContentScale.Fit,
+    placeholder: Painter = ColorPainter(FirefoxTheme.colors.layer3),
```
rahulsainani, line 52: `Painter` is not marked as stable by the compose compiler. That means it will recompose even if the inputs have not changed for this function. 

I think a `placeholder: @Composable () -> Unit` would be good here with the default set to the existing `DefaultPlaceholder` composable

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -467,6 +472,23 @@ private fun ProductRecommendation(
...
+                            Box(
+                                modifier = Modifier
+                                    .size(productRecommendationImageSize)
+                                    .background(
+                                        color = FirefoxTheme.colors.layer3,
+                                        shape = RoundedCornerShape(8.dp),
+                                    ),
+                            ) {
+                                Image(
+                                    imageVector = ImageVector.vectorResource(id = R.drawable.ic_file_type_image),
```
rahulsainani, line 485: I'm curious if there was a specific reason to use this over `painterResource`, the docs mention:
```
This function is intended to be used for when low-level ImageVector-specific functionality is required. 
For simply displaying onscreen, the vector/bitmap-agnostic painterResource is recommended instead.
```

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
```diff
@@ -435,6 +439,7 @@ private enum class Highlight(
     ),
 }
 
+@Suppress("LongMethod")
```
rahulsainani, line 442: 💡 Let's extract `ImagePlaceholder` as a composable, and possibly not add this

fenix/app/src/main/java/org/mozilla/fenix/compose/Image.kt
```diff
@@ -35,6 +35,8 @@ import org.mozilla.fenix.theme.FirefoxTheme
  * bounds defined by the width and height.
  * @param contentScale Optional scale parameter used to determine the aspect ratio scaling to be used
  * if the bounds are a different size from the intrinsic size of the [Painter].
+ * @param placeholder Optional placeholder image used when loading or an error occurs.
+ * By default set to a solid color in [DefaultImagePlaceholder].
```
rahulsainani, line 39:
```suggestion
 * @param placeholder composable displayed when loading.
 * By default set to a solid color in [DefaultImagePlaceholder].
```

I think if we should have a separate param for `fallback`, the recommendations usecase uses the same composable in both cases but the API should be more clear with good default params IMO

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysis.kt
DreVla: I thought it would have a certain impact when scaling the vector drawable, but upon reading the [doc](https://developer.android.com/jetpack/compose/graphics/images/compare#image-vector) more carefully, it seems that it is mostly used when other customizations are needed. Changed back to painterResource.

fenix/app/src/main/java/org/mozilla/fenix/compose/Image.kt
DreVla: I am blind.. Added fallback param 👍

rahulsainani (approved):
Looks good! ✅

PR #4749, opened 2023-12-07
Bug 1868778 - Make sure the correct process type for extensions process crashes is send in "crash" pings

willdurand:
I have a patch in GeckoView to change the "process type" of crashes originating from the extensions process. This PR is required to send the same "process type" value (`content`) in the Glean _crash_ pings for these extensions process crashes, with or without the GV patch applied.

There are two reasons for this:

1. we want to allow users to submit crash reports for extensions process crashes. Currently, GV sends crash reports with `processType: FOREGROUND_CHILD`, which kinda tells `lib-crash` to do nothing because [some "view" is supposed to take care of the user prompt](https://searchfox.org/mozilla-mobile/rev/a4e93f7618084f607e89c14c9277e47d72798e7a/firefox-android/android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt#224-229). I think the assumption is that only "tab" processes are foreground child processes, which was probably true until we enabled the extensions process (which is another child content process, currently reported as a "foreground child" process).
2. the extensions process is a content (child) process as per https://firefox-source-docs.mozilla.org/dom/ipc/process_model.html#webextensions, and we want to keep that in the Glean _crash_ pings. I think the assumption that any "background child" process is a "utility" process is no longer correct - especially given that a "utility process" seem to be a specific type of process as per https://firefox-source-docs.mozilla.org/ipc/utility_process.html#utility-process - but that is probably something to address later. One example is GPU process crashes that are already reported as "background" child processes.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1868778


android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/service/GleanCrashReporterService.kt
```diff
@@ -269,15 +269,34 @@ class GleanCrashReporterService(
...
+        // crashes are occurring in other processes (like GPU and extensions) for which users shouldn't notice anything
+        // (because there shouldn't be any noticeable impact in the app and the processes will be recreated
+        // automatically).
+        val processType = when (crash.processType) {
+            Crash.NativeCodeCrash.PROCESS_TYPE_MAIN -> "main"
+
+            Crash.NativeCodeCrash.PROCESS_TYPE_BACKGROUND_CHILD -> {
+                // The extensions process is a content process as per:
+                // https://firefox-source-docs.mozilla.org/dom/ipc/process_model.html#webextensions
+                if (crash.remoteType == "extension") {
```
rocketsroger, line 285: nit: I suggest we use `when` here as well.

rocketsroger (approved):
Looks good.  Small nit.  Thanks

afranchuk (approved):
Looks good to me!

PR #4750, opened 2023-12-07
Bug 1865304 - Custom Tabs RFC demo supplement

t-p-white:
Demo for https://github.com/mozilla-mobile/firefox-android/pull/4732

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1865304

### DIFF
--- a/fenix/app/build.gradle
+++ b/fenix/app/build.gradle
@@ -68,7 +68,7 @@ android {
 
         // This allows overriding the target activity for MozillaOnline builds, which happens
         // as part of the defaultConfig below.
-        def targetActivity = "HomeActivity"
+        def targetActivity = "FenixActivity"
 
         // Build flag for "Mozilla Online" variants. See `Config.isMozillaOnline`.
         if (project.hasProperty("mozillaOnline") || gradle.hasProperty("localProperties.mozillaOnline")) {
--- a/fenix/app/src/androidTest/java/org/mozilla/fenix/glean/BaselinePingTest.kt
+++ b/fenix/app/src/androidTest/java/org/mozilla/fenix/glean/BaselinePingTest.kt
@@ -33,7 +33,7 @@ import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.mozilla.fenix.GleanMetrics.GleanBuildInfo
-import org.mozilla.fenix.HomeActivity
+import org.mozilla.fenix.FenixActivity
 import org.mozilla.fenix.R
 import org.mozilla.fenix.helpers.HomeActivityTestRule
 import org.mozilla.fenix.helpers.MockWebServerHelper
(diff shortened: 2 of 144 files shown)


fenix/app/src/main/java/org/mozilla/FenixActivity.kt
```diff
@@ -0,0 +1,13 @@
...
+import androidx.navigation.NavDestination
+import androidx.navigation.NavDirections
+import mozilla.components.support.utils.SafeIntent
+import org.mozilla.fenix.BrowserDirection
+
+interface FenixActivity {
+    fun getBreadcrumbMessage(destination: NavDestination): String
+    fun getIntentSource(intent: SafeIntent): String?
+    fun getIntentSessionId(intent: SafeIntent): String?
+    fun getNavDirections(from: BrowserDirection, customTabSessionId: String?): NavDirections?
```
MatthewTighe, line 12: Is there a plan for how these functions will be called from `ExternalAppBrowserActivity`? It looks like this was the main point of the inheritance - AFAICT these functions are only called from within `HomeActivity`, which means we would be missing that inherited functionality in `ExternalAppBrowserActivity` unless we explicitly replicate it. For example, `ExternalAppBrowserActivity` used to inherit this line from HomeActivity and only overwrite the `::getBreadcrumbMessage`. 

```
// in home activity
lifecycle.addObserver(
    BreadcrumbsRecorder(
        components.analytics.crashReporter,
        navHost.navController,
        ::getBreadcrumbMessage,
    ),
)
```

MatthewTighe, line 12: I think of these, `getNavDirections` is probably the scariest since it is called from `openToBrowser`. I hadn't considered this, but any reference to `(activity as HomeActivity).openToBrowser` as called throughout the app could imply that the `activity` is actually an `ExternalBrowserActivity`. Have you happened to investigate whether that's the case anywhere?

t-p-white: Unless I'm misunderstanding the comment, [these functions](https://github.com/mozilla-mobile/firefox-android/blob/825175e8acaddd47c1899ea5169c74049ce61043/fenix/app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserActivity.kt#L48)  were already being overridden in `ExternalAppBrowserActivity` but they haven't had to change in the file as it's still using those existing functions but getting them from `FenixActivity` instead.

t-p-white: Agreed! Good question, I'm not sure that it _would_ happen right now, but it's possible. 

My preference would be to actually strip out these methods to prevent this, as again it's likely to cause more confusion:

- `openToBrowserAndLoad`
- `openToBrowser`
- `load`
- `handleRequestDesktopMode`

I'll add this as a step in the RFC. I've added a very noddy prototype as a quick demo (commit 3) of my thoughts.

MatthewTighe: Ah, I was assuming the step of removing the `HomeActivity` inheritance from `ExternalAppBrowserActivity`. I didn't see that the example was still doing that 😅

t-p-white:
No longer up-to-date, please see individual step PRs as list in the RFC.

PR #4755, opened 2023-12-07
Bug 1868886: Fix Nightly Font Telemetry

tomrittervg:
Unfortunately, the data we got back was flawed; and I hadn't noticed: the device information was present, but the font data was just populating the errors array instead of the fonts array. The FileInputStream on live devices doesn't support reset, so it's been replaced with two separate FileInputStreams.

Also bump the integer to cause another resubmission from Nightly users, and include the submission integer.

### Pull Request checklist

- [X] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [X] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1868886

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/fonts/FontEnumerationWorker.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/metrics/fonts/FontEnumerationWorker.kt
@@ -89,6 +89,7 @@ class FontEnumerationWorker(
         val submission = JSONObject()
 
         run {
+            submission.put("submission", kDesiredSubmissions)
             submission.put("brand", Build.BRAND)
             submission.put("device", Build.DEVICE)
             submission.put("hardware", Build.HARDWARE)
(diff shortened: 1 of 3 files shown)


fenix/app/src/main/java/org/mozilla/fenix/components/metrics/fonts/FontEnumerationWorker.kt
```diff
@@ -89,6 +89,7 @@ class FontEnumerationWorker(
         val submission = JSONObject()
 
         run {
+            submission.put("submission", kDesiredSubmissions)
```
jonalmeida, line 92: Good idea!

jonalmeida (approved):
This looks good to me and the fix seems fine to land for now.

As mentioned offline, we are shipping usage of an API to production that only works in tests. So we should file a follow-up issue to track re-writing the tests in a way that matches what we are shipping.

jonalmeida:
Choosing to land this on Friday so that we get weekend data back on Monday. Since the change seems safe as we're no longer sharing a stream to read, we should be fine.

PR #4756, opened 2023-12-07
Bug 1858948 - Add shopping NotAvailable product state

calumozilla:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1858948

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/EnumMapper.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/EnumMapper.kt
@@ -5,8 +5,8 @@
 package org.mozilla.fenix.shopping.middleware
 
 /**
- * Converts a string to an enum value, ignoring case. If the string does not match any of the
- * enum values, the default value is returned.
+ * Converts a string to an enum value, ignoring case and replacing spaces with underscores.
+ * If the string does not match any of the enum values, the default value is returned.
  */
 inline fun <reified T : Enum<T>> String.asEnumOrDefault(defaultValue: T? = null): T? =
-    enumValues<T>().firstOrNull { it.name.equals(this, ignoreCase = true) } ?: defaultValue
+    enumValues<T>().firstOrNull { it.name.equals(this.replace(" ", "_"), ignoreCase = true) } ?: defaultValue
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
@@ -27,6 +27,10 @@ private fun ProductAnalysis.toProductReview(): ProductReviewState =
         } else {
             ProductReviewState.Error.GenericError
         }
+    } else if (deletedProductReported) {
+        ProductReviewState.Error.ProductAlreadyReported
+    } else if (deletedProduct) {
+        ProductReviewState.Error.ProductNotAvailable
     } else if (notEnoughReviews && !needsAnalysis) {
         ProductReviewState.Error.NotEnoughReviews
     } else {
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
@@ -56,24 +56,7 @@ class ReviewQualityCheckNetworkMiddleware(
         scope.launch {
             when (action) {
                 FetchProductAnalysis, RetryProductAnalysis -> {
-                    val productAnalysis = reviewQualityCheckService.fetchProductReview()
-                    val productReviewState = productAnalysis.toProductReviewState()
-
-                    // Here the ProductReviewState should only updated after the analysis status API
-                    // returns a result. This makes sure that the UI doesn't show the reanalyse
-                    // button in case the product analysis is already in progress on the backend.
-                    if (productReviewState.isAnalysisPresentOrNoAnalysisPresent() &&
-                        reviewQualityCheckService.analysisStatus()?.status.isPendingOrInProgress()
-                    ) {
-                        store.updateProductReviewState(productReviewState, true)
-                        store.dispatch(ReviewQualityCheckAction.RestoreReanalysis)
-                    } else {
-                        store.updateProductReviewState(productReviewState)
-                    }
-
-                    if (productReviewState is ProductReviewState.AnalysisPresent) {
-                        store.updateRecommendedProductState()
-                    }
+                    store.onFetch()
                 }
 
                 ReviewQualityCheckAction.ReanalyzeProduct,
@@ -81,6 +64,13 @@ class ReviewQualityCheckNetworkMiddleware(
                 ReviewQualityCheckAction.RestoreReanalysis,
                 -> {
                     store.onReanalyze()
+                }
+
+                ReviewQualityCheckAction.ReportProductBackInStock -> {
+                    val status = reviewQualityCheckService.reportBackInStock()
+                    if (status == ReportBackInStockStatusDto.NOT_DELETED) {
+                        store.onFetch()
+                    }
                 }
 
                 ReviewQualityCheckAction.ToggleProductRecommendation -> {
@@ -101,6 +91,27 @@ class ReviewQualityCheckNetworkMiddleware(
                     reviewQualityCheckService.recordRecommendedProductImpression(action.productAid)
                 }
             }
+        }
+    }
+
+    private suspend fun Store<ReviewQualityCheckState, ReviewQualityCheckAction>.onFetch() {
+        val productAnalysis = reviewQualityCheckService.fetchProductReview()
+        val productReviewState = productAnalysis.toProductReviewState()
+
+        // Here the ProductReviewState should only updated after the analysis status API
+        // returns a result. This makes sure that the UI doesn't show the reanalyse
+        // button in case the product analysis is already in progress on the backend.
+        if (productReviewState.isAnalysisPresentOrNoAnalysisPresent() &&
+            reviewQualityCheckService.analysisStatus()?.status.isPendingOrInProgress()
+        ) {
+            updateProductReviewState(productReviewState, true)
+            dispatch(ReviewQualityCheckAction.RestoreReanalysis)
+        } else {
+            updateProductReviewState(productReviewState)
+        }
+
+        if (productReviewState is ProductReviewState.AnalysisPresent) {
+            updateRecommendedProductState()
         }
     }
 
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
@@ -57,6 +57,13 @@ interface ReviewQualityCheckService {
      * Sends an impression attribution event for a given product aid.
      */
     suspend fun recordRecommendedProductImpression(productAid: String)
+
+    /**
+     * Reports that a product is back in stock.
+     *
+     * @return [ReportBackInStockStatusDto] if the request succeeds, null otherwise.
+     */
+    suspend fun reportBackInStock(): ReportBackInStockStatusDto?
 }
 
 /**
@@ -195,6 +202,23 @@ class DefaultReviewQualityCheckService(
             }
         }
     }
+
+    override suspend fun reportBackInStock(): ReportBackInStockStatusDto? = withContext(Dispatchers.Main) {
+        suspendCoroutine { continuation ->
+            browserStore.state.selectedTab?.let { tab ->
+                tab.engineState.engineSession?.reportBackInStock(
+                    url = tab.content.url,
+                    onResult = {
+                        continuation.resume(it.asEnumOrDefault<ReportBackInStockStatusDto>())
+                    },
+                    onException = {
+                        logger.error("Error reporting product back in stock", it)
+                        continuation.resume(null)
+                    },
+                )
+            }
+        }
+    }
 }
 
 /**
@@ -232,3 +256,23 @@ data class AnalysisStatusProgressDto(
     val status: AnalysisStatusDto,
     val progress: Double,
 )
+
+/**
+ * Enum that represents the status returned from reporting a product is back in stock.
+ */
+enum class ReportBackInStockStatusDto {
+    /**
+     * Report created.
+     */
+    REPORT_CREATED,
+
+    /**
+     * Product is already reported to be back in stock.
+     */
+    ALREADY_REPORTED,
+
+    /**
+     * Product was not actually marked as deleted.
+     */
+    NOT_DELETED,
+}
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckAction.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckAction.kt
@@ -225,5 +225,5 @@ sealed interface ReviewQualityCheckAction : Action {
     /**
      * Triggered when the user reports a product is back in stock.
      */
-    object ReportProductBackInStock : TelemetryAction
+    object ReportProductBackInStock : NetworkAction, UpdateAction, TelemetryAction
 }
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
@@ -90,6 +90,21 @@ sealed interface ReviewQualityCheckState : State {
                  * Denotes a generic error has occurred.
                  */
                 object GenericError : Error
+
+                /**
+                 * Denotes a product is not available.
+                 */
+                object ProductNotAvailable : Error
+
+                /**
+                 * Denotes current user reported a product is back in stock.
+                 */
+                object ThanksForReporting : Error
+
+                /**
+                 * Denotes another user has already reported the product is back in stock.
+                 */
+                object ProductAlreadyReported : Error
             }
 
             /**
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStore.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStore.kt
@@ -197,5 +197,17 @@ private fun mapStateForUpdateAction(
                 }
             }
         }
+
+        ReviewQualityCheckAction.ReportProductBackInStock -> {
+            state.mapIfOptedIn {
+                if (it.productReviewState is ProductReviewState.Error.ProductNotAvailable) {
+                    it.copy(
+                        productReviewState = ProductReviewState.Error.ThanksForReporting,
+                    )
+                } else {
+                    it
+                }
+            }
+        }
     }
 }
--- a/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysisError.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysisError.kt
@@ -29,6 +29,7 @@ import org.mozilla.fenix.theme.FirefoxTheme
  * Product analysis error UI
  *
  * @param error The error state to display.
+ * @param onReportBackInStockClick Invoked when the user clicks on the report back in stock button.
  * @param productRecommendationsEnabled The current state of the product recommendations toggle.
  * @param productVendor The vendor of the product.
  * @param isSettingsExpanded Whether or not the settings card is expanded.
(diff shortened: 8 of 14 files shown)


fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/EnumMapper.kt
```diff
@@ -5,8 +5,8 @@
...
 
 /**
- * Converts a string to an enum value, ignoring case. If the string does not match any of the
- * enum values, the default value is returned.
+ * Converts a string to an enum value, ignoring case and replacing spaces with underscores.
+ * If the string does not match any of the enum values, the default value is returned.
  */
 inline fun <reified T : Enum<T>> String.asEnumOrDefault(defaultValue: T? = null): T? =
-    enumValues<T>().firstOrNull { it.name.equals(this, ignoreCase = true) } ?: defaultValue
+    enumValues<T>().firstOrNull { it.name.equals(this.replace(" ", "_"), ignoreCase = true) } ?: defaultValue
```
rahulsainani, line 12: Nice! Let's add tests for this

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
```diff
@@ -29,6 +29,10 @@ private fun ProductAnalysis.toProductReview(): ProductReviewState =
         }
     } else if (notEnoughReviews && !needsAnalysis) {
         ProductReviewState.Error.NotEnoughReviews
+    } else if (deletedProduct) {
```
rahulsainani, line 32: Can this be true and `notEnoughReviews` is also true? If so, in that case, this'll map to `NotEnoughReviews`, should that happen or `NotAvailable`?

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
```diff
@@ -159,6 +159,17 @@ sealed interface ReviewQualityCheckState : State {
...
                 }
             }
+
+            /**
+             * Denotes a product is not available.
+             *
+             * @property reported Whether or not current user has reported the product is back in stock.
+             * @property anotherUserReported Whether or not another user has reported the product is back in stock.
+             */
+            data class NotAvailable(
```
rahulsainani, line 169: This should be a child of the Error sealed interface, similar to `NotEnoughReviews` and `UnsupportedProductTypeError`

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckAction.kt
```diff
@@ -214,9 +219,4 @@ sealed interface ReviewQualityCheckAction : Action {
      * Triggered when the No analysis card is displayed to the user.
      */
     object NoAnalysisDisplayed : TelemetryAction
-
-    /**
-     * Triggered when the user reports a product is back in stock.
-     */
-    object ReportProductBackInStock : TelemetryAction
```
rahulsainani, line 221: Let's use the existing action

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
```diff
@@ -216,3 +240,28 @@ enum class AnalysisStatusDto {
...
+
+    /**
+     * Product was not actually marked as deleted.
+     */
+    NOT_DELETED,
+
+    /**
+     * Any other status.
+     */
+    OTHER,
```
rahulsainani, line 266: Since this API has 3 states and we handle them specifically, we don't need to define this `Other` state IMO

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
rahulsainani: Also, Looking at the [designs](https://www.figma.com/file/VE5xayz3yOSB5qOATUU9Im/Mobile-Shopping-Experience-MVP?type=design&node-id=5384-23225&mode=design&t=l1duuWVGcjQZAxvB-0) we want to represent 3 different states, 

`ProductNotAvailable` -> When `deleted_product` is returned true by the analysis.  Shows the button also. 
`ThanksForReporting` -> When button is clicked on `ProductNotAvailable`
`ProductAlreadyReported` -> When `deleted_product_reported` is returned true by the analysis api.

With this, they become objects in the Error sealed interface, similar to other Error states and the UI is 1:1 mapped.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
rahulsainani: Based on this [comment](https://github.com/mozilla-mobile/firefox-android/pull/4756/files#r1422911249), we should also check for when `deletedProductReported` is true. The above question/clarification applies for this also.

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStore.kt
```diff
@@ -161,5 +161,17 @@ private fun mapStateForUpdateAction(
                 }
             }
         }
+
+        ReviewQualityCheckAction.ReportBackInStock -> {
+            state.mapIfOptedIn {
+                if (it.productReviewState is ProductReviewState.NotAvailable) {
+                    it.copy(
+                        productReviewState = it.productReviewState.copy(reported = true),
```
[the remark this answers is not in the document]
rahulsainani: Check [this](https://github.com/mozilla-mobile/firefox-android/pull/4756#discussion_r1422912955) comment and the comment about representing 3 states.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/EnumMapper.kt
calumozilla: Added a check in EnumMapperTest

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
calumozilla: O good point, moved up the `else if (deletedProduct)` before checking `notEnoughReviews` and also added check for `deletedProductReported`

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
calumozilla: Sure, removed `OTHER` and set the default for `asEnumOrDefault` to `NOT_DELETED`, although I'm not sure if that's the best default between the 3.

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckAction.kt
calumozilla: Updated to use action `ReportProductBackInStock`

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckState.kt
calumozilla: Updated the 3 states to be Error states

fenix/app/src/main/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStore.kt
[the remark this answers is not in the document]
calumozilla: Reopening the bottom sheet now shows correct state

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysisError.kt
```diff
@@ -129,6 +88,95 @@ fun ProductAnalysisError(
     }
 }
 
+@Suppress("LongMethod")
```
calumozilla, line 91: had to suppress longMethod, maybe the ReviewQualityCheckInfoType could be set to Info to start?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
rahulsainani: We don't need a default since API returns one of the 3 strings/enum values. If it's something else, then the service will return null which is handled in the middleware.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
```diff
@@ -83,6 +83,13 @@ class ReviewQualityCheckNetworkMiddleware(
                     store.onReanalyze()
                 }
 
+                ReviewQualityCheckAction.ReportProductBackInStock -> {
+                    val status = reviewQualityCheckService.reportBackInStock()
+                    if (status == ReportBackInStockStatusDto.REPORT_CREATED) {
+                        store.updateProductReviewState(ProductReviewState.Error.ThanksForReporting)
+                    }
```
rahulsainani, line 90: This is fine for now but we should consider a future improvement here. 

When the user clicks the button, nothing really happens on the ui till the network request succeeds with the right response and then the UI transitions to thanks for reporting. Maybe we should add a loading state here while the request is inflight.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
```diff
@@ -195,6 +202,23 @@ class DefaultReviewQualityCheckService(
...
         }
     }
+
+    override suspend fun reportBackInStock(): ReportBackInStockStatusDto? = withContext(Dispatchers.Main) {
+        suspendCoroutine { continuation ->
+            browserStore.state.selectedTab?.let { tab ->
+                tab.engineState.engineSession?.reportBackInStock(
+                    url = tab.content.url,
+                    onResult = {
+                        continuation.resume(it.asEnumOrDefault(ReportBackInStockStatusDto.NOT_DELETED))
```
rahulsainani, line 212: Let's not set a default value in `asEnumOrDefault`

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
```diff
@@ -232,3 +256,23 @@ data class AnalysisStatusProgressDto(
...
     val progress: Double,
 )
+
+/**
+ * Enum that represents the status returned from reporting a product is back in stock.
+ */
+enum class ReportBackInStockStatusDto {
+    /**
+     * Created reported.
+     */
```
rahulsainani, line 266: 😄 typo?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
rahulsainani: Or we should treat this as a success by default and update to `ThanksForReporting` immediately, and only revert back to the previous state in case of error.

rahulsainani: It seems like that's already happening in the Store reducer, which now makes me wonder why do we need `store.updateProductReviewState(ProductReviewState.Error.ThanksForReporting)` ?

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysisError.kt
rahulsainani: I see, let's keep this when logic in the above function in that case. 

We don't want to set a var and then overwrite in, it's better to be declarative and have one state.

I think Triple has fulfilled it's usecase here, let's create a data class with 5 properties that are required for the `InfoCard` composable. Then each error state maps to that object which is a param for `ReviewQualityCheckInfoCard`.

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
```diff
@@ -1282,6 +1282,55 @@ class ReviewQualityCheckStoreTest {
             assertTrue(observedState.contains(expectedProgressState3))
         }
 
+    @Test
+    fun `GIVEN analysis succeeds WHEN deletedProduct is true and deletedProductReported is false THEN product not available is shown`() =
```
rahulsainani, line 1286: GIVEN deletedProduct is true and deletedProductReported is false WHEN report product back in stock is clicked THEN thanks for reporting card is shown

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
```diff
@@ -1282,6 +1282,55 @@ class ReviewQualityCheckStoreTest {
...
+                                deletedProductReported = false,
+                            )
+                        },
+                        reanalysis = AnalysisStatusDto.PENDING,
+                        statusProgress = {
+                            AnalysisStatusProgressDto(
+                                status = AnalysisStatusDto.COMPLETED,
+                                progress = 100.0,
+                            )
+                        },
```
rahulsainani, line 1304: 🔍 These two might not need to be set for this test, seem unrelated

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ProductAnalysisMapper.kt
rahulsainani: 👌

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
rahulsainani, line 1286:
```suggestion
    fun `GIVEN deletedProduct is true and deletedProductReported is false WHEN report product back in stock is clicked THEN thanks for reporting card is shown`() =
```

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
```diff
@@ -1282,6 +1282,55 @@ class ReviewQualityCheckStoreTest {
...
+                    ),
+                    networkChecker = FakeNetworkChecker(isConnected = true),
```
rahulsainani, line 1306: Let's set report to report created so we know this only should happen when report created is returned. On returning another response, UI should react differently. OOC, what do we show when not deleted is returned?

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
calumozilla: Removed default value `it.asEnumOrDefault<ReportBackInStockStatusDto>()`

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckService.kt
calumozilla: lol updated doc to `report created`

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
calumozilla: Ya this isn't necessary, updated to `{ null }`

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysisError.kt
calumozilla: I moved the `when` logic back into the above function, but didn't add the data class, since we would still have to check if it's `Error.ProductNotAvailable` to create a ReviewQualityCheckInfoCard with button. It only seemed marginally better. But if you have something else in mind, maybe I can add as a follow up, if we're trying to get this in before code freeze.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
calumozilla: hm ya we do update to `ThanksForReporting` immediately, I removed it here

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
calumozilla: Updated to return REPORT_CREATED for this test, also added a test to return NOT_DELETED which will return no analysis state since that's what desktop is doing

fenix/app/src/main/java/org/mozilla/fenix/shopping/ui/ProductAnalysisError.kt
rahulsainani: That's fair, we can discuss that approach after landing this 👍

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
rahulsainani: I'm not sure we should not update the state to `NoAnalysisPresent()` directly here. By just doing this, we are expecting that the report product api responses are immediately in sync with the analysis api.  This case should hardly ever happen. 

I think for the `productReviewState` we should use the analysis api as the source of truth  i.e fire the Fetch action that will take care of fetching and mapping, so the mapper does the usual checks and maps to the correct state - which could be `NoAnalysisPresent`. wdyt?

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
rahulsainani: Thanks! 

For not deleted, see [this](https://github.com/mozilla-mobile/firefox-android/pull/4756#discussion_r1425158603).

rahulsainani (approved):
This is looking good and works really well! 🎉 

Please check this [comment](https://github.com/mozilla-mobile/firefox-android/pull/4756#discussion_r1425158603). It's the only thing I'm unsure of right now. It's an implementation "idea" that shouldn't change the behavior and since It's a rare edge case, we can even do this as a follow up.

fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNetworkMiddleware.kt
calumozilla: Hm okay, but won't fetching and calling the mapper still map to one of the errors Error.ProductAlreadyReported or ProductReviewState.Error.ProductNotAvailable?

rahulsainani:
```
if (status == ReportBackInStockStatusDto.NOT_DELETED) {
    // fetch and map
}
```

Since this will happen when the report api returns not deleted, I doubt that the analysis api will return a response that leads to Error.ProductAlreadyReported or ProductReviewState.Error.ProductNotAvailable. 

If that does indeed happen, it would also happen if we close and open the sheet as that would trigger the fetch and map again.

fenix/app/src/test/java/org/mozilla/fenix/shopping/store/ReviewQualityCheckStoreTest.kt
```diff
@@ -1283,6 +1284,115 @@ class ReviewQualityCheckStoreTest {
...
+
+            val expected = ReviewQualityCheckState.OptedIn(
+                productReviewState = ReviewQualityCheckState.OptedIn.ProductReviewState.Error.ProductNotAvailable,
+                productRecommendationsPreference = false,
+                productRecommendationsExposure = true,
+                productVendor = ProductVendor.BEST_BUY,
+            )
+            assertEquals(expected, tested.state)
+
+            val noAnalysisTested = ReviewQualityCheckStore(
```
rahulsainani, line 1367: As a future improvement, we can use 
```
productAnalysis = { count -> 
// initial analysis with deleted product
// subsequent analysis with no analysis 
 }  
```
as fake response instead of creating two different stores, so we're testing on the same instance of the store. 

Alternatively you can always set an initialState and then just test the logic from here on.

calumozilla: Hey, coming back to this, how would the state count update?

PR #4791, opened 2023-12-11
Bug 1869252 - GPC TestRail matching

AndiAJ:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/androidTest/java/org/mozilla/fenix/ui/EnhancedTrackingProtectionTest.kt
```diff
@@ -80,6 +80,8 @@ class EnhancedTrackingProtectionTest {
             verifyEnhancedTrackingProtectionLevelSelected("Standard (default)", true)
             verifyStandardOptionDescription()
             verifyStrictOptionDescription()
+            verifyGPCTextWithSwitchWidget()
+            verifyGPCSwitchEnabled(false)
             selectTrackingProtectionOption("Custom")
             verifyCustomTrackingProtectionSettings()
             scrollToElementByText("Standard (default)")
```
jjSDET, line 87: My concern with this test is that it's trying to do too much.

It seems to me there's value in having this broken into smaller tests for more atomic measurements for better reporting, stability, and maintenance/extension.

The typical best practice is to have tests that run in the following order (though still operate as stand-alone tests):
1. Navigate to component and verify all elements are present that should be present
2. Navigate to component and verify all elements are responsive to user interactions
3. Navigate to component and verify a single setting's functionality
4. Repeat step 3 for each setting.

Though this results in more tests, the measurements from each test are more meaningful as they represent a distinct root-cause that is immediately obvious, resulting in easier debugging.

This approach can also build on itself to programmatically create permutations of your tests based on your test data.

Your test data would be: 1. your selectors, 2. selector behavior, 3. feature/setting state, 4. expected outcome based on feature/setting state.

Verifying the elements exist on the page and are also responsive depending on the type of element could be condensed into a single test, though having a separate suite that verifies elements on pages/components is an ideal test suite for self-discoverability, followed by a self-discoverable test suite for verifying the ui components at least visually respond to user feedback.

Both of those suites would likely be the 2nd and 3rd UI test suites run before more complex tests that include possible collisions of functionality and state as this appears to do.

These are considerations for how we can be breaking our tests up so once we start implementing a self-discoverable pattern, the underlying helpers will already be present to support them.

This need not be refactored during this PR, but we should create a Jira ticket or work item to make sure to start addressing overly ambitious tests in the future.

jjSDET (approved):
LGTM 👍

sv-ohorvath (approved):

PR #4795, opened 2023-12-11
Bug 1869134 - Don't crash on SyncScopedKeyMissingInServerResponse

bendk:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1869134

### DIFF
--- a/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
+++ b/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
@@ -31,6 +31,7 @@ import mozilla.components.service.fxa.AccountOnDisk
 import mozilla.components.service.fxa.AccountStorage
 import mozilla.components.service.fxa.FxaAuthData
 import mozilla.components.service.fxa.FxaDeviceSettingsCache
+import mozilla.components.service.fxa.FxaSyncScopedKeyMissingException
 import mozilla.components.service.fxa.Result
 import mozilla.components.service.fxa.SecureAbove22AccountStorage
 import mozilla.components.service.fxa.ServerConfig
@@ -756,7 +757,23 @@ open class FxaAccountManager(
             return
         }
 
-        val accessToken = account.getAccessToken(SCOPE_SYNC)
+        val accessToken = try {
+            account.getAccessToken(SCOPE_SYNC)
+        } catch (e: FxaSyncScopedKeyMissingException) {
+            // We received an access token, but no sync key which means we can't really use the
+            // connected FxA account.  Throw an exception so that the account transitions to the
+            // `AuthenticationProblem` state.  Things should be fixed when the user re-logs in.
+            //
+            // This used to be thrown when the android-components code noticed the issue in
+            // `asSyncAuthInfo()`.  However, the application-services code now also checks for this
+            // and throws its own error.  To keep the flow above this the same, we catch the
+            // app-services exception and throw the android-components one.
+            //
+            // Eventually, we should remove AccessTokenUnexpectedlyWithoutKey and have the higher
+            // functions catch `FxaSyncScopedKeyMissingException` directly
+            // (https://bugzilla.mozilla.org/show_bug.cgi?id=1869862)
+            throw AccessTokenUnexpectedlyWithoutKey()
+        }
         val tokenServerUrl = if (accessToken != null) {
             // Only try to get the endpoint if we have an access token.
             account.getTokenServerEndpointURL()
--- a/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Exceptions.kt
+++ b/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Exceptions.kt
@@ -37,6 +37,12 @@ typealias FxaOriginMismatchException = mozilla.appservices.fxaclient.FxaExceptio
 typealias FxaNoExistingAuthFlow = mozilla.appservices.fxaclient.FxaException.NoExistingAuthFlow
 
 /**
+ * Thrown when a scoped key was missing in the server response when requesting the OLD_SYNC scope.
+ */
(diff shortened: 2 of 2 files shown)


bendk:
Looks like this is a crasher in beta, should we uplift the fix?

Also, should we merge https://github.com/mozilla-mobile/firefox-android/pull/4608/ before the next release day?

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -256,12 +256,7 @@ open class FxaAccountManager(
                 // All good, request a sync.
                 AccountState.Authenticated -> {
                     // Make sure auth cache is populated before we try to sync.
-                    try {
-                        maybeUpdateSyncAuthInfoCache()
-                    } catch (e: AccessTokenUnexpectedlyWithoutKey) {
-                        crashReporter?.submitCaughtException(
-                            AccountManagerException.MissingKeyFromSyncScopedAccessToken("syncNow"),
-                        )
+                    if (!maybeUpdateSyncAuthInfoCache("syncNow")) {
```
bendk, line 259: One difference is that this code path will be taken whenever `getAccessToken()` fails, but I think that's okay.

In practice, I think the only way it will fail is if the sync key is missing.  In theory, there could be other ways for it to fail, but handling those as "access token key errors" seems fine.

mavduevskiy: Does `getCcessToken` throw the exception as well? My impression is that the real culprit is [here](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/Types.kt#80).

We are also changing the logic down the line:
`if (accessToken != null && tokenServerUrl != null) {
...
}  esle {
..
return false
}`

So if `accessToken` is null or `tokenServerUrl` we will during the cache update, now we are firing `processQueue(Event.Account.AccessTokenKeyError)`. Two questions come to my mind, if it's intentional and if the refactoring belongs to a patch called "don't crash on Don't crash on SyncScopedKeyMissingInServerResponse" :)

bendk: `FxaClient.getAccessToken()` from application-services can throw .  `FirefoxAccount.getAccessToken()`  from android-components wraps the call with `handleFxaExceptions()` so some exceptions will result in `null` being returned instead of throwing -- but only those where `FxaException.shouldPropogate()` return `false`.

The intention is to keep the same behavior as before in Fx120:
- **Fx120:**
    - The application-services code didn't check for missing sync keys
    - The android-components code would throw `AccessTokenUnexpectedlyWithoutKey` when it noticed the sync key was missing.
    - This codepath would be taken
- **Fx121 (current):**
   - The application service code does check for missing sync keys and throws `FxaSyncScopedKeyMissingException` which crashes
- **Fx121 (this PR)**: 
  - The application service code does check for missing sync keys and throws `FxaSyncScopedKeyMissingException` but it no longer crashes because `FxaException.shouldPropagate` returns `false` now.
  - `maybeUpdateSyncAuthInfoCache` notices the `null` access token and returns `false`
  - This code sees the `false` return and continues down the same codepath as before.

One issue with that is that if `FxaClient.getAccessToken()` throws a different exception, or if `FxaClient.getTokenServerEndpointURL` throws then it will be handled exactly like a missing key error, which would be a behavior change.  I think that there are some codepaths that do that, but we should never hit them  in practice.  I also don't see why the new behavior would be bad.

I wish there was a way to limit the changes more, but this was the best one I could think of.  The main alternative that I can see is to not add `FxaSyncScopedKeyMissingException` to the list of exceptions that shouldn't propagate and catch them in `maybeUpdateSyncAuthInfoCache`  instead.  However, that means if any other parts of the code calls `getAccessToken` they would need to catch that exception or crash.

mavduevskiy: Thanks for explanation! The patch now makes much more sense.

I am actually leaning more towards the alternative solution that you mentioned. 
1) It will be a smaller change (we only need to change the error type we are catching); 
2) we won't be introducing the risk of different branching of business logic (previously a `null` token or `tokenServerUrl` [wouldn't](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#769) throw and exception, it would just log the error, and the catch block with [this](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt#265) action won't be executed; in the patch `null` values return `false` and will trigger the code that used to be part of the catch block);
3) we won't be introducing the risk of misreading the exceptions and take something different as a missing key error (debugging something like that will be a nightmare);
4) The code will better represent what's actually happening: we are expecting the exception to be thrown and we are handling that case.

The last part also helps with understanding what errors we should expect from A-C. We might want to update the [doc](https://searchfox.org/mozilla-mobile/source/application-services/components/fxa-client/android/src/main/java/mozilla/appservices/fxaclient/FxaClient.kt#256) as well.

bendk: Makes sense to me.  I just updated the PR based on that.

bendk: Also good point on the docstring.  I made a PR for that here: https://github.com/mozilla/application-services/pull/5987

mavduevskiy (approved):
🚢 it! well, in this case up🏋️ as well :)

bendk:
@Mergifyio backport releases_v121

PR #4796, opened 2023-12-11
Bug 1868993 - Addons local file installation's cacheDir is not cleaned.

Amejia481:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1868993


android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -69,7 +74,11 @@ class AddonFilePicker(
         }
     }
 
-    internal fun convertToFileUri(uri: Uri): String = uri.toFileUri(context, "XPIs").toString()
+    internal fun removeTemporaryFiles() {
+        File(context.cacheDir, XPI_DIR).deleteRecursively()
```
willdurand, line 78: Why do we need the recursive behavior here? Can we use `delete()` instead?

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
```diff
@@ -92,4 +94,46 @@ class AddonFilePickerTest {
...
+            onError = any(),
+        )
+
+        onSuccessCallbackCapture.value.invoke(mock())
+
+        verify(filePicker).removeTemporaryFiles()
+    }
+
+    @Test
+    fun `WHEN there is an error installing THEN remove the temporary files`() {
```
willdurand, line 120:
```suggestion
    fun `WHEN there is an error during installation THEN remove the temporary files`() {
```

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
```diff
@@ -92,4 +94,46 @@ class AddonFilePickerTest {
             onError = any(),
         )
     }
+
+    @Test
+    fun `WHEN after the addon is installed correctly THEN remove the temporary files`() {
```
willdurand, line 99:
```suggestion
    fun `WHEN an addon is successfully installed THEN remove the temporary files`() {
```

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
Amejia481: As we are deleting a directory `delete()` doesn't work.

Amejia481:
Thanks for the review!
I updated the PR with your suggestions!

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
willdurand: Wait, why are we deleting a directory here? Could we delete the file instead?

Amejia481: If we delete the actual file, we won't be able deleting the previous stored files, side note: the directory is created every time we create a file if it's not required.

willdurand: I think I am fine with the trade-off, I'd rather delete a file than an entire directory given users can clear their cache to clear everything and this feature landed recently.

Amejia481: 👍🏽  I updated the patch!

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
```diff
@@ -52,13 +56,15 @@ class AddonFilePicker(
     }
 
     internal fun handleUriSelected(uri: Uri?) {
-        uri?.let {
+        uri?.let { safeUri ->
             val fileUri = convertToFileUri(uri)
```
willdurand, line 60: Can you help me understand why we have:

- `uri`
- `safeUri`
- `fileUri`

in these two lines?

Amejia481: Sure, 

`uri` -> It's the null able uri we get from Andriod after selecting the file.
`safeUri` -> it's non null version, which comes from `let`.
`fileUri`-> it's a string file `uri` version that comes from converting `uri` a file version.

willdurand: Why don't we pass `safeUri` to `convertToFileUri()`?

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
```diff
@@ -104,4 +106,46 @@ class AddonFilePickerTest {
...
+        verify(addonManager).installAddon(
+            url = any<String>(),
+            installationMethod = eq(InstallationMethod.FROM_FILE),
+            onSuccess = onSuccessCallbackCapture.capture(),
+            onError = any(),
+        )
+
+        onSuccessCallbackCapture.value.invoke(mock())
+
+        verify(filePicker).removeTemporaryFile(any())
```
willdurand, line 128: Let's verify that we call `removeTemporaryFile()` with the safe URI instead of `any()`.

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
```diff
@@ -104,4 +106,46 @@ class AddonFilePickerTest {
...
+        verify(addonManager).installAddon(
+            url = any<String>(),
+            installationMethod = eq(InstallationMethod.FROM_FILE),
+            onSuccess = any(),
+            onError = onErrorCapture.capture(),
+        )
+
+        onErrorCapture.value.invoke(mock())
+
+        verify(filePicker).removeTemporaryFile(any())
```
willdurand, line 149: ditto

willdurand (approved):
r+wc, thanks

android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/AddonFilePicker.kt
Amejia481: We are implicitly doing it :) , as we are inside of the `let?.` block we are warranty that `uri` will be no-null so the compiler does an implicit cast. I update the code to make it explicit :)  
image

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
Amejia481: Updated!

android-components/components/feature/addons/src/test/java/mozilla/components/feature/addons/ui/AddonFilePickerTest.kt
Amejia481: Updated!

PR #4798, opened 2023-12-11
Bug 1864072 - Create initial drawer UI

MozillaNoah:
https://github.com/mozilla-mobile/firefox-android/assets/87384386/053fc60a-55fe-427d-bd0f-a272c36eea38

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864072

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
@@ -1,0 +1,36 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.debugsettings.ui
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * The debug drawer UI.
+ */
+@Composable
+fun DebugDrawer() {
+    Box(
+        modifier = Modifier
+            .fillMaxSize()
+            .background(color = FirefoxTheme.colors.layer1),
+    ) {
+        Text(text = "Debug drawer")
+    }
+}
+
+@Composable
+@LightDarkPreview
+private fun DebugDrawerPreview() {
+    FirefoxTheme {
+        DebugDrawer()
+    }
+}
--- a/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
@@ -4,21 +4,31 @@
 
 package org.mozilla.fenix.debugsettings.ui
 
-import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.padding
+import androidx.compose.material.DrawerValue
+import androidx.compose.material.ModalDrawer
 import androidx.compose.material.Snackbar
 import androidx.compose.material.SnackbarHost
 import androidx.compose.material.SnackbarHostState
+import androidx.compose.material.rememberDrawerState
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.CompositionLocalProvider
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.setValue
+import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.unit.LayoutDirection
 import androidx.compose.ui.unit.dp
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.filter
 import org.mozilla.fenix.R
 import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.compose.button.FloatingActionButton
@@ -30,7 +40,25 @@ import org.mozilla.fenix.theme.FirefoxTheme
 @Composable
 fun DebugOverlay() {
     val snackbarState = remember { SnackbarHostState() }
-    val scope = rememberCoroutineScope()
+    // The separate boolean in conjunction with `drawerState` is to allow the drawer animations
+    // to still occur even with the show/hide logic documented below.
+    val drawerState = rememberDrawerState(initialValue = DrawerValue.Closed)
+    var drawerEnabled by remember { mutableStateOf(false) }
+
+    LaunchedEffect(drawerEnabled) {
+        if (drawerEnabled) {
+            drawerState.open()
+        }
+    }
+
+    LaunchedEffect(drawerState) {
+        snapshotFlow { drawerState.currentValue }
+            .distinctUntilChanged()
+            .filter { it == DrawerValue.Closed }
+            .collect {
+                drawerEnabled = false
+            }
+    }
 
     Box(
         modifier = Modifier.fillMaxSize(),
@@ -41,11 +69,36 @@ fun DebugOverlay() {
                 .align(Alignment.CenterStart)
                 .padding(start = 16.dp),
             onClick = {
-                scope.launch {
-                    snackbarState.showSnackbar("Show debug drawer")
-                }
+                drawerEnabled = true
             },
         )
+
+        // ModalDrawer utilizes a Surface, which blocks ALL clicks behind it, preventing the app
+        // from being interactable. This cannot be overridden in the Surface API, so we must hide
+        // the entire drawer when it is closed.
+        if (drawerEnabled) {
+            val currentLayoutDirection = LocalLayoutDirection.current
+            val sheetLayoutDirection = when (currentLayoutDirection) {
+                LayoutDirection.Rtl -> LayoutDirection.Ltr
+                LayoutDirection.Ltr -> LayoutDirection.Rtl
+            }
+
+            // Force the drawer to always open from the opposite side of the screen. We need to reset
+            // this below with `drawerContent` to ensure the content follows the correct direction.
+            CompositionLocalProvider(LocalLayoutDirection provides sheetLayoutDirection) {
+                ModalDrawer(
+                    drawerContent = {
+                        CompositionLocalProvider(LocalLayoutDirection provides currentLayoutDirection) {
+                            DebugDrawer()
+                        }
+                    },
+                    drawerBackgroundColor = FirefoxTheme.colors.layer1,
+                    scrimColor = FirefoxTheme.colors.scrim,
+                    drawerState = drawerState,
+                    content = {},
+                )
+            }
+        }
 
         // This must be the last element in the Box
         SnackbarHost(
@@ -63,11 +116,7 @@ fun DebugOverlay() {
 @LightDarkPreview
 private fun DebugOverlayPreview() {
     FirefoxTheme {
-        Box(
-            modifier = Modifier
-                .fillMaxSize()
-                .background(color = FirefoxTheme.colors.layer1),
-        ) {
+        Box(modifier = Modifier.fillMaxSize()) {
             DebugOverlay()
         }
     }
--- a/fenix/app/src/main/res/layout/activity_home.xml
+++ b/fenix/app/src/main/res/layout/activity_home.xml
@@ -1,26 +1,31 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<org.mozilla.fenix.perf.HomeActivityRootLinearLayout
+
+<!-- DO NOT INCLUDE ANYTHING ELSE INSIDE THIS FRAME LAYOUT -->
+<!-- This FrameLayout is solely to allow `debugOverlay` to be presented globally in Fenix. -->
+<!-- This is not to be used outside of the scope of the debug overlay and the Debug Drawer feature. -->
+<!-- PERFORMANCE NOTE: Anything outside of `HomeActivityRootLinearLayout` will not be measured for performance. -->
+<FrameLayout
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
-    android:layout_height="match_parent"
-    android:orientation="vertical"
-    android:id="@+id/rootContainer"
-    tools:context=".HomeActivity">
+    android:layout_height="match_parent">
 
-    <ViewStub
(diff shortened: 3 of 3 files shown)


fenix/app/src/main/res/layout/activity_home.xml
```diff
@@ -1,26 +1,26 @@
 <!-- This Source Code Form is subject to the terms of the Mozilla Public
    - License, v. 2.0. If a copy of the MPL was not distributed with this
    - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
-<org.mozilla.fenix.perf.HomeActivityRootLinearLayout
+<FrameLayout
```
MozillaNoah, line 4: I needed to make the root View the FrameLayout instead, in order to get the drawer UI over the top of the toolbar in the screens that use it (Settings, history, etc.)

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -41,12 +67,38 @@ fun DebugOverlay() {
...
+            val sheetLayoutDirection = when(currentLayoutDirection) {
+                LayoutDirection.Rtl -> LayoutDirection.Ltr
+                LayoutDirection.Ltr -> LayoutDirection.Rtl
+            }
+
+            // Force the drawer to always open from the opposite side of the screen
+            CompositionLocalProvider(LocalLayoutDirection provides sheetLayoutDirection) {
+                ModalDrawer(
+                    drawerContent = {
+                        // Ensure the drawer's content still respects the device's direction setting
```
MozillaNoah, line 89: Leaving the comments for now, but open to cleaning some/all of these up before landing

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -30,7 +39,24 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
     val snackbarState = remember { SnackbarHostState() }
-    val scope = rememberCoroutineScope()
+    val drawerState = remember { DrawerState(initialValue = DrawerValue.Closed) }
+    var drawerEnabled by remember { mutableStateOf(false) }
+
+    LaunchedEffect(drawerEnabled) {
+        if (drawerEnabled) {
+            drawerState.open()
+        }
+    }
```
MozillaNoah, line 49: because of the weirdness with Surface mentioned in the comment, I needed to add this launched effect in order to prevent some open animation weirdness on the first open

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MatthewTighe: I don't think this one is necessary or could at least be combined more with the one above `CompositionLocalProvider`. The rest seem useful to me

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -30,7 +39,24 @@ import org.mozilla.fenix.theme.FirefoxTheme
 @Composable
 fun DebugOverlay() {
     val snackbarState = remember { SnackbarHostState() }
-    val scope = rememberCoroutineScope()
+    val drawerState = remember { DrawerState(initialValue = DrawerValue.Closed) }
+    var drawerEnabled by remember { mutableStateOf(false) }
```
MatthewTighe, line 43: I'm wondering if we need both of these states or if we could just reuse one. It seems like their intent is roughly the same

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -41,12 +67,38 @@ fun DebugOverlay() {
...
-                    snackbarState.showSnackbar("Show debug drawer")
-                }
+                drawerEnabled = true
             },
         )
 
+        // ModalDrawer utilizes a Surface, which blocks ALL clicks behind it, preventing the app
+        // from being interactable. This cannot be overridden in the Surface API, so we must hide
+        // the entire drawer when it is closed. Whenever the app is Compose from
+        // the root layer, we can remove this.
```
MatthewTighe, line 77: It seems fine to keep this even when the app is fully compose, right? I would guess we probably want to skip these layout passes entirely when the drawer is not visible in either case

fenix/app/src/main/res/layout/activity_home.xml
MatthewTighe: This makes me a little wary still, though thinking about it I don't think it should have any real impact on the perf metrics that `HomeActivityRootLinearLayout` tracks. 

I think we should include some strongly worded comments about not adding anything that isn't strictly debug drawer related to this view, though.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -41,12 +67,38 @@ fun DebugOverlay() {
...
+                    drawerBackgroundColor = FirefoxTheme.colors.layer1,
+                    scrimColor = FirefoxTheme.colors.scrim,
+                    drawerState = drawerState,
+                    content = {},
+                )
+            }
+        }
+
         // This must be the last element in the Box
         SnackbarHost(
```
MatthewTighe, line 103: Do we still need this?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MozillaNoah: If the app is fully compose, then we I suppose there's two paths for the future of this feature's UI:
1. Make the drawer accessible from the root level and ANY feature can display one (which was kinda what I was thinking by including that comment)
2. Keep the `DebugOverlay` its own distinct layer, even after 100% migration, and have it act independently without any implementation details of what is behind it.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MozillaNoah: I'll see if they can be consolidated

fenix/app/src/main/res/layout/activity_home.xml
MozillaNoah: For sure. The other upside is it removes the ComposeView from `HomeActivityRootLinearLayout`, so the overlay can't be accidentally included in any of the performance calculations now.

I'll add a comment to this file explaining what's up and a note on how to maybe handle this if this whole file is moved to Compose

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MozillaNoah: I'll update the comment though to not assume any specific direction we take with getting Fenix 100% compose

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MozillaNoah: Oh, the separate boolean was because of [this comment above](https://github.com/mozilla-mobile/firefox-android/pull/4798#discussion_r1423019686). 

Essentially, I was seeing a little animation weirdness when telling the drawer UI to be added back to the tree AND then open itself (on the very first open). In order to solve this, I added the boolean and the launched effect around it, so the boolean would be updated, the UI would be added back to the tree, and then in a coroutine (the launched effect) the drawer would be told to open.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -30,7 +40,25 @@ import org.mozilla.fenix.theme.FirefoxTheme
 @Composable
 fun DebugOverlay() {
     val snackbarState = remember { SnackbarHostState() }
-    val scope = rememberCoroutineScope()
+    val drawerState = rememberDrawerState(initialValue = DrawerValue.Closed)
+    var drawerEnabled by remember { mutableStateOf(false) }
+
+    LaunchedEffect(drawerEnabled) {
+        Log.d("bondbond", "drawerEnabled update - $drawerEnabled")
```
MatthewTighe, line 47: They call me Bond. Bond Bond.

MatthewTighe (approved):
LGTM, but let's please remember to keep an eye on perf regressions with the change to the activity's view hierarchy. Might also make sense to branch off this for work for the rest of this week and wait until next week (new nightly cycle) to land this

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MatthewTighe: That makes sense, but maybe a quick comment to that effect for future readers?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MozillaNoah: 😂 😂 😂

MozillaNoah:
> LGTM, but let's please remember to keep an eye on perf regressions with the change to the activity's view hierarchy. Might also make sense to branch off this for work for the rest of this week and wait until next week (new nightly cycle) to land this

Thankfully, there's work in the oven that does not immediately depend on this (data store refactor, redux stand up, and tab manager), so I'm cool with deferring landing this until next Monday. I might still post a demo of this though.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
t-p-white: Should we make a generic component?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
```diff
@@ -0,0 +1,38 @@
...
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Debug drawer content.
```
t-p-white, line 17: nit: maybe something like 
```suggestion
 * The debug drawer UI.
```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
```diff
@@ -0,0 +1,38 @@
...
+    ) {
+        Text(text = "Debug drawer")
+    }
+}
+
+@Composable
+@LightDarkPreview
+private fun DebugDrawerPreview() {
+    FirefoxTheme {
+        Box {
```
t-p-white, line 34: is `Box` needed here?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -30,7 +40,25 @@ import org.mozilla.fenix.theme.FirefoxTheme
...
+        if (drawerEnabled) {
+            drawerState.open()
+        }
+    }
+
+    LaunchedEffect(drawerState) {
+        snapshotFlow { drawerState.currentValue }
```
t-p-white, line 54: is it possible the above `drawerState.open()` is not completed before we run this block?

fenix/app/src/main/res/layout/activity_home.xml
t-p-white: Little late to this, so for my understanding we're asserting that we are currently excluding the debug drawer from the performance profiling?

t-p-white: Little late to this, so for my understanding we're asserting that we are currently excluding the debug drawer from the performance profiling?
Do we have a good understanding of the ramifications of making this change?

t-p-white (commented):
Just a couple of minor comments, main point being around the `launchedEffect`s

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MozillaNoah: Potentially! We already have a [ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1814198) for it, actually. I think GL started on it a long time ago, and probably just needs to be taken over and restarted.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MozillaNoah: It's not! Good catch

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MozillaNoah: Good question

All calls to `drawerState.open()` should be completed before we get into the actual meat of the launched effect with `drawerState`. 
1. The snapshotFlow is being derived from `currentValue`, so it is only being triggered when the drawer state changes and has settled. There is actually a separate API `targetValue` which would allow us to listen to changes to what state the drawer is being moved to. If we were _just_ listening and acting upon all changes to drawerState (without a snapshotFlow on the parameter we cared about), then, yes, this would have been called preemptively.
2. Thanks to snapshotFlow + `distinctUntilChanged` + the `if`,  we're only listening for state changes where it has changed from `open` to `closed`. 

I'll add a quick polish to the snapshot flow, so now it'll be:
```
snapshotFlow { drawerState.currentValue }
            .distinctUntilChanged()
            .filter { it == DrawerValue.Closed }
            .collect {
                drawerEnabled = false
            }
```
With the filter, it's a little cleaner that we're only wanting state emissions of drawer state going from open to closed.

fenix/app/src/main/res/layout/activity_home.xml
MozillaNoah: Adding a note now.

It's only being excluded from any of the automatic profiling that currently exists on `HomeActivity`. We can still do anything manual or set up new profile markers just for the debug drawer in the future.

MozillaNoah:
Just pushed the suggested changes from yall and rebased with main so now this PR has the updated fab icon ☺️

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
MozillaNoah: There's a chance that the Snackbar is on the official roadmap for 2024, if the grid design system project gets prioritized, but we can still try to work on this sooner than that.

MozillaNoah:
This might also get cleaned up a little as part of the redux stand up. [Here's a diff of what I have so far](https://github.com/MozillaNoah/firefox-android/commit/4bd1e3506940b34240b1fde5c3367cdfd3f476af).

MatthewTighe:
> so I'm cool with deferring landing this until next Monday

Honestly, depending on your holiday schedule it could even wait til the New Year if you're at all worried about firefighting later this month. Up to you though!

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
t-p-white: Thanks @MozillaNoah, appreciate the info here! Looks great 🙂 

I think you can simplify:
```koltin
.distinctUntilChanged()
.filter { it == DrawerValue.Closed }
```
to
```kotlin
.distinctUntilChangedBy { it == DrawerValue.Closed }
```

t-p-white (approved):

MozillaNoah:
I've created a follow-up PR for this, where I add the Redux plumbing and move the debug drawer open boolean to the `State`

https://github.com/mozilla-mobile/firefox-android/pull/4842/commits/8e6a8d4e5dca771923adcbd192c8abc261886a04

PR #4810, opened 2023-12-12
Bug 1855983 - Translations UI Initial Basic Translate Refined Translate w/Separate First Run Flow.

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
```diff
@@ -516,6 +516,14 @@ class Settings(private val appContext: Context) : PreferencesHolder {
         default = true,
     )
 
+    /**
+     * Indicates if the user has completed successfully first translation.
+     */
+    var isFirstTranslationSuccessfully: Boolean by booleanPreference(
```
ohall-m, line 522: I'd prefer the naming to be something like, `showFirstTimeTranslation` or `isFirstTimeTranslation` just to make it a little more clear the intention here and elsewhere.

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2302,6 +2302,65 @@
 
     <!-- Translations feature-->
 
+    <!-- Translations bottom sheet -->
+    <!-- Title of the translations bottom sheet-->
+    <string name="translations_bottom_sheet_title">Translate this page?</string>
+    <!-- Title of the translations bottom sheet. It will appear only once when the user uses the first-time translation feature.
+    The first parameter is the name of the app defined in app_name (for example: Fenix) -->
+    <string name="translations_bottom_sheet_title_first_time">Try private translations in %1$s</string>
+    <!-- Translations bottom sheet info message. It will appear under the title.It will appear only once when the user uses the first-time translation feature.
```
ohall-m, line 2311:
```suggestion
    <!-- Translations bottom sheet info message. It will appear under the title. It will appear only once when the user uses the first-time translation feature.
```

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2302,6 +2302,65 @@
...
+     The first parameter is the name of the app defined in app_name (for example: Fenix). -->
+    <string name="translation_option_bottom_sheet_about_translations">About translations in %1$s</string>
+
+    <!-- Translations settings screen -->
+    <!-- Title of the settings screen that will appear on the toolbar.-->
+    <string name="translation_settings_toolbar_title">Translations</string>
+    <!-- Offer to translate when possible the switch label from the settings screen. -->
+    <string name="translation_settings_offer_to_translate">Offer to translate when possible</string>
+    <!-- Always download the switch label from the settings screen. -->
+    <string name="translation_settings_always_download">Always download languages  in data saving mode</string>
```
ohall-m, line 2354:
```suggestion
    <string name="translation_settings_always_download">Always download languages in data saving mode</string>
```

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2302,6 +2302,65 @@
...
+    <!-- Screen header presenting the translation settings preferences.  -->
+    <string name="translation_settings_translation_preference">Translation preferences</string>
+    <!--  The label of the button that will redirect the user to Automatic translations screen.  -->
```
ohall-m, line 2357: nit: Would prefer if we added "the" before the screen name here and below.

```suggestion
    <!--  The label of the button that will redirect the user to the Automatic translations screen.  -->
```

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
ohall-m: I think the default should also be `true` and we will toggle it to false once the first translation is complete when we work on integration.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
```diff
@@ -68,7 +68,7 @@ fun TranslationOptionsDialog(
             TextListItem(
                 label = stringResource(
                     id = R.string.translation_option_bottom_sheet_about_translations,
-                    formatArgs = arrayOf(stringResource(R.string.app_name)),
+                    formatArgs = arrayOf(stringResource(R.string.firefox)),
```
ohall-m, line 71: Nice catch!

ohall-m (commented):
Thanks for getting this all setup! I just had a few naming type suggestions.

I think this might also be a good place to make a few other visual refinements too, or else make note of them to follow-up in on in [1868991](https://bugzilla.mozilla.org/show_bug.cgi?id=1868991) depending on how much the change involves or whichever you prefer. 

Some visual refinement finalization suggestions:
* Add the BETA tag above "Translate this page"
* "Not now" needs to be [purple for dark theme](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=8366-17477&mode=design&t=qxoZBArTLc47F74Z-0) and also purple for light theme
* The lines under the language in dark mode should be [grey](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=8366-17477&mode=design&t=qxoZBArTLc47F74Z-0)
* [Landscape mode](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7039-32721&mode=design&t=qxoZBArTLc47F74Z-0) should span the page and pop-up more.

Delphine:
Need to do a final pass tomorrow morning PT. thanks

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2302,6 +2302,65 @@
...
+    <string name="translation_settings_automatic_translation">Automatic translation</string>
+    <!--  The label of the button that will redirect the user to never translate sites screen.  -->
```
gabrielluong, line 2359:
```suggestion
    <!-- The label of the button that will redirect the user to never translate sites screen. -->
```
This appears elsewhere

Delphine:
I've reviewed the strings and they look good (pending another update as per comment [here](https://github.com/mozilla-l10n/android-l10n/pull/718#discussion_r1427074134)).

However, many comments should be updated as they are currently more developer oriented rather than localizer oriented (for example, referencing to "bottom sheet" won't make sense to a localizer). Comments should be clear for a person who is translating and has no developer background. Thanks!

Delphine (changes requested):
I've reviewed the strings and they look good (pending another update as per comment [here](https://github.com/mozilla-l10n/android-l10n/pull/718#discussion_r1427074134)).

However, many comments should be updated as they are currently more developer oriented rather than localizer oriented (for example, referencing to "bottom sheet" won't make sense to a localizer). Comments should be clear for a person who is translating and has no developer background. Thanks!

iorgamgabriel:
> I've reviewed the strings and they look good (pending another update as per comment [here](https://github.com/mozilla-l10n/android-l10n/pull/718#discussion_r1427074134)).
> 
> However, many comments should be updated as they are currently more developer oriented rather than localizer oriented (for example, referencing to "bottom sheet" won't make sense to a localizer). Comments should be clear for a person who is translating and has no developer background. Thanks!

I made the changes.

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2349,11 +2403,13 @@
     <string name="download_languages_header_learn_more_preference">Learn more</string>
     <!-- The subhead of the download language preference screen will appear above the pivot language. -->
     <string name="download_languages_available_languages_preference">Available languages</string>
-    <!-- A text in parenthesis that will appear next to the pivot language -->
-    <string name="download_languages_default_system_language_require_preference">required</string>
+    <!-- A text in parenthesis that will appear next to the pivot language. -->
+    <string name="download_languages_default_system_language_require_preference">(required)</string>
```
Delphine, line 2407: This string should be something link `%1$s (required)` where the placeholder is replaced with the language name.

Also, string needs a new ID since it's changing

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2349,11 +2403,13 @@
...
+    <!-- A text in parenthesis that will appear next to a general language. The first parameter is the language file size, for example, "(3.91 KB)". -->
+    <string name="download_languages_general_language_preference">(%1$s)</string>
```
Delphine, line 2409: This should be something like `%1$s (%2$s)` where the first parameter is the Language, and the second is the file size

Delphine (changes requested):
Thanks for the updates. Just requesting two changes to the strings.xml file, and strings should be good to go.

Reminder that today is string freeze for v122 and the current [string PR](https://github.com/mozilla-l10n/android-l10n/pull/718) I will expose to localizers is getting quite long - so best if we can land remaining strings soon (it's also been open for a while now). Thanks!

fenix/app/src/main/res/values/strings.xml
iorgamgabriel: I can't do this . There are two different views.

fenix/app/src/main/res/values/strings.xml
iorgamgabriel: The same.

fenix/app/src/main/res/values/strings.xml
ohall-m: Is there any way to adjust the composable to help keep the language and additional information strings together on this one and on the one below? 

Do we need to research alternative options?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -13,63 +15,148 @@ import androidx.compose.foundation.layout.height
...
+
+        if (showFirstTimeTranslation) {
+            TranslationsDialogInfoMessage(onLearnMoreClicked, learnMoreUrl)
+        }
 
         Spacer(modifier = Modifier.height(14.dp))
 
-        Column {
+        when (orientation) {
+            Configuration.ORIENTATION_LANDSCAPE -> {
```
ohall-m, line 89: If it is an easy change here, could you also adjust the span of the popup in landscape too? On the screen on my test device (Pixel 4a), there is about a 15% gap on each side where I can see the content behind it, instead of it covering  [edge-to-edge](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7039-32721&mode=design&t=doDKIsA4zuDjY0hy-0).

Additionally, could you adjust the height of the pop-up in landscape some too? On my test device, the landscape pop-up starts at the same height as the toolbar instead of showing the full content, so content is hidden a bit.

Otherwise, we can just follow-up in [1868991](https://bugzilla.mozilla.org/show_bug.cgi?id=1868991).

ohall-m (commented):
Thanks for the refinements! I don't have any blocking changes. I think just researching a solution to having the strings be in the same view or an alternative to facilitate localization and this patch will be good to go.

fenix/app/src/main/java/org/mozilla/fenix/utils/Settings.kt
ohall-m: Thanks for the update!

fenix/app/src/main/res/values/strings.xml
iorgamgabriel: I made the changes.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
iorgamgabriel: I made the changes only for  the height of the pop-up in landscape.

ohall-m: Thanks, height looks good! I'll add the span issue as something for us to follow up in [1868991](https://bugzilla.mozilla.org/show_bug.cgi?id=1868991).

fenix/app/src/main/res/values/strings.xml
ohall-m: Thanks @iorgamgabriel for looking into this more and finding some options! This change looks good from my perspective, but I'm don't know enough for localization.

@Delphine, the current changes now have `download_languages_default_system_language_require_preference` staying as `required`. 

It then substitutes into `download_languages_language_item_preference` as the second parameter in `%1$s (%2$s)`. 

So for a pivot language it would look like: `English (required)` and on a non-pivot language it would look like `German (5.5 MB)`.

Will that meet the criteria?

ohall-m (approved):
LGTM for engineering!

Please hold for localization review and ensure criteria is met before merging!

fenix/app/src/main/res/values/strings.xml
Delphine: Thanks for looking into a workaround for this. It will work, but the comment for `download_languages_language_item_preference` needs to be updated to clearly reflect that the second variable with either be replaced by a size, or by "required" (which lives under `download_languages_default_system_language_require_preference`).

I'd like to bring to your attention that team members who handle reviews will be unavailable for the holidays until January 2nd, starting tomorrow. Is it possible to address the issue today, so that I can merge all the strings from the existing v122 string PR [here](https://github.com/mozilla-l10n/android-l10n/pull/718/files)? Otherwise, there's a concern that the translations may not be completed for the next 2+ weeks. Thank you.

ohall-m: Thanks, @Delphine, for all the reviews on this!

> Is it possible to address the issue today

It is currently outside of @iorgamgabriel's regular working hours, so it will probably be tomorrow before the comment on `download_languages_language_item_preference` can be updated on this patch.

Would it help on your side if I made a patch for certain strings that could land today? Maybe the ones with comments in the quarantine bug?

> there's a concern that the translations may not be completed for the next 2+ weeks

For the translations feature, Engineering is right now targeting release version 124, so we are landing a lot of the strings early. I think a two week wait is fine for our side, but let me know if you'd like us to address anything to make things smoother for the import or anyway we could be helpful.

Delphine: @ohall-m Thank you for bringing this to my attention. It appears that some of the fixed strings have already been merged into the main branch, specifically within the more comprehensive string [PR in our l10n repo](https://github.com/mozilla-l10n/android-l10n/pull/718/files). This pull request encompasses v122 strings, and as a result, all of these strings are currently held back from being merged as well.

Given that the string freeze for v122 was last Friday, I wanted to highlight this potential risk in our current situation

ohall-m: Sorry about all of the issues around the translation strings, we will work on this going forward. Please let me know what you recommend or what helps. 

>all of these strings are currently held back from being merged as well.

If I make a patch today that solves [this](https://github.com/mozilla-l10n/android-l10n/pull/718/files#r1424506570) and [this](download_language_header_preference) will that solve those 122 strings being held up? I assume a correction patch would also need an uplift into 122 as well? Or possibly just letting those strings ride 122 and we will deprecate them as required in 123? 

Just let me know what action to take. I'm not sure what the recommended solution is.

ohall-m: Just summarizing to close the loop here for clarity!

 @iorgamgabriel, just a quick summary:
* Could you adjust the string comment on this patch for `download_languages_language_item_preference` to add more information that it can be a file size or `required` as the second parameter, as mentioned above?
* Re-base to remove changes to `download_languages_default_system_language_require_preference` and `download_language_all_languages_item_preference` since they landed in #4899 and are no longer needed.

I landed #4899 to unblock the l10n repo for the current batch. The #4899 patch only covers the quarantined string comments and nothing else.

iorgamgabriel: Done

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2292,6 +2346,10 @@
     <string name="download_languages_available_languages_preference">Available languages</string>
     <!-- Text that will appear beside a core or pivot language package name to show that the language is necessary for the translation feature to function. -->
     <string name="download_languages_default_system_language_require_preference">required</string>
+    <!-- A text for download language preference item.
+    The first parameter is the language name, for example, "Spanish".
+    The second parameter is the language file size, for example, "(3.91 KB) or "required" if the language package name is pivot. -->
```
ohall-m, line 2351: nit:

```suggestion
    The second parameter is the language file size, for example, "(3.91 KB)" or "required" if the language package name is a pivot language. -->
```

iorgamgabriel: Done

PR #4835, opened 2023-12-14
Bug 1872753- handle RemoteSettings serialization to disk with kotlinx

sarah541:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1872753


android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -100,21 +128,56 @@ class RemoteSettingsClient(
...
                 RemoteSettingsResult.Success(response)
             } catch (e: IOException) {
+                val str = e.message
+                println("catch $str")
+
                 RemoteSettingsResult.DiskFailure(e)
             }
         }
+
+    fun parseJsonToResponse(jsonString: String): RemoteSettingsResponse {
```
MatthewTighe, line 147: I'm not sure we want these functions to remain public. I would suggest considering this from the perspective of the consumer. Will we be expecting consumers to have JSON that they want to parse to a `RemoteSettingsResponse`?

MatthewTighe (commented):
Some comments while this is still draft:

I do think this represents an opportunity to look into whether we can easily serialize data types across the Rust boundary. If we can't use kotlinx.serialization here, we could look into collaborating with app-services to figure that out. It might turn out that we can do something cool like use UniFFI to make them serializable automatically, which could help with this in the future.

Given the challenges we ran into with this in the other patch, I think it's also important that we spend some time fleshing out the test suite a little bit more. I would suggest taking some time to read through the example test I sent, and try to do something similar for the rest of the functions here. I think by trying to avoid mocks as much as possible, we can be more confident that the code we land will behave the way we expect it to.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
sarah541: Sorry for all the nits. This was never meant to be the final version. I just threw around two ways in two commits if we had to manually go through this route .

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -64,14 +67,39 @@ class RemoteSettingsClient(
...
+    suspend fun write(response: RemoteSettingsResponse): RemoteSettingsResult =
+        withContext(Dispatchers.IO) {
+            try {
+                AtomicFile(file).writeString {
+                    buildString {
+                        append("{\"records\": [")
+                        response.records.joinTo(this, separator = ",") { writeRecordAsString(it) }
+                        append("],\"lastModified\": ${response.lastModified.toLong()}}")
+                    }
+                }
```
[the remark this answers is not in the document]
sarah541: Will leave responses below to describe what was implemented.

sarah541:
I have added classes that will map RemoteSettingsResponse class of the remote api to SerialzableRemoteSettingsResponse for easier serialization and deserialization using kotlinx.serialization library. RemoteSettingsResponse is a complex structure and in order to parse and use it the correct way, we wanted classes to be serialized in a customized way hence this approach was taken

sarah541:
I have removed writeIfLastModified function as that is already available in the RemoteSettingsResponse and can be utilized by user where needed.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -176,3 +153,105 @@ sealed class RemoteSettingsResult {
...
+
+private fun RemoteSettingsRecord.toSerializable(): SerializableRemoteSettingsRecord {
+    return SerializableRemoteSettingsRecord(
+        id = this.id,
+        lastModified = this.lastModified,
+        deleted = this.deleted,
+        attachment = this.attachment?.toSerializable(),
+        fields = this.fields,
+    )
+}
```
MatthewTighe, line 212: After cycling on this several times, I'm sorry that we still ended up in a place with all this boilerplate. I'm personally the most happy with this result, but I think that this does indicate that there's a need for a better JSON bridge between us and A-S. 

Just to reframe our original issue, please correct me if my memory is wrong:

1. From A-S we got a fairly ergonomic type for use in Kotlin.
2. However, because we need to serialize that type so that we can save it to disk, we have to figure out a way to do that manually.
3. We tried using `JSONObject` to do this manually and found it to be a little messy, in that we had to extract every field using `getString`-type methods.
4. We tried using `kotlinx.serialization` to do it and ended up in a place where we have to write a lot of boilerplate to convert between `@Serializable` types.

Am I missing anything there?

android-components/components/support/remotesettings/src/test/java/mozilla/components/support/remotesettings/RemoteSettingsClientTest.kt
```diff
@@ -34,21 +38,131 @@ class RemoteSettingsClientTest {
...
+        val result = runBlocking { RemoteSettingsClient(tempFile, collectionName = "test").write(response) }
 
         require(result is RemoteSettingsResult.Success) { "Result should be Success" }
         assertEquals(response, result.response)
 
         tempFile.delete()
     }
 
+    @get:Rule
+    val tempFolder = TemporaryFolder()
```
MatthewTighe, line 58: Can we move this rule to the top of the class? That's typically where they are.

android-components/components/support/remotesettings/src/test/java/mozilla/components/support/remotesettings/RemoteSettingsClientTest.kt
```diff
@@ -34,21 +38,131 @@ class RemoteSettingsClientTest {
...
+        }
+
+        val readString = file.readText()
+        val writeString = Json.encodeToString(SerializableRemoteSettingsResponse.serializer(), response.toSerializable())
+
+        assertEquals(readString, writeString)
+    }
+
+    @Test
+    fun `GIVEN non-empty records WHEN read is called THEN result content is same as write content`() {
```
MatthewTighe, line 119:
```suggestion
    fun `GIVEN non-empty records WHEN read is called THEN result has top-level fields and arbitrarily nested fields`() {
```

android-components/components/support/remotesettings/src/test/java/mozilla/components/support/remotesettings/RemoteSettingsClientTest.kt
```diff
@@ -34,21 +38,131 @@ class RemoteSettingsClientTest {
...
+        val response = RemoteSettingsResponse(records, 125614567U)
+
+        runBlocking {
+            client.write(response)
+        }
+
+        val readString = file.readText()
+        val writeString = Json.encodeToString(SerializableRemoteSettingsResponse.serializer(), response.toSerializable())
+
+        assertEquals(readString, writeString)
```
MatthewTighe, line 115: I think we would be better served by something like:
```suggestion
        runBlocking {
            client.write(response)
            val result = client.read()
            assertEquals(response, result)
        }
```

because (as consumers of this API) we don't actually care about how the content is written to the file. We only care that we can successfully write something and then read it back later.

This might allow you to make all the `@Serializable` types `private`, as well.

MatthewTighe (commented):
This is looking great! Just a couple small changes and I think it'll be ready to land

MatthewTighe:
Ah, one more thing. Could we get a more descriptive commit and PR message? Something like `Bug 1872753 - handle RemoteSettings serialization to disk with kotlinx`

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -176,3 +153,105 @@ sealed class RemoteSettingsResult {
...
+        val jsonString = decoder.decodeString()
+        return JSONObject(jsonString)
+    }
+}
+
+/**
+ * Data class representing serializable version of RemoteSettingsResponse.
+ */
+@Serializable
+data class SerializableRemoteSettingsResponse(
```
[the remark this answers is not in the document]
sarah541: Makes sense! Thank you

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -176,3 +153,105 @@ sealed class RemoteSettingsResult {
...
+ * Data class representing serializable version of RemoteSettingsRecord.
+ */
+@Serializable
+data class SerializableRemoteSettingsRecord(
+    val id: String,
+    val lastModified: ULong,
+    val deleted: Boolean,
+    val attachment: SerializableAttachment?,
+    @Serializable(with = JSONObjectSerializer::class)
+    val fields: JSONObject,
```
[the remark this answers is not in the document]
sarah541: Yeah ! Plus I won't wanna go for a generic <String, String> when it could be any data type

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
sarah541: That's exactly it. Tbh, either way. `JSONObject` or Serialization looks like a lot of boilerplate code to me. 
My thinking? If we could have `JSONObject` being returned from the APIs instead of a new class i.e `RemoteSettingsResponse`?

android-components/components/support/remotesettings/src/test/java/mozilla/components/support/remotesettings/RemoteSettingsClientTest.kt
sarah541: I have a weird caching issue in my Android Studio, that was causing this to fail intermittently, that's why I modified it to this way. I am guessing I might have to reinstall my studio and will push this as stated above and cross my fingers on CI passing this.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -176,3 +153,104 @@ sealed class RemoteSettingsResult {
...
+ * Data class representing serializable version of RemoteSettingsRecord.
+ */
+@Serializable
+private data class SerializableRemoteSettingsRecord(
+    val id: String,
+    val lastModified: ULong,
+    val deleted: Boolean,
+    val attachment: SerializableAttachment?,
+    @Serializable(with = JSONObjectSerializer::class)
+    val fields: JSONObject,
```
MatthewTighe, line 176: Shucks, I'm just finding that kotlinx does actually have a `JsonObject` [type](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-object/). 

I think it would be fine to keep the patch as is, but replacing the type here might let us get rid of the custom `JSONObjectSerializer`.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -66,54 +79,31 @@ class RemoteSettingsClient(
      */
     suspend fun write(response: RemoteSettingsResponse): RemoteSettingsResult = withContext(Dispatchers.IO) {
         try {
-            val jsonObject = JSONObject()
-            jsonObject.put("response", response)
-            AtomicFile(file).writeString { jsonObject.toString() }
+            val jsonString = Json.encodeToString(
+                SerializableRemoteSettingsResponse.serializer(),
+                response.toSerializable(),
+            )
```
MatthewTighe, line 85: We could add another catch here for [serialization exceptions](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serialization-exception/) if we wanted to avoid crashing.

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
```diff
@@ -66,54 +79,31 @@ class RemoteSettingsClient(
...
+        try {
+            if (!file.exists()) {
+                RemoteSettingsResult.DiskFailure(FileNotFoundException("File not found"))
+            } else {
+                val jsonString = file.readText()
+                val response = Json.decodeFromString<SerializableRemoteSettingsResponse>(jsonString)
+                RemoteSettingsResult.Success(response.toRemoteSettingsResponse())
             }
+        } catch (e: IOException) {
+            RemoteSettingsResult.DiskFailure(e)
```
MatthewTighe, line 106: We could add another `catch` here for a [serialization exception](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-serialization-exception/) if we want to avoid crashes for malformed data. Should be pretty low risk since we're also the ones writing to disk though.

MatthewTighe (approved):
This LGTM! I have some optional suggestions that may be useful. If you decide to take them just let me know and I'll have a final glance over the finished product

android-components/components/support/remotesettings/src/main/java/mozilla/components/support/remotesettings/RemoteSettingsClient.kt
sarah541: I am guessing we still need to convert JsonObject to JSONObject , because both of the are a part of different library? I can look into this as a follow up

PR #4842, opened 2023-12-14
Bug 1864073 - Stand up debug drawer Store and State

MozillaNoah:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864073

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerAction.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerAction.kt
@@ -1,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.debugsettings.store
+
+import mozilla.components.lib.state.Action
+
+/**
+ * [Action] implementation related to [DebugDrawerStore].
+ */
+sealed class DebugDrawerAction : Action {
+
+    /**
+     * [DebugDrawerAction] fired when the user opens the drawer.
+     */
+    object DrawerOpened : DebugDrawerAction()
+
+    /**
+     * [DebugDrawerAction] fired when the user closes the drawer.
+     */
+    object DrawerClosed : DebugDrawerAction()
+}
--- a/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
@@ -17,9 +17,7 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -29,24 +27,34 @@ import androidx.compose.ui.unit.LayoutDirection
 import androidx.compose.ui.unit.dp
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.filter
+import mozilla.components.lib.state.ext.observeAsState
 import org.mozilla.fenix.R
 import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.compose.button.FloatingActionButton
+import org.mozilla.fenix.debugsettings.store.DebugDrawerAction
+import org.mozilla.fenix.debugsettings.store.DebugDrawerState
+import org.mozilla.fenix.debugsettings.store.DebugDrawerState.DrawerStatus
+import org.mozilla.fenix.debugsettings.store.DebugDrawerStore
 import org.mozilla.fenix.theme.FirefoxTheme
 
 /**
  * Overlay for presenting Fenix-wide debugging content.
+ *
+ * @param debugDrawerStore [DebugDrawerStore] used to listen for changes to [DebugDrawerState] and
+ * dispatch any [DebugDrawerAction]s.
  */
 @Composable
-fun DebugOverlay() {
+fun DebugOverlay(
+    debugDrawerStore: DebugDrawerStore,
+) {
+    val drawerStatus by debugDrawerStore.observeAsState(initialValue = DrawerStatus.Closed) { state ->
+        state.drawerStatus
+    }
     val snackbarState = remember { SnackbarHostState() }
-    // The separate boolean in conjunction with `drawerState` is to allow the drawer animations
-    // to still occur even with the show/hide logic documented below.
     val drawerState = rememberDrawerState(initialValue = DrawerValue.Closed)
-    var drawerEnabled by remember { mutableStateOf(false) }
 
-    LaunchedEffect(drawerEnabled) {
-        if (drawerEnabled) {
+    LaunchedEffect(drawerStatus) {
+        if (drawerStatus == DrawerStatus.Open) {
             drawerState.open()
         }
     }
@@ -56,7 +64,7 @@ fun DebugOverlay() {
             .distinctUntilChanged()
             .filter { it == DrawerValue.Closed }
             .collect {
-                drawerEnabled = false
+                debugDrawerStore.dispatch(DebugDrawerAction.DrawerClosed)
             }
     }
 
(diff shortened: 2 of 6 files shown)


MatthewTighe (approved):
Think this looks solid for now. We did talk offline about potentially passing explicit properties and callbacks into `DebugOverlay` instead of the `Store` directly. I think we eventually want to end up there if we want to ship this in A-C or as a library, but this seems like good starting place to iterate quickly. State hoisting is pretty easy, at the end of the day

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/redux/DebugDrawerStore.kt
```diff
@@ -0,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.debugsettings.redux
```
gabrielluong, line 5: Could we call the package `store` instead of `redux`? I assume this is kinda how we do this https://searchfox.org/mozilla-mobile/search?q=Store.kt&path=. This is more to not mistaken this for ReduxJS.

MozillaNoah: Sure thing

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerAction.kt
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * [Action] implementation related to [DebugDrawerStore].
+ */
+sealed class DebugDrawerAction : Action {
+
+    /**
+     * [DebugDrawerAction] fired when the user opens or closes the drawer.
+     */
+    data class DrawerToggled(val isOpen: Boolean) : DebugDrawerAction()
```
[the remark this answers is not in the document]
MozillaNoah: This is how the underlying `ModalDrawer` `DrawerState` composable handles it, and it has a separate enum with third state for half-expanded for when it's a bottom a bottom sheet instead.

I love me some transparent enums, but I don't feel super strongly either way, so happy to go with either approach!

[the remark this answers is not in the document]
MozillaNoah: Another alternative would be to just have separate actions for opening and closing the drawer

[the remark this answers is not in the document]
t-p-white: Oo I like that approach of having separate actions 🙂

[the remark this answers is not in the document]
MozillaNoah: yup yup. I'll push those changes as the next iteration of this and we can see how it looks/feels

MozillaNoah:
Recent push:
- Splits the drawer toggle action into two separate open/close actions
- Removes the state boolean in favor of using an enum
- Adds unit tests

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -29,24 +27,31 @@ import androidx.compose.ui.unit.LayoutDirection
...
     val snackbarState = remember { SnackbarHostState() }
-    // The separate boolean in conjunction with `drawerState` is to allow the drawer animations
-    // to still occur even with the show/hide logic documented below.
     val drawerState = rememberDrawerState(initialValue = DrawerValue.Closed)
-    var drawerEnabled by remember { mutableStateOf(false) }
 
-    LaunchedEffect(drawerEnabled) {
-        if (drawerEnabled) {
+    LaunchedEffect(drawerValue) {
+        if (drawerValue is DebugDrawerAction.DrawerOpened) {
```
MatthewTighe, line 54: Should this be `DebugDrawerState.DrawerValue.Opened`?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -29,24 +27,31 @@ import androidx.compose.ui.unit.LayoutDirection
...
+ *
+ * @param debugDrawerStore [DebugDrawerStore] used to listen for changes to [DebugDrawerState] and
+ * dispatch any [DebugDrawerAction]s.
  */
 @Composable
-fun DebugOverlay() {
+fun DebugOverlay(
+    debugDrawerStore: DebugDrawerStore,
+) {
+    val drawerValue by debugDrawerStore.observeAsState(initialValue = DebugDrawerAction.DrawerClosed) { state -> state.drawerValue }
```
MatthewTighe, line 49: Should this be `initialValue = DebugDrawerState.DrawerValue.Closed`?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugOverlay.kt
```diff
@@ -69,14 +74,14 @@ fun DebugOverlay() {
...
-                drawerEnabled = true
+                debugDrawerStore.dispatch(DebugDrawerAction.DrawerOpened)
             },
         )
 
         // ModalDrawer utilizes a Surface, which blocks ALL clicks behind it, preventing the app
         // from being interactable. This cannot be overridden in the Surface API, so we must hide
         // the entire drawer when it is closed.
-        if (drawerEnabled) {
+        if (drawerValue is DebugDrawerAction.DrawerClosed) {
```
MatthewTighe, line 84: Should this be `DebugDrawerState.DrawerValue.Closed`?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerAction.kt
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * [DebugDrawerAction] fired when the user opens the drawer.
+     */
+    object DrawerOpened : DebugDrawerAction()
+
+    /**
+     * [DebugDrawerAction] fired when the user closes the drawer.
+     */
+    object DrawerClosed : DebugDrawerAction()
+}
```
MatthewTighe, line 23: I won't ask you to change it again, but I do think this is an interesting case study in the verbosity/clarity spectrum. IMO this is tending a little more towards boilerplate-y

MozillaNoah: I'm by no means committed to this. In case you didn't catch the convo before I collapsed it, this change is a result of this [comment thread](https://github.com/mozilla-mobile/firefox-android/pull/4842#discussion_r1430430963). Since this is still a draft technically, I felt it was at least worth exploring (wording withstanding).

MozillaNoah: @t-p-white what's your take having seen the before/after now? 

Personally, I'm leaning towards going back to one action + a boolean to keep things simple and concise

t-p-white: @MozillaNoah Thanks for trying this out 👍 

Personally, I prefer this method of having the separate actions and using enums, I'm not sold on a Boolean as an appropriate type for open/close - as you found an enum is also how the `Material` `DrawerState` defines the values, so aligning with that design is no bad thing. IMO this just removes any ambiguity with each object having a single responsibility and should make it more open for extension, for example if you want to add a parameter in the future for just 'open' this method makes that an easy change

t-p-white: > enum is also how the Material DrawerState defines the values
Actually, couldn't we just use the existing `DrawerValue` enum?

MozillaNoah: We could. The only trade-off is that the `DebugDrawerState` would now be coupled to that Material class/library and it isn't extensible, but maybe that's overengineering it when we only know what we need right now anyway.

t-p-white: Yeah that's a good point, certainly doesn't hurt to roll our own 👍

MatthewTighe: I think we want to keep platform and external library types out of our state unless absolutely necessary to avoid coupling.  

I'm also not set on using a `Boolean` necessarily, as I do think there is value in well-defined types. I could've clarified further with my last comment, but I do think a single `Toggled` action that includes the updated state simplifies callsites and the reducer. 

Further, the nested types become fairly wordy, but that could be changed with a manual import if you wanted. For example:
```
DebugDrawerState.DrawerValue.Closed

MozillaNoah: I'll keep what we have for now. Maybe this can be a 5 minute topic in show and tell?

image

image


This is about as clean, single responsibility as it gets

t-p-white: Sounds good @MozillaNoah! So I guess this comes down to:
1. Separate Open/Close actions as you have it here
2. A toggle drawer action that takes the Open/Close enum type
?

My personal preference is having the explicit Open/Close actions:
1.  it's super clear to any users what the action is doing
2. I think this lends itself better for extension

But I can see we already have 'toggle' type actions in the code.

t-p-white (approved):
Once the conflicts & 'max line length' lint issue is resolved this lgtm 😎 Thanks for the discussion

PR #4844, opened 2023-12-14
Bug 1869225 - Catch `SuggestApiException`s.

linabutler:
This commit adds a `handleSuggestExceptions` helper to `FxSuggestStorage` that catches and reports `SuggestApiException`s from the Suggest Rust component. `handleSuggestExceptions` works like the existing `handle{Fxa, Places}Exceptions` helpers.

`FxSuggestIngestionWorker`'s retry behavior is unchanged: if ingestion fails, the exception will now be reported, but the operation will still be rescheduled using `WorkManager`'s retry mechanism, as before.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1869225

### DIFF
--- a/android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
+++ b/android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
@@ -5,27 +5,35 @@
 package mozilla.components.feature.fxsuggest
 
 import android.content.Context
+import androidx.annotation.VisibleForTesting
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.withContext
+import mozilla.appservices.suggest.SuggestApiException
 import mozilla.appservices.suggest.SuggestIngestionConstraints
 import mozilla.appservices.suggest.SuggestStore
 import mozilla.appservices.suggest.Suggestion
 import mozilla.appservices.suggest.SuggestionQuery
+import mozilla.components.concept.base.crash.CrashReporting
+import mozilla.components.support.base.log.logger.Logger
 import java.io.File
 
 /**
  * A coroutine-aware wrapper around the synchronous [SuggestStore] interface.
  *
- * @property context The Android application context.
+ * @param context The Android application context.
+ * @param crashReporter An optional [CrashReporting] instance for reporting unexpected caught
+ * exceptions.
  */
 class FxSuggestStorage(
     context: Context,
+    private val crashReporter: CrashReporting? = null,
 ) {
     // Lazily initializes the store on first use. `cacheDir` and using the `File` constructor
     // does I/O, so `store.value` should only be accessed from the read or write scope.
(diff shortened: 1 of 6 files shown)


linabutler:
(I'll go ahead and add some tests next, but does the overall shape and direction look okay?)

jonalmeida:
> (I'll go ahead and add some tests next, but does the overall shape and direction look okay?)

Thanks for this! Yeah, this looks like the right direction.

linabutler:
Great, thanks; this is ready for review now! 👀

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
```diff
@@ -35,25 +43,37 @@ class FxSuggestStorage(
...
+        handleSuggestExceptions("ingest", false) {
+            withContext(writeScope.coroutineContext) {
+                try {
+                    store.value.ingest(constraints)
+                    true
+                } catch (e: SuggestApiException) {
+                    false
+                }
+            }
         }
```
jonalmeida, line 77:
```suggestion
        withContext(writeScope.coroutineContext) {
            handleSuggestExceptions("ingest", false) {
                store.value.ingest(constraints)
                true
            }
        }
```

If I'm not mistaken, you can flip these nested methods and then you don't need a second try-catch because the `handleSuggestExceptions` already has this in there.

jonalmeida: In fact, I think we would end up dropping the exception and not hitting the crash reporter because we return false on line 74. 🤔

jonalmeida (approved):
In case I'm mistaken above, r+ so you can land this sooner and respond after.

android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
linabutler: > I think we would end up dropping the exception and not hitting the crash reporter because we return false on line 74.

I did this intentionally—my thinking was, `ingest()` is a background operation that can fail for any number of reasons, like connectivity issues, and we don't necessarily want to report all of them—but I actually like your suggestion better!

We have a ticket on file for the Rust component to improve the exception hierarchy—right now, everything's a `SuggestApiException.Other`, so it doesn't distinguish between HTTP errors, SQLite errors, interruptions—but until then, maybe it's better to report all of them for now.

WDYT?

jonalmeida: Yeah, it's fine reporting them all for now. We've done this in the past with places and push rust components. Those errors are reported as "infos" so they don't trigger any crash alerts. Eventually, we were able to ignore the noisy ones that were not actionable and only report the valuable ones.

_Narrator: why even throw if we are going to ignore them? 👀_

linabutler:
@Mergifyio backport releases_v122

PR #4845, opened 2023-12-14
Bug 1867717 - Add DownloadsFeature to AddonPopupBaseFragment

Amejia481:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1867717


android-components/components/feature/webcompat/src/main/java/mozilla/components/feature/webcompat/WebCompatFeature.kt
```diff
@@ -14,7 +14,7 @@ object WebCompatFeature {
     private val logger = Logger("mozac-webcompat")
 
     internal const val WEBCOMPAT_EXTENSION_ID = "webcompat@mozilla.org"
-    internal const val WEBCOMPAT_EXTENSION_URL = "resource://android/assets/extensions/webcompat/"
+    internal const val WEBCOMPAT_EXTENSION_URL = "resource://android/assets/extensions/extension/"
```
Amejia481, line 17: This was just for testing

Amejia481:
This is ready for review :)

willdurand:
I don't think I can review this patch, this is a bit out of my league. @jonalmeida is that something you could look at please?

Amejia481:
@jonalmeida would you mind taking a look at this ? :)

fenix/app/src/test/java/org/mozilla/fenix/browser/DownloadDialogUtilsTest.kt
```diff
@@ -0,0 +1,297 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+class DownloadDialogUtilsTest {
+
+    @Test
+    fun `GIVEN final status and openInApp WHEN calling handleOnDownloadFinished and try to open the file fails THEN show an error message`() {
+        val downloadUtils = spyk(DownloadDialogUtils)
```
jonalmeida, line 29: Can we avoid using a spy? It is very easy to misuse them and if we are relying on them then I believe we need to spend more time thinking about how we structure our code to be testable. For example, finding a way to delegate out the `FenixSnackbar` code in a test.

In fact, on lines 47 to 53, we should not be using the spy and are no longer executing the real code path needed to catch real regressions.

```
        every {
            downloadUtils.shouldShowCompletedDownloadDialog(
                any(),
                any(),
                any(),
            )
        } returns true
```

fenix/app/src/main/java/org/mozilla/fenix/addons/AddonPopupBaseFragment.kt
```diff
@@ -51,6 +63,35 @@ abstract class AddonPopupBaseFragment : Fragment(), EngineSession.Observer, User
                 owner = this,
                 view = view,
             )
+
+            downloadsFeature.set(
```
jonalmeida, line 67: Is this truly safe? DownloadsFeature holds some kind of state (in my brief look) and how does that work when there are two instances of it (`BrowserFragment` and `AddonPopupBaseFragment`)?

jonalmeida (commented):
I have the bandwidth for this any more, can you relay it to another teammate? 🙂

Amejia481:
> I do not have the bandwidth for this any more, but I have left two questions to consider. Can you relay it to another teammate? 🙂

Sure, it's not a problem!

fenix/app/src/main/java/org/mozilla/fenix/addons/AddonPopupBaseFragment.kt
Amejia481: We are instantiate`DownloadsFeature` in both `BrowserFragment` and `AddonPopupBaseFragment`  same happens with [promptsFeature](https://github.com/mozilla-mobile/firefox-android/pull/4845/files/3f98238951f3e470475863120997d5011516357f#diff-c1b68c76c9740c7e30d694d89f40819528bb55134ae00706bbe825b22efcb52cR39).

fenix/app/src/test/java/org/mozilla/fenix/browser/DownloadDialogUtilsTest.kt
Amejia481: Thanks for bringing this up I removed the spy.

rocketsroger (approved):
Tested and works great.  🚢

PR #4851, opened 2023-12-15
Bug 1870231 - Remove no-ops from ExternalAppBrowserActivity

t-p-white:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


MozillaNoah (approved):

MozillaNoah:
Per [this discussion in the RFC](https://github.com/mozilla-mobile/firefox-android/pull/4732/files/b76fcd28f9fc4b1980d99b15c46a360520465014#r1433222942), should this actually land after steps 2-4?

t-p-white:
> Per [this discussion in the RFC](https://github.com/mozilla-mobile/firefox-android/pull/4732/files/b76fcd28f9fc4b1980d99b15c46a360520465014#r1433222942), should this actually land after steps 2-4?

Thanks @MozillaNoah, please see reply to the linked comment. I've now gated those function calls in `HomeActivity` by checking if the `Activity` is an `ExternalAppBrowserActivity`

fenix/app/src/test/java/org/mozilla/fenix/settings/account/AuthCustomTabActivityTest.kt
```diff
@@ -0,0 +1,60 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
t-p-white, line 1: Same tests taken from `ExternalAppBrowserActivityTest`

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -695,7 +697,13 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         startupPathProvider.onIntentReceived(intent)
     }
 
-    open fun handleNewIntent(intent: Intent) {
+    @VisibleForTesting
+    @Suppress("UndocumentedPublicFunction") // only visible for tests.
```
gabrielluong, line 701:
```suggestion
    @Suppress("UndocumentedPublicFunction") // only visible for tests.
```
We should avoid this suppressing UndocumentedPublicFunction. Could we change the visibility of the this function to `internal` instead?

t-p-white: As discussed offline, adding kDoc and removing the suppression. Follow-up [task](https://bugzilla.mozilla.org/show_bug.cgi?id=1873009) created to investigate if we want should adjust detekt rules and make this not fail for VisibleForTesting

gabrielluong (approved):

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -695,7 +697,12 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
         startupPathProvider.onIntentReceived(intent)
     }
 
-    open fun handleNewIntent(intent: Intent) {
+    @VisibleForTesting
+    internal fun handleNewIntent(intent: Intent) {
+        if (this is ExternalAppBrowserActivity) {
+            return
+        }
```
MatthewTighe, line 704: 🚗 it feels a little code smelly to me for an `open` type to have knowledge of its subclasses. maybe `if (this !is HomeActivity)` could work instead?

MozillaNoah: Agreed, but it is temporary. Wouldn't this always be false though because of type inheritance, since an `ExternalAppBrowserActivity`'s super type is `HomeActivity`?

t-p-white: > Wouldn't this always be false though because of type inheritance, since an ExternalAppBrowserActivity's super type is HomeActivity

Yep, this. Agree it's not ideal but as this gets further decoupled we'll be able to refine these checks

t-p-white: in fact this check will just be changed to check a boolean if `ExternalAppBroswerActivity` is no more

MatthewTighe: Do you mean that we'll be using a single activity and deliberately tracking whether we are in custom tab mode, or something like that?

t-p-white: Yeah exactly that - essentially using a flag in the `Activity` to track if is 'external' or not

rvandermeulen:
@t-p-white `e: file:///builds/worker/checkouts/vcs/fenix/app/src/test/java/org/mozilla/fenix/settings/account/AuthCustomTabActivityTest.kt:18:47 Unresolved reference: BrowsingModeManager`

t-p-white:
@rvandermeulen strange, last time I checked on this PR all tasks bar the queue had passed

PR #4854, opened 2023-12-15
Bug 1870232 - Delegate the Activity's getNavDirections() functionality

t-p-white:
Step 2 from [Bug 1865306 - RFC Custom Tabs](https://github.com/mozilla-mobile/firefox-android/pull/4732/files)

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1870232

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
@@ -33,7 +33,6 @@ import androidx.appcompat.widget.Toolbar
 import androidx.core.app.NotificationManagerCompat
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.lifecycle.lifecycleScope
-import androidx.navigation.NavDirections
 import androidx.navigation.fragment.NavHostFragment
 import androidx.navigation.ui.AppBarConfiguration
 import androidx.navigation.ui.NavigationUI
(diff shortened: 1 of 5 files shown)


fenix/app/src/main/java/org/mozilla/fenix/HomeActivityNavDirections.kt
```diff
@@ -0,0 +1,163 @@
...
+            "Tried to navigate to ExternalAppBrowserFragment from $from",
+        )
+    }
+}
+
+private fun getHomeNavDirections(
+    from: BrowserDirection,
+): NavDirections = when (from) {
+    BrowserDirection.FromGlobal ->
+        NavGraphDirections.actionGlobalBrowser(null)
```
MozillaNoah, line 90: how come these directions no longer need `customTabSessionId`?

t-p-white: > the home directions never needed it in the first place?
That's exactly it 👍 

> Is it because of the when inside of Activity.getNavDirections
Also yes, so the idea being that the `Activity`s don't have responsibility in determining which `NavDirections` are required

MozillaNoah (approved):
Now that we're in a new release cycle, I'm cool with approving this 👍

t-p-white:
> Now that we're in a new release cycle, I'm cool with approving this 👍

Thanks @MozillaNoah! I've moved the code from the separate 'navigation' file to be in the Activity extension file based on feedback on another review if you could take a look

MozillaNoah (approved):

fenix/app/src/main/java/org/mozilla/fenix/HomeActivityNavDirections.kt
jonalmeida: ^ can we add this to a comment above this method. It looks suspicious and would save a future reader some time if the 'why' was described.

t-p-white: Nice suggestion @jonalmeida, instead I've updated the parameter default to null so we don't have to do this anymore 🙂

t-p-white:
Latest push updates the `nav_graph` to set the default arg to `null` so we no longer need to provide this explicitly 🙂 

cc @MozillaNoah

PR #4859, opened 2023-12-15
Bug 1864709 - Add tab tools and tab counter UI

MozillaNoah:
This PR:
- Stands up the TabTools composable + preview parameters
- Stands up the tab counter UI
- Plumbs in BrowserStore with observations to get the tab counts 

Standard view
image

Tab count with inactive tabs disabled
image

If anyone has issues with the previews not rendering (getting an IDE error), I had to use the non-canary version of Android Studio.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864709


fenix/app/src/main/res/values/strings.xml
```diff
@@ -2363,4 +2363,20 @@
     <string name="download_languages_item_content_description_not_downloaded_state">Download</string>
     <!-- Content description (not visible, for screen readers etc.): For a language list item that is selected. -->
     <string name="download_languages_item_content_description_selected_state">Selected</string>
+
+    <!-- Debug drawer -->
+    <!-- The title of the Tab manager feature in the Debug Drawer. -->
+    <string name="debug_drawer_tab_manager_title" tools:ignore="UnusedResources">Tab manager</string>
+    <!-- The section header for the tab count section in the Tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_title">Tab count</string>
```
MozillaNoah, line 2371: this is copy I've written, so certainly open to feedback here

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+                privateTabs = generateFakeTabs(isPrivate = true),
+                inactiveTabsEnabled = true,
+            ),
+        )
+}
+
+@Composable
+@LightDarkPreview
+private fun TabManagerPreview(
+    @PreviewParameter(TabManagerPreviewModelParameterProvider::class) model: TabManagerPreviewModel,
```
MatthewTighe, line 189: This is interesting!

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+}
+
+private data class TabManagerPreviewModel(
+    val normalTabs: List<TabSessionState> = emptyList(),
+    val inactiveTabs: List<TabSessionState> = emptyList(),
+    val privateTabs: List<TabSessionState> = emptyList(),
+    val inactiveTabsEnabled: Boolean = true,
+)
+
+private class TabManagerPreviewModelParameterProvider : PreviewParameterProvider<TabManagerPreviewModel> {
```
MatthewTighe, line 163: I'm wondering if in the future (when this is in a module and we can use `internal`) if it would make sense to group all the preview utilities into a separate file. I think it could help de-clutter the Composable files.

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2363,4 +2363,20 @@
...
+    <!-- The normal tab count category in the tab count section in the tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_normal">Normal</string>
+    <!-- The inactive tab count category in the tab count section in the tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_inactive">Inactive</string>
+    <!-- The private tab count category in the tab count section in the tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_private">Private</string>
+    <!-- The total tab count category in the tab count section in the tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_total">Total</string>
+    <!-- A footnote to indicate inactive tabs when are disabled. -->
+    <string name="debug_drawer_tab_manager_inactive_tabs_disabled">* Inactive tabs are disabled</string>
```
MatthewTighe, line 2381: Does the `*` notation make sense if translate these?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+import org.mozilla.fenix.tabstray.ext.isNormalTabInactive
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Tab manager UI for [DebugDrawer] that displays the tab counts and allows easy  bulk opening of tabs.
+ */
+@Composable
+fun TabManager(
+    store: BrowserStore,
+    inactiveTabsEnabled: Boolean,
```
MatthewTighe, line 45: I don't think that we need to change anything here, but it strikes me as a little odd that we'd pass the `Store` and a parameter that we're presumably extracting in a layer above. Wondering whether a Composable should either be extracting every value itself or none

MatthewTighe (commented):
I didn't see a separate bug for it, so I think it would be fine to wire this into the drawer as part of this patch. LMK what you decide there and I'll cycle back for a final review

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.ext.maxActiveTime
+import org.mozilla.fenix.tabstray.ext.isNormalTabInactive
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Tab manager UI for [DebugDrawer] that displays the tab counts and allows easy  bulk opening of tabs.
+ */
+@Composable
+fun TabManager(
```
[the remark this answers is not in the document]
MozillaNoah: The thought was to initially have it synced up with the copy of the feature in the drawer feature list. And there's also the `@Composable` annotation to assist with this. "Tabs Overview" I fear might read as something that gives a high level but not an actual tool.

What if we call the feature "Tab Manager" in the copy, but name the Composable "DebugDrawerTabTools"?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
MozillaNoah: That would definitely be a boon. The preview parameters are super helpful but they inherently add some boilerplate bloat to the files.

fenix/app/src/main/res/values/strings.xml
MozillaNoah: [It looks like we have at least one official string with this](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/values/strings.xml#1609).

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
MozillaNoah: so the thought process here is:
1. This is weird problem of using `BrowserStore`, which is an AC store and not a Fenix store, so it just contains the data for tabs in no UI-organizational manner, nor does it track the state of settings related to tabs.
2. The store is passed-in for just values right now, but it'll be needed to dispatch tab creation actions in a follow-up ticket.
3. I'm still thinking through how we're ultimately going to pass-in the drawer feature entries (and how that class/enum will be structured), and passing-in a single store seemed like the easy win for parameter simplification when those feature entries have to be initialized at runtime. This also ensures the observations only occur while the feature is open.
4. Yes, inactiveTabsEnabled, is obtained from settings, ultimately. We can't use `components.settings` at this level though, as it breaks previews, so we can't pass `Settings` as a dependency.

Thinking back on `TabsTray` and its parameter explosion, my goal for this screen is to have the root, feature-scoped composable node be the controlling data node, instead of having `DebugDrawer` or its caller be manager of the stuff specific to the tab manager/overview (e.g. tab stats and tab creation). 

Having a Store + an odd boolean is not ideal, but I can't think of an alternative besides:
- Creating a store/state for this sub feature
- Hoisting ALL of these observations to a level that may conflict with trying to keep the Debug Drawer ignorant of its contents.

MozillaNoah: Another alternative would be creating an interstitial composable, so this top level handles the store + settings observations and then the `content` composable just gets raw data parameters.

MatthewTighe: Thanks for the thoughts! I do have some responses, but they are strictly in the realm of theory and future considerations. Still happy with whatever we choose to do for now here as we will learn more as our goals change

> Thinking back on TabsTray and its parameter explosion

I would still like to find a way to avoid this if possible, and I'm wondering if something like a `TabManagerContract` should be expected by the top-level API. This removes any opinion about the backing architecture and data model, in the case that we want to expose it to consumers that aren't Fenix. Then, consumers can define their own way to scope their state to the contract

Something like:

```kotlin
// public API
data class TabManagerContract(
    val tabs: State<List<TabSessionState>>
    val inactiveTabs: State<..>
    val inactiveTabsEnabled: Boolean
    val addTab: (name: String) -> Unit
)

// Fenix API
fun toTabManagerContract(browserStore, settings): TabManagerContract

// if we wanted to get fancy and use context receivers instead
context(Settings)
fun BrowserStore.toTabManagerContract(): TabManagerContract

// or, more agnostic
fun BrowserStore.toTabManagerContract(inactiveTabsEnabled: Boolean): TabManagerContract
```

It's hard to imagine a world where anyone using the Debug Drawer isn't _also_ using the `BrowserStore`. However, something like `inactiveTabsEnabled` is more likely to live in an app-level Store, which we would not have access to if we were publishing this in A-C or otherwise. 

These thoughts also handily ignore touching on observability and smart recompositions 😅

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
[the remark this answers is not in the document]
MatthewTighe: I like the idea of keeping the copy/Composable in sync, and "Tab tools" strikes a chord with me for sure, but we don't need to cycle on this too much. Follow your heart

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
MozillaNoah: So the core assumption I've been formulating my thoughts around this has been: the debug drawer will be as framework agnostic as possible but the tab manager/tools is something specific to Fenix. So we have the following scenario:

- debug drawer is framework agnostic as possible 
- debug drawer sub features that have the potential to be used by other apps would follow the same
- debug drawer features that are Fenix specific can have our framework as dependencies

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2363,4 +2363,20 @@
     <string name="download_languages_item_content_description_not_downloaded_state">Download</string>
     <!-- Content description (not visible, for screen readers etc.): For a language list item that is selected. -->
     <string name="download_languages_item_content_description_selected_state">Selected</string>
+
+    <!-- Debug drawer -->
+    <!-- The title of the Tab manager feature in the Debug Drawer. -->
+    <string name="debug_drawer_tab_manager_title" tools:ignore="UnusedResources">Tab manager</string>
+    <!-- The section header for the tab count section in the Tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_title">Tab count</string>
+    <!-- The normal tab count category in the tab count section in the tab manager. -->
```
t-p-white, line 2372: nit: 
```suggestion
    <!-- The active tab count category in the tab count section in the tab manager. -->
```

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2363,4 +2363,20 @@
...
+    <string name="debug_drawer_tab_manager_tab_count_normal">Normal</string>
```
t-p-white, line 2373: nit: ```suggestion
    <string name="debug_drawer_tab_manager_tab_count_activel">Active</string>
```

File: fenix/app/src/main/res/values/strings.xml [hunk 9]
```diff
@@ -2363,4 +2363,20 @@
...
+    <!-- The inactive tab count category in the tab count section in the tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_inactive">Inactive</string>
+    <!-- The private tab count category in the tab count section in the tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_private">Private</string>
+    <!-- The total tab count category in the tab count section in the tab manager. -->
+    <string name="debug_drawer_tab_manager_tab_count_total">Total</string>
+    <!-- A footnote to indicate inactive tabs when are disabled. -->
```
t-p-white, line 2380:
```suggestion
    <!-- A footnote to indicate when inactive tabs are disabled. -->
```

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2363,4 +2363,20 @@
     <string name="download_languages_item_content_description_not_downloaded_state">Download</string>
     <!-- Content description (not visible, for screen readers etc.): For a language list item that is selected. -->
     <string name="download_languages_item_content_description_selected_state">Selected</string>
+
+    <!-- Debug drawer -->
+    <!-- The title of the Tab manager feature in the Debug Drawer. -->
+    <string name="debug_drawer_tab_manager_title" tools:ignore="UnusedResources">Tab manager</string>
+    <!-- The section header for the tab count section in the Tab manager. -->
```
t-p-white, line 2370:
```suggestion
    <!-- The title for the tab count section in the Tab manager. -->
```

fenix/app/src/main/res/values/strings.xml
t-p-white, line 1: nit: Sorry, super pedantic, could we use just one format for references to tab manager, so either `Tab manager` or `tab manager`

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+import mozilla.components.lib.state.ext.observeAsState
+import org.mozilla.fenix.R
+import org.mozilla.fenix.compose.Divider
+import org.mozilla.fenix.compose.annotation.LightDarkPreview
+import org.mozilla.fenix.ext.maxActiveTime
+import org.mozilla.fenix.tabstray.ext.isNormalTabInactive
+import org.mozilla.fenix.theme.FirefoxTheme
+
+/**
+ * Tab manager UI for [DebugDrawer] that displays the tab counts and allows easy  bulk opening of tabs.
```
t-p-white, line 40:
```suggestion
 * Tab manager UI for [DebugDrawer] that displays the tab counts and allows easy bulk opening of tabs.
```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+    val tabs by store.observeAsState(initialValue = emptyList()) { state -> state.tabs }
+    val privateTabCount = remember(tabs) { tabs.filter { it.content.private }.size }
+    val inactiveTabCount = remember(tabs) {
+        if (inactiveTabsEnabled) {
+            tabs.filter { it.isNormalTabInactive(maxActiveTime) }.size
+        } else {
+            0
+        }
+    }
+    val normalTabCount = remember(tabs) { tabs.size - privateTabCount - inactiveTabCount }
```
t-p-white, line 56:
```suggestion
    val activeTabCount = remember(tabs) { tabs.size - privateTabCount - inactiveTabCount }
```

t-p-white, line 56: nit: could store `tab.size` as it's also used below

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+                color = FirefoxTheme.colors.textPrimary,
+                style = FirefoxTheme.typography.caption,
+            )
+        }
+    }
+}
+
+@Composable
+private fun TabCountRow(
+    text: String,
```
t-p-white, line 133: nit:
```suggestion
    tabType: String,
```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+        Text(
+            text = count,
+            color = FirefoxTheme.colors.textSecondary,
+            style = FirefoxTheme.typography.headline6,
+        )
+    }
+}
+
+private data class TabManagerPreviewModel(
+    val normalTabs: List<TabSessionState> = emptyList(),
```
t-p-white, line 157:
```suggestion
    val activeTabs: List<TabSessionState> = emptyList(),
```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
t-p-white, line 163:
```suggestion
private class TabManagerPreviewParameterProvider : PreviewParameterProvider<TabManagerPreviewModel> {
```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+}
+
+@Composable
+@LightDarkPreview
+private fun TabManagerPreview(
+    @PreviewParameter(TabManagerPreviewModelParameterProvider::class) model: TabManagerPreviewModel,
+) {
+    val browserStore = BrowserStore(
+        initialState = BrowserState(
+            tabs = model.normalTabs + model.inactiveTabs + model.privateTabs,
```
t-p-white, line 193: nit: 
```suggestion
            tabs = with(model) { normalTabs + inactiveTabs + privateTabs }
```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
```diff
@@ -0,0 +1,226 @@
...
+            text = stringResource(R.string.debug_drawer_tab_manager_tab_count_normal),
+            count = normalTabCount.toString(),
+        )
+
+        TabCountRow(
+            text = stringResource(R.string.debug_drawer_tab_manager_tab_count_inactive),
+            count = if (inactiveTabsEnabled) {
+                inactiveTabCount.toString()
+            } else {
+                "0*"
```
t-p-white, line 100: could we just not show the `inactive` tabs count if it's disabled? or replace '0*' with something like `N/A`?

fenix/app/src/main/res/values/strings.xml
t-p-white: My comment above might mean this is not required

t-p-white (commented):
Nice work Noah, looks great! Mostly naming nits from me, nothing blocking, but I would be interested in your thoughts on not showing `inactive` tabs at all if it's disabled.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
MozillaNoah: It's certainly an option. My thought process was that it'll be more clear to QA, or also really anyone using this feature for the first few times, that the tab counter will _always_ include the inactive tab line item, so it isn't hidden and they'd have to check secret settings to see if inactive tabs are even enabled. 

That being said, this was certainly more of a catch for when the copy was "Normal" versus "Active" tabs. Now that the copy has changed, I'll make this next iteration show/hide the entire inactive section and we can see what that looks like.

MozillaNoah:
Recent push has:
- All of the comment fixes
- The change from "Normal" to "Active" tabs
- Removes the inactive tabs footnote and hides the tab count when disabled
- Stands up a `content` composable (this greatly simplified and optimized the preview code) and added `Settings` as an explicit dependency for the root node.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
MozillaNoah: Update: screenshots have this change

fenix/app/src/main/res/values/strings.xml
MozillaNoah: collapsing this thread in favor of https://github.com/mozilla-mobile/firefox-android/pull/4859#discussion_r1431578629

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabManager.kt
MatthewTighe: Yeah, that makes sense to me!

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/tabs/TabTools.kt
```diff
@@ -0,0 +1,215 @@
...
+ * Tab Tools UI for [DebugDrawer] that displays the tab counts and allows easy bulk-opening of tabs.
+ *
+ * @param store [BrowserStore] used to obtain the tab counts and fire any tab creation actions.
+ * @param settings Used to obtain whether the inactive tabs feature is enabled.
+ */
+@Composable
+fun TabTools(
+    store: BrowserStore,
+    settings: Settings,
+) {
```
MatthewTighe, line 46: Yeah, this makes sense to me.

MatthewTighe (approved):
LGTM!

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2304,4 +2304,18 @@
...
     <!-- Content description (not visible, for screen readers etc.): For a language list item that is selected. -->
     <string name="download_languages_item_content_description_selected_state">Selected</string>
+
+    <!-- Debug drawer -->
+    <!-- The title of the Tab Tools feature in the Debug Drawer. -->
+    <string name="debug_drawer_tab_tools_title" tools:ignore="UnusedResources">Tab Tools</string>
+    <!-- The title of the tab count section in Tab Tools. -->
+    <string name="debug_drawer_tab_tools_tab_count_title">Tab count</string>
+    <!-- The active tab count category in the tab count section in Tab Tools. -->
+    <string name="debug_drawer_tab_tools_tab_count_normal">Active</string>
```
t-p-white, line 2314: nit: perhaps in another PR could `debug_drawer_tab_tools_tab_count_normal` also be updated to `debug_drawer_tab_tools_tab_count_active`?

MozillaNoah: dang. yeah. thanks for catching this.

https://bugzilla.mozilla.org/show_bug.cgi?id=1872881

PR #4860, opened 2023-12-15
Bug 1862263 - Translations Error Support

ohall-m:
This patch defines different possible translation errors, how they map from GeckoView, and how they map to errors we display.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1862263

### DIFF
--- a/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslationUtils.kt
+++ b/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslationUtils.kt
@@ -1,0 +1,61 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+package mozilla.components.browser.engine.gecko.translate
+
+import mozilla.components.concept.engine.translate.TranslationError
+import org.mozilla.geckoview.TranslationsController.TranslationsException
+
+/**
+ * Utility file for translations functions related to the Gecko implementation.
+ */
+object GeckoTranslationUtils {
+    /**
+     * Convenience method for mapping a [TranslationsException] to the Android Components defined
+     * error type of [TranslationError].
+     *
+     * @param throwable The engine throwable that occurred during translating. Ordinarily should be
+     * a [TranslationsException].
+     */
+    fun fromGeckoErrorToTranslationError(throwable: Throwable): TranslationError =
+        if (throwable is TranslationsException) {
+            when ((throwable).code) {
+                TranslationsException.ERROR_UNKNOWN ->
+                    TranslationError.UnknownError(throwable)
+
+                TranslationsException.ERROR_ENGINE_NOT_SUPPORTED ->
+                    TranslationError.EngineNotSupportedError(throwable)
+
+                TranslationsException.ERROR_COULD_NOT_TRANSLATE ->
+                    TranslationError.CouldNotTranslateError(throwable)
+
+                TranslationsException.ERROR_COULD_NOT_RESTORE ->
+                    TranslationError.CouldNotRestoreError(throwable)
+
+                TranslationsException.ERROR_COULD_NOT_LOAD_LANGUAGES ->
+                    TranslationError.CouldNotLoadLanguagesError(throwable)
+
+                TranslationsException.ERROR_LANGUAGE_NOT_SUPPORTED ->
+                    TranslationError.LanguageNotSupportedError(throwable)
+
+                TranslationsException.ERROR_MODEL_COULD_NOT_RETRIEVE ->
+                    TranslationError.ModelCouldNotRetrieveError(throwable)
+
+                TranslationsException.ERROR_MODEL_COULD_NOT_DELETE ->
+                    TranslationError.ModelCouldNotDeleteError(throwable)
+
+                TranslationsException.ERROR_MODEL_COULD_NOT_DOWNLOAD ->
+                    TranslationError.ModelCouldNotDownloadError(throwable)
+
+                TranslationsException.ERROR_MODEL_LANGUAGE_REQUIRED ->
+                    TranslationError.ModelLanguageRequiredError(throwable)
+
+                TranslationsException.ERROR_MODEL_DOWNLOAD_REQUIRED ->
+                    TranslationError.ModelDownloadRequiredError(throwable)
+
+                else -> TranslationError.UnknownError(throwable)
+            }
+        } else {
+            TranslationError.UnknownError(throwable)
+        }
+}
--- a/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
+++ b/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
@@ -1,0 +1,124 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.engine.translate
+
+/**
+ * The types of translation errors that can occur. Has features for determining telemetry error
+ * names and determining if an error needs to be displayed.
+ *
+ * @param errorName The translation error name. The expected use is for telemetry.
+ * @param displayError Signal to determine if we need to specifically display an error for
+ * this given issue. (Some errors should only silently report telemetry or simply revert to the
+ * prior UI state.)
+ * @param cause The original throwable before it was converted into this error state.
+ */
+sealed class TranslationError(
+    val errorName: String,
+    val displayError: Boolean,
+    override val cause: Throwable?,
+) : Throwable(cause = cause) {
+
+    /**
+     * Default error for unexpected issues.
+     *
+     * @param cause The original throwable that lead us to the unknown error state.
+     */
+    class UnknownError(override val cause: Throwable) :
+        TranslationError(errorName = "unknown", displayError = false, cause = cause)
+
+    /**
+     * Default error for unexpected null value received on a non-null translations call.
+     */
+    class UnexpectedNull :
+        TranslationError(errorName = "unexpected-null", displayError = false, cause = null)
+
+    /**
+     * Translations engine does not work on the device architecture.
+     *
+     * @param cause The original throwable before it was converted into this error state.
+     */
+    class EngineNotSupportedError(override val cause: Throwable?) :
+        TranslationError(errorName = "engine-not-supported", displayError = false, cause = cause)
+
+    /**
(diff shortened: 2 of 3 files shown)


ohall-m:
I followed the [WebRequest](https://searchfox.org/mozilla-mobile/search?q=geckoerror&path=firefox-android&case=false&regexp=false) example.

Looking for advice on how best to use this mapper to avoid crossing an abstraction layer when I need to know what the throwable was. 

For example, if `isTranslationsEngineSupported`  had an error, it will return a Gecko error code. Would the best place to be to [intercept it here](https://searchfox.org/mozilla-mobile/rev/bba84847ffbb25a9865b21912792b4186fe13c09/firefox-android/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#698) and make a new exception? Or is there another pattern for this?

Should it be instead of just an enumerated type, of  `TranslationError`, should this be something like `class TranslationException(errorCode: TranslationError) : Exception(message)` that is created `onError`?

jonalmeida:
> I followed the [WebRequest](https://searchfox.org/mozilla-mobile/search?q=geckoerror&path=firefox-android&case=false&regexp=false) example.
> 
> Looking for advice on how best to use this mapper to avoid crossing an abstraction layer when I need to know what the throwable was.
> 
> For example, if `isTranslationsEngineSupported` had an error, it will return a Gecko error code. Would the best place to be to [intercept it here](https://searchfox.org/mozilla-mobile/rev/bba84847ffbb25a9865b21912792b4186fe13c09/firefox-android/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#698) and make a new exception? Or is there another pattern for this?
> 
> I would make something like `class TranslationException(errorCode: TranslationError) : Exception(message)` that is created `onError` that maps the Gecko Error to the Translation Error code?

That could work as well, although I would modify that signature to include the original `throwable` if there is helpful information in there coming from Gecko.

Is the idea to put the `errorCode` in the existing `message` property in the Exception or is it something separate? I think both directions would be fine too.

Alternatively, there is prior art with `onCheckForFormDataException` where we bubble up the exception to the BrowserStore, so that we can get telemetry back with the `TelemetryMiddleware` in Fenix.

ohall-m:
>Is the idea to put the errorCode in the existing message property in the Exception or is it something separate? 

I'm not entirely sure yet, still thinking it through. I think I'll need `errorCode` as a property somewhere for telemetry and probably to know for certain which UI error message to show, but that is as far as I've thought out. (I think right now all the UI error messages are close to 1:1 per call, but I have a suspicion the erroring could become more specific down the road.)

>Alternatively, there is prior art with onCheckForFormDataException where we bubble up the exception to the BrowserStore, so that we can get telemetry back with the TelemetryMiddleware in Fenix.

Awesome, thanks for the example! I'll take a look at that too for ideas!

ohall-m:
Will follow up on the earlier discussion mapping the errors as they come from the API in [bug 1872703](https://bugzilla.mozilla.org/show_bug.cgi?id=1872703).

Would like to land this just as a data model for the UI bug #4974 to start using.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationErrorDisplay.kt
```diff
@@ -0,0 +1,54 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.concept.engine.translate
+
+/**
+ * The types of translation errors we display to the user.
+ *
+ * @param errorName The error name of the translation error to display.
+ */
+enum class TranslationErrorDisplay(val errorName: String) {
```
Amejia481, line 12: Maybe we could transform `TranslationError` into a [sealed class](https://kotlinlang.org/docs/sealed-classes.html), so we could have a generic `TranslationError` with multiple children or groups : 


``` kotlin
sealed class TranslationError(val errorName: String) {
 data class NoDisplay(override val caller: errorName) : TranslationError()
 data class UnableToTranslate(override val caller: errorName) : TranslationError()
 data class UnableToLoadLanguage(override val caller: errorName) : TranslationError()
 data class LanguageNotSupported(override val caller: errorName) : TranslationError()
}

```

Amejia481: What do you think?

Amejia481: Also what is the purpose of `errorName`? Is it going to be used for showing a message to the user?

ohall-m: Thanks for taking a look!

Sounds like a great idea, I'll give it a try and update!

>Also what is the purpose of errorName?  

Envisioning it as what telemetry will report as the error reason. Something like [how this is used](https://searchfox.org/mozilla-mobile/rev/b4fff7f0b3d73d23a8baf7711dc8e0163518d2a1/firefox-android/fenix/app/metrics.yaml#567-570) or a pre-specified value. (I did the mapping incorrectly in the past like [this](https://searchfox.org/mozilla-mobile/rev/b4fff7f0b3d73d23a8baf7711dc8e0163518d2a1/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/share/SaveToPDFMiddleware.kt#107).)

ohall-m: (Could probably just have had `errorName` on `TranslationError` and not needed on `TranslationErrorDisplay`, now that you mention it, though. I'll fix in the refactor at any rate. 😅)

ohall-m: Refactored to a sealed class. 🦭 Please let me know if you think this works! I inherited from `Throwable`. 

I am thinking in [bug 1872703](https://bugzilla.mozilla.org/show_bug.cgi?id=1872703) I'll go through to all the translations functions, for example, [`isTranslationsEngineSupported`](https://searchfox.org/mozilla-mobile/rev/bba84847ffbb25a9865b21912792b4186fe13c09/firefox-android/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#698) and change the `onError` something like `onError(GeckoTranslationUtils.fromGeckoErrorToTranslationError(throwable))`. 

I might also add another named error for unexpected nulls instead of using [`translationsUnexpectedNull`](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/GeckoEngine.kt#676) to use in that bug.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
```diff
@@ -0,0 +1,76 @@
...
+ * names and determining if an error needs to be displayed.
+ *
+ * @param errorName The translation error name. The expected use is for telemetry.
+ * @param displayError Signal to determine if we need to specifically display an error for
+ * this given issue. (Some errors should only silently report telemetry or simply revert to the
+ * prior UI state.)
+ *
+ */
+
+sealed class TranslationError(val errorName: String, val displayError: Boolean) : Throwable() {
```
jonalmeida, line 18: `displayError` is a nice addition.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Default error for unexpected issues.
+     *
+     * @param originalThrowable The original throwable that lead us to the unknown error state.
+     */
+    class UnknownError(val originalThrowable: Throwable? = null) : TranslationError("unknown", false)
```
jonalmeida, line 25: Should the `originalThrowable` just be an override for the Throwable's `cause`?

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationErrorDisplay.kt
jonalmeida: The above sounds good to me.

jonalmeida (approved):
Your direction looks good to me. Just left a few nits to consider.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
ohall-m: Ohh, yes, that makes sense! Do you think all of the errors should get a `cause` in that case?

Amejia481 (approved):
LGTM!

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
ohall-m: Seemed like I should add it to all of them since it could add value down the road for debugging. 

Added cause as required on `UnknownError`, cause as null on `UnexpectedNull` (since we don't really know how it happened), and cause as a non-default optional on the others, but with the cause specified in the mapper.

android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslationUtils.kt
```diff
@@ -0,0 +1,61 @@
...
+object GeckoTranslationUtils {
+    /**
+     * Convenience method for mapping a [TranslationsException] to the Android Components defined
+     * error type of [TranslationError].
+     *
+     * @param throwable The engine throwable that occurred during translating. Ordinarily should be
+     * a [TranslationsException].
+     */
+    fun fromGeckoErrorToTranslationError(throwable: Throwable): TranslationError =
+        if (throwable is TranslationsException) {
```
jonalmeida, line 21: @ohall-m this is styling feedback that I only thought off a bit too late, so feel free to consider it as a future change or ignore!

In Kotlin, you can add type-safe utility methods with extension functions that are identical to this as static java methods, but might be easier to use:

```kotlin
fun TranslationsException.intoTranslationError(): TranslationError {
  return when (code) {
    // ...
  }
}
```

This could mean one less type-check if you add it straight at the caller.

ohall-m: Ohh, thank you, that's neat! I miss out on a lot of cool Kotlin features. 

I'll be revisiting in [bug 1872703](https://bugzilla.mozilla.org/show_bug.cgi?id=1872703) to connect this to all the translations calls. 

I'll adjust it over there very shortly since this bug just landed. I'll leave a note on the bug!

PR #4324, opened 2023-11-01
Bug 1861732 - Split-out "terms of use" and "privacy policy" text links

MozillaNoah:
| Before | After | 
| -- | -- |
| image | image |

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1861732

PR #4874, opened 2023-12-18
Bug 1857850 - [a11y] Improve experience when accessing links in text

DreVla:
This patch aims to improve a11y experience when accessing links present in bodies of text. When we have a link text element, we now indicate that there are links present in it and upon double tapping, if there is only one link, that one will be opened, if there are multiple ones, a dialog is open from where the user can access any of the links present in the text. This behavior tries to follow the default Talkback a11y one where the user can access links from a text in case they are annotated as URLs.

This solution makes all the links in a text accessible and also improves accessibility for lower version devices.

Review checker was chosen to showcase thise due the text with two links, but this solution applies to all places wher `LinkText` is used.

| Talkback | Our Solution |
| ------------- | ------------- |
| <video src="https://github.com/mozilla-mobile/firefox-android/assets/32488956/e821be8e-d246-4a2c-8892-fd11ba5b07a7">  | <video src="https://github.com/mozilla-mobile/firefox-android/assets/32488956/450469fc-2db8-4b3a-a0ce-d0096e952c41">|

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1857850

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
@@ -7,10 +7,22 @@ package org.mozilla.fenix.compose
 import androidx.annotation.VisibleForTesting
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.ClickableText
+import androidx.compose.material.Card
+import androidx.compose.material.Text
+import androidx.compose.material.TextButton
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.contentDescription
 import androidx.compose.ui.semantics.onClick
 import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.AnnotatedString
@@ -20,6 +32,9 @@ import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.window.Dialog
+import org.mozilla.fenix.R
 import org.mozilla.fenix.theme.FirefoxTheme
 
 /**
@@ -67,6 +82,12 @@ fun LinkText(
         linkTextDecoration,
     )
 
+    val showDialog = remember { mutableStateOf(false) }
+    val linksAvailable = stringResource(id = R.string.a11y_links_available)
+
+    if (showDialog.value) {
+        LinksDialog(linkTextStates) { showDialog.value = false }
+    }
     // When using UrlAnnotation, talkback shows links in a separate dialog and
     // opens them in the default browser. Since this component allows the caller to define the
     // onClick behaviour - e.g. to open the link in in-app custom tab, here StringAnnotation is used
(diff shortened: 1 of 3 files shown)


DreVla:
The reasons we cannot use UrlAnnotation is that if fenix is not set as default browser, Talkback opens the URL from the text in whatever browser is the default one.

https://github.com/mozilla-mobile/firefox-android/assets/32488956/83890127-2bc0-42e7-ad7e-641fceb4e29f

DreVla:
@anna-yeddi can to take a look from an a11y perspective and let us know if this would be a good approach?

anna-yeddi:
> @anna-yeddi can to take a look from an a11y perspective and let us know if this would be a good approach? 

This looks like an expected behavior. Just make sure the TalkBack would announce (after reading the entire block of text with links) that `actions available, double-tap and long press` or similar, basically hinting the user that there are, in fact, additional actions available for them to access.

DreVla:
> > @anna-yeddi can to take a look from an a11y perspective and let us know if this would be a good approach?
> 
> Just make sure the TalkBack would announce (after reading the entire block of text with links) that `actions available, double-tap and long press` or similar, basically hinting the user that there are, in fact, additional actions available for them to access.

Yep, the way this is handled is that when there are links in the text, after the whole text has been read, it is followed by `links available. Double tap to activate`

DreVla:
@Delphine considering that we reverted to previous strings that now have `moz:RemovedIn`, can we just remove that? Or should we just create new ones?

Delphine:
@DreVla Sorry, I'm not sure that I have all the context by looking at the string file in this PR.
Which strings are you asking to remove (and have been labeled as `moz:RemovedIn`)?

Note that I'm off until January 8, so this might have to wait until then (unless someone else can help in the meantime). Thanks

DreVla:
> @DreVla Sorry, I'm not sure that I have all the context by looking at the string file in this PR. Which strings are you asking to remove (and have been labeled as `moz:RemovedIn`)?
> 
> Note that I'm off until January 8, so this might have to wait until then (unless someone else can help in the meantime). Thanks

No rush, in [this](https://github.com/mozilla-mobile/firefox-android/pull/4324) PR we have changed the terms of use & privacy policy string from being one text that has two links into a simple text followed by two separate links. This was done for a11y reasons. This patch aims to revert that change because we found a solution to access multiple links present in one body of text using Talkback. You can find the strings I am referring to [here](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/values/strings.xml#2180-2190).  In this patch, we reverted to using the strings marked with `moz:RemovedIn="121"` so I am unsure if we should just delete the `RemovedIn` or simply add new strings that are basically the same ones as the ones marked to be removed.

anna-yeddi:
> LE: After some modifications, the additional text was changed from links available. Double tap to activate to actions available. Double tap to activate

Oh, sorry, if this was customized it's not what I meant. Whatever is the default system hint - is the one that we want to hear (this would be a confirmation of the programmatic availability of these controls). Apologies for the confusion

anna-yeddi:
> > LE: After some modifications, the additional text was changed from links available. Double tap to activate to actions available. Double tap to activate
> 
> Oh, sorry, if this was customized it's not what I meant. Whatever is the default system hint - is the one that we want to hear (this would be a confirmation of the programmatic availability of these controls). Apologies for the confusion 

@DreVla tagging you for visibility

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2179,17 +2179,17 @@
     <!-- Clickable text from the contextual onboarding card that links to review quality check support article. -->
     <string name="review_quality_check_contextual_onboarding_learn_more_link">Learn more</string>
     <!-- Caption text to be displayed in review quality check contextual onboarding card above the opt-in button. First parameter is the Fakespot product name. Following parameters are for clickable texts defined in review_quality_check_contextual_onboarding_privacy_policy and review_quality_check_contextual_onboarding_terms_use. In the phrase "Fakespot by Mozilla", "by" can be localized. Does not need to stay by. -->
-    <string name="review_quality_check_contextual_onboarding_caption" moz:RemovedIn="121" tools:ignore="UnusedResources">By selecting “Yes, try it” you agree to %1$s by Mozilla’s %2$s and %3$s.</string>
+    <string name="review_quality_check_contextual_onboarding_caption">By selecting “Yes, try it” you agree to %1$s by Mozilla’s %2$s and %3$s.</string>
```
DreVla, line 2182: Since we went back to using the initial strings, I have removed `moz:RemovedIn="121" tools:ignore="UnusedResources` from them and added it to _2 strings.

ewachowiak:
For the single link experience ("Learn more" from this example), is it accurate that Voiceover would read the entire paragraph and then say "Links available, double tap to activate" - referring to comment above that says:

> LLE: Reverted to `links available. Double tap to activate` since this is also the default Talkback message.

1. If there is only one link available, would Voiceover read "Link available" singular?
2. I have a concern that the user may not know what the link is without additional context. In this case, the link is at the end of the paragraph, but a link could be anywhere in the paragraph, and even coming at the end it may not be obvious. Therefore a user won't know what type of destination to expect when they "double tap to activate." Is there a way to use alt text or to somehow describe the link more clearly? e.g. "[Voiceover reads entire paragraph]. Link available: Learn more about Review Checker. Doubletap to activate." In this case, "Learn more about Review Checker" is the link's alt text.

CC @anna-yeddi Let me know if you disagree with the above concerns.

anna-yeddi:
Thank you both for the thoughtful investigation! These are a screen reader provided behaviors and are expected in this case, we do not need to edit them and we cannot do it since we have no control over these announcements.

We could add an accessible hint with extra information but it is not necessary and some users would have hints disabled so I don't think we need/want to rely on it.

fenix/app/src/main/res/values/strings.xml
Delphine: This approach should work for l10n!

fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
```diff
@@ -67,6 +82,14 @@ fun LinkText(
         linkTextDecoration,
     )
 
+    val showDialog = remember { mutableStateOf(false) }
+    val actionsAvailable = stringResource(id = R.string.a11y_links_available)
+
+    if (showDialog.value) {
+        LinksDialog(
+            linkTextStates,
+        ) { showDialog.value = false }
```
rahulsainani, line 91: 🔍 Could this be in one line?

fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
```diff
@@ -76,19 +99,78 @@ fun LinkText(
...
+            contentDescription = "$annotatedString $actionsAvailable"
         },
         onClick = { charOffset ->
             onTextClick(annotatedString, charOffset, linkTextStates)
         },
     )
 }
 
+@Composable
+internal fun LinksDialog(
```
rahulsainani, line 121: I think this can be private

fenix/app/src/main/java/org/mozilla/fenix/compose/LinkText.kt
```diff
@@ -67,6 +82,14 @@ fun LinkText(
         linkTextDecoration,
     )
 
+    val showDialog = remember { mutableStateOf(false) }
+    val actionsAvailable = stringResource(id = R.string.a11y_links_available)
```
rahulsainani, line 86: 💡 `linksAvailable` could be a better name

rahulsainani (approved):
Looks really good! 💯 Just a function visibility modifier comment and some nits.

PR #4900, opened 2023-12-19
Bug 1870912 - hook up the `FxaStateMachineChecker` code

bendk:
This PR hooks up the `FxaStateMachineChecker` code so that we can do a "dry-run" test of the new app-services logic against the current android-components logic.  This checks if the behavior would be the same, without actually having the app-services code perform any `FxaClient` calls or other side effects.

I'm flagging this `do not land`  because #4506 should land first and also because I want to test it out myself a bit more before landing.  I still think it's useful for visibility though.  It makes the plan laid out in the [RFC](docs.google.com/document/d/1Us3Y6mjkCEYIqb-BCfRS15IX6jcoWhRkpHMVI0rKukg/edit) a bit more concrete.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1870912


bendk:
You can see the tests with `adb logcat | grep fxa_client::state_machine::checker`.

Here's an example run where everything matched up.  In the app I started in the connected state, logged out, and logged back in.  (I cut off the left side of the logs to make it more readable).
```
breadcrumb: fxa-state-machine-checker: Initialize -> GetAuthState
breadcrumb: fxa-state-machine-checker: GetAuthStateSuccess -> EnsureDeviceCapabilities
breadcrumb: fxa-state-machine-checker: EnsureDeviceCapabilitiesSuccess -> Complete
breadcrumb: fxa-state-machine-checker: Disconnect -> Disconnect
breadcrumb: fxa-state-machine-checker: DisconnectSuccess -> Complete
breadcrumb: fxa-state-machine-checker: BeginOAuthFlow -> BeginOAuthFlow
breadcrumb: fxa-state-machine-checker: BeginOAuthFlowSuccess -> Complete
breadcrumb: fxa-state-machine-checker: CompleteOAuthFlow -> CompleteOAuthFlow
breadcrumb: fxa-state-machine-checker: CompleteOAuthFlowSuccess -> InitializeDevice
breadcrumb: fxa-state-machine-checker: InitializeDeviceSuccess -> Complete
```

Here's an example of a failure (caused because I didn't hook things up properly in the a-c code).   In the app, I tried to connect.  The error happened because when the `android-components` was about to call `FxaClient.completeOAuthFlow()` but the app-services state machine wasn't in the `CompleteOAuthFlow` state.

```
breadcrumb: fxa-state-machine-checker: Initialize -> GetAuthState
breadcrumb: fxa-state-machine-checker: GetAuthStateSuccess -> Complete
breadcrumb: fxa-state-machine-checker: BeginOAuthFlow -> BeginOAuthFlow
breadcrumb: fxa-state-machine-checker: BeginOAuthFlowSuccess -> Complete
report fxa-state-machine-checker: State mismatch: Complete vs CompleteOAuthFlow
```

bendk:
That last log ends with `report:`, which indicates that we would report that to Sentry (along with all the previous breadcrumbs).  Once we report one error, the checking stops to avoid spamming Sentry.

android-components/components/feature/accounts/src/test/java/mozilla/components/feature/accounts/FirefoxAccountsAuthFeatureTest.kt
```diff
@@ -71,6 +73,10 @@ class TestableFxaAccountManager(
 
 @RunWith(AndroidJUnit4::class)
 class FirefoxAccountsAuthFeatureTest {
+    val entryPoint: FxAEntryPoint = mock<FxAEntryPoint>().apply {
```
Mukuljangir372, line 76: Should we rename `entryPoint` to `mockEntryPoint` as `FxAEntryPoint` is a mocked class. What do you think?

Mukuljangir372 (approved):

mavduevskiy (changes requested):
That's a good start. I really appreciate the approach you chose for making sure that the two state machines work in sync.

There are a few changes to the code structure I would like to suggest. 

Firstly, I would encourage keeping as little changes in `FxaAccountManager` as possible. We could create a wrapper for `FxaStateMachineChecker` where we could explain the idea behind this AC class and some of its behaviour. We could put the mapping of AC to AS that happen in `handleASPublicEvent` and `checkASPublicState` events there.

Secondly, we could create a similar mapping function for `internalStateSideEffects` which would map AC progress events to AS level counterparts instead of putting those directly into `internalStateSideEffects`. We could also chain `checkInternalState` with `handleInternalEvent` inside that wrapper. 

Basically, I am suggesting to wrap the temporary code into a class that would accept events from AC, then map those to AS and feed to the checker. That way `FxaAccountManager` will have the least amount of extra code and the code we need for testing will be contained within one class. It makes for easier understanding, managing and removing once we no longer needed it.

One note about the comments: I would encourage you to think about comments as a lower abstraction layer than code itself. Reading the code is most of the time self-explanatory, and most of the logic can be understood pretty fast. Explaining something of that calibre in a comment isn't really worth it (and more often then not it still does only half of the job, reading the code is still necessary). But some things (like architecture, edge cases, oddities) might take hours to figure out, and those are good candidates to be explain on a conceptual level. In the patch a good example of something not that obvious would be reporting to sentry inside `checkInternalState` methods in AS. 

One thing that is not really clear to me – with public states we change the state first, then verify it, but with inner states we do the reverse – I wonder why? Making a state transition and then verifying that the new state matches the expected one makes more sense to me.

bendk:
Pushed a new commit that moves some of the state checking to a separate module/class, but like I mentioned above I can't figure out a way to isolate it more.  I think that would require refactoring the android-components state machine more, which doesn't seem worth it to me.  I also made the comment more concise.  The last change was to move the `checkEvent` call to after we've checked that the event was valid -- i.e. it resulted in a state transition.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/AppServicesStateMachineChecker.kt
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * The app-services internal states correspond the results of internal firefox account method
+     * calls.  Call this before after making a call.
+     */
+    internal fun handleInternalEvent(event: FxaStateCheckerEvent) {
+        rustChecker.handleInternalEvent(event)
+    }
+}
+
+public val appServicesStateMachineChecker = AppServicesStateMachineChecker()
```
mavduevskiy, line 105: it looks like the intention here is to have a singleton. if it's the case, let's just change `AppServicesStateMachineChecker` to an object and use it directly.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
```diff
@@ -551,14 +558,32 @@ open class FxaAccountManager(
         via: Event,
     ): Event? = when (forState.progressState) {
         ProgressState.Initializing -> {
+            appServicesStateMachineChecker.checkInternalState(FxaStateCheckerState.GetAuthState)
```
mavduevskiy, line 561: It looks like all the branches take in consideration current state, via event and the result progress state and call `AppServicesStateMachineChecker` respectively. If we refactor the checker to have something like

```
internal fun validateProgressState(
        forState: State.Active,
        via: Event,
        result: Event,
    ) {
        when (forState.progressState) {
            ProgressState.Initializing -> {
                rustChecker.checkInternalState(FxaStateCheckerState.GetAuthState)
                when (result) {
                    Event.Progress.AccountNotFound -> {
                        rustChecker.handleInternalEvent(
                            FxaStateCheckerEvent.GetAuthStateSuccess(FxaRustAuthState.DISCONNECTED),
                        )
                    }
                    Event.Progress.AccountNotFound -> {
                        rustChecker.handleInternalEvent(
                            FxaStateCheckerEvent.GetAuthStateSuccess(FxaRustAuthState.CONNECTED),
                        )
                    }
                    else -> {}
                }
            }

            else -> {}
        }
    }

we could move the branching logic there and reduce the change to the `FxaAccountManager.kt` itself.

The exceptions are `Event.Account.BeginEmailFlow` and `Event.Account.BeginPairingFlow`. 

I wonder if we could make it more neat? I would move out of `FxaAccountManager` class as much code as possible. Wonder what you think.

mavduevskiy (commented):
works nicely! I have a question about the code organization, but otherwise we should be landing it soon.

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/FxaAccountManager.kt
bendk: This makes sense to me.  I moved all the logic I could into that class and the only exceptions left were the complete flow events, there's much less changes to `FxaAccountManager` now.

mavduevskiy (approved):
looks good! I might consider adding unit tests in case it might raise your condifence about the change. but since it's easy to see its behaviour using logger and it's a temporary code that doesn't affect the logic, landing this without tests is fine with me as well

bendk:
I didn't add tests to this for a few reasons:
  - This code is temporary and will be removed in the near future, so tests provide less benefit.
  - I couldn't think of a way to write tests that didn't just copy the logic from the code
  - The ultimate test for this is the Sentry events that we'll be getting back.  Those should identify errors in the code.

PR #4965, opened 2023-12-27
Bug 1872184 - Rate limit auth recovery checks

bendk:
Don't schedule multiple auth recovery checks when they happen in the same time period.  If multiple operations run at once when the user has a bad token, then we shouldn't queue up multiple checks at the same time.

The exception to this is syncing.  When we're syncing and hit an auth error, we always want to schedule an auth recovery check so that either the sync restarts or the user enters the authentication problems state. Also, there should only be one sync scheduled at once, so there's less chance of multiple recovery checks being queued at once.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1872184


Mukuljangir372 (approved):

android-components/components/service/firefox-accounts/src/main/java/mozilla/components/service/fxa/manager/GlobalAccountManager.kt
```diff
@@ -54,7 +58,18 @@ internal object GlobalAccountManager {
             // First error, start our count.
             authErrorCountWithinWindow = 1
         } else if (timeSinceLastAuthErrorMs <= AUTH_CHECK_CIRCUIT_BREAKER_RESET_MS) {
-            // Error within the reset time window, increment the count.
```
mavduevskiy, line 57: I would suggesting keeping the original comment above the increment line, but adding a single comment above the sync check explaining the case when we want to quit the execution. One more thing, a code comment is better when it's a high level description. Mentioning implementation details like `If this is for an operation wrapped by `handleFxaExceptions`, then quit` might make a comment obsolete when somebody decided to change the name of the function or refactor the general flow of execution.

mavduevskiy (changes requested):
looking good! a small nit about the comment, and we probably should write a test for checking `forSync` / `!forSync` cases

mavduevskiy (approved):
looks good, 🚢it

PR #5011, opened 2024-01-03
Bug 1872885 - Move debug drawer store management from `DebugOverlay` to `FenixOverlay`

MozillaNoah:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1872885


fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/FenixOverlay.kt
```diff
@@ -0,0 +1,45 @@
...
+                debugDrawerStore.dispatch(DebugDrawerAction.DrawerClosed)
+            },
+        )
+    }
+}
+
+@LightDarkPreview
+@Composable
+private fun FenixOverlayPreview() {
+    FenixOverlay()
```
MozillaNoah, line 44: An additional positive side effect from creating `FenixOverlay` is now the entire feature will be interactable within the previews, meaning Tab Tools and everything else we eventually plug in here will be fully functional!

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/FenixOverlay.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.mozilla.fenix.debugsettings.store.DrawerStatus
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.theme.Theme
+
+/**
+ * Overlay for presenting Fenix-wide debugging content.
+ */
+@Composable
+fun FenixOverlay() {
+    val debugDrawerStore = remember { DebugDrawerStore() }
```
MatthewTighe, line 23: Are there any perf considerations for remembering a complex type like this?

MatthewTighe (approved):

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/FenixOverlay.kt
MozillaNoah: Only in the positive sense that it'll help save on recompositions. Other than having `DebugDrawerStore` as a parameter for `FenixOverlay` and passed-in here via `HomeActivity`, this is the only way to ensure that store does not get nuked during a random recomp. 

Thinking aloud though, this may need to be re-examined in the lifecycle case though (like when the device is rotated) 🤔

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DrawerStatus.kt
```diff
@@ -0,0 +1,20 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.debugsettings.store
+
+/**
+ * Possible values of the debug drawer's physical state.
+ */
+enum class DrawerStatus {
```
t-p-white, line 10: nit: `DrawerState`?

t-p-white: this would affect some of the variable names

t-p-white (approved):
lgtm 😎 with 1 nit

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DrawerStatus.kt
MozillaNoah: I've been hesitant to use `DrawerState` to avoid any naming overload confusion, since we have `DrawerState` from `ModalDrawer` and then `DebugDrawerState` from the redux.

t-p-white: Ah okay, gotcha! Based on the desc for `DebugDrawerState` `UI state of the debug drawer feature.` it seems like this enum is a part of this?

MozillaNoah: It is, but after the PR, it's now a value passed into `DebugOverlay` and then managed outside of it (so if this is uplifted to AC, it can be used with any framework). So in other words it's a dependency of the feature but then managed in Fenix via `DebugDrawerState`.

t-p-white: 👍 

In that case, I'm wondering if we could do with updating some of the naming around this, as I'd say this really is a 'state' over a 'status' and we're only using that because it's already used? Maybe `DebugDrawerState` should become something like `DebugDrawerUiState` or `DebugDrawerUiData`? 

`OnboardingPageUiData.kt` maybe a useful reference

MozillaNoah: I hear you. I think I need to simmer on this, but I'll add a note to my todo to passively think about this in the next week or so. If you, or anyone else randomly reading this thread, come up with a way to keep our store/state naming convention while having an indicative name for the drawer open/close enum (that doesn't overlap with the Composable library naming) please send them my way!

PR #5012, opened 2024-01-03
Bug 1864074 - Create debug drawer landing page UI

MozillaNoah:
Screenshots below. Also debating whether to add the "Debug Drawer" verbiage somewhere, either in a toolbar above the header or as part of it.

image

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864074


fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
```diff
@@ -6,31 +6,118 @@ package org.mozilla.fenix.debugsettings.ui
...
+
+    val appName: String
+    val appVersion: String
+    if (inComposePreview) {
+        appName = "App Name Preview"
+        appVersion = "100.00.000"
+    } else {
+        appName = LocalContext.current.appName
+        appVersion = LocalContext.current.appVersionName
+    }
```
MozillaNoah, line 53: Our extension function, `Context.appName` is broken in previews 💔

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
```diff
@@ -6,31 +6,118 @@ package org.mozilla.fenix.debugsettings.ui
...
 
 /**
  * The debug drawer UI.
  */
 @Composable
-fun DebugDrawer() {
-    Box(
+fun DebugDrawer(
+    navigationItems: List<String>,
+    onNavigationItemClick: (String) -> Unit,
```
MozillaNoah, line 41: need a second opinion on these parameter names. The only other potential candidate I had was something along the lines of "debugDestinations", but navigation items seemed to fit the bill better.

t-p-white: how about something like `menuItems`? Then that would read as Debug drawer menu items

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
```diff
@@ -6,31 +6,118 @@ package org.mozilla.fenix.debugsettings.ui
...
+        Divider()
+
+        LazyColumn(
+            modifier = Modifier.fillMaxSize(),
+            state = lazyListState,
+        ) {
+            items(navigationItems) { item ->
+                TextListItem(
+                    label = item,
+                    onClick = {
```
t-p-white, line 85: nit: could be inlined

t-p-white (commented):
1. Could the app name and version all be on the same line? Seems like we could utilize the horizontal space. 
2. Are we planning on having this feature only in debug? If so, I wonder how value the app name is adding.

> Also debating whether to add the "Debug Drawer" verbiage somewhere, either in a toolbar above the header or as part of it.

3. I like this idea, perhaps even as a (centred?) title for the drawer. I'm wondering if this should be presented user facing as a 'debug menu' or 'debug tools'

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MatthewTighe: +1 `menuItems`. then there isn't an implied opinion about whether they need to actually navigate, or if they're top-level.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
```diff
@@ -6,31 +6,118 @@ package org.mozilla.fenix.debugsettings.ui
...
-fun DebugDrawer() {
-    Box(
+fun DebugDrawer(
+    navigationItems: List<String>,
+    onNavigationItemClick: (String) -> Unit,
+) {
+    val lazyListState = rememberLazyListState()
+
+    val appName: String
+    val appVersion: String
```
MatthewTighe, line 46: Git revision, A-C version, and geckoview version may all be useful things to add here as well

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MatthewTighe: Could also introduce a type here so we don't end up needing to do String lookups between the two. Something like:

`data class DebugDrawerMenuItem(val label: String, val onClick: () -> Unit)`

MatthewTighe (approved):
Yeah, this LGTM for now. I am assuming with some of this stuff that implementation details may end up having an effect on the final version, so I think we're fine to iterate quickly here

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MatthewTighe: Or in a details subpage

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MozillaNoah: I'm a fan of `menuItems`! I'm actually going to make a class or an enum for menu items as part of the [navigation middleware ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1864075), so we can either do the latter half-stop refactoring now or as part of the [navigation middleware integration](https://bugzilla.mozilla.org/show_bug.cgi?id=1864076).

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MozillaNoah: I'm thinking this will be a subpage. If I include Gecko version, etc., it'll be specific to browsers and not aligned with the generic use case we're working towards.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MozillaNoah: I'll clean this up as part of the work to [integrate the navigation logic](https://bugzilla.mozilla.org/show_bug.cgi?id=1864076).

MozillaNoah:
> 1. Could the app name and version all be on the same line? Seems like we could utilize the horizontal space.
> 
>     2. Are we planning on having this feature only in debug? If so, I wonder how value the app name is adding.
> 
> 
> > Also debating whether to add the "Debug Drawer" verbiage somewhere, either in a toolbar above the header or as part of it.
> 
>     3. I like this idea, perhaps even as a (centred?) title for the drawer. I'm wondering if this should be presented user facing as a 'debug menu' or 'debug tools'

1. Sure, I'll see what that looks like with the next push of changes
2. This will be available in all flavors of Fenix. So this is to help ensure we know what version/flavor someone is working with, especially in the scenario a user submits a screen recording.
3. Ok awesome! I already have a ticket to [add a toolbar for the sub pages](https://bugzilla.mozilla.org/show_bug.cgi?id=1864078), so what I'll do then is just add the toolbar to the root then and have something along the lines of "Debug Tools" as the header.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MozillaNoah: Just kidding. This will be cleaned up in the next push

MozillaNoah:
Recent push includes:
- Renaming `navigationItems` to `menuItems`
- Creating `DebugDrawerMenuItem` data class to handle the UI side of navigation elements.
- Moving the version number to the same line as app name
- Changed the scrolling behavior, so now the header will be included in the LazyColumn and will be scrolled.

MozillaNoah:
Recent push moves the landing page UI to its own Composable, to help with future navigation work.

The screenshot has also been updated

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawerLandingScreen.kt
```diff
@@ -0,0 +1,129 @@
...
+                    text = appVersion,
+                    color = FirefoxTheme.colors.textSecondary,
+                    style = FirefoxTheme.typography.headline5,
+                )
+            }
+
+            Divider()
+        }
+
+        items(menuItems) { menuItem ->
```
rahulsainani, line 87: 💡 It would be performant to have an `id` and set that as key for the each item --> https://developer.android.com/jetpack/compose/performance/bestpractices#use-lazylist-keys

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
rahulsainani: 💡 It would help if `AppInfo` model or something is passed on and preview can pass in these strings and the top level composable can pass in the names from context.

t-p-white:
Thanks for trying out the suggestions @MozillaNoah. Appreciate the 'recent push' summaries that you have been providing for PRs on this project 🙏

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawerLandingScreen.kt
MozillaNoah: good catch

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/ui/DebugDrawer.kt
MozillaNoah: I thought about this, but it kinda just pushes the problem upstream, but maybe that's OK. I'll see how it looks

MozillaNoah: I'll make a note to check this out after the [state hoisting PR](https://github.com/mozilla-mobile/firefox-android/pull/5011) lands.

MozillaNoah:
Recent push contains:
- Changing `LandingScreen` to `Home`
- Adding the key/id to the lazy column items

t-p-white (approved):

PR #5029, opened 2024-01-04
Bug 1860472 - Periodically clean the Uploads cacheDir.

Amejia481:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1860472

### DIFF
--- a/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
+++ b/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
@@ -25,6 +25,7 @@ import mozilla.components.support.base.feature.OnNeedToRequestPermissions
 import mozilla.components.support.base.feature.PermissionsFeature
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.ktx.android.content.isPermissionGranted
+import mozilla.components.support.ktx.android.net.getFileName
 import mozilla.components.support.ktx.android.net.isUnderPrivateAppDirectory
 import mozilla.components.support.utils.ext.getParcelableExtraCompat
 
@@ -40,6 +41,7 @@ internal var captureUri: Uri? = null
 /**
  * @property container The [Activity] or [Fragment] which hosts the file picker.
  * @property store The [BrowserStore] this feature should subscribe to.
+ * @property fileUploadsDirCleaner a [FileUploadsDirCleaner] to clean up temporary file uploads.
  * @property onNeedToRequestPermissions a callback invoked when permissions
  * need to be requested before a prompt (e.g. a file picker) can be displayed.
  * Once the request is completed, [onPermissionsResult] needs to be invoked.
@@ -48,6 +50,7 @@ internal class FilePicker(
     private val container: PromptContainer,
     private val store: BrowserStore,
     private var sessionId: String? = null,
+    private var fileUploadsDirCleaner: FileUploadsDirCleaner,
     override val onNeedToRequestPermissions: OnNeedToRequestPermissions,
 ) : PermissionsFeature {
 
@@ -72,7 +75,11 @@ internal class FilePicker(
             val hasPermission = container.context.isPermissionGranted(type.permission)
             // The captureMode attribute can be used if the accepted types are exactly for
             // image/*, video/*, or audio/*.
-            if (hasPermission && type.shouldCapture(promptRequest.mimeTypes, promptRequest.captureMode)) {
+            if (hasPermission && type.shouldCapture(
+                    promptRequest.mimeTypes,
+                    promptRequest.captureMode,
+                )
+            ) {
                 type.buildIntent(container.context, promptRequest)?.also {
                     saveCaptureUriIfPresent(it)
                     container.startActivityForResult(it, FILE_PICKER_ACTIVITY_REQUEST_CODE)
@@ -135,8 +142,7 @@ internal class FilePicker(
     }
 
     private fun getActivePromptRequest(): PromptRequest? =
-        store.state.findCustomTabOrSelectedTab(sessionId)?.content?.promptRequests?.lastOrNull {
-                prompt ->
+        store.state.findCustomTabOrSelectedTab(sessionId)?.content?.promptRequests?.lastOrNull { prompt ->
             prompt is File
         }
 
@@ -194,6 +200,9 @@ internal class FilePicker(
                 if (sanitizedUris.isEmpty()) {
                     request.onDismiss()
                 } else {
+                    sanitizedUris.map {
+                        enqueueForCleanup(container.context, it)
+                    }
                     request.onMultipleFilesSelected(container.context, sanitizedUris)
                 }
             }
@@ -202,6 +211,7 @@ internal class FilePicker(
             uri?.let {
                 // We want to verify that we are not exposing any private data
                 if (!it.isUnderPrivateAppDirectory(container.context)) {
+                    enqueueForCleanup(container.context, it)
                     request.onSingleFileSelected(container.context, it)
                 } else {
                     request.onDismiss()
@@ -219,6 +229,12 @@ internal class FilePicker(
     fun askAndroidPermissionsForRequest(permissions: Set<String>, request: File) {
         currentRequest = request
         onNeedToRequestPermissions(permissions.toTypedArray())
+    }
+
+    private fun enqueueForCleanup(context: Context, uri: Uri) {
+        val contentResolver = context.contentResolver
+        val fileName = uri.getFileName(contentResolver)
+        fileUploadsDirCleaner.enqueueForCleanup(fileName)
     }
 
     companion object {
--- a/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
+++ b/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
@@ -1,0 +1,86 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.prompts.file
+
+import androidx.annotation.VisibleForTesting
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.DelicateCoroutinesApi
+import kotlinx.coroutines.Dispatchers.IO
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import mozilla.components.concept.engine.prompt.PromptRequest.File.Companion.DEFAULT_UPLOADS_DIR_NAME
+import mozilla.components.support.base.log.logger.Logger
+import java.io.File
+import java.io.IOException
+
+/**
+ * A storage implementation for organizing temporal uploads metadata to be clean up.
+ */
+@OptIn(DelicateCoroutinesApi::class)
+class FileUploadsDirCleaner(
+    private val scope: CoroutineScope = GlobalScope,
+    private val cacheDirectory: () -> File,
+) {
+    private val logger = Logger("FileUploadsDirCleaner")
+
+    private val cacheDir by lazy { cacheDirectory() }
+
+    @VisibleForTesting
+    internal var fileNamesToBeDeleted: List<String> = emptyList()
+
+    @VisibleForTesting
+    internal var dispatcher: CoroutineDispatcher = IO
+
+    /**
+     * Enqueue the [fileName] for future clean up.
+     */
+    internal fun enqueueForCleanup(fileName: String) {
+        fileNamesToBeDeleted += (fileName)
+        logger.info("File $fileName added to the upload cleaning queue.")
+    }
+
+    /**
+     * Remove all the temporary file uploads.
+     */
+    internal fun cleanRecentUploads() {
+        scope.launch(dispatcher) {
+            val cacheUploadDirectory = File(getCacheDir(), DEFAULT_UPLOADS_DIR_NAME)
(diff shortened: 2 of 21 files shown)


jonalmeida:
Can you add context into this solution's design? Reading the bug, there is a different conversation there as compared to what is happening here.

What catches my eye in particular is that we have GV that copies and uploads files and another system that would periodically delete the upload directory without informing or using GV (i.e. two systems working on the same files independently). What happens if you leave your device to upload large files and put it into an idle state, we would delete be deleting the upload files while they are being used and cause data corruption.

Can you clarify why you abandoned the solution in [comment 6](https://bugzilla.mozilla.org/show_bug.cgi?id=1860472#c6) which might be safer, albeit I haven't considered jackyzy's solution at this time.

Amejia481:
> Can you add context into this solution's design? Reading the bug, there is a different conversation there as compared to what is happening here.
> 
> What catches my eye in particular is that we have GV that copies and uploads files and another system that would periodically delete the upload directory without informing or using GV (i.e. two systems working on the same files independently). What happens if you leave your device to upload large files and put it into an idle state, we would delete be deleting the upload files while they are being used and cause data corruption.
> 
> Can you clarify why you abandoned the solution in [comment 6](https://bugzilla.mozilla.org/show_bug.cgi?id=1860472#c6) which might be safer, albeit I haven't considered jackyzy's solution at this time.

Thank you so much for taking a look :) 
Sorry my bad, I should added more context around the ticket. I was chatting with @rahulsainani and others in the last "Android Engineers Sync/Show-and-Tell", I've doing some [research](https://matrix.to/#/!hDHZJZOdaLkmRzgxyB:mozilla.org/$I4HUeGWxx6ApJ-HAzBpqLyulzzObvnd2-fkEObXXXUg?via=mozilla.org&via=igalia.com&via=matrix.org), unfortunately I haven't found many hints, the goal of this patch is to reduce the pain of the idle space used for uploads, while we continue researching for final solution. 

Some of the findings:

- We have partial support for PDF [content type URIs](https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/ContentInputStream.java#42)

- The whole interface [nsIFilePicker.idl](https://searchfox.org/mozilla-central/source/widget/nsIFilePicker.idl) is based on `nsIFile` and `nsIURI`, which means the we have to provide an [nsIFile](https://searchfox.org/mozilla-central/source/mobile/android/components/geckoview/FilePickerDelegate.sys.mjs#80) as part of the contract.

Amejia481:
> What catches my eye in particular is that we have GV that copies and uploads files and another system that would periodically delete the upload directory without informing or using GV (i.e. two systems working on the same files independently). What happens if you leave your device to upload large files and put it into an idle state, we would delete be deleting the upload files while they are being used and cause data corruption.

@rahulsainani and I were chatting about possible ways to avoid issues, by only deleting files that are a day older than the actual day. What do you think about this approach ?

Amejia481:
@jonalmeida feel free to bring any new ideas :)

jonalmeida:
@Amejia481 sorry, I'm not sure I follow: why does the solution in [comment 6](https://bugzilla.mozilla.org/show_bug.cgi?id=1860472#c6) not work? That is, why are we no longer trying to create a GV API that will have full control and understanding of the inprogress state of uploads when we decide to clear it?

Amejia481:
> @Amejia481 sorry, I'm not sure I follow: why does the solution in [comment 6](https://bugzilla.mozilla.org/show_bug.cgi?id=1860472#c6) not work? That is, why are we no longer trying to create a GV API that will have full control and understanding of the inprogress state of uploads when we decide to clear it?

We are going to continue moving forward with [comment 6](https://bugzilla.mozilla.org/show_bug.cgi?id=1860472#c6), but in the meantime we would like to put a work around for clearing up the space for users :)

Amejia481:
@jonalmeida I added updated commit with the suggestions from https://matrix.to/#/!hDHZJZOdaLkmRzgxyB:mozilla.org/$AeS2h2QRtlnIAY9JEVFYoJkPR1zArA_UqHTzTCCRXqc?via=mozilla.org&via=igalia.com&via=matrix.org.

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/GlobalUploadsDependencyProvider.kt
```diff
@@ -0,0 +1,32 @@
...
+package mozilla.components.feature.prompts.file
+
+import androidx.annotation.VisibleForTesting
+import mozilla.components.browser.state.store.BrowserStore
+
+/**
+ * Provides global access to the dependencies needed for purging temporal uploaded files.
+ */
+/** @suppress */
+object GlobalUploadsDependencyProvider {
```
rahulsainani, line 14: I think we can remove this and directly inject `browserStore` instance in `FileUploadsDirCleaner`

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
```diff
@@ -0,0 +1,146 @@
...
+/**
+ * A implementation which uses WorkManager APIs to periodically clean the temporal uploads directory.
+ */
+internal class FileUploadsDirCleanerWorker(
+    val context: Context,
+    params: WorkerParameters,
+) : CoroutineWorker(context, params) {
+    private val logger = Logger("FileUploadsDirCleanerWorker")
+
+    @Suppress("TooGenericExceptionCaught")
```
rahulsainani, line 122: Do we need this?

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
```diff
@@ -0,0 +1,146 @@
...
+    fun unregister()
+}
+
+/**
+ * An implementation of [FileUploadsDirCleaner] that uses the work manager api for scheduling checks.
+ * @property applicationContext The application context.
+ * @param frequency (Optional) indicates how often checks should be performed, defaults
+ * to once per day.
+ */
+@Suppress("LargeClass")
```
rahulsainani, line 46: Do we need this?

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
```diff
@@ -0,0 +1,146 @@
...
+    @VisibleForTesting
+    internal fun getWorkerConstrains(): Constraints {
+        var builder =
+            Constraints.Builder()
+                .setRequiresBatteryNotLow(true)
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            builder = builder.setRequiresDeviceIdle(true)
+        }
+        return builder.build()
```
rahulsainani, line 94:
```suggestion
        return Constraints.Builder().apply {
            setRequiresBatteryNotLow(true)
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
                setRequiresDeviceIdle(true)
            }
        }.build()
```

jonalmeida:
> @jonalmeida I added updated commit with the suggestions from https://matrix.to/#/!hDHZJZOdaLkmRzgxyB:mozilla.org/$AeS2h2QRtlnIAY9JEVFYoJkPR1zArA_UqHTzTCCRXqc?via=mozilla.org&via=igalia.com&via=matrix.org.

To summarize this thread since it's not as easily accessible to share threads outside of Matrix today: mconley suggested that we consider clearing the uploads when the page is unloaded (e.g. if you navigate away from a site, the uploaded cache is no longer needed).

---

Given this information, which I think is a great idea, we already know when we navigate away from a page changes with some combination of `UpdateUrlAction`, `UpdateLoadRequestAction`, and other actions. 

With this, couldn't we just clear the upload cache when that happens? I think this would simplify the process by not needing a work task or a full sqlite database to manage the uploads. Maybe a lighter version of those at least. What do you both think @Amejia481 @rahulsainani ?

Amejia481:
> `UpdateUrlAction`, `UpdateLoadRequestAction`

The idea sounds really straightforward, but it won't be a bit over kill, cleaning the directory every time we have a`UpdateUrlAction` or `UpdateLoadRequestAction`, as most of the times we won't have anything on the `uploads` directory?

jonalmeida:
Yes, as mentioned, some combination of those actions and your concept to use metadata would be good (can we just keep an in-memory list of the page changes + we can perform a one-time clear on a cold app start?). I think we need to always perform a clean for every one of those actions if we know there is nothing there to clean.

Amejia481:
> keep an in-memory list of the page changes + we can perform a one-time clear on a cold app start

Sounds good to me 👍🏽

rahulsainani:
@jonalmeida @Amejia481 Some combination of those events sound interesting, if that can be figured out, I think we can get rid of the database and hold that information in a memory cache. If the conditions meet, we should schedule a job for clearing uploads or delete it immediately. I'm concerned adding this on startup will impact the startup time, Since this is not needed for app startup, we shouldn't add it there. Happy to discuss it offline also!

Amejia481:
> I'm concerned adding this on startup will impact the startup time, Since this is not needed for app startup, we shouldn't add it there. Happy to discuss it offline also!

As the task runs in [background](https://github.com/mozilla-mobile/firefox-android/pull/5029/files#diff-7004b68195871c93ff618abf02b4c0ecb920101510810d45e2ee97ece50ee747R123) thread, we should be safe, what do you think? @rahulsainani @jonalmeida

Amejia481:
We agreed offline to go with solution similar to prior art clearing out storage components on startup with the [ThumbnailStorage](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/thumbnails/src/main/java/mozilla/components/browser/thumbnails/storage/ThumbnailStorage.kt#52), doing this on the IO dispatcher by default and also add this task to the [visualCompletenessQueue](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/components/PerformanceComponent.kt#16), already found a good combination when we should clean up, there is no need for `WorkManager` task.

Amejia481:
@jonalmeida @rahulsainani I updated the patch (Test are pending until confirming the approach)

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * A storage implementation for organizing temporal uploads metadata.
+ */
+object FileUploadsDirCleaner {
+
+    @VisibleForTesting
+    internal var fileNamesToBeDeleted: List<String> = emptyList()
+    private val scope = CoroutineScope(Dispatchers.IO)
+    val logger = Logger("UploadMetadataStorage")
```
rahulsainani, line 24: Could be private

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
```diff
@@ -0,0 +1,70 @@
...
+     */
+    fun enqueueForCleanup(fileName: String) {
+        fileNamesToBeDeleted += (fileName)
+        logger.info("File $fileName added to the upload cleaning queue.")
+    }
+
+    /**
+     * Remove all the temporary file uploads.
+     */
+    fun cleanRecentUploads(context: Context) {
```
rahulsainani, line 37: I think this and `cleanUploadsDirectory` could be suspend functions, removing the need for `CoroutineScope` as we are not cancelling it here. We can using withContext(Dispatchers.IO) to make sure the caller doesn't need to be explicit about the dispatcher.

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FilePicker.kt
```diff
@@ -221,6 +226,12 @@ internal class FilePicker(
         onNeedToRequestPermissions(permissions.toTypedArray())
     }
 
+    internal fun enqueueForCleanup(context:Context, uri: Uri){
```
rahulsainani, line 229: Could this be private?

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @VisibleForTesting
+    internal var fileNamesToBeDeleted: List<String> = emptyList()
+    private val scope = CoroutineScope(Dispatchers.IO)
+    val logger = Logger("UploadMetadataStorage")
+
+    /**
+     * Enqueue the [fileName] for future clean up.
+     */
+    fun enqueueForCleanup(fileName: String) {
```
rahulsainani, line 29: This and `cleanRecentUploads` seem to be only use in this module so maybe they can be internal

android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/file/FileUploadsDirCleaner.kt
```diff
@@ -0,0 +1,70 @@
...
+            }
+        }
+    }
+
+    /**
+     * Remove the file uploads directory if exists.
+     */
+    fun cleanUploadsDirectory(context: Context) {
+        scope.launch {
+            val cacheUploadDirectory = File(context.cacheDir, "/uploads")
```
rahulsainani, line 58: Maybe we can extract the directory names as private constants

jonalmeida (commented):
Leaving some general feedback here. The direction looks good, but I'll differ to Rahul's review for future iterations.

Thanks!

rahulsainani (approved):
Overall it's looking good! 

Just summarizing the offline chat to make `cleanUploadsDirectory` a suspend function and remove the dependency on `Context`. Understand the complexity with app lifecycle scope so let's think about removing `GlobalScope` if possible as an iteration. Thanks for making the changes and trying different approaches! 🙌

PR #5030, opened 2024-01-04
Bug 1864075 - Create navigation middleware for the debug drawer

MozillaNoah:
Here's what I'm envisioning for the navigation architecture (amongst this and other tickets).

## Constraints
- `DebugDrawer` needs to be ignorant of our Store/State/Middleware pattern.
- We (ideally) need to use the Compose navigation library to handle the screen transitions + back stack management.
- We need a data type wrapper for the navigation destinations to be usable in or out of Fenix.
- The debug drawer screens/destinations need to be constructed at runtime because of their dependencies (Stores, Settings, etc.)

## Plan of attack
- [Utilize the Compose navigation library](https://developer.android.com/jetpack/compose/navigation)
    - The `DebugDrawer` composable will host the `NavHost` composable
    - The navigation middleware will make calls to the navController, which will in turn update the `NavHost`
- A `Toolbar` composable will be added to `DebugDrawer` to handle the screen titles and back navigation.
- A data class `DebugDrawerDestination` will serve as the data model to encapsulate:
    - The route used by the nav controller via the navigation library
    - The title of the destination (used for the toolbar as well as the feature list inside of `DebugDrawer`)
    - The Compose content block, which is rendered when it's on-screen.
- A helper function will generate all of these destinations for Fenix (`generateDebugDrawerDestinations`)
- Create an `action` for every destination (keeping everything 1 to 1) as well as for navigating back
- The `NavHost` will have its nav graph constructed at runtime from a collection of `DebugDrawerDestination` passed-in to `DebugDrawer`

## Follow-ups
- The UI sync-up to this logic will happen in [Bug 1864076](https://bugzilla.mozilla.org/show_bug.cgi?id=1864076) 
- The Toolbar for back navigation will be added as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1864078.
- The actual tab tools feature will be added in https://bugzilla.mozilla.org/show_bug.cgi?id=1871098

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1864075


MozillaNoah:
Added some unit tests

fenix/app/src/test/java/org/mozilla/fenix/debugsettings/DebugDrawerNavigationMiddlewareTest.kt
```diff
@@ -0,0 +1,60 @@
+package org.mozilla.fenix.debugsettings
```
gabrielluong, line 1: License please

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerStore.kt
```diff
@@ -24,5 +24,6 @@ private fun reduce(state: DebugDrawerState, action: DebugDrawerAction): DebugDra
     return when (action) {
         is DebugDrawerAction.DrawerOpened -> state.copy(drawerStatus = DebugDrawerState.DrawerStatus.Open)
         is DebugDrawerAction.DrawerClosed -> state.copy(drawerStatus = DebugDrawerState.DrawerStatus.Closed)
+        is DebugDrawerAction.NavigateTo, DebugDrawerAction.OnBackPressed -> state // handled by middleware
```
gabrielluong, line 27:
```suggestion
        is DebugDrawerAction.NavigateTo, DebugDrawerAction.OnBackPressed -> state // handled by [DebugDrawerNavigationMiddleware]
```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt
```diff
@@ -0,0 +1,40 @@
...
+) : Middleware<DebugDrawerState, DebugDrawerAction> {
+
+    override fun invoke(
+        context: MiddlewareContext<DebugDrawerState, DebugDrawerAction>,
+        next: (DebugDrawerAction) -> Unit,
+        action: DebugDrawerAction,
+    ) {
+        when (action) {
+            is DebugDrawerAction.NavigateTo.LandingScreen -> {
+                navController.navigate(route = DEBUG_DRAWER_ROUTE_LANDING_SCREEN)
```
gabrielluong, line 29: Please excuse my complete lack of understanding here. I am wondering why we wanto to do navigation via middleware compare to the Compose Navigation examples of defining a `NavHost` and just navigating directly.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/FenixDebugDrawerDestinations.kt
```diff
@@ -0,0 +1,35 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
t-p-white, line 1: nit: Can we drop the `Fenix` prefix? Perhaps this could also be called `DebugDrawerNavigation`

t-p-white: If it becomes `DebugDrawerNavigation` maybe `DebugDrawerDestination` & this file can be consolidated

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/FenixDebugDrawerDestinations.kt
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * The navigation route for [TabTools].
+ */
+const val DEBUG_DRAWER_ROUTE_TAB_TOOLS = "tab_tools"
+
+/**
+ * Creates a list of [DebugDrawerDestination]s for the debug drawer features in Fenix.
+ */
+fun generateDebugDrawerDestinations(
+    // Composable content dependencies will go here (e.g. browser store, the list of menu items, etc.)
```
t-p-white, line 23: could add a todo here

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt
MozillaNoah: Kinda similar reasons why we would have, historically, done the navigation from within Interactors/Controllers.
- Keep the navigation calls outside of the UI (the nav controller will still be connected to a `NavHost` in follow-up work noted above)
- Have something that's testable

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/FenixDebugDrawerDestinations.kt
```diff
@@ -0,0 +1,35 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.debugsettings.navigation
+
+import org.mozilla.fenix.R
+
+/**
+ * The navigation route for [DebugDrawerLandingScreen].
+ */
+const val DEBUG_DRAWER_ROUTE_LANDING_SCREEN = "landing_screen"
```
t-p-white, line 12: is the 'landing screen' verbiage already used? I wonder if something like 'main menu' or 'drawer home' would be more consistent with the current naming?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerAction.kt
```diff
@@ -20,4 +20,25 @@ sealed class DebugDrawerAction : Action {
...
+
+    /**
+     * [DebugDrawerAction] fired when a navigation event occurs for a specific destination.
+     */
+    sealed class NavigateTo : DebugDrawerAction() {
+
+        /**
+         * [NavigateTo] action fired when the debug drawer needs to navigate to the landing screen.
+         */
+        object LandingScreen : NavigateTo()
```
t-p-white, line 32: as mentioned in previous comment

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/FenixDebugDrawerDestinations.kt
MozillaNoah: I initially opted to not do this, since I think we've been historically hesitant to committing todos. Regardless, this is temporary and was more of an explainer for peopl reviewing it.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/FenixDebugDrawerDestinations.kt
MozillaNoah: It's not set in stone; I do have [another PR using it](https://github.com/mozilla-mobile/firefox-android/pull/5012), but I definitely wanted to get yalls thoughts on what to name this screen. The name though will only be visible in the code (nothing user facing) so:
- The composable name
- Route 
- Random variable naming

MozillaNoah: so maybe:
- DebugDrawerHome for the composable
- "home" for the route

MozillaNoah:
Recent push contains:
- Change from `Fenix` to `DebugDrawerNavigation`
- Change from `...LandingScreen` to `Home` (this will be changed in the other PR as well
- The missing license on the test

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt
```diff
@@ -0,0 +1,40 @@
...
+                navController.navigate(route = DEBUG_DRAWER_ROUTE_HOME)
+            }
+            is DebugDrawerAction.NavigateTo.TabTools -> {
+                navController.navigate(route = DEBUG_DRAWER_ROUTE_TAB_TOOLS)
+            }
+            is DebugDrawerAction.OnBackPressed -> {
+                navController.popBackStack()
+            }
+            else -> {} // no-op
+        }
```
t-p-white, line 38: nit:
```suggestion
        when (action) {
            is DebugDrawerAction.NavigateTo.Home -> navController.navigate(route = DEBUG_DRAWER_ROUTE_HOME)
            is DebugDrawerAction.NavigateTo.TabTools -> navController.navigate(route = DEBUG_DRAWER_ROUTE_TAB_TOOLS)
            is DebugDrawerAction.OnBackPressed -> navController.popBackStack()
            else -> {} // no-op
        }
```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerStore.kt
```diff
@@ -24,5 +24,8 @@ private fun reduce(state: DebugDrawerState, action: DebugDrawerAction): DebugDra
     return when (action) {
         is DebugDrawerAction.DrawerOpened -> state.copy(drawerStatus = DebugDrawerState.DrawerStatus.Open)
         is DebugDrawerAction.DrawerClosed -> state.copy(drawerStatus = DebugDrawerState.DrawerStatus.Closed)
+        is DebugDrawerAction.NavigateTo,
+        DebugDrawerAction.OnBackPressed,
+        -> state // handled by [DebugDrawerNavigationMiddleware]
```
t-p-white, line 29: nit:
```suggestion
        is DebugDrawerAction.NavigateTo, DebugDrawerAction.OnBackPressed -> state // handled by [DebugDrawerNavigationMiddleware]

```

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerDestination.kt
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * A navigation destination for sub-features within the Debug Drawer.
+ *
+ * @property route The navigation route of the destination.
+ * @property labelId The string ID of the destination's title.
+ * @property content The destination's [Composable].
+ */
+data class DebugDrawerDestination(
+    val route: String,
+    @StringRes val labelId: Int,
```
t-p-white, line 19:
```suggestion
    @StringRes val title: Int,
```
?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/FenixDebugDrawerDestinations.kt
t-p-white: that's fair, the only reason I'd prefer a todo is that it makes it harder for the comment to get forgotten, but this is not a blocker

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerNavigation.kt
```diff
@@ -0,0 +1,36 @@
...
+
+package org.mozilla.fenix.debugsettings.navigation
+
+import org.mozilla.fenix.R
+import org.mozilla.fenix.debugsettings.tabs.TabTools
+
+/**
+ * The navigation route for [DebugDrawerHome].
+ */
+const val DEBUG_DRAWER_ROUTE_HOME = "home"
```
t-p-white, line 13: nit: I think we can also drop the `DEBUG_DRAWER...` as they are imported from `DebugDrawerNavigation`

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/FenixDebugDrawerDestinations.kt
t-p-white: >  DebugDrawerHome for the composable
    "home" for the route
    
Yeah I think this is good. I'm flip-flopping from 'home' to 'main' to 'parent' but don't have a strong preference for either

t-p-white (approved):
Thanks for the updates @MozillaNoah, this looks great. Just nits from me (all optional suggestions)

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt
```diff
@@ -0,0 +1,40 @@
...
+            is DebugDrawerAction.NavigateTo.Home -> {
+                navController.navigate(route = DEBUG_DRAWER_ROUTE_HOME)
+            }
+            is DebugDrawerAction.NavigateTo.TabTools -> {
+                navController.navigate(route = DEBUG_DRAWER_ROUTE_TAB_TOOLS)
+            }
+            is DebugDrawerAction.OnBackPressed -> {
+                navController.popBackStack()
+            }
+            else -> {} // no-op
```
MatthewTighe, line 37: @boek and I were talking about this, and using `else` blocks in middlewares is something that we should probably avoid moving forward. Instead, we should just include the full list of no-op actions. This allows us to take full advantage of the compile-time exhaustive checking we get from using sealed classes for the action hierarchies. This adds clarity and discoverability for future development because it requires any new actions to be explicitly handled everywhere they might need to be used .

MatthewTighe (commented):
just a quick 🚗 since it looks like this is well-covered by other reviewers

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt
MozillaNoah: That's fair and it's admittedly what I did in the Store, so I'll follow suit here as well 👍

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerNavigation.kt
MozillaNoah: Since they're outside of a class, they would then show up in the global namespace as `ROUTE_HOME`. Obviously, this wouldn't be the case if we are able to long term modularize everything.

ALTERNATIVELY, we could do something like:
```
enum class DebugDrawerRoute(val route: String) {
    Home("home"),
    TabTools("tab_tools"),
}
```

and then we could have the string ID included here, too.

I'll try this with the next push

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt
gabrielluong: Makes sense. My initial impression is that it seems unnecessarily complicated.
Some followup questions:

1. I am wondering if we really have all that much test coverage around our Compose code for the lambda/callback functions. I understand the idea is to ideally just dispatch actions in these callback and handle this all in our middlewares. It makes me wonder if navigation is really something we care about having test coverage (if we did have good test coverage for our composables).
2. Do we also have examples of this being done elsewhere already?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerStore.kt
MozillaNoah: I had to do this because of lint, but I'll try to find something cleaner

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt
MozillaNoah: hope that all makes sense. this is definitely the most technically & architecturally challenging part of this feature.

MozillaNoah:
Recent push contains:
- Some when statement clean up (and the suggestion from Matt to make sure we lean on the exhaustiveness check with sealed classes)
- `DebugDrawerRoute` is a new enum (to be used only in Fenix) which will contain our debug drawer routes/destinations and the display string IDs. This will now be easily transformed into a list of navigation destinations for the debug drawer API, and by using `DebugDrawerRoute.values().map{}` with an `when` it will now enforce that every new `DebugDrawerRoute` enum added in future must add an entry to the `debugDrawerDestinationsFromRoutes()` function.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt
gabrielluong: Thanks for taking the time for these responses, I do really appreciate. I am just not sure if I am still suffering from smooth brain or just being really far detached from the code since I have been doing for monorepo for so long.

MozillaNoah: sure thing [ReviewQualityCheckNavigationMiddleware](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/shopping/middleware/ReviewQualityCheckNavigationMiddleware.kt)

It uses a `UseCase` in this example.

I'm going to do a deep dive on all of this soon(tm), but I'm always happy to do a call sometime to catch up on the current state/future state with ya.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerDestination.kt
```diff
@@ -0,0 +1,21 @@
...
+import androidx.compose.runtime.Composable
+
+/**
+ * A navigation destination for screens within the Debug Drawer.
+ *
+ * @property route The navigation route of the destination.
+ * @property title The string ID of the destination's title.
+ * @property content The destination's [Composable].
+ */
+data class DebugDrawerDestination(
```
t-p-white, line 17: With the introduction of the enum can this just take the `DebugDrawerRoute` to get these values? In which case is this data class still needed, could the enum replace this completely?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerNavigation.kt
t-p-white: Nice, this works well. I'm wondering if it can replace `DebugDrawerDestination` althogther?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerDestination.kt
```diff
@@ -0,0 +1,21 @@
...
+    val route: String,
```
t-p-white, line 18: can this just take a `DebugDrawerRoute` param now?

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerRoute.kt
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * The navigation routes for screens within the Debug Drawer.
+ *
+ * @property route The navigation route of the destination.
+ * @property title The string ID of the destination's title.
+ */
+enum class DebugDrawerRoute(val route: String, @StringRes val title: Int) {
+    /**
+     * The navigation route for [DebugDrawerHome].
```
t-p-white, line 19:
```suggestion
     * The navigation route for [Home].
```

t-p-white (commented):
With the introduction of the `DebugDrawerRoute` enum, can that completely replace the `DebugDrawerDestination` data class? e.g.

```kotlin
/**
 * The navigation routes for screens within the Debug Drawer.
 *
 * @property route The navigation route of the destination.
 * @property title The string ID of the destination's title.
 */
enum class DebugDrawerRoute(val route: String, @StringRes val title: Int, val content: @Composable () -> Unit,) {
    /**
     * The navigation route for [Home].
     */
    Home(
        route = "home",
        title = R.string.debug_drawer_title,
        {},
    ),

    /**
     * The navigation route for [TabTools].
     */
    TabTools(
        route = "tab_tools",
        title = R.string.debug_drawer_tab_tools_title,
        {}
    ),
}
```
which would also remove the `debugDrawerDestinationsFromRoutes` func

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerRoute.kt
MozillaNoah: This is intentionally `DebugDrawerHome`, so we can CMD+click into the source code (I think I just need to add the import for it).

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerRoute.kt
```diff
@@ -0,0 +1,51 @@
...
+package org.mozilla.fenix.debugsettings.navigation
+
+import androidx.annotation.StringRes
+import org.mozilla.fenix.R
+import org.mozilla.fenix.debugsettings.tabs.TabTools
+
+/**
+ * The navigation routes for screens within the Debug Drawer.
+ *
+ * @property route The navigation route of the destination.
```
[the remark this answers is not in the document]
MozillaNoah: In our direct use case, it is. The navigation library can technically take a URI as its navigation input, `route`. The route can be just a simple string, but it can also have a data model in it. It can also support deep linking. So I'm calling it `route` to keep the naming in sync with the library, but also not confuse it any of our data model abstractions.

MozillaNoah:
> With the introduction of the `DebugDrawerRoute` enum, can that completely replace the `DebugDrawerDestination` data class? e.g.
> 
> which would also remove the `debugDrawerDestinationsFromRoutes` func

Good question. So it ultimately comes down to the separation of Fenix stuff from the debug drawer specific stuff, if there ever comes a time we upstream this component to AC.

`DebugDrawerDestination` is the overall data class that will be the cornerstone of the debug drawer API, and it's intentionally isolated from Fenix and its architecture.

The `DebugDrawerRoute` enum serves as a descriptive interstitial abstraction for capturing:
- The routing in Fenix
- The strings
- The ordering of the items inside the home screen table of contents

`debugDrawerDestinationsFromRoutes`, as you point out, acts as the transformation from the `DebugDrawerRoute` enum to a list of `DebugDrawerDestination`. The crucial thing inside this function though is the stand-up of the composables. All of the composables for the screens/destinations have some sort of dependency that lives outside of the debug drawer and also is something only derived at runtime, not compile time. We unfortunately cannot define enums with the Compose blocks at compile time because of this. 

So to solve this, `debugDrawerDestinationsFromRoutes` will be called within `FenixOverlay` during its first set up to generate the list of `DebugDrawerDestination` to send to the debug drawer code. `FenixOverlay` will be responsible for gathering all of the runtime dependencies of the sub-screens (e.g. `BrowserStore` and a the inactive tabs settings for `TabTools`).

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerRoute.kt
MozillaNoah: We could always typealias it, if we wanted

MozillaNoah:
Hey @t-p-white, I'm going to land this today to keep development moving forward, but happy to keep the discussion going on things. Some of it will make more sense in the next two PRs.

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerRoute.kt
t-p-white: Gotcha, I think the import is fine 🙂

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerRoute.kt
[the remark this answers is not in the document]
t-p-white: 👍 
```
The navigation route of the destination.
``` 
Can we do something to clarify what a `route` is, perhaps including what you've mentioned here around the URI?

t-p-white:
Sure thing @MozillaNoah, thanks for the comment updates 🙂

fenix/app/src/main/java/org/mozilla/fenix/debugsettings/navigation/DebugDrawerRoute.kt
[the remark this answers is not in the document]
MozillaNoah: Sure thing! I'll look into updating the docs with `route` in the plumbing PR.

PR #5043, opened 2024-01-05
Bug 1873260 - Move Behavior classes to ui-widget component

mavduevskiy:
### Why?
The goal is to reuse `CoordinatorLayout.Behavior` classes that enable scrolling for navigation toolbar.
### How?
Currently, the implementation is tied to the `BrowserToolbar` class though it never actually uses any of its methods and treats it just as a `View`. We could easily reuse the Behaviour for the `NavigationToolbar`.

Also, `ToolbarBehaviorController` was dependant on `Toolbar` interface, which – in addition to scrolling behaviour methods – has a lot of toolbar methods. I extracted scrolling behaviour part of the interface into `ScrollableToolbar` and made `ToolbarBehaviorController` use the new lighter version instead.

### What else is going on?
`Behavior` classes are moved into the `concept-engine`, since those stopped depending on anything `Toolbar` related, but are functionality built around `EngineView`. The renaming to adjust to the new implementation and location.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1873260


mavduevskiy:
@titooan
The implementation ended up being slightly different from what you guys initially suggested. I moved behaviour classes into the `ui:widget` component, not `feature:session`. Moving it to `feature:session` would brake the components [hierarchy](https://mozac.org/contributing/architecture), making the `browser:toolbar` component depend on the `feature:session` (features should be the highest level of abstraction). I also moved behaviour class from session lower into `ui:widget` layer, as was advised by @jonalmeida in our preliminary conversations.

titooan:
Looks good to me! 

I would just update the commit name for something a bit more specific, like "Move Behavior classes to ui-widget component" for example, what do you think?

mavduevskiy:
@titooan 
Updated! yep, I agree, mine was a very generic one

titooan (approved):

PR #5046, opened 2024-01-05
Bug 1873280 - Update Firebase Cloud Messaging to version 23.4.0

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1873280

### DIFF
--- a/android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
+++ b/android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
@@ -91,7 +91,7 @@ object Versions {
     }
 
     object Firebase {
-        const val messaging = "23.3.1"
+        const val messaging = "23.4.0"
     }
 }
 


jonalmeida:
> The SDK now calls messageHandled() after a message has been handled successfully.

I'm not sure where `messageHandled` is yet, at least in the existing SDK.

jonalmeida (approved):
> > The SDK now calls messageHandled() after a message has been handled successfully.
> 
> I'm not sure where `messageHandled` is yet, at least in the existing SDK.

Yup, it's a new method [deep in it's RPC API](https://github.com/firebase/firebase-android-sdk/commit/879c84bb4ae02b9b6923d3ae0620c97c63bae10d), but doesn't say what a consumer should do about this.

I'm not too worried about it, but let's land this as request QA verification for Send Tab and WebPush as a preventative measure. What do you think?

rvandermeulen:
Will do!

PR #5048, opened 2024-01-05
Bug 1870335 - Merge content descriptions in the Add-ons details screen.

Amejia481:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1870335


willdurand:
My understanding of the bug is that we should make each row in the "table" below the add-on's description announceable. While this PR updates the labels in the first column, it doesn't allow users with Talkback to navigate between each row more easily. I suppose something like this would be more like what Anna described:

```diff
diff --git a/fenix/app/src/main/res/layout/fragment_add_on_details.xml b/fenix/app/src/main/res/layout/fragment_add_on_details.xml
index 61bff2fcde..b0e34ba78b 100644
--- a/fenix/app/src/main/res/layout/fragment_add_on_details.xml
+++ b/fenix/app/src/main/res/layout/fragment_add_on_details.xml
@@ -28,6 +28,7 @@
             tools:text="@tools:sample/lorem/random" />
 
         <TextView
+            android:focusable="true"
             android:id="@+id/author_label"
             style="@style/AboutHeaderContentText"
             android:layout_width="wrap_content"
@@ -49,7 +50,6 @@
             app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toEndOf="@+id/author_label"
             app:layout_constraintTop_toBottomOf="@+id/details"
-            android:importantForAccessibility="no"
             tools:text="@tools:sample/full_names" />
 
         <View
```

Also, I don't think links should be marked as "no important for a11y".

Also, also, I think we should check with Anna what would work best here. E.g., should the entire row be a control? Some rows have "read-only" information but some others have a label and a link, and the "Homepage" only has a link. We should determine how we want users to navigate between each of these.

willdurand:
> should the entire row be a control?

From: https://bugzilla.mozilla.org/show_bug.cgi?id=1870351

> If the recommendation in the [bug 1870335](https://bugzilla.mozilla.org/show_bug.cgi?id=1870335) would be implemented, then the entire row would be made a control, thus the touch targets would span the whole height of the row, and it would drastically improve the touch target experience.

So I guess, that's what we should aim for.

Amejia481:
> My understanding of the bug is that we should make each row in the "table" below the add-on's description announceable. While this PR updates the labels in the first column, it doesn't allow users with Talkback to navigate between each row more easily

Could you help me to clarify here? I think the patch address navigation issue and announces each section as a row (Please ignore the rating section, it needs to be addressed 😅 ). 

**Without the fix**:

https://github.com/mozilla-mobile/firefox-android/assets/773158/bfe39480-f3d4-430e-9199-909653256a4b

**With the Fix**:

https://github.com/mozilla-mobile/firefox-android/assets/773158/b93aa5b7-b138-4aef-a19f-3ce3169ee0bf


> Also, I don't think links should be marked as "no important for a11y".

I'll address this!

Amejia481:
With improvements:
(Navigating from down to up to better visualization as the add-on long description occupies a lot of the screen)

https://github.com/mozilla-mobile/firefox-android/assets/773158/5e80061d-dcd9-41b4-834b-95e001f4bc97

willdurand:
@Amejia481 Thanks, in the emulator things worked differently for me. This looks better now (thanks!) but we should get someone from a11y to make sure we're fixing this part correctly. For instance, the control surface is still relatively small.

@anna-yeddi hello, could you help us here? thanks!

anna-yeddi:
> @Amejia481 Thanks, in the emulator things worked differently for me. This looks better now (thanks!) but we should get someone from a11y to make sure we're fixing this part correctly. For instance, the control surface is still relatively small.
> 
> @anna-yeddi hello, could you help us here? thanks!

Apologies for the delay and thank you for the ping, @Amejia481 ! 

The patch is not exactly what I was expecting, but it does achieve the goals of reducing the number of swipes needed to navigate the page and of bundling the key-value pairs together. If the buttons are working when you double-tap them (I assume they do, but just to confirm), an R+ from me on the a11y side.

Amejia481:
> If the buttons are working when you double-tap them (I assume they do, but just to confirm

Yes, they do :)

PR #5073, opened 2024-01-08
Bug 1873470 - Removed unused func `reportWebAppUsed`

t-p-white:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1873470

### DIFF
--- a/android-components/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
+++ b/android-components/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
@@ -237,15 +237,6 @@ class WebAppShortcutManager(
     }
 
     /**
-     * Updates the usedAt timestamp of the web app this url is associated with.
-     *
-     * @param manifest the manifest to update
-     */
-    suspend fun reportWebAppUsed(manifest: WebAppManifest): Unit? {
-        return storage.updateManifestUsedAt(manifest)
-    }
-
-    /**
      * Possible install states of a Web App.
      */
     enum class WebAppInstallState {


MozillaNoah (approved):
do we happen to know when this was last used?

t-p-white:
@MozillaNoah afaict it was never actually used. [This is the original PR](https://github.com/mozilla-mobile/android-components/pull/5724/files) and I couldn't find any subsequent uses

MozillaNoah:
> @MozillaNoah afaict it was never used by Fenix or Focus (which I'm assuming makes this okay to remove). [This is the original PR](https://github.com/mozilla-mobile/android-components/pull/5724/files) and I couldn't find any subsequent uses

Sounds good. Thanks for doing the due diligence with this.

android-components/components/feature/pwa/src/main/java/mozilla/components/feature/pwa/WebAppShortcutManager.kt
```diff
@@ -236,15 +236,6 @@ class WebAppShortcutManager(
         return storage.recentManifestsCount(activeThresholdMs = activeThresholdMs)
     }
 
-    /**
-     * Updates the usedAt timestamp of the web app this url is associated with.
-     *
-     * @param manifest the manifest to update
-     */
-    suspend fun reportWebAppUsed(manifest: WebAppManifest): Unit? {
```
gabrielluong, line 244: Often this is done to surface the underlying API even if it's not used in Fenix or Focus. I think this would be considered a breaking change and would require a changelog entry for AC since this is affecting a public API.

My general recommendation is that we do nothing here because there may be external consumers of AC that are using this even if we aren't using it ourselves.

t-p-white: Thanks @gabrielluong, good call out. Closing this PR and the related bug.

cc @MozillaNoah

PR #5080, opened 2024-01-08
Bug 1859192 - Always intercept app links if scheme is not supported by engine

rocketsroger:
No reason to block app links when our engine can't support it even if the user clicked cancel last time.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1859192


android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt
```diff
@@ -113,7 +113,7 @@ class AppLinksInterceptor(
         val redirect = useCases.interceptedAppLinkRedirect(uri)
         val result = handleRedirect(redirect, uri)
 
-        if (inUserDoNotIntercept(uri, redirect.appIntent)) {
+        if (engineSupportedSchemes.contains(uriScheme) && inUserDoNotIntercept(uri, redirect.appIntent)) {
```
t-p-white, line 116: Just looking at the comments on the bug, do we still want to do `inUserDoNotIntercept` at all?

rocketsroger: yes, but we don't want to not intercept any scheme our engine doesn't support.  Example here is if the user tap on tel:/ link and tapped cancel, if they tap on it again, we should still intercept.  Since our browser doesn't support that link, if we don't intercept, nothing happens.

t-p-white (approved):

PR #5097, opened 2024-01-10
Bug 1828927 - Fix apply wallpaper anr

rahulsainani:
### How
– Add wallpaper URL in debug builds so this issue can be reproduced in debug builds and is highlighted by strict mode using this snippet in app build.gradle. 

```
if (isDebug) {
        buildConfigField 'String', 'WALLPAPER_URL', '"https://assets.mozilla.net/mobile-wallpapers/android"'
    } else {
        try {
            def token = new File("${rootDir}/.wallpaper_url").text.trim()
            buildConfigField 'String', 'WALLPAPER_URL', '"' + token + '"'
            println "(Added from .wallpaper_url file)"
        } catch (FileNotFoundException ignored) {
            buildConfigField 'String', 'WALLPAPER_URL', '""'
            println("--")
        }
    }
```

– Use Fragment view's `lifecycleScope` instead of `runBlocking` so the main thread is not blocked. Also, in case the app enters paused state or changes configuration while the wallpapers are being loaded, this would cancel the previous load, which was the cause of many [ANRs](https://mozilla.sentry.io/issues/?groupStatsPeriod=auto&page=0&project=6295546&query=level%3Afatal+is%3Aunresolved+applyWallpaper&referrer=issue-list&sort=freq&statsPeriod=90d).
– Move `context.filesDir` call on IO.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1828927

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -49,7 +49,9 @@ import kotlinx.coroutines.Dispatchers.Main
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 import mozilla.components.browser.state.selector.findTab
 import mozilla.components.browser.state.selector.normalTabs
@@ -74,6 +76,7 @@ import mozilla.components.lib.state.ext.consumeFlow
 import mozilla.components.lib.state.ext.consumeFrom
 import mozilla.components.service.glean.private.NoExtras
 import mozilla.components.support.base.feature.ViewBoundFeatureWrapper
+import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.ui.colors.PhotonColors
 import org.mozilla.fenix.GleanMetrics.HomeScreen
 import org.mozilla.fenix.GleanMetrics.Homepage
@@ -119,7 +122,6 @@ import org.mozilla.fenix.messaging.FenixNimbusMessagingController
 import org.mozilla.fenix.messaging.MessagingFeature
 import org.mozilla.fenix.nimbus.FxNimbus
 import org.mozilla.fenix.perf.MarkersFragmentLifecycleCallbacks
-import org.mozilla.fenix.perf.runBlockingIncrement
 import org.mozilla.fenix.search.toolbar.DefaultSearchSelectorController
 import org.mozilla.fenix.search.toolbar.SearchSelectorMenu
 import org.mozilla.fenix.tabstray.TabsTrayAccessPoint
@@ -132,6 +134,8 @@ import java.lang.ref.WeakReference
 @Suppress("TooManyFunctions", "LargeClass")
 class HomeFragment : Fragment() {
     private val args by navArgs<HomeFragmentArgs>()
+
+    private val logger = Logger("HomeFragment")
 
     @VisibleForTesting
     internal lateinit var bundleArgs: Bundle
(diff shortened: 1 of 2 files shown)


jonalmeida:
Assigning Matt based on [comment 2](https://bugzilla.mozilla.org/show_bug.cgi?id=1828927#c2) in the bug.

MatthewTighe (approved):
This change does introduce noticeable visual jank on lower-end devices, which is why this code was badly written in the first place. IIRC we were asked to intentionally eliminate that when this code was first written. Some videos below demonstrate the issue on a Moto G5 release build. I do think the issue was noticeable on higher-end devices back then, but I'm not able to reproduce it on my Samsung Fold 3 now.

I would argue that the ANR rate is the greater of the two evils these days, so I'm happy to let this patch land. I do think that we should get either product or UX sign-off on this decision, though I have no problem shipping to nightly to verify the fix in the meantime. 


https://github.com/mozilla-mobile/firefox-android/assets/10427470/f156701d-71d4-4d22-91ee-68e6698a9390






https://github.com/mozilla-mobile/firefox-android/assets/10427470/2e766630-5303-4830-b2c8-314b5a1410ae


https://github.com/mozilla-mobile/firefox-android/assets/10427470/b04de22f-4bf1-41ed-8d0a-f749d51ceeed


https://github.com/mozilla-mobile/firefox-android/assets/10427470/0a1e8b1f-c31d-4a81-8532-5b6e810abebd

rahulsainani:
@MatthewTighe Thank you for sharing the videos! 👌 

Since details matter here, I don't think you mean [jank](https://developer.android.com/studio/profile/jank-detection) here as it means blocking the UI thread, ignoring user inputs. Longer Janks result in ANR, we're trying to fix one here. The slowness that you mention is from loading the bitmap from disk, not applying the wallpaper. With this change, main thread is not blocked when IO is happening, that means user can do interact with the app when they go to the home screen, even if wallpaper loading takes longer. You can test it by adding `delay(5000)` after `loadBitmap(wallpaper)` to experience what users might be experiencing when they try going to Home screen from any other screen. 

Fixing the loading speed of wallpaper would be the way to fix the visual defect you're talking about. There's lots of optimisations that image loading libraries do and we should try to learn from there.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1007,17 +1014,23 @@ class HomeFragment : Fragment() {
...
                         requireComponents.useCases.wallpaperUseCases.loadBitmap(wallpaper)
                     wallpaperImage?.let {
+                        logger.debug("loaded bitmap")
                         it.scaleToBottomOfView(binding.wallpaperImageView)
                         binding.wallpaperImageView.isVisible = true
                         lastAppliedWallpaperName = wallpaperName
+                        logger.debug("applied bitmap")
                     } ?: run {
+                        logger.debug("bitmap null, is coroutine active $isActive, ${lifecycle.currentState}")
+                        if (!isActive) return@run
```
rahulsainani, line 1033: As mode is updated on privacy mode to normal mode button click, the `observeWallpaperUpdates` gets triggered. Then the activity is recreated as part of themeManager logic, that results in this coroutine being cancelled as the view is destroyed, so we check here to make sure if the coroutine is active. If not, nothing to do.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -1051,11 +1064,16 @@ class HomeFragment : Fragment() {
     }
 
     private fun observeWallpaperUpdates() {
-        consumeFrom(requireComponents.appStore) {
-            val currentWallpaper = it.wallpaperState.currentWallpaper
-            if (currentWallpaper.name != lastAppliedWallpaperName) {
-                applyWallpaper(wallpaperName = currentWallpaper.name, orientationChange = false)
-            }
+        consumeFlow(requireComponents.appStore, viewLifecycleOwner) { flow ->
```
rahulsainani, line 1067: Updated this to make the flow more lifecycle aware, to stop when the lifecycler owner is destroyed based on consumeFlow's implementaiton.

MatthewTighe (approved):
Updates look good to me 👍

PR #5116, opened 2024-01-11
Bug 1874885 - Fix reset to default locale when a test crashes

sv-ohorvath:
More details here: https://mozilla-hub.atlassian.net/browse/MTE-2089
Both tests will have the system locale changed before they run, using the fun runWithSystemLocaleChanged(locale: Locale, testRule: ActivityTestRule<HomeActivity>, testBlock: () -> Unit) method.

After the tests crashes, [the finally statement](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/app/src/androidTest/java/org/mozilla/fenix/helpers/AppAndSystemHelper.kt#L272) is not executed and the setSystemLocale(defaultLocale) is never called.

The afterActivityFinished and beforeActivityLaunched methods seemed like the best places to have this, but there are 2 bugs which got in the way:
1 is an app bug, for which I can’t use the before method: if I change the locale then open the app, the Settings button is not translated immediately, you need to restart the app to see the change. That’s because that button (and a few others) come from GV and it doesn’t work instantly. (there is a bug logged for this somewhere) So this will still generate failures in other tests that use that button.
2nd it seems to be an Android bug which won’t be fixed: https://github.com/android/android-test/issues/498 - the afterActivityFinished methods are not executed if the activity doesn’t finish “peacefully”. It worked if the test passed, or failed, but doesn’t work if the test crashes.
So the only other way is to add this helper method in the tearDown of every test class where we change the locale.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1874885


fenix/app/src/androidTest/java/org/mozilla/fenix/ui/NavigationToolbarTest.kt
```diff
@@ -50,6 +51,7 @@ class NavigationToolbarTest {
     @After
     fun tearDown() {
         mockWebServer.shutdown()
+        AppAndSystemHelper.resetSystemLocaleToEnUS()
     }
```
jjSDET, line 55: I'd like to start moving away from using `@After` and any `tearDown` functions and instead include everything in a `@Before` hook in a `setUp` function.

This will always run even if the runner or device crashes and the `@After` hook isn't called.

To enable this, there will need to me more checks in the `setUp`, but an example might be:

```
@Before
fun setUp() {
    # first run functions to cover a possible missed `@After` hook on runner crash

    # calling shutdown on a server that isn't running typically doesn't throw an error
    # but verify if this is the case as we might need to `catch` it
    mockWebServer.shutdown()
    AppAndSystemHelper.resetSystemLocaleToEnUS()
    # Include any other checks that need to replace calls in existing @After hooks
    ...
    
    # Now include your standard @Before hooks for your test setUp
    mockWebServer.start()
    # include anything else you need
    ...
}

File: fenix/app/src/androidTest/java/org/mozilla/fenix/ui/SearchTest.kt [hunk 2]
```diff
@@ -92,6 +93,7 @@ class SearchTest {
     @After
     fun tearDown() {
         searchMockServer.shutdown()
+        AppAndSystemHelper.resetSystemLocaleToEnUS()
```
jjSDET, line 96: same comment as above. I'd like to convert this into an `@Before` hook instead.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/SettingsGeneralTest.kt
```diff
@@ -51,6 +52,7 @@ class SettingsGeneralTest {
     @After
     fun tearDown() {
         mockWebServer.shutdown()
+        AppAndSystemHelper.resetSystemLocaleToEnUS()
```
jjSDET, line 55: same comment as above

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/SettingsSearchTest.kt
```diff
@@ -64,6 +65,7 @@ class SettingsSearchTest {
     @After
     fun tearDown() {
         mockWebServer.shutdown()
+        resetSystemLocaleToEnUS()
```
jjSDET, line 68: same comment as above

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/NavigationToolbarTest.kt
jjSDET: If implementing this pattern is difficult or impractical for any reason, then please let me know because it's an indication we need to modify our test `setup/teardown` strategy from a lower level that might require migrating some code and disconnecting some low-level hooks that, IMHO, aren't appropriate for UI testing enablement.

jjSDET (changes requested):

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/NavigationToolbarTest.kt
sv-ohorvath: I've done some test runs and noticed [Bug 1812183](https://bugzilla.mozilla.org/show_bug.cgi?id=1812183) when I switch the device language before the test starts. Due to the mentioned bug, for the language change to take effect, you need to restart the app. So the only way we can avoid the bug and also make sure we're resseting the language after the test crashes, the only way left is inside the @After methods.

jjSDET: okay sounds good for now, thanks for the context.

there's a larger concern we'll have to find a good way to address, and that's implementing workarounds for known bugs.

when we do that, we lose the value of the measurement of a test in general, or at least dilute that value, by either requiring the test to be flaky if we know the first run will always fail, like for `SearchSuggest` or we run the risk of avoiding addressing the bug until later than we'd like. 

we'll figure it out later.

jjSDET (approved):
LGTM

PR #5141, opened 2024-01-12
Bug 1812843 - Increased the width for a CustomTabsToolbarFeature action button

t-p-white:
This bug was initially raised for the "Tweet" button.

Original image of bug provided:

"Tweet" has since be rebranded as "Post" but the problem is still the same.

Example of current Fenix:

Example in Chrome:

Example in Edge:

---

This patch aims to follow suit with the Chrome & Edge style by:

1.  increasing the width (doubling) to create a rectangle instead of a square - this will cater better for text
2. adding a small margin to the start of the button to prevent the URL from being too close, see example below

With margin:

Without margin:

---

With action button:

No action button:

### Pull Request checklist

### After merge

3. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
4. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1812843


t-p-white:
I don't love that the margin has to get passed as much as it needs to with this approach, open to alternative suggestions

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -254,16 +254,16 @@ class BrowserToolbar @JvmOverloads constructor(
...
      *
      * If there is not enough room to show all icons then some icons may be moved to an overflow
      * menu.
      *
      * Related:
      * https://developer.mozilla.org/en-US/Add-ons/WebExtensions/user_interface/Browser_action
      */
-    override fun addBrowserAction(action: Toolbar.Action) {
-        display.addBrowserAction(action)
+    override fun addBrowserAction(action: Toolbar.Action, marginStart: Int?) {
```
MozillaNoah, line 265: does this signature also need `marginStart: Int? = null`?

MozillaNoah (commented):
So two thoughts with alternative approaches:
- Expand the margin API so it's all 4 directions and not just `start`.
- Make it an optional API in `Toolbar.ActionButton` or `Toolbar.Action`, and then function signatures would not need to be changed.

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
t-p-white: As this is an overriding function:
1.  will inherit the default value from the interface
2. it cannot specify its own default

t-p-white:
Thanks for the conversation @MozillaNoah 🙂  I've updated the PR with a different solution which I think is a cleaner approach than my original

android-components/components/browser/toolbar/src/main/res/layout/mozac_browser_toolbar_displaytoolbar.xml
```diff
@@ -29,10 +29,12 @@
         android:layout_width="0dp"
```
t-p-white, line 29: we need to change the margin depending on whether the action button is displayed or not, see details for demo

MozillaNoah (approved):
Nice use of `layout_goneMarginEnd`. I totally forgot that was a thing with ConstraintLayout.

t-p-white:
@MozillaNoah since your last review, I had to also update the `CustomTabsToolbarFeatureTest` to account for the update to the width value

PR #5145, opened 2024-01-12
Bug 1873897 - Update Testrail Milestone Release Trigger and Slack Message

jjSDET:
This is Phase 2 of hardening the entry-point into taskcluster/testrail integration.

The goal of this is to decouple the previous implementation, which lived inside of `ui-test-apk` and make a dedicated task for handling the `testrail milestone creation` and `slack notifications`.

Summary of changes:
- added notification task `notify-release-testrail`
- split the previous script into `testrail_main.py`, `testrail_utils.py`, and `testrail_api.py`
- removed the duplicated files from `fenix` and `focus` and moved them to `taskcluster/scripts`
- added `slack_notify.py`
- removed previous code added to `ui-test-apk` and `ui-test.sh` for testrail/slack notifications

A successful Milestone creation and Slack message look like this:

![Screenshot 2024-01-30 at 1 08 18 AM](https://github.com/mozilla-mobile/firefox-android/assets/107960801/9f7cf46c-f4e7-479a-9ba3-625001afa77c)

![Screenshot 2024-01-30 at 1 12 36 AM](https://github.com/mozilla-mobile/firefox-android/assets/107960801/d199ae0c-b943-49ff-9373-60bd352dd7ba)


The goal is to create a new TestRail Milestone at the beginning of the testing cycle for each version, beta 1 through RC, that contains the Test Runs for our automated tests.

This need to happen at the beginning of the testing cycle and is a pre-condition of sign-off for release and promote, which limited the release hooks I could use, as far as I could tell.

Because of that, I decided to create a task, `notify-release-testrail`, that is triggered on any `push` to `releases_v`.

What this does is check if the current version already exists as a Milestone that's been created using the TestRail API. If it doesn't, then this should be the first push for that version, and a Milestone is created with a Slack notification using the Taskcluster `Notify.slack` API/service that goes out to #mobile-testeng notifying the release testers the new testing cycle is ready.

On all subsequent pushes to a release branch, the TestRail API is called to verify the Milestone already exists and will exit without creating a Milestone or Test Run and won't send a Slack notificaiton.

I added `retries` to the API calls to TestRail in case they fail, but don't know a more robust solution should the TestRail server/instance be down.

Ideally this task won't be build blocking and may need additional checks to swallow exceptions, send a slack notification of the failure, but still exit gracefully.


### Pull Request checklist
<!-- Before submitting the PR, please address each item -->

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1873897

### DIFF
--- a/android-components/components/browser/engine-system/src/main/res/values-zh-rCN/strings.xml
+++ b/android-components/components/browser/engine-system/src/main/res/values-zh-rCN/strings.xml
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <!-- Text for the title of an alert dialog displayed by a web page. %1$s will be replaced with the URL of the current page (displaying the dialog). -->
-    <string name="mozac_browser_engine_system_alert_title">域名为 %1$s 的页面说：</string>
+    <string name="mozac_browser_engine_system_alert_title">域名为 %1$s 的页面提示：</string>
     <!-- Text for the message of an auth dialog displayed by a web page.
     %1$s will be replaced by the hostname or a description of the protected area/site, %2$s will be replaced with the URL of the current page (displaying the dialog). -->
-    <string name="mozac_browser_engine_system_auth_message">%2$s 要求您输入用户名和密码。该网站说：“%1$s”</string>
+    <string name="mozac_browser_engine_system_auth_message">%2$s 要求您输入用户名和密码。该网站提示：“%1$s”</string>
     <!-- Text for the message of an auth dialog displayed by a web page. %1$s will be replaced with the URL of the current page (displaying the dialog). -->
     <string name="mozac_browser_engine_system_auth_no_realm_message">%1$s 要求您输入用户名和密码。</string>
 </resources>
--- a/android-components/components/feature/addons/src/main/res/values-am/strings.xml
+++ b/android-components/components/feature/addons/src/main/res/values-am/strings.xml
@@ -22,6 +22,8 @@
      then we will show another collapsed entry saying "Access your data on 2 other domains". This entry it's for the plural case, when the add-on is accessing more than one extra domain.
      %1$d will be replaced by an integer indicating the number of additional domains for which this web extension is requesting permission. -->
     <string name="mozac_feature_addons_permissions_extra_domains_description_plural" tools:ignore="PluralsCandidate">በ%1$d ሌሎች ጎራዎች ላይ የእርስዎን ውሂብ ይድረሱበት</string>
+    <!-- Content description for a permission item, for an installed add-on details page. %1$s will be replaced with the permission text name, %2$d will be replaced with the permission position in the list and %3$d will be replaced with the total number of permissions in the list.Taking as an example "Read and modify privacy settings" the whole text will read Read and modify privacy settings, 2 of 6. -->
+    <string name="mozac_feature_addons_permissions_content_description_item">%1$s፣ %2$d ከ%3$d</string>
     <!-- Description for tabs add-on permission. -->
     <string name="mozac_feature_addons_permissions_tabs_description">የአሳሽ ትሮችን ይድረሱ</string>
     <!-- Description for unlimited_storage permission. -->
@@ -75,7 +77,7 @@
     <!-- The author of an add-on. -->
     <string name="mozac_feature_addons_author">ደራሲ</string>
     <!-- The authors of an add-on. -->
-    <string name="mozac_feature_addons_authors" moz:removedIn="120" tools:ignore="UnusedResources">ደራሲያን</string>
+    <string name="mozac_feature_addons_authors" moz:removedIn="123" tools:ignore="UnusedResources">ደራሲያን</string>
     <!-- The last date that the add-on was updated. -->
     <string name="mozac_feature_addons_last_updated">መጨረሻ የተሻሻለው</string>
     <!-- The developer website (Homepage) of the add-on. -->
--- a/android-components/components/feature/addons/src/main/res/values-azb/strings.xml
+++ b/android-components/components/feature/addons/src/main/res/values-azb/strings.xml
@@ -22,6 +22,8 @@
      then we will show another collapsed entry saying "Access your data on 2 other domains". This entry it's for the plural case, when the add-on is accessing more than one extra domain.
      %1$d will be replaced by an integer indicating the number of additional domains for which this web extension is requesting permission. -->
     <string name="mozac_feature_addons_permissions_extra_domains_description_plural" tools:ignore="PluralsCandidate">باشقا %1$d دامنه‌‌لرده دیتالاریزا ال چاتما</string>
+    <!-- Content description for a permission item, for an installed add-on details page. %1$s will be replaced with the permission text name, %2$d will be replaced with the permission position in the list and %3$d will be replaced with the total number of permissions in the list.Taking as an example "Read and modify privacy settings" the whole text will read Read and modify privacy settings, 2 of 6. -->
+    <string name="mozac_feature_addons_permissions_content_description_item">%3$d -دن %2$d و %1$s</string>
     <!-- Description for tabs add-on permission. -->
     <string name="mozac_feature_addons_permissions_tabs_description">موروچو تاغلارینا ال چاتما</string>
     <!-- Description for unlimited_storage permission. -->
@@ -148,6 +150,8 @@
     <string name="mozac_feature_addons_addons_manager">تاخیلان مودیریتی</string>
     <!-- The title of the "crash" notification in the add-ons manager -->
     <string name="mozac_feature_addons_manager_notification_title_text">تاخیلان‌لار گئچیجی اولاراق گوجدن سالینمیش.</string>
+    <!-- The content of the "crash" notification in the add-ons manager -->
+    <string name="mozac_feature_addons_manager_notification_content_text">بیر ویا نئچه تاخیلان دایاندیریلدی، بودا سیستمیزی ثابیتسیز ائله‌دی.</string>
     <!-- Button to re-enable the add-ons in the "crash" notification -->
     <string name="mozac_feature_addons_manager_notification_restart_button">تاخیلان‌لاری یئنی‌دن باشلات</string>
     <!-- Button in the add-ons manager that opens AMO in a tab -->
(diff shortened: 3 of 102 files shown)


taskcluster/ci/ui-test-apk/kind.yml
```diff
@@ -7,12 +7,13 @@ loader: taskgraph.loader.transform:loader
 transforms:
     - android_taskgraph.transforms.gradle_optimization:transforms
     - android_taskgraph.transforms.test:transforms
+    - android_taskgraph.transforms.slack_notify:transforms
+    - taskgraph.transforms.notify:transforms
```
ahal, line 11: Ah, when we were chatting over zoom I didn't realize the test tasks themselves were doing the notification. That's totally fine, just a misunderstanding on my part.

In this case, to only show the notification on release graphs, you'd add something like:
```
if config.params.get("shipping_phase") not in ("ship", "promote"):
    yield task
    continue
    
# logic to add `notify` section
```

to your transform function. That way, it'll only notify when it runs on release graphs.

ahal: I will just mention, that in case you wanted to have a single notification that runs after *all* test tasks have finished, then you'd want to have a dedicated "notify" task that depends on all the tasks that it should be notifying about.

Here's an example of one such task in the `mozilla-vpn-client` repo:
https://github.com/mozilla-mobile/mozilla-vpn-client/blob/main/taskcluster/kinds/release-notify/kind.yml

taskcluster/ci/ui-test-apk/kind.yml
```diff
@@ -42,265 +41,14 @@ task-defaults:
...
-                        {
-                            "type": "mrkdwn",
-                            "text": ":testops-notify: created by <https://mozilla-hub.atlassian.net/wiki/spaces/MTE/overview|Mobile Test Engineering>"
-                        }
-                    ]
-                  }
-                ]
 
 tasks:
-    focus-arm-debug:
```
ahal, line 101: Why are these tasks being deleted? Is that a separate issue?

ahal (changes requested):

taskcluster/ci/ui-test-apk/kind.yml
jjSDET: No no, it was just temporary because initially I didn't know how to stop them all from being triggered and was trying to only trigger one task for testing purposes.

They're all being added back.

taskcluster/ci/ui-test-apk/kind.yml
jjSDET: i tried setting this up using
```
worker-type: succeed
worker:
    implementation: succeed
```
but i'm getting an assertion error from `set_implementation` because i guess `implementation` is already being set in the worker somewhere else...i'm guessing from the other `notify` tasks?

so i'm guessing i'm inheriting some stuff I don't know about?

also, in my new task, i'm using a python script, which needs some dependencies that might not be in the docker image used for `succeed`, but i can't even get that far to see if it works; so i've just been using a different work-type, the same as here, `b-android`.

some help here would be greatly appreciated.

taskcluster/ci/release-notify-testrail/kind.yml
```diff
@@ -0,0 +1,72 @@
...
+              path: .testrail_credentials.json
+              json: true
+    routes:
+        - notify.slack-channel.G016BC5FUHJ.on-failed
+    scopes:
+        - queue:route:notify.slack-channel.G016BC5FUHJ # notify mobile-alerts-sandbox on failure
+        - notify:slack-channel:G016BC5FUHJ
+        - queue:route:notify.slack-channel.C02KDDS9QM9 # notify mobile-testeng on success
+        - notify:slack-channel:C02KDDS9QM9
+
```
jjSDET, line 36: wanted to check with you @ahal , do these `notify` keys need to be indented under `queue`? not really sure if i need all of this since i'm not using `.on-failed` or `.on-succeed` here.

I'm only trying to add the necessary scopes so the python script can send a message to either channel depending on if the script passes or fails, not this task.

ahal: No this is right, they should be at the same level.

taskcluster/ci/release-notify-testrail/kind.yml
```diff
@@ -0,0 +1,72 @@
...
+    worker:
+        docker-image: {in-tree: ui-tests}
+        max-run-time: 120
+        env:
+            TESTRAIL_PROJECT_ID: '59' # Fenix Browser
+            TESTRAIL_TEST_SUITE_ID: '49319' # Test Automation Release Milestone
+    run:
+        use-caches: false
+        using: run-commands
+        secrets:
```
ahal, line 24: TIL firefox-android has this.. nifty! We should upstream this into Taskgraph proper.

taskcluster/ci/release-notify-testrail/kind.yml
```diff
@@ -0,0 +1,72 @@
...
+tasks:
+    create-milestone-focus:
+        dependencies:
+            ui-test-apk: ui-test-apk-focus-arm-beta
+        description: Create Testrail Milestone for Focus
+        run-on-tasks-for: [github-push]
+        run-on-git-branches: [releases_v]
+        run:
+            pre-commands:
+                # get-secrets is called from '..' directory so we need to cd into any directory to make it work
```
ahal, line 46: Doh, that's annoying. Would it be simple to fix that rather than add this workaround?

taskcluster/ci/release-notify-testrail/kind.yml
```diff
@@ -0,0 +1,72 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - taskgraph.transforms.notify:transforms
```
ahal, line 8: Since these tasks don't have the `notify` config anymore, you can remove these transforms.

taskcluster/ci/release-notify-testrail/kind.yml
```diff
@@ -0,0 +1,72 @@
...
+        run:
+            pre-commands:
+                # get-secrets is called from '..' directory so we need to cd into any directory to make it work
+                - ["cd", "fenix"]
+            commands:
+                - [python3, "../taskcluster/scripts/testrail_main.py"]
+        worker:
+            env:
+                SHIPPING_PRODUCT: fenix
+                TESTRAIL_PRODUCT_TYPE: Firefox
```
ahal, line 71: I noticed these tasks are pretty similar save for a few minor changes. You could probably simplify this into a single definition that then gets split out in a transform.. but honestly this is more than good enough and if you really want to go further with this, you can do so in a follow-up.

ahal (approved):
Thanks!

I didn't look at the testrail scripts too closely as I'm not familiar with that. You might want to get someone else on your team to review. The slack script and task definitions look good though!

There's a couple nits / questions but nothing blocking. Nice work!

taskcluster/scripts/lib/testrail_api.py
```diff
@@ -0,0 +1,80 @@
+import os
```
AaronMT, line 1: We should add a top level Mozilla license in these 
```
# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```

As well some descriptors as to what this script does

taskcluster/scripts/lib/testrail_api.py
```diff
@@ -0,0 +1,80 @@
...
+        :param delay: Delay between retries in seconds.
+        """
+        for attempt in range(max_retries):
+            try:
+                return api_call(*args)
+            except Exception as e:
+                if attempt == max_retries - 1:
+                    raise  # Reraise the last exception
+                time.sleep(delay)
+
```
AaronMT, line 80: nit: newline

taskcluster/scripts/lib/testrail_utils.py
```diff
@@ -0,0 +1,41 @@
+from datetime import datetime
```
AaronMT, line 1: Same as above

taskcluster/scripts/slack_notifier.py
```diff
@@ -0,0 +1,135 @@
+import json
```
AaronMT, line 1: Same as above. As well since this lives in `taskcluster/scripts` I would add how this file can be used if people would like to start using it.

taskcluster/scripts/testrail_main.py
```diff
@@ -0,0 +1,68 @@
+"""
```
AaronMT, line 1: Same as above (License)

taskcluster/scripts/testrail_main.py
```diff
@@ -0,0 +1,68 @@
...
+    release_type = get_release_type(release_version)
+
+    # Build milestone information
+    milestone_name = build_milestone_name(testrail_product_type, release_type, release_version)
+    milestone_description = build_milestone_description(milestone_name)
+    
+    options = get_taskcluster_options()
+    success_channel_id = 'C02KDDS9QM9' # mobile-testeng 
+    error_channel_id = 'G016BC5FUHJ' # mobile-alerts-sandbox
+    try:
```
AaronMT, line 45: nit: newline above `try`

taskcluster/scripts/testrail_main.py
```diff
@@ -0,0 +1,68 @@
...
+    release_version = get_release_version()
+    release_type = get_release_type(release_version)
+
+    # Build milestone information
+    milestone_name = build_milestone_name(testrail_product_type, release_type, release_version)
+    milestone_description = build_milestone_description(milestone_name)
+    
+    options = get_taskcluster_options()
+    success_channel_id = 'C02KDDS9QM9' # mobile-testeng 
+    error_channel_id = 'G016BC5FUHJ' # mobile-alerts-sandbox
```
AaronMT, line 44: nit: I would capitalize/set these constants top level out of main

AaronMT (changes requested):
Looks great; just a couple of minor things to the python scripts

taskcluster/ci/release-notify-testrail/kind.yml
jjSDET: done!

taskcluster/ci/release-notify-testrail/kind.yml
jjSDET: that's a great idea, thank you. I need to get more familiar with everything in general and that sounds like a great exercise as well. i'll create a follow-up story to do that.

taskcluster/scripts/testrail_main.py
jjSDET: done

taskcluster/ci/release-notify-testrail/kind.yml
```diff
@@ -0,0 +1,71 @@
...
+
+task-defaults:
+    description: Sends Slack message to release testers that Testrail Milestone was created.
+    worker-type: b-android
+    worker:
+        docker-image: {in-tree: ui-tests}
+        max-run-time: 120
+        env:
+            TESTRAIL_PROJECT_ID: '59' # Fenix Browser
+            TESTRAIL_TEST_SUITE_ID: '49319' # Test Automation Release Milestone
```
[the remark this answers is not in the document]
jjSDET: I created a new test suite in their project to use just for this called `Test Automation Release Milestone`. Based on our discussion from earlier, I'd like to use something more dynamic so all the test cases are updated automatically and accurate based on what tests are currently disabled/enabled at the time of the test run.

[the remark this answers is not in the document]
isabelrios: If you are not modifying things, you should use the [Full Functional test suite](https://testrail.stage.mozaws.net/index.php?/suites/view/3192&group_by=cases:section_id&group_order=asc&group_id=385792) they use as the source of truth. That's the one where they add / remove / udpate manual tests and their references to automated tests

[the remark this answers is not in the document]
jjSDET: i'd have to extend the functionality of `testrail_api.py` for that, which I can do. For that case, I'd need to add a method to `filter_request` or something to add the filters to return only test cases that have stuff like `Automation Status: Completed, Sub-Type: Smoke&Sanity` or whatever we want. 

I'll create a follow-up for this to add that because it'll definitely come up again, especially regarding auto-updating the test cases for each run based on the data-alignment work that's been done.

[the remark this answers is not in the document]
jjSDET: As a workaround for now, I created the stand-alone test suite to use until I explored filtering with the TestRail API.

Does that work as a temporary solution and I can include that in Phase 4 or 5?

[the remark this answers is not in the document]
isabelrios: That sounds perfect to me! Thanks!

AaronMT (approved):

taskcluster/ci/release-notify-testrail/kind.yml
jjSDET: I a _little_ bit of time looking at it and I think the pathing was changed for all (most/some ?) of the `transform scripts` because of the mono-repo, and I was too afraid there was something else it was necessary for somewhere that I'm not accounting for.

I'm too afraid to modify any of the transforms yet since they are applied everwhere, or at least in some ways, I'm not 100% confident about.

It was definitely more of a skills/confidence gap, but might be simple to fix. The risk just seemed too high to me.

PR #5148, opened 2024-01-12
Bug 1855979 - Basic Translation Beginning

ohall-m:
The goal of this bug is to get a very basic initial `translate` interaction using the default determined page language and default determined user language and begin starting `Store` and `Interactor` files.

This bug also sets up `sessionId` as a navigation parameter for translations.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1855979


fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -115,7 +125,20 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
                                             from = BrowserDirection.FromTranslationsDialogFragment,
                                         )
                                     },
-                                    onTranslateButtonClick = {},
+                                    onTranslateButtonClick = {
```
ohall-m, line 128: This is just a very basic default translate interaction to get the necessary files started. (Originally planned to also populate the dropdowns here, but it felt like the PR was growing too fast.)

Split work to [Bug 1874441](https://bugzilla.mozilla.org/show_bug.cgi?id=1874441) to add the dropdown data and populate the request off of those choices.

Split work to [1874477](https://bugzilla.mozilla.org/show_bug.cgi?id=1874477) to manage the pending states that'll occur during translating.

ohall-m:
Still thinking about a few things, but any early comments are welcome!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsInteractor.kt
```diff
@@ -0,0 +1,38 @@
...
+     * @param options Optional options to specify and additional criteria for the translation.
+     */
+    fun onTranslate(
+        tabId: String?,
+        fromLanguage: String,
+        toLanguage: String,
+        options: TranslationOptions?,
+    ) {
+        logger.info("Requesting a translation.")
+        translationUseCase.invoke(tabId, fromLanguage, toLanguage, options)
```
iorgamgabriel, line 36: I think the interactor should only call some methods from a controller, and the implementation of the methods should be done in the controller. So here, I think a controller is needed. Just like AddLoginInteractor and SavedLoginsStorageController.

iorgamgabriel: I think something like this it should look like : 
 interactor = TranslationsInteractor(
            TranslationsController(
                translationUseCase = requireComponents.useCases.sessionUseCases.translate,
                browserStore = requireComponents.core.store,
                tabId = args.sessionId,
            ),
        ) 
in Fragment.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -115,7 +125,20 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
...
+                                        store.getDetectedLanguages()?.let {
```
iorgamgabriel, line 129: I don't think it is okay to call methods from the store directly from fragments.You should call in fragment a method from interactor, from interector to controller, and then in controller the store.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsStore.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Manages coordinating the retrieval of data for use in the views.
+ *
+ * @param browserStore The browser store to request information from.
+ * @param tabId The tab id requesting data.
+ */
+class TranslationsStore(
+    private val browserStore: BrowserStore,
+    private val tabId: String,
```
iorgamgabriel, line 20: If we call methods only from BrowserStore, I don't think a translation store is needed. We can do it with a controller. If a TranslationsStore is need we should extend class Store<S : State, A : Action>.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsInteractor.kt
ohall-m: Thanks for taking a look at this! 

That makes sense to me for sure! I had started this as a controller in a draft and refactored it out after reading this [architecture document](https://github.com/mozilla-mobile/firefox-android/blob/93c3ebad015317fcefd831b66c43cc64d763d616/fenix/docs/architecture-overview.md#interactor) misinterpreting something. I'll move back to controller, that seemed right to me at the time too.

Rereading it now, I think I interpreted:
>Note that prior to the introduction of Controllers, Interactors handled the responsibilities of both objects. You may still find this pattern in some parts of the codebase, but it is being actively refactored out.

as "don't add new `Controllers`", but on rereading I see `actively refactored out` is talking about `Interactors handled the responsibilities of both objects`.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: Thanks for mentioning this, was wondering about that!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsInteractor.kt
ohall-m: Updated to a `controller` pattern, and thanks for the really clear login example!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: Updated, so that the `translate` will revert to some sensible use cases for nulls to get this out of the fragment.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsStore.kt
ohall-m: Removed it for the controller pattern! I'll also keep that need to extend in mind if we need to add a `TranslationsStore` later.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsController.kt
```diff
@@ -0,0 +1,84 @@
...
+        if (fromLanguage != null && toLanguage != null) {
+            logger.info("Requesting a translation.")
+            translationUseCase.invoke(tabId, fromLanguage, toLanguage, options)
+            return
+        }
+
+        // Fallback to find defaults
+        var defaultFromLanguage = fromLanguage
+        var defaultToLanguage = toLanguage
+        var detectedLanguages = getDetectedLanguages()
```
iorgamgabriel, line 66: I think here it should be val .

iorgamgabriel (approved):
LGTM

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsController.kt
iorgamgabriel: I tested with 
interactor.onTranslate(
                                            tabId = args.sessionId,
                                            fromLanguage = "en",
                                            toLanguage = "es",
                                            null,
                                        )
and it is working.

ohall-m: Updated, thanks for catching the `val`!

Amejia481:
Sorry trying to prioritize https://github.com/mozilla-mobile/firefox-android/pull/2910, I'll add the label after it lands.

PR #5150, opened 2024-01-12
Bug 1875187 - Avoid race condition setting mode before opening a tab

MatthewTighe:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1875187

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/home/intent/ReEngagementIntentProcessor.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/home/intent/ReEngagementIntentProcessor.kt
@@ -15,8 +15,6 @@ import org.mozilla.fenix.HomeActivity
 import org.mozilla.fenix.NavGraphDirections
 import org.mozilla.fenix.R
 import org.mozilla.fenix.browser.browsingmode.BrowsingMode
-import org.mozilla.fenix.components.AppStore
-import org.mozilla.fenix.components.appstate.AppAction
 import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.onboarding.ReEngagementNotificationWorker
 import org.mozilla.fenix.onboarding.ReEngagementNotificationWorker.Companion.isReEngagementNotificationIntent
@@ -32,7 +30,6 @@ import org.mozilla.fenix.utils.Settings
 class ReEngagementIntentProcessor(
     private val activity: HomeActivity,
     private val settings: Settings,
-    private val appStore: AppStore,
 ) : HomeIntentProcessor {
 
     override fun process(intent: Intent, navController: NavController, out: Intent): Boolean {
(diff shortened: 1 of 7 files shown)


sv-ohorvath:
@MatthewTighe I saw today this test which seems to be affected by the same issue. Can you take a look, please? https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/6854215613285633784/executions/bs.9ca07fa2ef429fc4

MatthewTighe:
The `RecentlyClosed` test definitely seems related, but I don't think the `Bookmarks` test follows a codepath I'd expect to be problematic. I'll have to look into it more.

MatthewTighe:
@sv-ohorvath Still not sure that the bookmarks test is related. It looks like that test is failing to find the 3 dot menu in the toolbar. Is that right?

In the meantime, I've updated the patch to fix the bugs for top sites and recently closed. My hope is to land it before the cut next week so we can avoid these bugs in release.

fenix/app/src/main/java/org/mozilla/fenix/home/intent/ReEngagementIntentProcessor.kt
```diff
@@ -49,12 +46,12 @@ class ReEngagementIntentProcessor(
                         navController.nav(null, directions, options)
                     }
                     else -> {
-                        appStore.dispatch(AppAction.OpenTabInBrowser(BrowsingMode.Private))
```
MatthewTighe, line 52: This was the only other place using this, so I think it makes sense to just remove it now that we have the ability to

gabrielluong: I think we need to also remove this in https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/test/java/org/mozilla/fenix/home/intent/ReEngagementIntentProcessorTest.kt#89

gabrielluong:
@MatthewTighe I was ready to approve this PR, but seeing that we are still uncertain about the bookmark tests. I am thinking we should instead back out the regressing changes and reland after the cut happens. There is really no time to verify that your changes here has solve all the lingering issues if we landed this today since it is soft freeze. I think it would be wise to avoid any uncertainty and risk and just buy ourselves a bit more time.

MatthewTighe:
Original patch reverted, but I am still curious for more details about the bookmarks test case failure if you are able to provide them @sv-ohorvath. We will plan to address this patch and the bookmarks bug after the start of the new nightly cycle next week

gabrielluong:
@MatthewTighe Do you think we should close this PR?

MatthewTighe:
Yeah, I'll close this for now and include it in the re-ordered patches for https://bugzilla.mozilla.org/show_bug.cgi?id=1861459

PR #5171, opened 2024-01-15
Bug 1871408 - Translations UI Add Messages Under On Toggles

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1871408

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
@@ -17,6 +17,9 @@ import androidx.compose.material.Icon
 import androidx.compose.material.IconButton
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.snapshots.SnapshotStateList
+import androidx.compose.runtime.toMutableStateList
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.painterResource
@@ -49,9 +52,42 @@ fun TranslationOptionsDialog(
 ) {
     TranslationOptionsDialogHeader(onBackClicked)
 
+    val translationOptionsListState = remember {
+        translationOptionsList.toMutableStateList()
+    }
+
+    EnabledTranslationOptionsItems(
+        translationOptionsListState,
+    )
+
     LazyColumn {
-        items(translationOptionsList) { item: TranslationSwitchItem ->
-            TranslationOptions(translationSwitchItem = item)
+        items(translationOptionsListState) { item: TranslationSwitchItem ->
+
+            val translationSwitchItem = TranslationSwitchItem(
+                textLabel = item.textLabel,
+                description = if (item.isChecked) item.description else null,
+                importance = item.importance,
+                isChecked = item.isChecked,
+                isEnabled = item.isEnabled,
+                hasDivider = item.hasDivider,
+                onStateChange = { checked ->
+                    // If the item has the same importance, only one switch should be enabled.
+                    val iterator = translationOptionsListState.iterator()
+                    iterator.forEach {
+                        if (it != item && it.importance == item.importance && it.isChecked) {
+                            it.isChecked = false
+                        }
+                    }
+
+                    val index = translationOptionsListState.indexOf(item)
+                    translationOptionsListState[index] = translationOptionsListState[index].copy(
+                        isChecked = checked,
+                    )
+                },
+            )
+            TranslationOptions(
+                translationSwitchItem = translationSwitchItem,
+            )
         }
 
         item {
@@ -79,14 +115,43 @@ fun TranslationOptionsDialog(
     }
 }
 
+/**
+ * If the item with the highest importance is checked, all other items should be disabled.
+ * If all items are unchecked, all of them are enabled.
+ * If the item with the highest importance is unchecked and a item with importance 1 is checked ,
+ * the item with importance 0 is disabled.
+ * If the item with importance 0 is checked all the items are enabled.
+ */
 @Composable
-private fun TranslationOptions(translationSwitchItem: TranslationSwitchItem) {
+private fun EnabledTranslationOptionsItems(
+    translationOptionsListState: SnapshotStateList<TranslationSwitchItem>,
+) {
+    val itemCheckedWithHighestImportance =
+        translationOptionsListState.sortedByDescending { listItem -> listItem.importance }
+            .firstOrNull { it.isChecked }
+
+    if (itemCheckedWithHighestImportance == null || itemCheckedWithHighestImportance.importance == 0) {
+        translationOptionsListState.forEach {
+            it.isEnabled = true
+        }
+    } else {
+        translationOptionsListState.forEach {
+            it.isEnabled = it.importance >= itemCheckedWithHighestImportance.importance
+        }
+    }
+}
+
+@Composable
+private fun TranslationOptions(
+    translationSwitchItem: TranslationSwitchItem,
+) {
     SwitchWithLabel(
+        label = translationSwitchItem.textLabel,
+        description = translationSwitchItem.description,
+        enabled = translationSwitchItem.isEnabled,
         checked = translationSwitchItem.isChecked,
         onCheckedChange = translationSwitchItem.onStateChange,
-        label = translationSwitchItem.textLabel,
-        modifier = Modifier
-            .padding(start = 72.dp, end = 16.dp),
+        modifier = Modifier.padding(start = 72.dp, end = 16.dp),
     )
 
     if (translationSwitchItem.hasDivider) {
@@ -106,8 +171,7 @@ private fun TranslationOptionsDialogHeader(
     ) {
         IconButton(
             onClick = { onBackClicked() },
-            modifier = Modifier
-                .size(24.dp),
+            modifier = Modifier.size(24.dp),
         ) {
             Icon(
                 painter = painterResource(id = R.drawable.mozac_ic_back_24),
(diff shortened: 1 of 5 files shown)


iorgamgabriel:
https://github.com/mozilla-mobile/firefox-android/assets/8118371/6301a817-0c78-4b64-8ff5-8d409a9b09a1

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
```diff
@@ -49,9 +52,45 @@ fun TranslationOptionsDialog(
...
+    EnabledTranslationOptionsItems(
+        translationOptionsListState,
+    )
+
     LazyColumn {
-        items(translationOptionsList) { item: TranslationSwitchItem ->
-            TranslationOptions(translationSwitchItem = item)
+        items(translationOptionsListState) { item: TranslationSwitchItem ->
+            val description =
+                stringResource(id = R.string.translation_option_bottom_sheet_switch_description)
```
ohall-m, line 66: For the two in the language section, e.g.,  "Always translate Spanish" and "Never translate Spanish", could you use the description "Overrides offer to translate"? 

[String Figma](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7202%3A122161&mode=dev) I'm basing this off of.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
```diff
@@ -49,9 +52,45 @@ fun TranslationOptionsDialog(
...
+    }
+
+    EnabledTranslationOptionsItems(
+        translationOptionsListState,
+    )
+
     LazyColumn {
-        items(translationOptionsList) { item: TranslationSwitchItem ->
-            TranslationOptions(translationSwitchItem = item)
+        items(translationOptionsListState) { item: TranslationSwitchItem ->
```
ohall-m, line 64: Could you also go ahead and add logic for always showing the descriptions while in [landscape](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7796%3A15675&mode=dev) too?

ohall-m (commented):
Looks good to me! Just a string change and suggestion for landscape mode.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
```diff
@@ -80,13 +119,41 @@ fun TranslationOptionsDialog(
 }
 
 @Composable
-private fun TranslationOptions(translationSwitchItem: TranslationSwitchItem) {
+private fun EnabledTranslationOptionsItems(
```
ohall-m, line 122: Could you add some comments here too? Maybe a quick overview of how the logic works or behavior?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
```diff
@@ -80,13 +119,41 @@ fun TranslationOptionsDialog(
...
+    translationOptionsListState: SnapshotStateList<TranslationSwitchItem>,
+) {
+    val itemCheckedWithHighestImportance =
+        translationOptionsListState.sortedByDescending { listItem -> listItem.importance }
+            .firstOrNull { it.isChecked }
+
+    // If no items are checked, then all the switches should be enabled.
+    if (itemCheckedWithHighestImportance == null || itemCheckedWithHighestImportance.importance == 0) {
```
ohall-m, line 130: I'm still thinking about how we will connect all of these calls and manage state on this screen. I think this will set us up to manage the view state in a nice way. 

What are your thoughts on organizing those calls?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
```diff
@@ -8,14 +8,20 @@ package org.mozilla.fenix.translations
  * TranslationSwitchItem that will appear on Translation screens.
  *
  * @property textLabel The text that will appear on the switch item.
+ * @property description An optional description text below the label.
+ * @property importance Based on this, the translation switch item is enabled or disabled.
```
ohall-m, line 12: nit:  From what I understand, `importance` only controls showing the `description`, is that right? Maybe adjust the language here since `importance` reads like it could impact the switch's `isEnabled` state too.

ohall-m: Reread the PR: I do see some coupling on `importance` with other states. I think we might want to keep them decoupled, what do you think? 

Might also depend on  how we'd like the controller to send the initial state over.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
iorgamgabriel: In landscape it works the same like in portrait . I put a comment in Figma; I don't think the design is correct.
https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7796:15675&mode=dev#650459280

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
iorgamgabriel: Based on importance, I enabled and disabled the views, or I showed the description.

iorgamgabriel: The logic here is complicated because of the design. It happens a lot of times, almost at random. The description is different for every view, and the enabled state is the same.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
ohall-m: Thanks for that background! Yeah, I agree the design is complicated here with a whole lot to consider. 

I wonder if we should consider if there is way to decouple the logic from the view and have a utility-like function that pre-calculates the state so that the view is purely consuming a data model?

Where it is so complicated, there is a lot to follow on how the view is making the determinations on if it should be switched or enabled. Also, it would be nice to be able to test this logic. 

We can also revisit how all of this works when connecting to the functionality too and have some of the controller logic setup.

I'm wondering maybe something like:

```
initialSettingState : PagePreferences = {
  globalAlways = true,
  languageAlways = true,
  languageNever = false,
  siteNever = false,
}
```


```
toggleSettings : TranslationSwitchItem = {
  textLabel,
  description,
  checked,
  enabled, 
  anything else we need for each view toggle to know
}
```

```
initialSettingState : PageSwitchSettings = {
  globalAlways : TranslationSwitchItem,
  languageAlways : TranslationSwitchItem,
  languageNever : TranslationSwitchItem,
  siteNever : TranslationSwitchItem,
}
```

```
computePageSwitchSettingsState(preferences: PagePreferences) : PageSwitchSettings {

  // Logic on how the initial settings determines toggle state in one place

  return PageSwitchSettings
}
```

Just thinking out loud! I think this might also be beneficial when we are connecting to the functionality too.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
ohall-m: Thanks for checking with the design team!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationOptionsDialog.kt
ohall-m: (Conversation happened at [other comment](https://github.com/mozilla-mobile/firefox-android/pull/5171#discussion_r1454148224))

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSwitchItem.kt
ohall-m: We talked in Slack - we are going to follow-up on this when connecting to the functionality, so we have the full context.

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2301,6 +2301,10 @@
     <string name="translation_option_bottom_sheet_never_translate_in_language">Never translate %1$s</string>
     <!-- Toggle switch label that allows a user to set the setting if they would like the browser to never translate the site the user is currently visiting. -->
     <string name="translation_option_bottom_sheet_never_translate_site">Never translate this site</string>
+    <!-- Toggle switch description that will appear under the label if the switch is checked.If the switch label is "translation_settings_automatic_never_translate_sites" -->
```
ohall-m, line 2304: I'm not sure if pontoon can connect string names like `translation_settings_automatic_never_translate_sites` to the string, maybe substitute in the comment to be safe?

```suggestion
    <!-- Toggle switch description that will appear under the "Never translate these sites" settings toggle switch to provide more information on how this setting interacts with other settings. -->
```

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2301,6 +2301,10 @@
...
+    <string name="translation_option_bottom_sheet_switch_never_translate_site_description">Overrides all other settings</string>
+    <!-- Toggle switch description that will appear under the label if the switch is checked.If the switch label is "translation_option_bottom_sheet_never_translate_in_language" or "translation_option_bottom_sheet_always_translate_in_language" -->
```
ohall-m, line 2306:
```suggestion
    <!-- Toggle switch description that will appear under the "Never translate" and "Always translate" toggle switch settings to provide more information on how these  settings interacts with other settings. -->
```

ohall-m (approved):
LGTM!

Please wait until 124 begins before merging because of new strings.

PR #5172, opened 2024-01-15
Bug 1872704 - Fix HomeActivity leak by ExtensionsProcessDisabledController

rahulsainani:
### Context
`ExtensionsProcessDisabledController` extends from `ExtensionsProcessDisabledPromptObserver` that wasn't previously cancelling the coroutine when the `LifecycleObserver`/`HomeActivity` was being destroyed. This was by design to observe the state changes even when the app was in foreground. However, by passing a reference to `HomeActivity`, it held the reference to the destroyed activity, preventing it from being GC'ed. 

### Solution
The solution is to break the Controller into two - 
1. ExtensionsProcessDisabledForegroundController - This still holds the reference to the Activity to show a dialog, but it cancels the coroutine onStop and starts in `onStart`. As a result the destroyed activity can be GC'ed. No behaviour changes. 
2. ExtensionsProcessDisabledForegroundController - This doesn't hold the reference to the Activity and continue to observe the state changes even when the app is background, and performs the same logic as before. 

### Way to recreate activity/detect leak
One of the ways is to simply change from browsing mode from normal to private mode or back. This recreates the activity and leak can be easiliy detected by LeakCanary. Details about leak in the bug.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1872704


fenix/app/src/main/java/org/mozilla/fenix/addons/ExtensionsProcessDisabledForegroundController.kt
```diff
@@ -20,35 +18,30 @@ import mozilla.components.support.webextensions.ExtensionsProcessDisabledPromptO
...
  * @param context to show the AlertDialog
  * @param browserStore The [BrowserStore] which holds the state for showing the dialog
  * @param appStore The [AppStore] containing the application state
  * @param builder to use for creating the dialog which can be styled as needed
  * @param appName to be added to the message. Optional and mainly relevant for testing
- * @param onKillApp called when the app is backgrounded and extensions process is disabled
  */
-class ExtensionsProcessDisabledController(
+class ExtensionsProcessDisabledForegroundController(
     @UiContext context: Context,
```
Amejia481, line 34: Thanks for putting the patch!
In general the solution looks good, but could we address the issue by just using a `WeakReference` of `Context`?

rahulsainani: Thanks @Amejia481 ! In theory yes, I didn't try that and test it though.

Amejia481: I really like your suggestion 💪🏽 , but it will be nice if we could give a try, as it maybe a bit simpler and it's pattern that more widely use in the code base :)

rahulsainani: Sounds reasonable! I also like the solution since it separates two different actions we take based on foreground or background, but I also went with the aim of keeping changes to the minimum, I'll give that a shot!

Amejia481 (approved):
LGTM ✅ 
@rahulsainani and I chatted offline, and I agreed to keep the actual solution, so we are making sure we are cancelling the the scope, whenever is not needed anymore.

rahulsainani:
Thanks a lot for the review @Amejia481 ! And the discussion! 🙌 

After thinking more, I'm leaning towards having two different controllers over WeakReference since using weak reference only masks the problem with the design. It would prevent specific instances of memory leaks by allowing objects to be garbage collected, but it doesn't address the underlying issue—the reason why those objects are being retained longer than they should be. I think it's better to understand why objects are being retained too long and to fix the issue by using proper lifecycle management.

PR #5175, opened 2024-01-15
Bug 1874733 - Add exported flag to broadcast receiver for notification dismissal.

mcarare:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1874733


android-components/components/feature/media/src/main/java/mozilla/components/feature/media/middleware/RecordingDevicesMiddleware.kt
```diff
@@ -192,12 +193,15 @@ class RecordingDevicesMiddleware(
         private val processRecordingState: () -> Unit,
     ) : BroadcastReceiver() {
         override fun onReceive(context: Context, intent: Intent) {
-            ThreadUtils.postToMainThreadDelayed(
-                {
-                    processRecordingState.invoke()
-                },
-                NOTIFICATION_REMINDER_DELAY_MS,
-            )
+            val callingAppInfo = context.packageManager.getNameForUid(Binder.getCallingUid())
```
mcarare, line 196: This is for protecting against receiving the call from other apps, but maybe it's an overkill.

mcarare:
Also tagging @Amejia481 due to this being possibly related to [Restrictions to implicit and pending intents](https://developer.android.com/about/versions/14/behavior-changes-14#safer-intents) in Android 14.

Also tried using explicit intent but that does not seem to work.

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/middleware/RecordingDevicesMiddleware.kt
```diff
@@ -157,7 +158,7 @@ class RecordingDevicesMiddleware(
         context.registerReceiverCompat(
             broadcastReceiver,
             IntentFilter(ACTION_RECORDING_DEVICES_NOTIFICATION_DISMISSED),
-            ContextCompat.RECEIVER_NOT_EXPORTED,
+            ContextCompat.RECEIVER_EXPORTED,
```
Amejia481, line 161: Checking at line 137 to 140: 

```kotlin
        val intent =
            context.packageManager.getLaunchIntentForPackage(context.packageName)?.apply {
                flags = Intent.FLAG_ACTIVITY_NEW_TASK
} ?: throw IllegalStateException("Package has no launcher intent")

```

This looks safe as we are passing the `packageName`.

jonalmeida: Not a big deal right now, but don't those lines also mean that other applications can observe and receive updates when a user swipes away our notification? Shouldn't the intent be non-exported in that case (so that it matches this registerReceiverCompat as well)?

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/middleware/RecordingDevicesMiddleware.kt
jonalmeida: No, I don't think so. I didn't see this comment until after I posted the above comment, so I guess I had the same concerns.

android-components/components/feature/media/src/main/java/mozilla/components/feature/media/middleware/RecordingDevicesMiddleware.kt
mcarare: It's the other way around, the receiver can receive broadcasts from other apps not just our own, when the exported flag is used.

To answer the other questions, the intent received is the one with `ACTION_RECORDING_DEVICES_NOTIFICATION_DISMISSED`, that does not have a package set. I tried to set the package to make it an explicit intent, but it is still not received. Without investigating further I would say that it is because the system actually sends the intent set for the delete intent, not our app.

That means that the receiver has to receive broadcasts from anywhere, thus the reason for the filtering in `onReceive`. 

Side note, the documentation says that for broadcasts received exclusively from the system you don't have to specify a flag. From my testing and several bugs filed on the Google issue tracker, it seems that causes crashes even if the docs say they shouldn't.

Amejia481 (approved):

PR #5177, opened 2024-01-15
Bug 1874771 - Only run 'github-push' Taskcluster tasks against main a…

ahal:
…nd release branches

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1874771

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -147,7 +147,7 @@ tasks:
                   $if: >
                       tasks_for in ["action", "cron"]
                       || (isPullRequest && pullRequestAction in ["opened", "reopened", "synchronize"])
-                      || (tasks_for == "github-push" && head_ref[:10] != "refs/tags/" && short_head_ref != "staging.tmp" && short_head_ref != "trying.tmp" && short_head_ref[:8] != "mergify/")
+                      || (tasks_for == "github-push" && (short_head_ref == "main" || short_head_ref[:9] == "releases_"))
                   then:
                       $mergeDeep:
                           - $if: 'tasks_for != "action"'


jcristau (changes requested):
This will prevent merging PRs created by github actions.

ahal:
> This will prevent merging PRs created by github actions.

How so? If a PR is created, then that PR will still get a `github-pull-request` graph. Currently those PRs are getting both `github-push` and `github-pull-request` graphs, and there's a race condition over which task ends up getting displayed in the Github checks UI (the later task overwrites the earlier task in the UI)

ahal:
Ftr, I'm not opposed to leaving tasks on as level 1.. but in my experience with VPN, if your workflow involves creating pull-requests off of branches in the canonical repo, you probably want to disable `on-push` tasks in those branches

jcristau:
> > This will prevent merging PRs created by github actions.
> 
> How so? If a PR is created, then that PR will still get a `github-pull-request` graph. Currently those PRs are getting both `github-push` and `github-pull-request` graphs, and there's a race condition over which task ends up getting displayed in the Github checks UI (the later task overwrites the earlier task in the UI)

I'll try to break down the different cases:
- for PRs created by collaborators, the pull-request graph is just fine.  We can get rid of any push graph here. (IMO they should push to a fork instead of the main repo, but that is not actually enforced)
- PRs created by humans that aren't collaborators are necessarily coming off of a fork.  They don't run ui-tests, they don't get a push graph.  No change there from this patch.
- PRs created by github actions is where we need to be careful.  These are on a branch on the main repo.  Their PR graph is incomplete, because it's not authored by a collaborator. We currently rely on the push graph in this case to satisfy merge conditions.  So we do need to keep that graph running here, that's why I suggested on the bug to add those branch name patterns both to the .tc.yml filter and to branch protections, although I'm not sure if we can say "github actions can push here but not human users".

ahal:
Ah ok, thanks for the explanation. And I missed your comment on the bug. I just replied over there.

.taskcluster.yml
```diff
@@ -147,7 +147,7 @@ tasks:
                   $if: >
                       tasks_for in ["action", "cron"]
                       || (isPullRequest && pullRequestAction in ["opened", "reopened", "synchronize"])
-                      || (tasks_for == "github-push" && head_ref[:10] != "refs/tags/" && short_head_ref != "staging.tmp" && short_head_ref != "trying.tmp" && short_head_ref[:8] != "mergify/")
+                      || (tasks_for == "github-push" && (short_head_ref == "main" || short_head_ref[:9] == "releases_"))
```
hneiva, line 150: I'm assuming "releases_*" branches are protected in this repo?

hneiva (approved):

hneiva (changes requested):
I'll leave as requested changes according to previous conversation

PR #5178, opened 2024-01-15
Bug 1874768 - Hide default browser onboarding card if user already set us as default browser

rocketsroger:
Examples if the onboarding card is the first card.  We'll add the privacy notice with the link in the first card.  (ONLY in the first card)

Screenshots

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1874768


fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageState.kt
```diff
@@ -5,27 +5,29 @@
...
 data class OnboardingPageState(
     @DrawableRes val imageRes: Int,
     val title: String,
     val description: String,
-    val linkTextState: LinkTextState? = null,
     val primaryButton: Action,
     val secondaryButton: Action? = null,
     val onRecordImpressionEvent: () -> Unit = {},
+    val onPrivacyPolicyClick: (String) -> Unit = {},
+    val privacyPolicyUrl: String = SupportUtils.getMozillaPageUrl(SupportUtils.MozillaPage.PRIVATE_NOTICE),
```
rahulsainani, line 30: I think we should revisit this approach. These shouldn't be available on each page. If we are separating description and "caption", we can add a nullable caption model which contains the String and `LinkTextState`. We should only update the mapper logic where the first page has the `caption`. Also, the UI doesn't need to check the index to display the link, just check if the caption with link is available and display it.

fenix/app/onboarding.fml.yaml
```diff
@@ -25,8 +25,7 @@ features:
               card-type: default-browser
               title: juno_onboarding_default_browser_title_nimbus_2
               ordering: 10
-              body: juno_onboarding_default_browser_description_nimbus_2
-              link-text: juno_onboarding_default_browser_description_link_text
```
rahulsainani, line 29: If we are going to remove this and only use the app string for caption, I think we should also remove this field from the OnboardingCardData type in this file.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -56,13 +56,16 @@ private const val IMAGE_HEIGHT_RATIO_SMALL = 0.28f
...
+ * @param pagePosition The current page position.
  * @param modifier The modifier to be applied to the Composable.
  * @param onDismiss Invoked when the user clicks the close button. This defaults to null. When null,
  * it doesn't show the close button.
  */
 @Composable
+@Suppress("LongMethod")
 fun OnboardingPage(
     pageState: OnboardingPageState,
+    pagePosition: Int? = null,
```
rahulsainani, line 68: This is a common component, this shouldn't care about the page position. See [this](https://github.com/mozilla-mobile/firefox-android/pull/5178/files#r1453242485) for alternate approach

fenix/app/onboarding.fml.yaml
rocketsroger: Agreed, I'll do that.  Thanks

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageState.kt
rocketsroger: (Thanks for the offline explanation) Agreed, it will make things cleaner.  I'll update the implementation.

rocketsroger:
> Can we update the privacy notice string in all instances from "Learn more in our privacy notice" to "Firefox privacy notice"? The whole string will be a link. This is because the "Learn more" string spoke directly to the body copy that proceeded it, while the new copy will be more general and apply to any card.

Sure I'll update the string.

> RE: The video shown above. We will still include the notification permission card for this audience, yes? So if they have already set Fx to default before first run, they will first see the Sign In card and then the Notification card. TY!

Yes, I didn't change the algorithm on which card to show.  I'll update the recording with a newer version emulator to show this.

ewachowiak (approved):
Changes look good to me.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingMapper.kt
```diff
@@ -4,15 +4,15 @@
...
-import org.mozilla.fenix.compose.LinkTextState
 import org.mozilla.fenix.nimbus.OnboardingCardData
 import org.mozilla.fenix.nimbus.OnboardingCardType
-import org.mozilla.fenix.settings.SupportUtils
 
 /**
  * Returns a list of all the required Nimbus 'cards' that have been converted to [OnboardingPageUiData].
  */
 internal fun Collection<OnboardingCardData>.toPageUiData(
+    privacyCaption: Caption?,
```
rahulsainani, line 14: This should be non nullable `Caption`

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingScreen.kt
```diff
@@ -65,16 +64,21 @@ fun OnboardingScreen(
...
     onSkipFirefoxWidgetClick: () -> Unit,
-    onFinish: (pageType: OnboardingPageUiData) -> Unit,
+    onFinish: (pageType: OnboardingPageUiData?) -> Unit,
     onImpression: (pageType: OnboardingPageUiData) -> Unit,
 ) {
+    /* do not continue if there's no onboarding pages to display */
+    if (pagesToDisplay.isEmpty()) {
+        onFinish(null)
+        return
+    }
```
rahulsainani, line 80: I think this should be in the Fragment. We create pagesToDisplay list there and we can check if it's empty `onCreate` and just finish there. 

A tip 💡  we don't need this in compose but if we did, we should be doing this kind of check in an [effect handler](https://developer.android.com/jetpack/compose/side-effects) in composable like a `LaunchedEffect`, without returning.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -118,8 +118,16 @@ fun OnboardingPage(
 
                 DescriptionText(
                     description = pageState.description,
-                    linkTextState = pageState.linkTextState,
                 )
+
+                Spacer(modifier = Modifier.height(16.dp))
+
+                pageState.privacyCaption?.apply {
```
rahulsainani, line 125: 🔍 `let` or an `if` would suit this case better. [apply](https://kotlinlang.org/docs/scope-functions.html#apply) is good for builder pattern objects for configuration

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
```diff
@@ -150,21 +158,13 @@ fun OnboardingPage(
...
-            textAlign = TextAlign.Center,
-            style = FirefoxTheme.typography.body2,
-        )
-    }
+    Text(
+        text = description,
+        color = FirefoxTheme.colors.textSecondary,
+        textAlign = TextAlign.Center,
+        style = FirefoxTheme.typography.body2,
+    )
```
rahulsainani, line 167: 🔍 Maybe we can just inline this in the upper composable now that the logic is gone

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingScreen.kt
```diff
@@ -268,10 +266,10 @@ private fun defaultPreviewPages() = listOf(
         type = OnboardingPageUiData.Type.DEFAULT_BROWSER,
         imageRes = R.drawable.ic_onboarding_welcome,
         title = stringResource(R.string.juno_onboarding_default_browser_title_nimbus_2),
-        description = stringResource(R.string.juno_onboarding_default_browser_description_nimbus_2),
-        linkText = stringResource(R.string.juno_onboarding_default_browser_description_link_text),
+        description = stringResource(R.string.juno_onboarding_default_browser_description_nimbus_3),
         primaryButtonLabel = stringResource(R.string.juno_onboarding_default_browser_positive_button),
         secondaryButtonLabel = stringResource(R.string.juno_onboarding_default_browser_negative_button),
+        privacyCaption = null,
```
rahulsainani, line 272: ❓ lets show the caption in the preview?

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPageState.kt
```diff
@@ -13,19 +13,19 @@ import org.mozilla.fenix.compose.LinkTextState
...
  */
 data class OnboardingPageState(
     @DrawableRes val imageRes: Int,
     val title: String,
     val description: String,
-    val linkTextState: LinkTextState? = null,
     val primaryButton: Action,
     val secondaryButton: Action? = null,
     val onRecordImpressionEvent: () -> Unit = {},
+    val privacyCaption: Caption?,
```
rahulsainani, line 28: We could set null as a default so we don't have to set it in `UpgradeOnboarding` and `NotificationDialogScreen`. Also, maybe we can use the same param position as now deleted `linkTextState` so lamda is the last param

fenix/app/src/test/java/org/mozilla/fenix/onboarding/view/OnboardingMapperTest.kt
```diff
@@ -18,29 +16,24 @@ class OnboardingMapperTest {
             imageRes = R.drawable.ic_onboarding_welcome,
             title = "default browser title",
             description = "default browser body with link text",
-            linkTextState = LinkTextState(
-                text = "link text",
-                url = SupportUtils.getMozillaPageUrl(SupportUtils.MozillaPage.PRIVATE_NOTICE),
-                onClick = stringLambda,
```
rahulsainani, line 24: The `stringLamda` property below is now used, let's delete that

fenix/app/src/androidTest/java/org/mozilla/fenix/onboarding/view/OnboardingMapperTest.kt
```diff
@@ -71,12 +75,30 @@ class OnboardingMapperTest {
         )
     }
 
+    @Test
+    fun showDefaultBrowserPageFalse_showNotificationFalse_showAddWidgetTrue_pagesToDisplay_returnsSortedListOfAllConvertedPages_withoutNotificationPage() {
```
rahulsainani, line 79: We should add a test that asserts that caption is actually present on the first page always, regardless of the combinations

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingScreen.kt
rocketsroger: Good idea, I'll move the check.  Thanks

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingPage.kt
rocketsroger: Sure, I'll use let.  Thanks

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingScreen.kt
rocketsroger: Sounds good.  I'll add to all of the cards so it's taken into consideration when we change things.  Thanks

fenix/app/src/test/java/org/mozilla/fenix/onboarding/view/OnboardingMapperTest.kt
rocketsroger: Good catch.  Thanks

fenix/app/src/androidTest/java/org/mozilla/fenix/onboarding/view/OnboardingMapperTest.kt
rocketsroger: Good idea, I'll add some tests to make that happen.

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingMapper.kt
rocketsroger: Sure, I'll make it non nullable

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingScreen.kt
```diff
@@ -65,14 +65,13 @@ fun OnboardingScreen(
...
     onSkipDefaultClick: () -> Unit,
-    onPrivacyPolicyClick: (url: String) -> Unit,
     onSignInButtonClick: () -> Unit,
     onSkipSignInClick: () -> Unit,
     onNotificationPermissionButtonClick: () -> Unit,
     onSkipNotificationClick: () -> Unit,
     onAddFirefoxWidgetClick: () -> Unit,
     onSkipFirefoxWidgetClick: () -> Unit,
-    onFinish: (pageType: OnboardingPageUiData) -> Unit,
+    onFinish: (pageType: OnboardingPageUiData?) -> Unit,
```
rahulsainani, line 74: This could remain  `onFinish: (pageType: OnboardingPageUiData) -> Unit` since it won't callback with null here

rahulsainani (approved):
Looks good! Thanks for making the changes! 🚀

fenix/app/src/main/java/org/mozilla/fenix/onboarding/view/OnboardingScreen.kt
rocketsroger: Make sense, however, if we change it back we'll have to have two different `onFinish` functions. I think keeping it with the nullable signature is simpler.

rahulsainani: > If we change it back we'll have to have two different onFinish functions. 

🤔 I doubt that, can you please double check? This wouldn't affect the onFinish in `OnboardingFragment`

rocketsroger: Ah I see, we're not passing in `onFinish()` directly, but as a lambda that calls `onFinish`.  Sure I'll update.  Thanks

PR #5204, opened 2024-01-18
Bug 1875294 - Record breadbcrumbs before crashing with UnsatsisfiedLinkError

bendk:
Report breadcrumbs to Sentry before crashing with `UnsatsisfiedLinkError`:

  - The files inside the app directory, to detect if the library is present or not.
  - The files inside the APK, to detect if a file inside the APK didn't get installed correctly.
  - The name of the installer package, so that we can know if the APK was installed from a 3rd-party source.
  - The path to the APK.  I don't really see this being useful, but it's needed to determine all the other info, so we might as well report it.

The idea here is that hopefully one of these will detect that something was off with their install.  For example, maybe they installed an APK for the wrong arch or maybe `libjnidispatch` was not extracted correctly.  If not, at least we can rule these possibilities out as the root cause.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1875294


fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -521,13 +524,61 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+        breadcrumbs.add("Files inside APK: $apkFileList")
+
+        for (breadcrumb in breadcrumbs) {
+            components.analytics.crashReporter.recordCrashBreadcrumb(
+                Breadcrumb(
+                    category = "Startup",
+                    message = breadcrumb,
+                    level = Breadcrumb.Level.INFO,
+                ),
+            )
```
bendk, line 580: These should show up on Sentry events, but not Soccoro, is there a way to get this data inside those reports too?

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -521,13 +524,61 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+                // This shouldn't happen, but if it does it's better to ignore the exception from
+                // the breadcrumb code and rethrow the initial exception.
+            }
+            throw e
+        }
+    }
 
-        RustLog.enable()
+    private fun reportUnsatisfiedLinkErrorBreadcrumbs() {
+        val breadcrumbs = mutableListOf<String>()
```
rocketsroger, line 548: nit: I suggest we rename this to `breadcrumbStrings`

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -521,13 +524,61 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+
+        val apkDirFileList = apkDir?.let {
+            it.walk()
+                .filter { it != apkDir }
+                .map { it.relativeTo(apkDir).toString() }
+                .joinToString(", ")
+        }
+        breadcrumbs.add("Files inside dir: $apkDirFileList")
+
+        val apkFileList = ZipInputStream(FileInputStream(apkPath)).use {
```
rocketsroger, line 568: Does this have privacy issue?

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
```diff
@@ -521,13 +524,61 @@ open class FenixApplication : LocaleAwareApplication(), Provider {
...
+            generateSequence { it.nextEntry }.joinToString(", ")
+        }
```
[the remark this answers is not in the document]
rocketsroger: A little slower when we're about to crash is acceptable in my opinion.  Is there a way for us to filter those files so we can turn the file list into a list of true/false? (x file exists, y file exist etc)

rocketsroger (commented):
Looks good.  Have a privacy concern question.

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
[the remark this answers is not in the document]
bendk: I tested it my emulator (Pixel 6, x86-64).

It would definitely be possible to transform the file list, but maybe it should more like a diff:  X files were present in the APK, but not on the device -- Y files were present on the device, but not the APK.

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
bendk: I'm not sure how to judge that, but it feels like technical data to me and I think that can be shared on Sentry.  IMO, the most sensitive would be the installer package name, maybe some users wouldn't want us to know how they installed the app?  I personally don't think that's an issue, but I'm no expert.  Who is a good person to ask about this.

bendk: The installer package name is also captured in the [startup metrics](https://github.com/mozilla-mobile/firefox-android/blob/da34eb224eb391b21dd2e7ff58bdf082a98fffc4/fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt#L847-L854), maybe this means it's okay to share in a sentry report.

rocketsroger: Can this crash only happen on fresh install? or can it happen at any time when user starts the app?

fenix/app/src/main/java/org/mozilla/fenix/FenixApplication.kt
rocketsroger: Unfortunately, this will require some work to support.  Currently breadcrumbs is only supported for Socorro reports for caught exception.  For uncaught exceptions, work is needed to pass these breadcrumbs to the crash reporting service. (Or persist on disk and cleared after reporting)

rocketsroger: This is the [issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1792898) tracking this work.

bendk:
I'm updating this one to change the output so that:
  - It only includes files inside the `lib` directory
  - The APK contents are diffed against the filesystem and the differences are displayed.

Here's what the breadcrumb messages look like on my machine (with a successful install):

```
APK: /data/app/~~nHVwCTpcQDCgPFNJx4gUWw==/org.mozilla.fenix.debug-4RPJsYeg6OpqG1Uw_LGxIg==/base.apk
Installer package name: null
Files only inside lib/ dir: <none>
Files only inside APK lib/ dir: <none>
Files inside both lib/ dirs: lib/x86_64/libsoftokn3.so, lib/x86_64/libxul.so, lib/x86_64/libjnidispatch.so, lib/x86_64/libmozglue.so, lib/x86_64/libgkcodecs.so, lib/x86_64/libnss3.so, lib/x86_64/libnssckbi.so, lib/x86_64/libmozavcodec.so, lib/x86_64/libipcclientcerts.so, lib/x86_64/libfreebl3.so, lib/x86_64/liblgpllibs.so, lib/x86_64/libplugin-container.so, lib/x86_64/libmozavutil.so, lib/x86_64/libsentry-android.so, lib/x86_64/libsentry.so, lib/x86_64/libmegazord.so
```

rocketsroger (approved):
🚢

bendk:
What's the next step for shipping this?  I thought it was adding the needs-landing tag, is there some other step I should take?

PR #5224, opened 2024-01-19
Bug 1874441 - Connecting Dropdowns to Translate Button

ohall-m:
This patch connects the translations to and from dropdowns to the translation button and sets default dropdown selection values.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1874441


fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsController.kt
```diff
@@ -67,12 +67,12 @@ class TranslationsController(
 
         if (defaultFromLanguage == null) {
             defaultFromLanguage = detectedLanguages?.documentLangTag
-            logger.info("Setting translating to use a default 'from' of $defaultFromLanguage.")
+            logger.info("Setting translating to use a default 'from'.")
```
ohall-m, line 70: Had thoughts that these values maybe shouldn't be logged, is it too specific?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -128,12 +171,14 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
                                     onTranslateButtonClick = {
                                         interactor.onTranslate(
                                             tabId = args.sessionId,
-                                            fromLanguage = null,
-                                            toLanguage = null,
+                                            fromLanguage = fromSelected?.code,
```
ohall-m, line 174: Filed [Bug 1875534](https://bugzilla.mozilla.org/show_bug.cgi?id=1875534) to follow-up on the condition when the from and to language are the same.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -71,10 +79,41 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
             ),
         )
 
+        // Signalling user intention to translate
+        browserStore.dispatch(TranslationsAction.TranslateExpectedAction(args.sessionId))
```
iorgamgabriel, line 83: I don't know if this is okay to dispatch an action from a fragment. In all the project this is happening from a "feature" or from a Middleware.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
iorgamgabriel: In what class do we do all these validations? This will not be the only one needed in this fragment. @gabrielluong

iorgamgabriel: If a language is selected, do we still show it in the drop-down menu? I asked in figma. https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7202:119619&mode=design#677742095

iorgamgabriel: In what class do we modify the state of the views? For example, if the translation is successful, this fragment should look different. The first dropdown should disappear, the title should be different, and so should the buttons. 
Where do we put this logic ?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: I'm having trouble finding a good place to put this or alternatives. I need the `tabId` for setting the error, so I need a tab related action for the `id`. Unfortunately, `InitAction` doesn't have the ID and some of the things I tried seem fragile. Let me know what ideas you have!

Tested changing [this middleware action](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt#39) to [AddTabAction](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt#177), this works fine for first user session, but the action isn't re-triggered when restoring from an app close. I can probably use [RestoreAction](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt#230) in addition to `AddTabAction`, but I think that could be potentially a lot of calls if the user had a bunch of tabs. [SelectTabAction](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt#205) sort of works on restore, but the initial first tab won't have data because it was never selected.

`TranslateExpectedAction` triggers now before we get here based on the translation engine thinking we should translate. This action that is firing from this Fragment is more of a guard to make sure the data is coming on the case where the user accesses via the browser menu. 

**Maybe the action should be triggered from the browser menu controller [there](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt#410)?** This is the case I'm guarding against. WDYT?

(I plan on `TranslateExpectedAction` correlating to showing our toolbar icon as well. (I asked in [Figma](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7246:207835&mode=design#677883739), if we want to show the toolbar if the user accessed via browser menu. If setup and toolbar need decoupled, then it will be just a new action either way.)

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: > In what class do we do all these validations?  
> In what class do we modify the state of the views?

These are the big question right now about the overall architecture this project should be heading towards. We should take some time to formally write out a plan or diagram because there is so much involved. It seems especially tricky where almost everything is happening on the GeckoView engine layer and reporting on the browser store, so we need a pattern that supports that data model. Maybe we should try and do a Figma Jam or something together with anyone interested? 

> In what class do we modify the state of the views?

My super naive thought is: can we just observe everything in the fragment and update based on that? For example, I had tested around some with having  the `translationError`  be observed and set on `TranslationDialogBottomSheet`. It showed the error as I expected and seemed to be managing what I expected. 

I imagine this type of logic might become more profound with all of the additional criteria, but in general, the `browserStore` observed values seems to correlate fairly strongly to the UI needs.

I have a sense there is a reason not too do what I'm thinking out loud about and use formal `States`, but don't really understand the why aspect and would like to know more.

> In what class do we do all these validations?  

Would it be acceptable to do validation in the fragment as well when it is view related? E.g., observe the `from` and `to`, if they are the same then disable the button. I suppose the logic might get overwhelming.

Mostly want to get a sense of why certain things aren't a good idea too so we can start focusing in on a pattern.

Also, worth mentioning, the goal of this bug is only to give us a few more pieces lined out to work with as a starting point, so I expect whatever lands here will be refactored or change as we decide on a pattern.

MatthewTighe: This is a preliminary comment to more architectural advice (if desired), because I don't have a ton of time today but would love to talk more about all this. Have only done a cursory reading of some of the material so far. Here are some suggestions I would make to answer the following questions and in general for this feature to keep it in line with our Redux architectural goals:

The workflow for Redux is:
1. Dispatch an action
2. Handle side-effects and logic in middleware
3. Update state in a reducer
4. Observe updates to that state in the view and update accordingly

So, to answer the above:

> In what class do we do all these validations?

These can happen in the reducer (if they are pure computations) or in the middleware (if they are not, for example they require accessing disk state through `Settings`). In the case of middleware, this may result in the original action being stopped from entering the rest of reducer chain, though that does add some complexity.

> In what class do we modify the state of the views?

The views should modify themselves as they observe changes to `State`. I would recommend having State that isn't specific to a view, for example it would contain a `translationError` instead of a `translationErrorString`. 

> I have a sense there is a reason not too do what I'm thinking out loud about and use formal States, but don't really understand the why aspect and would like to know more.

I think I'd need a bit more detail here to comment either way. 

> E.g., observe the from and to, if they are the same then disable the button.

This is an interesting case. I'll think about this before I can answer definitively, but some questions we can ask about it:
1. Do we want to be able to unit test the fact the button is disabled? If so, does it make sense to keep it in `State`, or will we be computing it every time we want to inspect it? Could we expose the logic through something like an extension function and test that easily?

Finally, I see that the interactor/controller have landed. My suggestion would be to remove those completely, along with the `TranslateUseCase`. Instead, I would dispatch Actions directly from the callback passed to the view (like `onTranslateButtonClick` referenced in the diff of this comment). This is because the current workflow is: callback -> interactor -> controller -> usecase -> dispatch action. I don't see any reason why we can't remove the interactor -> controller -> usecase middlemen here.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationEngineState.kt
```diff
@@ -19,3 +19,81 @@ data class TranslationEngineState(
...
+ * @param candidateLanguages The language options available to select as a final initial value.
+ * @return The best determined "to" language or null if a determination cannot be made.
+ */
+fun TranslationEngineState.initialToLanguage(candidateLanguages: List<Language>?): Language? {
+    return determineInitialDropdownValue(
+        translationEngineState = this,
+        isFromLanguage = false,
+        candidateLanguages = candidateLanguages,
+    )
+}
```
MatthewTighe, line 36: IMO it would be more clear to handle each use case in each function instead of combining them in determineInitialDropdownValue
```suggestion
fun TranslationEngineState.initialToLanguage(candidateLanguages: List<Language>?): Language? = 
    candidateLanguages?.find { 
        it.code == (requestedTranslationPair?.toLanguage ?: detectedLanguages?.userPreferredLangTag) 
    }
```

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationEngineState.kt
```diff
@@ -19,3 +19,81 @@ data class TranslationEngineState(
...
+    if (code == null) {
+        code = if (isFromLanguage) {
+            translationEngineState.detectedLanguages?.documentLangTag
+        } else {
+            translationEngineState.detectedLanguages?.userPreferredLangTag
+        }
+    }
+
+    // Process code to a known language in our candidate language lists:
+    fun languageCodeToLanguage(bcpCode: String?, list: List<Language>?): Language? {
```
MatthewTighe, line 92: Just FYI, nesting functions inside each other is not a typical Kotlin structure. There are some examples of it in the codebase but I would ignore those 😅

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -355,15 +396,18 @@ private fun TranslationsDropdown(
 
         Spacer(modifier = Modifier.height(4.dp))
 
+        var initialValue = stringResource(R.string.translations_bottom_sheet_default_dropdown_selection)
+        if (initiallySelected?.localizedDisplayName != null) {
+            initialValue = initiallySelected.localizedDisplayName!!
+        }
```
MatthewTighe, line 402: You can get around the compiler's inability to avoid a force unwrap (`!!`) here by using a receiver block instead: 
```suggestion
        initiallySelected?.localizedDisplayName?.let {
            initialValue = it
        }
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -71,10 +79,41 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
...
+                    state ->
+                state.findTab(args.sessionId)?.translationsState?.translationEngineState
+                    ?.initialFromLanguage(translateFromLanguages)
+            }.value
+
+            val initialToSelected = browserStore.observeAsComposableState {
+                    state ->
+                state.findTab(args.sessionId)?.translationsState?.translationEngineState
+                    ?.initialToLanguage(translateToLanguages)
+            }.value
```
MatthewTighe, line 106: I would consider combining these into one `translationsState` observable and getting/computing the subproperties you need from there.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
MatthewTighe: > Filed Bug 1875534 to follow-up on the condition when the from and to language are the same.
> In what class do we do all these validations? This will not be the only one needed in this fragment.

IMO this is something that the view can handle directly, and is much easier in Compose. Something like:

```
Button(isEnabled = state.fromLanguage != state.ToLanguage)
```

is not necessarily about stressing unit tests for. 

> In what class do we modify the state of the views? For example, if the translation is successful, this fragment should look different. The first dropdown should disappear, the title should be different, and so should the buttons.

The views should control their own state in response to observations of feature state. In all of these cases, I would not suggest introducing another class to hold view state.

MatthewTighe (commented):
Bit more detail now, but still short on time. Will cycle back again tomorrow. Overall though, I think much of the design here is good and in the right direction.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationEngineState.kt
ohall-m: Updated, ty! Thanks for suggesting the idiomatic kotlin format too

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationEngineState.kt
ohall-m: Thanks for mentioning, good to know, updated!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
ohall-m: Ty, updated

ohall-m:
Thanks for taking a look and the great discussion! I'll focus on updating the architecture here in this patch or elsewhere today.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -126,14 +164,24 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
...
                                     onTranslateButtonClick = {
-                                        interactor.onTranslate(
-                                            tabId = args.sessionId,
-                                            fromLanguage = null,
-                                            toLanguage = null,
-                                            null,
-                                        )
+                                        fromSelected?.code?.let { fromLanguage ->
+                                            toSelected?.code?.let { toLanguage ->
+                                                TranslationsAction.TranslateAction(
```
ohall-m, line 169: Will follow-up on null case in [bug 1875534](https://bugzilla.mozilla.org/show_bug.cgi?id=1875534)

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: Any suggestions for this dispatch or is it okay for the time being? I can file a follow-up to look at this deeper too.

MatthewTighe: I can take a deeper look at this later today if you'd like, but this comment:

> Tested changing [this middleware action](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt#39) to [AddTabAction](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt#177), this works fine for first user session, but the action isn't re-triggered when restoring from an app close. I can probably use [RestoreAction](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt#230) in addition to AddTabAction, but I think that could be potentially a lot of calls if the user had a bunch of tabs. [SelectTabAction](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt#205) sort of works on restore, but the initial first tab won't have data because it was never selected.

has me wondering if we can react to several actions (`SelectTab`, `RestoreTab` would maybe cover it?) in the middleware. My current understanding of the workflow in the diff is that we would wait until a tab is active and the dialog is pulled up to start the work of determining which translations are available for the dropdown. Is that correct? Do we want to start that work earlier so we don't have to potentially wait for it when the dialog becomes visible?

MatthewTighe: But overall I think it would also be fine as is and we could revisit in a later bug if that's easiest for now

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -127,19 +135,29 @@ fun TranslationsDialogBottomSheet(
 
         Spacer(modifier = Modifier.height(14.dp))
 
-        if (translationError !is TranslationError.CouldNotLoadLanguagesError) {
+        if (translationError !is TranslationError.CouldNotLoadLanguagesError &&
+            translateFromLanguages != null && translateToLanguages != null
```
MatthewTighe, line 139: I'm wondering whether we should introduce a new `Error` type that can be used in `GeckoTranslateSessionDelegate::onTranslationStateChange` to avoid some of this nullability. Are there cases where the first part of this conditional is true and the latter parts are false?

MatthewTighe: Or perhaps we need `TranslationEngineState` to be an `Option` type? Like:

```
sealed class TranslationEngineState {
    data class Success(val candidateLanguages, from, to, etc) : TranslationEngineState
    data class Error(val error)  
}
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: > has me wondering if we can react to several actions (SelectTab, RestoreTab would maybe cover it?) 

I think there is some combination that would work. My main concern is fragility or missing some obscure case where neither action occurs. Also, I kinda wonder if `RestoreTab` could call this on a bunch of tabs and have a noticeable impact. I would need to dig a little deeper to feel confident about getting the right combination.

> My current understanding of the workflow in the diff is that we would wait until a tab is active and the dialog is pulled up to start the work of determining which translations are available for the dropdown. Is that correct?

Right, that is the main impact as far as the patch goes. There are some bigger pieces that need to harmonize too.

More background:
Two categories of translations:
* `Expected` - we will show a translate icon on the toolbar
   * Engine signals `TranslateExpectedAction`, which will probably correlate to showing icon on toolbar in [bug 1875817](https://bugzilla.mozilla.org/show_bug.cgi?id=1875817). (Originates [from delegate](https://searchfox.org/mozilla-mobile/rev/99576d6e69eac8ee1bf6c8a654d5fa96a6f6bf84/firefox-android/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslateSessionDelegate.kt#26-28) -> [dispatches here ](https://searchfox.org/mozilla-mobile/rev/99576d6e69eac8ee1bf6c8a654d5fa96a6f6bf84/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/EngineObserver.kt#473).)
* `Unexpected` - no translate icon on toolbar, user accessed via browser menu on a site in their preferred language or an unsupported language
   * Criteria to show icon on toolbar as well in this case will happen if the user fully performs a translation from that entry point

I'm thinking `TranslateExpectedAction` will have to become `TranslateFetchLanguagesAction` or `TranslateStateInit` or use one of those global actions based on new information in the [Figma](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7246:207835&mode=design#677883739).

I completely forgot I already filed a bug to research this further when I wrote the middleware with [bug 1873672](https://bugzilla.mozilla.org/show_bug.cgi?id=1873672), so maybe we can brainstorm some ideas here and execute over there.

>Do we want to start that work earlier so we don't have to potentially wait for it when the dialog becomes visible?

Ideally, would like to get the languages early; however it is not an especially long operation.  (IIRC, first time ever is engine -> GV -> network call, but then later calls will be engine -> GV -> cache.) If it is okay to dispatch initialization here, I think it could conceptually simplify things as well.

Tangent:
We'll need to do similar initialization for [page settings](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7202-99696&mode=design&t=w3GVOGCG9Ytv9Zmj-0) and global settings [1](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7231-206407&mode=design&t=w3GVOGCG9Ytv9Zmj-0), [2](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7231-206408&mode=design&t=w3GVOGCG9Ytv9Zmj-0), [3](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7231-206406&mode=design&t=w3GVOGCG9Ytv9Zmj-0) , so maybe need to consider those situations as well and think of a plan.

1. Should the main translate page, page settings, and some very deep global options be initialized together on the same action, even if users are progressively less likely to go into the deeper settings? Or separate actions that occur as that data is needed?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsController.kt
ohall-m: (Forgot use cases were used in reference browser, I assume they should stay for that reason.) #5279

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
ohall-m: > translateFromLanguages != null && translateToLanguages != null

I think the only scenario this will happen is if fetching the languages is late, but there is no error yet, then this can happen. 

> initialFrom and intitialTo

Very often we can have one and not the other:
* `intitialFrom` - we detected a compatible page language
   * null if we can't detect page language
   * null if page language is unsupported (have another [error case ](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7202-119890&mode=design&t=qr7VxyeKAmJhSnfM-0)that'll need to be integrated) 
* `initialTo` - user's preferred language
   * null if user's preferred language isn't supported
   * maybe can also be null for other reasons, like we can't determine it or the language code doesn't parse


> I'm wondering whether we should introduce a new Error type that can be used in GeckoTranslateSessionDelegate::onTranslationStateChange to avoid some of this nullability.

I could see sending an action that does something from `onTranslationStateChange`.

Mentioning directly some background here, because the naming was really tricky on that state and a lot happens there. The `TranslationEngineState` is fully controlled by the browser engine independently of AC and Fenix and we are just reporting back details.

`TranslationEngineState` generally changes many times:
* Before page load (I think, very early at any point) - lots of expected nulls
* After page load - some expected nulls
* Translate - almost everything full
* Translate Restore - looks like after page load
* Restoring page from a full shutdown

The [GeckoView code](https://searchfox.org/mozilla-central/rev/24f9846b7d8ea0819edee7aafc51a39753f1619a/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#1173,1184) kinda shows the relationships happening in that delegate nicely. The state comes in as a [message from toolkit](https://searchfox.org/mozilla-central/rev/24f9846b7d8ea0819edee7aafc51a39753f1619a/mobile/android/modules/geckoview/GeckoViewTranslations.sys.mjs#27,36,106-123) that performs the translation. 

One reason it is especially important is this translations engine can make decisions independent of Fenix based on some prefs we set. For example, I can set always translate French and right after page load, the page will translate without me touching Fenix UI. Fenix UI needs to reflect this browser state; however.

Other tangent:

I don't think we need that specific [TranslationEngineState.error](https://searchfox.org/mozilla-mobile/rev/99576d6e69eac8ee1bf6c8a654d5fa96a6f6bf84/firefox-android/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationEngineState.kt#11) the engine reports except for debugging. It is the error the browser reported and, IIRC, the error Desktop will visually show. `translationError` [here](https://searchfox.org/mozilla-mobile/rev/99576d6e69eac8ee1bf6c8a654d5fa96a6f6bf84/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt#33) is the one important to Fenix.

>Or perhaps we need TranslationEngineState to be an Option type? 

Sure, let me know what you prefer based on the addition info! I don't think I have enough background to evaluate what should be done here.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: >If a language is selected, do we still show it in the drop-down menu? 

Closing the loop here, answer was yes the language should be in the drop-down menu, but the design changed to included a checkmark next to the selected item. Filed new UI bug - [Bug 1876140](https://bugzilla.mozilla.org/show_bug.cgi?id=1876140)

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: Much neater, updated!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
MatthewTighe: Thanks for all the info! The `Option` type isn't seeming worth it based on all of that. Mostly just asking questions out loud to try and understand the domain better

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
MatthewTighe: I lost my progress on a comment here by clicking a link 😩 hopefully this captures it all.

> Also, I kinda wonder if RestoreTab could call this on a bunch of tabs and have a noticeable impact.

We can always filter down to the selected tab (according to BrowserState, not the RestoreAction) in the middleware and launch the `TranslationsAction.TranslateExpectedAction` accordingly. 

Though, I am also seeing that `EngineObserver::onTranslateExpected` dispatches this action. Are we just trying to account for the unexpected cases here?

My big concern here is that we end up in one of these scenarios:
1. UI (toolbar button, dropdown, etc) is shown but isn't interact-able because we haven't finished fetching data. 
2. UI is not shown and pops in once state is available for observations.

Both cases can be avoided by fetching data eagerly enough, so the risk of a race condition is highest if we wait until the dialog is shown. I am not seeing placeholder UI in the Figma for things like the dropdown either, so I imagine we have to have the data available pretty immediately after a patch load.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -46,6 +52,7 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
     private var behavior: BottomSheetBehavior<View>? = null
     private val args by navArgs<TranslationsDialogFragmentArgs>()
     private lateinit var interactor: TranslationsInteractor
+    private lateinit var browserStore: BrowserStore
```
MatthewTighe, line 55: nit:
```suggestion
    private val browserStore: BrowserStore by lazy { requireComponents.core.store }
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
MatthewTighe: Optionally, this is so deeply nested that I would consider defining a lambda at (or before, even) the `setContent` level. This could also wait until the null case bug though

MatthewTighe (approved):
This definitely seems to work as far as I can tell from a limited manual test. I don't want to block you by getting too bogged down in details, so I'll add an approval. I think we can figure out exactly how to handle things like action dispatches as follow-ups, especially as manual testing as we iterate informs our understanding. Happy to continue the conversations we're having here, in later patches, or offline though as you prefer

iorgamgabriel (approved):

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: 😩 Ohh, no, sorry you lost your original message! One nice thing about when we go to Phabricator is it saves draft comments and eagerly saves.
 
> We can always filter down to the selected tab (according to BrowserState, not the RestoreAction) in the middleware and launch the TranslationsAction.TranslateExpectedAction accordingly.

Ohh, that could be interesting and solve that issue!

> Are we just trying to account for the unexpected cases here?

Right, for the expected use-cases we should have already pulled the data almost right after page load. 

Since that is what I'm guarding, which is the browser menu use case, I wonder if I could launch the action [here](https://searchfox.org/mozilla-mobile/rev/07a850b0ef36fd7cd7aa339f6b14f40a91b6b4b2/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMenuController.kt#410). Not super early, but maybe enough time?

Something I hadn't thought of is, I wonder if I could do a check if this value is empty and do it `onTranslationStateChange`. (It occurs fairly often, but I could check if the stored value is empty and fetch as needed. The languages stay the same between navigation as well.)

Thanks for the scenarios to watch for too and all the brainstorming ideas! I'll give this the attention it deserves in [bug 1873672](https://bugzilla.mozilla.org/show_bug.cgi?id=1873672). I like the idea of `AddTabAction` + filtered to selected tab `RestoreTab`.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: Thanks! Added as a comment to make this refactor in the null case [bug 1875534](https://bugzilla.mozilla.org/show_bug.cgi?id=1875534).

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
ohall-m: Thanks for asking questions and brainstorming. It is helping me understand my own assumptions better for sure!

ohall-m:
Thanks everyone for the reviews and discussion! I think this patch will work as a basis for future work and a jumping off point to figure out the rest of the details. I'll land after I can rebase when #5279 lands. 

> Also realize that I hopped in

Thank you for hopping in! Really glad you are interested!

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2337,6 +2337,8 @@
     <string name="translations_bottom_sheet_translating_in_progress">Translating</string>
     <!-- Button content description (not visible, for screen readers etc.) for the translations dialog translate button that indicates a translation is currently in progress. -->
     <string name="translations_bottom_sheet_translating_in_progress_content_description">Translating in Progress</string>
+    <!-- Default dropdown option when initially selecting a language from the translations dialog language selection dropdown.-->
```
gabrielluong, line 2340:
```suggestion
    <!-- Default dropdown option when initially selecting a language from the translations dialog language selection dropdown. -->
```

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/state/TranslationEngineStateTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        )
+
+        val translationEngineState = TranslationEngineState(
+            detectedLanguages = detectedLanguages,
+            error = null,
+            isEngineReady = true,
+            requestedTranslationPair = translationPair,
+        )
+
+        assert(translationEngineState.initialToLanguage(mockCandidateLanguages) == null)
```
gabrielluong, line 105:
```suggestion
        assert(translationEngineState.initialToLanguage(mockCandidateLanguages) == null)
```
You could also simply this to use `assertNull`

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/state/TranslationEngineStateTest.kt
```diff
@@ -0,0 +1,121 @@
...
+        )
+
+        val translationEngineState = TranslationEngineState(
+            detectedLanguages = detectedLanguages,
+            error = null,
+            isEngineReady = true,
+            requestedTranslationPair = null,
+        )
+
+        assert(translationEngineState.initialToLanguage(mockCandidateLanguages)?.code == "es")
```
gabrielluong, line 37:
```suggestion
        assert(translationEngineState.initialToLanguage(mockCandidateLanguages)?.code == "es")
```
Perhaps use `assertEquals`

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/state/TranslationEngineStateTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import mozilla.components.concept.engine.translate.initialToLanguage
+import org.junit.Test
+
+val mockCandidateLanguages = listOf(
+    Language("de", "German"),
+    Language("fr", "French"),
+    Language("en", "English"),
+    Language("es", "Spanish"),
+)
+class TranslationEngineStateTest {
```
gabrielluong, line 21:
```suggestion
private val mockCandidateLanguages = listOf(
    Language("de", "German"),
    Language("fr", "French"),
    Language("en", "English"),
    Language("es", "Spanish"),
)

class TranslationEngineStateTest {
```

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/state/TranslationEngineStateTest.kt
```diff
@@ -0,0 +1,121 @@
...
+    @Test
+    fun `initialToLanguage and initialFromLanguage works as expected on the not translated case`() {
```
gabrielluong, line 23: Could you use the GIVEN WHEN THEN format for the test descriptions?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsBottomSheet.kt
```diff
@@ -127,25 +128,36 @@ internal fun TranslationsOptionsAnimation(
...
 @Composable
+@Suppress("LongParameterList")
 internal fun TranslationsDialog(
     learnMoreUrl: String,
     showFirstTimeTranslation: Boolean,
     translationError: TranslationError? = null,
+    translateFromLanguages: List<Language>?,
+    translateToLanguages: List<Language>?,
+    initialFrom: Language? = null,
+    initialTo: Language? = null,
```
gabrielluong, line 139: Could you reorder the param here so that the mandatory params come before the optional ones? We generally follow https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-component-api-guidelines.md#parameters-order, but the one exception is that we keep the callback lambdas along with the (optional) trailing @Composable lambda at the end of the param order.

gabrielluong (approved):

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/state/TranslationEngineStateTest.kt
MatthewTighe: Oh this is a good catch. `assert` won't actually fail the tests IIRC, since it isn't a JUnit API. You'll want to use `assertEquals`, `assertTrue`, `assertNull` etc imported from `org.junit`

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
MatthewTighe: > Since that is what I'm guarding, which is the browser menu use case, I wonder if I could launch the action here. Not super early, but maybe enough time?

Conceptually I think this spot makes more sense, but I'm not sure we'd get measurably different results. I think it would work to add your dispatch here for now. From a long-term architectural viewpoint, I imagine we would eventually want this to look like

```
ToolbarMenu.Item.Translate -> {
    // now this action can be re-used by the translation middleware _and_ a navigation middleware
    store.dispatch(ToolbarMenuAction.TranslateClicked)
}
```

> Something I hadn't thought of is, I wonder if I could do a check if this value is empty and do it onTranslationStateChange.

Which value do you mean? Not totally following here

ohall-m: Nice, I think a `TranslateClicked` would work well! Thanks for the idea.

> Which value do you mean? Not totally following here

I totally meant to remove that thought in an edit! Wasn't thinking clearly first thing in the morning. 😅

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
ohall-m: Updated, thanks!

fenix/app/src/main/res/values/strings.xml
ohall-m: Updated, ty

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/state/TranslationEngineStateTest.kt
ohall-m: Updated

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsBottomSheet.kt
ohall-m: Ty, updated

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/state/TranslationEngineStateTest.kt
ohall-m: Updated, ty

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
mcarare: I think the nullability can be handled y the button state. The translate button should not be enabled when one or both languages have no selection so it should not even end up being handled here (`onTranslateButtonClick`). That would allow removing nesting and having this whole block as:

```
                                    onTranslateButtonClick = {
                                        browserStore.dispatch(
                                            TranslationsAction.TranslateAction(
                                                tabId = args.sessionId,
                                                fromLanguage = fromSelected?.code!!,
                                                toLanguage = toSelected?.code!!,
                                                options = null,
                                            ),
                                        )
                                    },
```
Any thoughts against this? This would enforce the assumption no null values would end up being used in this lambda, and we do want to catch cases where this "contract" is broken.

PR #5239, opened 2024-01-22
Bug 1832147 - Keep the profiler presets in sync with their desktop counterparts

canova:
This PR is part one of [Bug 1832147](https://bugzilla.mozilla.org/show_bug.cgi?id=1832147).

This PR updates the profiler presets to reflect the current presets in the [desktop side](https://searchfox.org/mozilla-central/rev/c130c69b7b863d5e28ab9524b65c27c7a9507c48/devtools/client/performance-new/shared/background.jsm.js) and adds a comment at the top of the file to make sure that they are in sync all the time. It looks like it's easy to forget these two sides when we are updating the presets, so a comment on each side would make this clearer.

The second part is going to be another patch on the mozilla-central side that adds a comment to ensure the presets are in sync.

https://bugzilla.mozilla.org/show_bug.cgi?id=1571089
https://bugzilla.mozilla.org/show_bug.cgi?id=1832147
https://bugzilla.mozilla.org/show_bug.cgi?id=1571089
https://bugzilla.mozilla.org/show_bug.cgi?id=1832147
https://bugzilla.mozilla.org/show_bug.cgi?id=1571089


canova:
@jonalmeida hey, I can't add people as reviewers as I don't have the permission. Could you maybe look or redirect it to someone else please? :) This PR should be small enough and it syncs the presets with the desktop presets.

PR #5242, opened 2024-01-22
Bug 1856979 - Translations UI Post-Translate Popup

iorgamgabriel:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogState.kt
```diff
@@ -0,0 +1,70 @@
...
+    val translateFromLanguages: List<Language>,
+    val translateToLanguages: List<Language>,
+) : State
+
+/**
+ * The current state of the translation task.
+ */
+sealed class TranslationTaskState {
+    /**
+     * If translation task is in progress.
```
Mukuljangir372, line 31: Do you think we need these comments?

ohall-m:
I filed [Bug 1878091](https://bugzilla.mozilla.org/show_bug.cgi?id=1878091) to update syncing the state with the browser engine. I have a few things I don't want to lose the thread on first, but I'll try to prioritize it next week. (I don't consider it blocking this bug, but would improve manual testing this patch.)

iorgamgabriel:
1.This is the first flow when the user tries to translate a website for the first time.
Uploading screen-20240205-162322.mp4…

iorgamgabriel:
2.If the first translation is successful when the user enters again in the bottom sheet: 
- The text "For your privacy, translations never leave your device. New languages and improvements are coming soon! Learn more" should disappear, and the title "Try private translations in Firefox" should become "Page translated from "English"(for example) to "French"(for example).
- "Translate from" dropdown should disappear. 
- "Not Now" negative button should become "Show Original".
- The "Translate" button should be disabled until the user selects a different language from the "Translate to" dropdown.
https://github.com/mozilla-mobile/firefox-android/assets/8118371/93dd3906-628c-4c0d-93a9-5c68cff4bbf9

iorgamgabriel:
3.If the user presses "Show original" button (restore). The translation icon should remain the same( for example white).
-when the user enters again in the bottom sheet. He should see the title "Translate this page?" and both dropdowns with the second one ("Translate to") unselected and the "Translate button disabled.

https://github.com/mozilla-mobile/firefox-android/assets/8118371/fa0b136c-9452-4823-9c45-ad2970ac38dc

iorgamgabriel:
4.If a error at translation appears this flow should appear (https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?node-id=7246%3A208068&mode=dev) 
If the error type is TranslationError.CouldNotTranslateError, TranslationError.CouldNotLoadLanguagesError or TranslationError.LanguageNotSupportedError

iorgamgabriel:
I added a TranslationsDialogStore with TranslationsDialogState, TranslationsDialogMiddleware ,translationsDialogStateReducer and TranslationsDialogFeature.
- In TranslationsDialogFeature I observe the changes at translationsState from the BrowserStore and I dispatch actions on translationsDialogStore based on the changes.
- In TranslationsDialogMiddleware I only send TranslationsActions to the BrowserStore.
- In TranslationsDialogStateReducer I modify the TranslationsDialogState based on the actions from the TranslationsDialogFeature.
- The snackbar, the translation icon, the colour change, and the content description. I sent them with setFragmentResult in BrowserFragment.

ohall-m:
Can you re-base this patch? This patch that just landed, https://github.com/mozilla-mobile/firefox-android/pull/5431, might offer some insight during testing this patch. I'll test some tomorrow and take a deeper look at the code.

ohall-m (commented):
Thanks for all the work to figure out how to manage the translations states!

I looked at this patch more from a "what" and feature perspective. It does a great job of managing the complicated states and looks to me to be a good foundation for additional UI state refinements.

I also tested using `about:config` `browser.translations.alwaysTranslateLanguages` = es, which automatically translates Spanish language pages to the user default. The correct translate state appeared.

I think @gabrielluong and @MatthewTighe and @mcarare could say more from an architecture and "how" perspective.

I'll also review again tomorrow.

ohall-m:
>This pull request has conflicts when rebasing. Could you fix it @iorgamgabriel? 🙏

Thought this might be a good place to share:

 If you notice languages loading slower after switching the initialization call over to:
```
TranslationsAction.OperationRequestedAction(
  tabId = args.sessionId,
  operation = TranslationOperation.FETCH_SUPPORTED_LANGUAGES,
),
)
```

The slowness is because #5389 decoupled fetching languages from a different event. (In certain cases, they were pre-loaded before we arrived on the Fragment.) Still looking for the best time to call this in bug [1875817](https://bugzilla.mozilla.org/show_bug.cgi?id=1875817). 

The change is noticeable on main, so I'm assuming it might be noticeable on this patch when it changes over, just mentioning it for others reviewing this patch.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -211,34 +218,86 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
+                }
+            }
+        }
     }
 
     private fun initTranslationsAction(context: Context) {
         if (!context.settings().enableTranslations) {
             return
         }
+        if (translationsAction == null) {
```
MatthewTighe, line 261: Will this button always be available as long as translations are enabled, even if the page is detected to be in the users default language?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -67,45 +67,29 @@ private val ICON_SIZE = 24.dp
...
+ * @param onPositiveButtonClicked Invoked when the user clicks on the positive button.
+ * @param onNegativeButtonClicked Invoked when the user clicks on the negative button.
  * @param onFromSelected Invoked when the user selects an item on the from dropdown.
  * @param onToSelected Invoked when the user selects an item on the to dropdown.
  */
-@Composable
 @Suppress("LongParameterList")
+@Composable
 fun TranslationsDialogBottomSheet(
+    translationsDialogState: TranslationsDialogState,
```
MatthewTighe, line 83: Typically, I think we try to have Composables avoid using `State` parameters directly. I'm not sure I'm up on the latest of that discussion, @MozillaNoah may be able to comment more.

If we do keep this, can we compute the `translatedPageTitle` directly from the state?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -211,4 +174,165 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
...
+        }
+
+        val isTranslated = translationsDialogStore.observeAsComposableState { state ->
+            state.isTranslated
+        }.value
+
+        setTranslatedPageTitle(fromSelected = fromSelected, toSelected = toSelected)
+
+        TranslationsDialog(
+            translationsDialogState = TranslationsDialogState(
```
MatthewTighe, line 236: Do we need all these observations just to recreate a `TranslationsDialogState`, or can we just observe the `translationsDialogStore.state` directly and pass that down?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,219 @@
...
+    val isInProgress: Boolean = false,
+    val positiveButtonType: PositiveButtonType? = null,
+    val error: TranslationError? = null,
+    val dismissDialogState: DismissDialogState? = null,
+    val showFirstTime: Boolean = false,
+    val initialFrom: Language? = null,
+    val initialTo: Language? = null,
+    val fromLanguages: List<Language>? = null,
+    val toLanguages: List<Language>? = null,
+) : State
```
MatthewTighe, line 57: I would suggest considering whether this can be wrapped into the existing `TranslationEngineState` that exists in the browser store, and filtering in the `BrowserFragment` based on the currently selected tab.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -211,34 +218,86 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
+                            FenixSnackbar.make(
+                                view = binding.dynamicSnackbarContainer,
+                                duration = Snackbar.LENGTH_SHORT,
+                                isDisplayedWithBrowserToolbar = true,
+                            )
+                                .setText(requireContext().getString(R.string.translation_in_progress_snackbar))
+                                .show()
+                        }
+
+                        val updateContentDescriptionSelected = result.getString(
```
MatthewTighe, line 242: Moving the `State` added in this patch into the `BrowserStore` should allow us to access the state from the Store instead of needing to communicate it between fragments using a bundle.

MatthewTighe (commented):
Agreed that this would be an easier review if it were split into smaller pieces. We can always land multiple patches for a single bug too!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFeature.kt
```diff
@@ -0,0 +1,134 @@
...
+                        translationsDialogStore.dispatch(
+                            TranslationsDialogAction.UpdateFromSelectedLanguage(
+                                fromSelected,
+                            ),
+                        )
+                    }
+
+                    val toSelected =
+                        translationsState.translationEngineState?.initialToLanguage(
+                            translationsState.supportedLanguages?.fromLanguages,
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2322,6 +2326,10 @@
     <!-- Translation request dialog -->
     <!-- Title for the translation dialog that allows a user to translate the webpage. -->
     <string name="translations_bottom_sheet_title">Translate this page?</string>
+    <!-- Title for the translation dialog after a translation was done with success.
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2334,6 +2342,8 @@
     <string name="translations_bottom_sheet_translate_to">Translate to</string>
     <!-- Button text on the translations dialog to dismiss the dialog and return to the browser. -->
     <string name="translations_bottom_sheet_negative_button">Not now</string>
+    <!-- Button text on the translations dialog to restore the site language to its initial value. -->
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
iorgamgabriel: I don't know about this. The request was to show the translation toolbar button if the settings were enabled.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
iorgamgabriel: Done . I added translatedPageTitle in TranslationsDialogState.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsBottomSheet.kt
```diff
@@ -127,35 +126,27 @@ internal fun TranslationsOptionsAnimation(
     }
 }
 
-@Composable
 @Suppress("LongParameterList")
+@Composable
 internal fun TranslationsDialog(
+    translationsDialogState: TranslationsDialogState,
+    translatedPageTitle: String? = null,
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -211,34 +218,86 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
+                        R.drawable.mozac_ic_translate_24,
+                    )!!.apply {
+                        setTint(
+                            ContextCompat.getColor(
+                                context,
+                                R.color.fx_mobile_text_color_primary,
+                            ),
+                        )
+                    },
+                    imageSelected = AppCompatResources.getDrawable(
```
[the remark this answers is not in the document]
iorgamgabriel: I don't know until when we should keep the violet colour and the content description for the icon. I was thinking that if the user opens the bottom sheet again, a new translation process will happen. The icon should not be violet. 
If the user makes a translation with success, the icon becomes violet, then opens the bottom sheet again, then closes the bottom sheet without making a translation. What colour should the icon have? I think black.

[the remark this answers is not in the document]
ohall-m: I can definitely see that perspective! I asked in the Figma to double check, we can follow-up if we need to whenever we have more information from them. 

>If the user makes a translation with success, the icon becomes violet, then opens the bottom sheet again, then closes the bottom sheet without making a translation. What colour should the icon have? I think black.

I think the answer depends on if the intention of the violet color is "hey, something new happened and changed, take a look!" or "FYI, the page is in a translated state". The design team can give us clarity on what should happen.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
ohall-m: I have some more information overall about the toolbar icon:
* If the page is detected to be in a user preferred language -> Do not show toolbar icon - change in Bug [1875817](https://bugzilla.mozilla.org/show_bug.cgi?id=1875817)
* If the page is detected to be in a user preferred language  -> User visits translations from the browser menu
   * If they translate: Show purple translated state toolbar icon
   * If they do not translate: No change, no icon on toolbar
   
Some more information in this [thread](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7246:207835&mode=design#677883739) too (easily could have summarized wrong, etc.).

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
ohall-m: Thanks for taking a look at the architecture on this patch! 
>wrapped into the existing TranslationEngineState 

Could you give an example or psudocode to illustrate?

I don't have a sense of what the architecture should be here, but have some relevant details to guide the discussion:

* `TranslationEngineState`'s lifecycle is tightly coupled to the browser translations engine and a [delegate](https://searchfox.org/mozilla-mobile/rev/4d9cf33f0069fd4650a7598e6c07dcfcd86c86d1/firefox-android/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/translate/GeckoTranslateSessionDelegate.kt#55) fully controls the lifecycle of those data properties. I'd say `TranslationEngineState` and `TranslationsState` as they stand are not implementation specific. [TranslationsState](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt) is less coupled to the translations engine itself. [TranslationEngineState](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationEngineState.kt) is tightly coupled to the translations engine.

* My impression of `TranslationsDialogState` is that it is an implementation specific repackaging of the data model. It is pulling the raw data from the BrowserStore and preparing it for the Fenix implementation. For example:

    `TranslationsDialogState`:
    `isTranslated` -> `TranslationsState.isTranslated`
    `isInProgress` -> `TranslationsState.isTranslateProcessing`
    `positiveButtonType` -> No engine correlation
    `error` -> `TranslationsState.translationError`
    `dismissDialogState` -> No engine correlation
    `showFirstTime` -> Fenix controlled concept
    `initialFrom`-> `TranslationsState.translationEngineState.initialFromLanguage`
    `initialTo` ->`TranslationsState.translationEngineState.initialToLanguage`
    `fromLanguages` -> `TranslationsState.supportedLanguages.fromLanguages`
    `toLanguages` -> `TranslationsState.supportedLanguages.toLanguages`
    

* I'm not sure what I make of the two sources of truth for the state of things like `isTranslated` especially, that keeps catching my eye as something to examine.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
MatthewTighe: Cool, thanks! We can always update the behavior in a follow-up but was just curious

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
MatthewTighe: Ah, thanks for clarifying about `TranslationEngineState` vs `TranslationState`. 

> My impression of TranslationsDialogState is that it is an implementation specific repackaging of the data model.

Do we need to repackage it? There's no reason that all the properties that are replicated across the `TranslationsDialogState` and `TranslationState` need to be. We can observe them directly from the `BrowserStore`, which would be my suggestion so that we aren't doing extra work in the observations and re-dispatching of the state. That avoids the risk of keeping two sources of truth in sync, as you've mentioned.

As to the other properties in `TranslationDialogState` that are Fenix-specific:

Some of these are pretty simple derived properties of the state that already exists in `BrowserStore`. For example, `positiveButtonType` could become a `when` expression:
```
// TranslationDialogFragment
val translationState = browserStore.observeAsComposableState { state.findTab(<currentTabId>).translationState }
val positiveButtonType = when {
    translationsState.isTranslated && !translationsState.isTranslateProcessing -> PositiveButtonType.Enabled
    is TranslationError -> PositiveButtonType.Disabled
}
```

This isn't the exact logic of course, but just an example. We don't necessarily need to keep derived properties in a separate state class if they are just used to display different UI conditionally. I think a similar case could be made for `dismissDialogState`. `showFirstTime` is something that could be passed directly from the Fragment to the Composable IMO. 

The benefit to having a `TranslationDialogStore` is that we can actually unit test some of the derived state logic. However, I am not seeing unit tests for that here, so we should either do the simple thing and remove the additional Store or do the "correct" thing and write tests as well.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFeature.kt
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * Helper for observing Translation state from browserStore.
+ */
+class TranslationsDialogFeature(
+    val browserStore: BrowserStore,
+    private val translationsDialogStore: TranslationsDialogStore,
+    private val sessionId: String,
+    private val context: Context,
+) : LifecycleAwareFeature {
```
MatthewTighe, line 28: I would suggest implementing `AbstractBinding` instead of `LifecycleAwareFeature` because it does a bit more work and is more directly intended to bind a view to observations of a Store

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFeature.kt
```diff
@@ -0,0 +1,152 @@
...
+                        translationsState.translationEngineState?.initialFromLanguage(
+                            translationsState.supportedLanguages?.fromLanguages,
+                        )
+                    fromSelected?.let {
+                        translationsDialogStore.dispatch(
+                            TranslationsDialogAction.UpdateFromSelectedLanguage(
+                                fromSelected,
+                            ),
+                        )
+                    }
```
MatthewTighe, line 50: As mentioned in my [comment](https://github.com/mozilla-mobile/firefox-android/pull/5242#discussion_r1482166568), we could simplify a lot of this dramatically if we remove the replicated state from `TranslationDialogStore`.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,225 @@
...
+    val isInProgress: Boolean = false,
+    val positiveButtonType: PositiveButtonType? = null,
+    val error: TranslationError? = null,
+    val dismissDialogState: DismissDialogState? = null,
+    val initialFrom: Language? = null,
+    val initialTo: Language? = null,
+    val fromLanguages: List<Language>? = null,
+    val toLanguages: List<Language>? = null,
+    val translatedPageTitle: String? = null,
+) : State
```
MatthewTighe, line 57: As mentioned in my [comment](https://github.com/mozilla-mobile/firefox-android/pull/5242#discussion_r1482166568), I think we can observe most of these properties directly from the `BrowserStore` and leave this state for Fenix-specific items.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,225 @@
...
+
+@Suppress("UndocumentedPublicClass")
```
MatthewTighe, line 59: Let's remove this suppression and make sure we're documenting everything appropriately.

MatthewTighe (commented):
Another few comments today, but I expect this will need another cycle depending on which of my suggestions from [this comment](https://github.com/mozilla-mobile/firefox-android/pull/5242#discussion_r1482166568) are taken

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFeature.kt
iorgamgabriel: Added AbstractBinding and unittests.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
iorgamgabriel: This will not work. The request from figma is https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7231-206075&mode=design&t=MP4I9hWYk1TLaTp0-0
"Translating snackbar - This will appear if the bottom sheet is dismissed during a translation that hasn’t completed yet." 
I need to know somehow that the user closes the bottom sheet when translation is in progress.

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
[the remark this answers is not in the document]
ohall-m: Based on this [new conversation](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD?node-id=7202:97807&mode=design#693485387), it sounds like it should remain violet in the translated state and should also have the translated content description. 

I went ahead and filed [1879336](https://bugzilla.mozilla.org/show_bug.cgi?id=1879336) to look at this. (We can also use this bug to check all the color mode options are showing up as expected on the icon.)

android-components/components/concept/toolbar/src/main/java/mozilla/components/concept/toolbar/Toolbar.kt
```diff
@@ -383,13 +389,16 @@ interface Toolbar : ScrollableToolbar {
...
 
-        private fun updateViewState() {
+        private fun updateViewState(updateContentDescriptionSelected: String? = null) {
             view?.get()?.let {
                 it.isSelected = selected
-
                 if (selected) {
                     it.setImageDrawable(imageSelectedDrawable)
-                    it.contentDescription = contentDescriptionSelected
+                    if (updateContentDescriptionSelected != null) {
```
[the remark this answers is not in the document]
ohall-m: Filed  [1879336](https://bugzilla.mozilla.org/show_bug.cgi?id=1879336) to look at this with the `violet` bug.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
MatthewTighe: > However, I am not seeing unit tests for that here, so we should either do the simple thing and remove the additional Store or do the "correct" thing and write tests as well

Reflecting on this comment from yesterday, I didn't mean for this part to read facetious or anything, hope it didn't!

I do think having a `TranslationDialogStore` with some of these derived states in them will be useful long-term and lets us have better tests, so is the right thing to do. That said, I am still of the opinion that we don't need to replicate the state that's already hosted in `BrowserStore`.

ohall-m (commented):
Thanks for the updates and tests. I have just a few minor comments with things to look at. Nothing new came up with testing. I'd like to look through it again where it is a bigger patch too.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
ohall-m: @iorgamgabriel It looks like this patch went the direction of adding tests. Could you also add a comment explaining the reasoning behind that direction too? 

Based on the conversation we had in Slack, the reasoning you mentioned there made sense to me; however, I'm not the best contact to be reviewing Fenix architecture. Please post some additional information here as well for other reviewers with much more Fenix experience!

I'm more comfortable with the duplicated store portions based on having tests. If you can think of a way to extract `isTranslated` and `isInProgress` states, I'd be more comfortable, but it sounds like that might not be the best direction based on your research.

fenix/app/src/test/java/org/mozilla/fenix/translations/TranslationsDialogBindingTest.kt
```diff
@@ -0,0 +1,278 @@
...
+                    toLanguage = spanishLanguage.code,
+                ),
+            )
+
+            val supportLanguages = TranslationSupport(
+                fromLanguages = listOf(englishLanguage),
+                toLanguages = listOf(spanishLanguage),
+            )
+
+            browserStore.dispatch(
```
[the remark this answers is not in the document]
iorgamgabriel: I need it because  
val toSelected = translationsState.translationEngineState?.initialToLanguage(
                        translationsState.supportedLanguages?.toLanguages,
                    )
And the UpdateFromSelectedLanguage action gets called only when toSelected is not null.

fenix/app/src/test/java/org/mozilla/fenix/translations/TranslationsDialogBindingTest.kt
```diff
@@ -0,0 +1,278 @@
...
+                        R.string.translations_bottom_sheet_title_translation_completed,
+                        englishLanguage.localizedDisplayName,
+                        spanishLanguage.localizedDisplayName,
+                    ),
+                ),
+            )
+        }
+
+    @Test
+    fun `WHEN translate action is send to the browserStore THEN update translation dialog store based on operation`() =
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -116,121 +100,297 @@ fun TranslationsDialogBottomSheet(
...
+            onNegativeButtonClicked = onNegativeButtonClicked,
+            onPositiveButtonClicked = onPositiveButtonClicked,
         )
     }
 }
 
+/**
+ * Dialog content if the web page was translated.
+ *
+ * @param positiveButtonType Can be enabled,disabled or in progress.
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -116,121 +100,297 @@ fun TranslationsDialogBottomSheet(
...
+ * @param onPositiveButtonClicked Invoked when the user clicks on the positive button.
+ * @param onNegativeButtonClicked Invoked when the user clicks on the negative button.
+ */
 @Composable
-private fun TranslationsDialogContentInPortraitMode(
-    translateFromLanguages: List<Language>,
-    translateToLanguages: List<Language>,
-    initialFrom: Language? = null,
+private fun DialogContentTranslated(
+    positiveButtonType: PositiveButtonType? = null,
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: I added unit tests for TranslationsDialogBinding, TranslationsDialogMiddleware and TranslationsDialogReducer.
I tried to remove almost all the business logic from the fragment. 
I removed the part when the translation icon from the toolbar is violet at the end of the translation task (I will do it in other pr).

fenix/app/src/test/java/org/mozilla/fenix/translations/TranslationsDialogBindingTest.kt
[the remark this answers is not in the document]
ohall-m: Thanks for the explanation!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+    init {
+        dispatch(TranslationsDialogAction.FetchSupportedLanguages)
+    }
+}
+
+/**
+ * The current state of the Translations bottom sheet dialog.
+ *
+ * @property isTranslated The page is currently translated.
+ * @property isInProgress The page is currently attempting a translation.
```
ohall-m, line 38: nit: Slight recommendation for renaming `isInProgress` to `translationInProgress` or similar. Reason for recommendation is a translation restore can also be in progress and this value appears only tied to a translation in progress.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+    /**
+     * When FetchSupportedLanguages is dispatched, an [TranslationsAction.OperationRequestedAction]
+     * will be dispatched to the [BrowserStore]
+     */
+    object FetchSupportedLanguages : TranslationsDialogAction()
+
+    /**
+     * Invoked when the [TranslationsDialogStore] is added to the fragment.
+     */
+    object InitTranslationsDialog : TranslationsDialogAction()
```
ohall-m, line 74: nit: Recommend ordering the `InitTranslationsDialog` to be the first item in this list of Actions.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+     * Invoked when a translation error occurs during the translation process.
+     */
+    data class UpdateTranslationError(
+        val translationError: TranslationError? = null,
+    ) : TranslationsDialogAction()
+
+    /**
+     *  Updates translate from languages list.
+     */
+    data class UpdateTranslateFromLanguages(
```
ohall-m, line 96: I think there is a good chance the from and to languages are going to be refactored onto the browser outside of a tab based on how #5544 is heading. I don't think that requires any action on this patch, since the refactor will also fix it. Mentioning it mostly as a heads-up for the expected change.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+     * Dismiss the translation dialog fragment.
+     */
+    data class DismissDialog(
+        val dismissDialogState: DismissDialogState,
+    ) : TranslationsDialogAction()
+
+    /**
+     * Updates the dialog content to translation in progress status.
+     */
+    data class UpdateInProgress(
```
ohall-m, line 131: nit: Recommend considering if this should be renamed to be translation specific since a restore can also be in progress. I don't have a strong opinion, but I think it is something to consider.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+sealed class DismissDialogState {
+    /**
+     * The dialog should be dismissed.
+     */
+    object Dismiss : DismissDialogState()
+
+    /**
+     * This is the step when translation is in progress and the dialog is waiting to be dismissed.
+     */
+    object WaitingToBeDismiss : DismissDialogState()
```
ohall-m, line 180: nit:
```suggestion
    object WaitingToBeDismissed : DismissDialogState()
```

ohall-m (approved):
Thanks for getting this patch situated! Just had a few more slight suggestions.

From my perspective, during testing, the patch either provides the expected behavior or provides a clear foundation for achieving the expected behavior. I'll leave the final architecture review to other reviewers.

Please do not merge until 125 begins due to a few new strings and the fact it is a generally large patch. Please also hold merging until at least one other reviewer approves.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBinding.kt
```diff
@@ -0,0 +1,145 @@
...
+import org.mozilla.fenix.R
+
+/**
+ * Helper for observing Translation state from browserStore.
+ */
+class TranslationsDialogBinding(
+    val browserStore: BrowserStore,
+    private val translationsDialogStore: TranslationsDialogStore,
+    private val sessionId: String,
+    private val context: Context,
```
MatthewTighe, line 26: Let's remove this direct dependency on `Context`, as it is an anti-pattern to give classes access to a god-object like this. It implies that future additions to the class could have access to anything in the god object. Instead, you can compute the string using a lambda. Something like:

```
// constructor
private val getTranslatedPageTitle: (localizedFrom: String, localizedTo: String) -> String

// usage
translationsDialogStore.dispatch(
                        TranslationsDialogAction.UpdateTranslatedPageTitle(
                            getTranslatedPageTitle(from, to),
                        ),
)

// initialization
TranslationsDialogBinding(getTranslatedPageTitle = { from, to -> 
          context.context.getString(
                                R.string.translations_bottom_sheet_title_translation_completed,
                                from,
                                to
                            ),
)
```

sorry for the formatting there, in a bit of a rush today

MatthewTighe (approved):
Since the patch works and there are tests around everything, I am approving for now to unblock this given the deadlines we're operating under. I do think that from an architectural standpoint that we are introducing a bit of tech debt here by replicating some of this state across the two stores. I could have been more clear about expectations there, so that one is on me. In general, I think we need more comprehensive and clear guidelines and education opportunities for a lot of this Redux stuff, which I am hoping to work more on soon. 

In the meantime, I think we can prioritize getting this work done, and I am always happy to chat through implementation details before they arrive in a patch - it might help avoid long review cycles. It may be good to return to refactor some of these details at a later point once we've got some more well-documented and established patterns.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBinding.kt
```diff
@@ -0,0 +1,145 @@
...
+import mozilla.components.browser.state.selector.findTab
+import mozilla.components.browser.state.state.BrowserState
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.concept.engine.translate.initialFromLanguage
+import mozilla.components.concept.engine.translate.initialToLanguage
+import mozilla.components.lib.state.helpers.AbstractBinding
+import org.mozilla.fenix.R
+
+/**
+ * Helper for observing Translation state from browserStore.
```
gabrielluong, line 20:
```suggestion
 * Helper for observing Translation state from [BrowserStore].
```
We should just use the class reference if the wording just works out.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -67,45 +67,29 @@ private val ICON_SIZE = 24.dp
...
  *
+ * @param translationsDialogState The current state of the Translations bottom sheet dialog.
  * @param learnMoreUrl The learn more link for translations website.
- * @param showFirstTimeTranslation Whether translations first flow should be shown.
- * @param translateFromLanguages Translation menu items to be shown in the translate from dropdown.
- * @param translateToLanguages Translation menu items are to be shown in the translate to dropdown.
- * @param initialFrom The initial selection for the translate from dropdown.
- * @param initialTo The initial selection for the translate to dropdown.
- * @param translationError The type of translation errors that can occur.
+ * @param showFirstTime Whether translations first flow should be shown.
```
gabrielluong, line 72: Could we use a better name here? Does `showFirstTimeTranslation` no longer apply here? What about `showFirstTimeFlow`?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
```diff
@@ -116,121 +100,297 @@ fun TranslationsDialogBottomSheet(
...
 
+/**
+ * Dialog content will adapt based on the [TranslationsDialogState].
+ *
+ * @param translationsDialogState The current state of the Translations bottom sheet dialog.
+ * @param learnMoreUrl The learn more link for translations website.
+ * @param onLearnMoreClicked Invoked when the user clicks on the learn more button.
+ * @param onPositiveButtonClicked Invoked when the user clicks on the positive button.
+ * @param onNegativeButtonClicked Invoked when the user clicks on the negative button.
+ * @param onFromSelected Invoked when the user selects an item on the from dropdown.
```
gabrielluong, line 144:
```suggestion
 * @param onFromDropdownSelected Invoked when the user selects an item on the from dropdown.
```
Similar to `onNegativeButtonClicked` naming, we should be clear about what the UI element is.

fenix/app/src/test/java/org/mozilla/fenix/translations/TranslationsDialogMiddlewareTest.kt
```diff
@@ -0,0 +1,101 @@
...
+import mozilla.components.support.test.libstate.ext.waitUntilIdle
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.fenix.helpers.FenixRobolectricTestRunner
+
+@RunWith(FenixRobolectricTestRunner::class)
+class TranslationsDialogMiddlewareTest {
+
+    @Test
+    fun `ExpectedTranslations action from TranslationDialogStore MUST call OperationRequestedAction from BrowserStore`() =
```
gabrielluong, line 24: Can we use the GIVEN WHEN THEN format here?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
```diff
@@ -53,6 +54,9 @@ class TranslationsDialogFragment : BottomSheetDialogFragment() {
     private var behavior: BottomSheetBehavior<View>? = null
     private val args by navArgs<TranslationsDialogFragmentArgs>()
     private val browserStore: BrowserStore by lazy { requireComponents.core.store }
+    private val translationDialogBinding = ViewBoundFeatureWrapper<TranslationsDialogBinding>()
+    private lateinit var translationsDialogStore: TranslationsDialogStore
+    private var isInProgress: Boolean? = null
```
gabrielluong, line 59:
```suggestion
    private var isTranslationInProgress: Boolean? = null
```

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -211,6 +214,26 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
             owner = viewLifecycleOwner,
             view = binding.root,
         )
+
+        setTranslationFragmentResultListener()
+    }
+
+    private fun setTranslationFragmentResultListener() {
```
gabrielluong, line 221: I was wondering if we were able to just show the snackbar from within the bottom sheet fragment like https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/onboarding/WallpaperOnboardingDialogFragment.kt#132?

Just wondering if we could reduce the amount of code handling with emitting fragment result to show this snackbar if there is another way to do this.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBinding.kt
```diff
@@ -0,0 +1,145 @@
...
+            TranslationsDialogAction.UpdateTranslationInProgress(
+                false,
+            ),
+        )
+
+        translationsDialogStore.dispatch(
+            TranslationsDialogAction.UpdateTranslated(
+                true,
+            ),
+        )
```
gabrielluong, line 135: Perhaps the reducer `UpdateTranslated` should be setting the in-progress state to `false` when it is `true` to dispatching multiple actions to change a individual states?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+ * @property dismissDialogState Whether the dialog bottom sheet should be dismissed.
+ * @property initialFrom Initial "from" language, based on the translation state and page state.
+ * @property initialTo Initial "to" language, based on the translation state and page state.
+ * @property fromLanguages Translation menu items to be shown in the translate from dropdown.
+ * @property toLanguages Translation menu items to be shown in the translate to dropdown.
+ * @property translatedPageTitle Title of the bottom sheet dialogue if the page was translated.
+ */
+data class TranslationsDialogState(
+    var isTranslated: Boolean = false,
+    val translationInProgress: Boolean = false,
```
gabrielluong, line 50:
```suggestion
    val isTranslationInProgress: Boolean = false,
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+ * Positive button type from the translation bottom sheet.
+ */
+sealed class PositiveButtonType {
+    /**
+     * The download indicator will appear.
+     */
+    object InProgress : PositiveButtonType()
+
+    /**
+     * The button will be TertiaryButton disabled state.
```
gabrielluong, line 158:
```suggestion
     * The button is in a disabled state.
```
Let's avoid `TertiaryButton` or `PrimaryButton` here. This class should have no concept of what a primary and tertiary button is since that's all within the design system.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+    object WaitingToBeDismissed : DismissDialogState()
+}
+
+internal object TranslationsDialogReducer {
+    /**
+     * Reduces the translations dialog state from the current state and an action performed on it.
+     *
+     * @param state The current translations dialog state.
+     * @param action The action to perform.
+     * @return the new translations dialog state
```
gabrielluong, line 189:
```suggestion
     * @return The new [TranslationsDialogState].
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+            is TranslationsDialogAction.UpdateTranslatedPageTitle -> {
+                state.copy(translatedPageTitle = action.title)
+            }
+
+            is TranslationsDialogAction.TranslateAction,
+            TranslationsDialogAction.FetchSupportedLanguages,
+            TranslationsDialogAction.RestoreTranslation,
+            -> {
+                state
+            } // handled by [TranslationsDialogMiddleware]
```
gabrielluong, line 281:
```suggestion
            is TranslationsDialogAction.TranslateAction,
            TranslationsDialogAction.FetchSupportedLanguages,
            TranslationsDialogAction.RestoreTranslation,
            -> {
                // Handled by the [TranslationsDialogMiddleware].
                state
            } 
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+     * This is the step when translation is in progress and the dialog is waiting to be dismissed.
+     */
+    object WaitingToBeDismissed : DismissDialogState()
+}
+
+internal object TranslationsDialogReducer {
+    /**
+     * Reduces the translations dialog state from the current state and an action performed on it.
+     *
+     * @param state The current translations dialog state.
```
gabrielluong, line 187:
```suggestion
     * @param state The current [TranslationsDialogState].
```

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+     */
+    data class UpdateTranslatedPageTitle(val title: String) : TranslationsDialogAction()
+}
+
+/**
+ * Positive button type from the translation bottom sheet.
+ */
+sealed class PositiveButtonType {
+    /**
+     * The download indicator will appear.
```
gabrielluong, line 153:
```suggestion
     * The download indicator will appear.
```
Did you mean translating indicator?

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
```diff
@@ -0,0 +1,284 @@
...
+
+    /**
+     * Invoked when a translation error occurs during the translation process.
+     */
+    data class UpdateTranslationError(
+        val translationError: TranslationError? = null,
+    ) : TranslationsDialogAction()
+
+    /**
+     *  Updates translate from languages list.
```
gabrielluong, line 94:
```suggestion
     * Updates translate from languages list.
```
Might want to fix the spacing for all of these.

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2358,6 +2364,8 @@
     <string name="translation_error_language_not_supported_warning_text">Sorry, we don’t support %1$s yet.</string>
     <!-- Button text on the warning card when a language is not supported. The link will take the user to a page to a support page about translations. -->
     <string name="translation_error_language_not_supported_learn_more">Learn more</string>
+    <!-- Snackbar title shown if the user closes the Translation Request dialogue and a translation is in progress. -->
+    <string name="translation_in_progress_snackbar">Translating...</string>
```
gabrielluong, line 2368:
```suggestion
    <string name="translation_in_progress_snackbar">Translating…</string>
```
I think we should be using the "…" text encoding here just like in https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/values/strings.xml#1254. Very rarely is this ever gonna be 3 "." for localization purposes.

gabrielluong: I also saw we used the unicode in https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/values/strings.xml#1409. @Delphine what would you advise here?

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
iorgamgabriel: It doesn't work. I tried this. I need to show the snackbar in onDismiss.

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBinding.kt
```diff
@@ -0,0 +1,145 @@
...
+import mozilla.components.concept.engine.translate.initialFromLanguage
+import mozilla.components.concept.engine.translate.initialToLanguage
+import mozilla.components.lib.state.helpers.AbstractBinding
+import org.mozilla.fenix.R
+
+/**
+ * Helper for observing Translation state from browserStore.
+ */
+class TranslationsDialogBinding(
+    val browserStore: BrowserStore,
```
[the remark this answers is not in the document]
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBinding.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBottomSheet.kt
iorgamgabriel: Done

fenix/app/src/test/java/org/mozilla/fenix/translations/TranslationsDialogMiddlewareTest.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogFragment.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: Done

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogStore.kt
iorgamgabriel: DOne

fenix/app/src/main/res/values/strings.xml
iorgamgabriel: Done

PR #5247, opened 2024-01-22
Add some basic prototypes for demo

MatthewTighe:
This PR is meant to demonstrate some of the concepts discussed in the proposal for [RFC 0010](https://github.com/mozilla-mobile/firefox-android/pull/4126/files).

Namely, this shows:
1. A `BrowserNavigator` abstraction for middlewares to have a common use case for handling navigation to the browser and opening a tab.
2. A slight refactor of how the view gets refreshed when the History fragment is finished syncing.

For transient effects such as the view refresh shown in this prototype, I do think the code may be determined by the effect. For example, the [undo snackbar](https://searchfox.org/mozilla-mobile/rev/2655756f8d424489c4dda3486c7cec3bd408f347/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistoryStorageMiddleware.kt#62) for history needs to be an atomic sequence of work (barring more refactoring).

Finally, this prototype does not include navigation between screens using a `NavController` as there are existing examples in the app like the [HistoryNavigationMiddleware](https://searchfox.org/mozilla-mobile/rev/2655756f8d424489c4dda3486c7cec3bd408f347/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistoryNavigationMiddleware.kt#32) and the [DebugDrawerNavigationMiddleware](https://searchfox.org/mozilla-mobile/rev/2655756f8d424489c4dda3486c7cec3bd408f347/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/debugsettings/store/DebugDrawerNavigationMiddleware.kt#21).

I do also believe that `BrowserNavigator` has plenty of room for improvement, but this would be the first simple step in decoupling the implementation from `HomeActivity` and moving it into the `lib-state` world.

### GitHub Automation

Used by GitHub Actions.

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/browser/readermode/BrowserNavigator.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/browser/readermode/BrowserNavigator.kt
@@ -1,0 +1,173 @@
+package org.mozilla.fenix.browser.readermode
+
+import androidx.annotation.IdRes
+import androidx.navigation.NavController
+import androidx.navigation.NavDirections
+import mozilla.components.browser.state.action.ContentAction
+import mozilla.components.browser.state.search.SearchEngine
+import mozilla.components.browser.state.state.SessionState
+import mozilla.components.browser.state.store.BrowserStore
+import mozilla.components.concept.base.profiler.Profiler
+import mozilla.components.concept.engine.EngineSession
+import mozilla.components.concept.storage.HistoryMetadataKey
+import mozilla.components.feature.search.SearchUseCases
+import mozilla.components.feature.session.SessionUseCases
+import mozilla.components.feature.tabs.TabsUseCases
+import mozilla.components.support.ktx.kotlin.isUrl
+import mozilla.components.support.ktx.kotlin.toNormalizedUrl
+import org.mozilla.fenix.BrowserDirection
+import org.mozilla.fenix.R
+import org.mozilla.fenix.browser.browsingmode.BrowsingMode
+import org.mozilla.fenix.components.AppStore
+import org.mozilla.fenix.components.Components
+import org.mozilla.fenix.ext.alreadyOnDestination
+import org.mozilla.fenix.ext.nav
+import org.mozilla.fenix.utils.Settings
+
+class BrowserNavigator(
+    private val addTabUseCase: TabsUseCases.AddNewTabUseCase,
+    private val sessionUseCases: SessionUseCases,
+    private val searchUseCases: SearchUseCases,
+    private val appStore: AppStore,
+    private val browserStore: BrowserStore,
+    private val profiler: Profiler?,
+    private val settings: Settings,
+    private val getNavDirections: (BrowserDirection, String?) -> NavDirections?,
+    private val getNavController: () -> NavController,
+) {
+
+    constructor(
+        components: Components,
+        settings: Settings,
+        getNavDirections: (BrowserDirection, String?) -> NavDirections?,
+        getNavController: () -> NavController,
+    ) : this(
+        components.useCases.tabsUseCases.addTab,
+        components.useCases.sessionUseCases,
+        components.useCases.searchUseCases,
+        components.appStore,
+        components.core.store,
+        components.core.engine.profiler,
+        settings,
+        getNavDirections,
+        getNavController,
+    )
+
+    fun openToBrowserAndLoad(
+        searchTermOrURL: String,
+        newTab: Boolean,
+        from: BrowserDirection,
+        customTabSessionId: String? = null,
+        engine: SearchEngine? = null,
+        forceSearch: Boolean = false,
+        flags: EngineSession.LoadUrlFlags = EngineSession.LoadUrlFlags.none(),
+        requestDesktopMode: Boolean = false,
+        historyMetadata: HistoryMetadataKey? = null,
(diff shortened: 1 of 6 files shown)


fenix/app/src/main/java/org/mozilla/fenix/browser/readermode/BrowserNavigator.kt
```diff
@@ -0,0 +1,173 @@
...
+    private val browserStore: BrowserStore,
+    private val profiler: Profiler?,
+    private val settings: Settings,
+    private val getNavDirections: (BrowserDirection, String?) -> NavDirections?,
+    private val getNavController: () -> NavController,
+) {
+
+    constructor(
+        components: Components,
+        settings: Settings,
```
MozillaNoah, line 41: for further simplification: settings can technically be obtained from components

fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistoryUiEffectMiddleware.kt
```diff
@@ -0,0 +1,35 @@
+package org.mozilla.fenix.library.history.state
+
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareContext
+import org.mozilla.fenix.library.history.HistoryFragmentAction
+import org.mozilla.fenix.library.history.HistoryFragmentState
+
+class HistoryUiEffectMiddleware(
```
MozillaNoah, line 8: it feels a little weird to have a middleware just to trigger UI side effects that could just be a part of their state observation, but that is undoubtedly an implementation issue with History.

fenix/app/src/main/java/org/mozilla/fenix/ext/Fragment.kt
```diff
@@ -149,3 +151,13 @@ fun Fragment.registerForActivityResult(
         }
     }
 }
+
+fun Fragment.buildBrowserNavigator() =
```
MozillaNoah, line 155: +1 for the extension

fenix/app/src/main/java/org/mozilla/fenix/browser/readermode/BrowserNavigator.kt
```diff
@@ -0,0 +1,173 @@
...
+class BrowserNavigator(
+    private val addTabUseCase: TabsUseCases.AddNewTabUseCase,
+    private val sessionUseCases: SessionUseCases,
+    private val searchUseCases: SearchUseCases,
+    private val appStore: AppStore,
+    private val browserStore: BrowserStore,
+    private val profiler: Profiler?,
+    private val settings: Settings,
+    private val getNavDirections: (BrowserDirection, String?) -> NavDirections?,
+    private val getNavController: () -> NavController,
```
MozillaNoah, line 36: why a lambda instead of just the actual nav controller? so `BrowserNavigator` doesn't hold onto the reference?

MozillaNoah (commented):
Thanks for the demo! this helps makes sense for one of Fenix's complex navigation use cases.

If we ever get to 100% compose, we'll have to rethink the fragment navigation, but that is obviously a problem for way down the road.

fenix/app/src/main/java/org/mozilla/fenix/library/history/state/HistoryUiEffectMiddleware.kt
MatthewTighe: This was meant to demo callback handling through middleware. I think it many cases it would be more appropriate to launch a side-effect from observing a change to state, but I also don't think it is necessary to track things like `isMenuOpen` in the state for a screen. For example if we have the following action:

```
sealed class ScreenAction {
    data class ItemLongClicked(val item: Item) : ScreenAction
}
```

Handling the side-effect from a middleware makes more sense to me
```
class ScreenUiMiddleware(
    private val displayItemMenu: (Item) -> Unit
)  : Middleware<ScreenState, ScreenAction> {
    override fun invoke(
        context: MiddlewareContext<ScreenState, ScreenAction>,
        next: (ScreenAction) -> Unit,
        action: ScreenAction,
    ) {
        when (action) {
             is ItemLongClicked -> displayItemMenu(action.item)
         }
    }
}
```

t-p-white, line 8: 🤔 thinking out loud here, I wonder if there's any notable slow down in navigation based on the extra processing that has to happen using middleware

MozillaNoah: Anecdotally, in my testing with Debug Drawer, I have not seen nor felt any slowness (within previews or on-device) by having this additional overhead.

MatthewTighe: Generally speaking, the pattern has been fast enough that there is now an [RFC](github.com/mozilla-mobile/firefox-android/pull/5353) to allow Stores to run on the main thread. There is always risk that something in the reducer chain misbehaves and slows things down (for example a middleware doing expensive computations), but that would generally be considered a solvable bug rather than a side-effect of the design principle.

PR #5302, opened 2024-01-25
Bug 1876567 - Fix Middleware Issue with Processing Translations States

ohall-m:
This patch corrects an issue where `isTranslateProcessing` and `isRestoreProcessing` states were not updating due to not sending along pre-process actions in `EngineDelegateMiddleware`.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1876567


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddleware.kt
```diff
@@ -44,8 +44,8 @@ internal class EngineDelegateMiddleware(
             is EngineAction.ExitFullScreenModeAction -> exitFullScreen(context.store, action)
             is EngineAction.SaveToPdfAction -> saveToPdf(context.store, action)
             is EngineAction.PrintContentAction -> printContent(context.store, action)
-            is TranslationsAction.TranslateAction -> translate(context.store, action)
-            is TranslationsAction.TranslateRestoreAction -> translateRestoreOriginal(context.store, action)
+            is TranslationsAction.TranslateAction -> translate(context.store, next, action)
+            is TranslationsAction.TranslateRestoreAction -> translateRestoreOriginal(context.store, next, action)
```
gabrielluong, line 48: Could we reorder these `TranslationsAction` to be after the `EngineAction`? Just nice to keep all the `EngineAction` altogether.

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddlewareTest.kt
```diff
@@ -742,4 +745,67 @@ class EngineDelegateMiddlewareTest {
...
+            middleware = EngineMiddleware.create(
+                engine = engine,
+                scope = scope,
+            ),
+            initialState = BrowserState(
+                tabs = listOf(tab),
+            ),
+        )
+
+        assertEquals(false, store.state.findTab(tab.id)?.translationsState?.isTranslateProcessing)
```
gabrielluong, line 765: Use `assertFalse`. Same thing below

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddlewareTest.kt
```diff
@@ -742,4 +745,67 @@ class EngineDelegateMiddlewareTest {
...
+                toLanguage = "en",
+                options = null,
+            ),
+        ).joinBlocking()
+
+        dispatcher.scheduler.advanceUntilIdle()
+        store.waitUntilIdle()
+
+        verify(engineSession).requestTranslate(any(), any(), any())
+        assertEquals(true, store.state.findTab(tab.id)?.translationsState?.isTranslateProcessing)
```
gabrielluong, line 780: Use `assertTrue`. Same thing below

gabrielluong (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddleware.kt
```diff
@@ -44,8 +44,8 @@ internal class EngineDelegateMiddleware(
...
             is EngineAction.ClearDataAction -> clearData(context.store, action)
             is EngineAction.PurgeHistoryAction -> purgeHistory(context.state)
             else -> next(action)
```
gabrielluong, line 51: It's possible that this should just be a no-op and we should always just be calling `next(action)`. What do you think @MatthewTighe?

MatthewTighe: It looks generally like this `Middleware` intercepts actions and then dispatches a different action to complete the reducer chain. Since this is pre-existing, core browser functionality I would hesitate to update the existing code without an investigation. 

Visually, it might make the distinction between handling more clear and quickly communicated to readers to write the translate actions like:

```
is TranslationsAction.TranslateAction -> {
    next(action)
    translate(context.store, action)
}
is TranslationsAction.TranslateRestoreAction -> {
    next(action)
    translateRestoreOriginal(context.store, action)
}
```

One other thing to note is that the `context.store` reference in `Middleware::invoke` will actually update as a result of the reducer chain, so the `state` will be different after calling `next`. If you want the original state, you can capture it before calling next or handling the middleware work first (if it won't slow the first of the chain noticeably)

For example:
```
is TranslationsAction.TranslateAction -> {
    val currentState = context.store.state
    next(action)
    translate(state, action)
}
or 
is TranslationsAction.TranslateRestoreAction -> {
    translateRestoreOriginal(context.store, action)
    next(action)
}
```

gabrielluong: +1 Definitely worth a followup investigation for adding the `next(action)` outside of the `else`. No action needed here.

+1 to add the `next(action)` inside of the` is TranslationsAction.TranslateAction -> {}` instead of the `translate()` function.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddleware.kt
ohall-m: Sure, updated!

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddlewareTest.kt
ohall-m: Updated, is it okay to force unwrap in a test?

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddlewareTest.kt
ohall-m: Updated

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddleware.kt
ohall-m: > One other thing to note is that the context.store reference in Middleware::invoke will actually update as a result of the reducer chain, so the state will be different after calling next. If you want the original state, you can capture it before calling next or handling the middleware work first (if it won't slow the rest of the chain noticeably)

This is my intention:
* action(next) switches in inProgress on the store
* translate/restore will remove inProgress when complete

I think I probably want translate/restore to use the updated/most current store, is that right?

Made [bug 1876598](https://bugzilla.mozilla.org/show_bug.cgi?id=1876598). Please add to the bug or let me know if I misunderstood anything in the filing.

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/EngineDelegateMiddlewareTest.kt
MatthewTighe: Definitely. The test will just fail anyway if it crashes

PR #5322, opened 2024-01-26
Bug 1796209 - If `config.toolbarColor` is null use the `toolbar` 'menu' colour as a fallback

t-p-white:
The fix here is to use the main toolbar colours as a fallback (this will be consistent with the relevant icons) if no colour is defined in the config.

A quick way to test this is to comment out the following in `SupportUtils`:
```kotlin
//        .setDefaultColorSchemeParams(
//            CustomTabColorSchemeParams.Builder().setToolbarColor(context.getColorFromAttr(R.attr.layer1)).build(),
//        )
```
and access the 'privacy policy' from the onboarding link.

Original bug demo:

Patch fix demo:

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


t-p-white:
@jonalmeida - as the original creator of the bug, does this cover the bug as expected?

jonalmeida:
> @jonalmeida - as the original creator of the bug, does this cover the bug as reported?

The screenshots look good to me, but I'll leave the code review for someone else. Don't forget to test this in dark mode too!

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -74,7 +73,7 @@ class CustomTabsToolbarFeature(
     private var initialized: Boolean = false
     private val titleObserver = CustomTabSessionTitleObserver(toolbar)
     private val context get() = toolbar.context
-    internal var readableColor = Color.WHITE
+    internal var readableColor = -1
```
MozillaNoah, line 76: you might be able to do a `lateinit internal var` instead of using `-1`

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -114,11 +113,8 @@ class CustomTabsToolbarFeature(
     internal fun init(config: CustomTabConfig) {
         // Don't allow clickable toolbar so a custom tab can't switch to edit mode.
         toolbar.display.onUrlClicked = { false }
-
-        // If it's available, hold on to the readable colour for other assets.
-        if (updateToolbarBackground && config.toolbarColor != null) {
-            readableColor = getReadableTextColor(config.toolbarColor!!)
-        }
+        readableColor =
+            config.toolbarColor?.let { getReadableTextColor(it) } ?: toolbar.display.colors.menu
```
MozillaNoah, line 117: there's apparently some unit tests around this value. It looks like your CI is all green, but it might be worthwhile adding a unit test to cover the use path this fix is for.
[CustomTabsToolbarFeatureTest](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeatureTest.kt)

MozillaNoah (approved):
About to test locally to confirm, but the changes look good (just a comment about adding a test). Taking another look at it, the custom tab theming code needs some TLC.

MozillaNoah:
Testing locally, the bug looks resolved. However I did notice a new issue where the colors in the custom tab activity do not get updated when switching between light and dark mode.

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
t-p-white: I considered this, but it would mean also making it nullable - which is not what we expect. However, the fact we both thought about this is perhaps indicative of a code smell. I've updated this to pass the `readableColor` value in as a param instead of having the class property

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
t-p-white: Good catch, updated tests 👍

t-p-white:
> However I did notice a new issue where the colors in the custom tab activity do not get updated when switching between light and dark mode.

I can't recreate this specifically for this use case - we have a number of open bugs around the custom tab's implementation though which i'm working through - e.g. bugzilla.mozilla.org/show_bug.cgi?id=1813913.

MozillaNoah (approved):
re-approving, as I'm not reproducing what I saw on Friday

t-p-white:
@MozillaNoah [PR up](https://github.com/mozilla-mobile/firefox-android/pull/5368) for the potential bug you saw

PR #5353, opened 2024-01-29
Bug 1877307 - RFC: Introduce a Store for UI components

jonalmeida:
### Deadline:
~~Wednesday, February 21th, 2024~~
Wednesday, March 13th, 2024 (updated)

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1877307


docs/rfcs/0011-customizable-store-dispatcher.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Motivation
+
+Android embedders use the main thread for UI, user-facing, or touch handling work. For example, notifying UI components when IO from storage layers have completed, an engine's tasks that can happen on a separate thread, or global-level state updates for different components to observe.
+
+When components dispatch actions to the store, these actions are performed on an independant single thread dispatcher in the `Store` to avoid overloading the main thread with heavy work that might be performed during the `reduce` or in a `Middleware`. In practice, these actions have been short and fast so they do not cause overhead. Work done in a `Middleware` that can be slow, like I/O,  is put onto separate Dispatchers. The performance optimization to switch to a `Store` thread, requires that components which are always run on the main thread, to ensure synchronisation is now kept between the main thread and the store thread for observers of the `State`.
```
MatthewTighe, line 18: > In practice, these actions have been short and fast so they do not cause overhead.

Are there any profiles or measurements that have shown this to be true?

docs/rfcs/0011-customizable-store-dispatcher.md
```diff
@@ -0,0 +1,71 @@
...
+    reducer = AppStoreReducer::reduce,
+    runOnMain = true, // a boolean or enum value to signify the dispatcher used.
+)
+```
+
+Another advantage of this change is that `Store`s would be able to run on the test framework's default dispatcher. This reduces the number of intermittent test failures we have seen where the `Store` is still processing actions async to an assertion. Fewer usages of `store.waitUntilIdle()` are required after every `dispatch` is called in a test, as a result.
+
+## Drawbacks
+
+* Mistakenly doing work on the main thread - we could end up performing large amounts of work on the main thread unintentionally if we are not careful. This could be because of a large number of small tasks, a single large task, a blocking task, or a combination. This could be mitigated by keeping the `Store` dispatcher as the default, and catching performance regressions in the specific features they are used in.
```
MatthewTighe, line 63: Is there something we could do to mitigate this risk by introducing restrictions on `Middleware`? They are the most likely culprit for long-running operations since they handle side-effects.

docs/rfcs/0011-customizable-store-dispatcher.md
```diff
@@ -0,0 +1,71 @@
...
+    dispatcher = Dispatchers.Main // choose a preferable dispatcher.
+)
+
+// Configurable set of pre-defined options to choose.
+class AppStore(
+    initialState: AppState = AppState(),
+) : Store<AppState, AppAction>(
+    initialState = initialState,
+    reducer = AppStoreReducer::reduce,
+    runOnMain = true, // a boolean or enum value to signify the dispatcher used.
```
MatthewTighe, line 55: Of these options, I think an enum value would be preferable here, but what benefit does this then provide over using the finite set already defined on `Dispatchers`? It seems as potentially error-prone to me.

That is to say, my preference would be to allow for `Dispatchers` directly.

MozillaNoah:
This sounds pretty cool. Could you add to the RFC a (maybe) hypothetical, follow-up use case of using a non-store dispatcher in practice? In other words, do you have a feature/use-case in mind that would immediately take use of this change (whether it's a refactor of an existing feature or something new in the pipeline)?

Assuming this RFC is accepted, what are your proposed next steps for implementing and consuming this? Are there any technical design docs (or plans to make some)?

MatthewTighe (commented):
Love this suggestion. My preference would be to allow for `Dispatchers` directly as parameters to Stores.

In Fenix, our long-term plan is to continue migrating components to `lib-state`. Does this introduce any risk that these shared thread pools get overwhelmed by work as applications scale? As an alternative, what about sharing `CoroutineScope`s directly between Stores to take advantage of cooperative computing?

Finally, a couple tangential things to add in case they are worth wrapping into this RFC or follow-up work:

1. It may be helpful to add `Store.scope` to the `MiddlewareContext`, so that middleware can inherit test scopes and scopes tied to the lifecycle of the Store.
2. This is more unrelated, but while you're in here thinking about things. The `dispatch` method defined directly on the `MiddlewareContext` behaves differently than calling `context.store.dispatch`. Is it worth renaming (or some other change) in order to make the difference more apparent?

jonalmeida:
> This sounds pretty cool. Could you add to the RFC a (maybe) hypothetical, follow-up use case of using a non-store dispatcher in practice? In other words, do you have a feature/use-case in mind that would immediately take use of this change (whether it's a refactor of an existing feature or something new in the pipeline)?

@MozillaNoah All the stores in Fenix that are meant for specific UI screens (e.g. all the fragment stores) would benefit from having the Store thread be the same as the main thread. Do not change this for the `BrowserStore` however without performing very careful testing, as it's the at the core of the application.

> Assuming this RFC is accepted, what are your proposed next steps for implementing and consuming this? Are there any technical design docs (or plans to make some)?

I am considering writing a blog post about this on mozac.org once this lands on it's usage. It would be up to your team to decide to make these changes within Fenix.

> In Fenix, our long-term plan is to continue migrating components to `lib-state`. Does this introduce any risk that these shared thread pools get overwhelmed by work as applications scale? As an alternative, what about sharing `CoroutineScope`s directly between Stores to take advantage of cooperative computing?

@MatthewTighe these are precisely the reasons I would not want to allow any dispatcher to be used. Misuse of which dispatcher to use can lead to very gnarly bugs that will be difficult to debug. We may create a dispatcher that is starved, a dispatcher with shared threads amongst multiple other dispatchers, etc. that would lead to these kinds of bugs that will be a pain to deal with.

The benefits today are from making it easier to have a Store and frontend code running on the same UI thread, so I'd recommend starting with that. If there are benefits to having it on any other dispatcher, we can expand this scope in the future.

> Finally, a couple tangential things to add in case they are worth wrapping into this RFC or follow-up work:
> 
>   1. It may be helpful to add `Store.scope` to the `MiddlewareContext`, so that middleware can inherit test scopes and scopes tied to the lifecycle of the Store.
> 
>   2. This is more unrelated, but while you're in here thinking about things. The `dispatch` method defined directly on the `MiddlewareContext` behaves differently than calling `context.store.dispatch`. Is it worth renaming (or some other change) in order to make the difference more apparent?

Yeah, these are tangential, so please file follow-ups.

docs/rfcs/0011-customizable-store-dispatcher.md
jonalmeida: Aside from the mitigation already mentioned, I can't see any obvious ways to do this programatically. Each application of the `Store` would have to know what work they are performing and how large it is to avoid this.

So it would be up to them to know what work is considered non-performant. In Fenix, there is process to can performance regressions and that has been used to find bugs where too many actions were being dispatched (e.g. download progress was [too noisy](https://github.com/mozilla-mobile/android-components/pull/8197/files#diff-51a4f62b18b097b25a5482b31e10be7850f9a23d1c097fecaa7fb2d4f6674bc1R635-R641).

docs/rfcs/0011-customizable-store-dispatcher.md
jonalmeida: The boolean/naming is to show to an example of some kind of option, this is not the final name/version of it.

See the comment below about using any `Dispatcher`.

docs/rfcs/0011-customizable-store-dispatcher.md
```diff
@@ -0,0 +1,71 @@
...
+)
+```
+
+Another advantage of this change is that `Store`s would be able to run on the test framework's default dispatcher. This reduces the number of intermittent test failures we have seen where the `Store` is still processing actions async to an assertion. Fewer usages of `store.waitUntilIdle()` are required after every `dispatch` is called in a test, as a result.
```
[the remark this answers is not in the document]
jonalmeida: I'm not sure I understand what you are asking here in regards to this RFC.

> If we are only concerned about the result of the reducer, then `store.dispatch().joinBlocking()` should suffice and can be used across threads.

From my understanding, `joinBlocking` will block the running of the test for that particular action alone and then continue with that test. With `waitUntilIdle`, we would wait until all work with that Store scope would be completed, so actions that are dispatched as side-effects of the original one. 

You can use `waitUntilIdle` in all cases today, but you can't use `joinBlocking` for all cases that `waitUntilIdle` is used.

> this change would not be sufficient to alleviate this concern without allowing the middleware to inherit the dispatcher of the Store.

Isn't this something that already exists if true? We are not trying to make sure all Middlewares inherit the same scope. You can always pass in a separate scope or dispatcher for work that you don't want to be done in the Store scope.

docs/rfcs/0011-customizable-store-dispatcher.md
```diff
@@ -0,0 +1,71 @@
...
+title: Allow customisation of the Store dispatcher
+permalink: /rfc/0011-customizable-store-dispatcher
+---
+
+* Start date: 2024-01-29
+* RFC PR: [#5353](https://github.com/mozilla-mobile/firefox-android/pull/5353)
+
+## Summary
+
+In most applications of the `Store`, it is preferable to have reducers perform work on the main thread. This helps to simplify the levels of async tasks that need to be synchronised with other async tasks that are already on the main thread.
```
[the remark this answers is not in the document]
jonalmeida: > Is there an example of how staying on a single thread helps synchronize these tasks vs synchronizing them across threads?

I mentioned this in the 'Motivation' section below, is there something I can do to make this more clear?

"Android embedders use the main thread for UI, user-facing, or touch handling work. For example, notifying UI components when IO from storage layers have completed, an engine's tasks that can happen on a separate thread, or global-level state updates for different components to observe."

> Is the goal to get results directly from dispatch, or will observation still be the primary method of reading state?

You will still get the results from observations. Maybe I need to rephrase this paragraph better.. The actions will all happen on the same thread async, but when working with other systems that also have separate threads, there are then more than one-level of async work that needs to be waited on.

An example of this that @titooan  and I have run into recently, is where GeckoView has it's own Gecko Main thread for performing async work that also touches into the Gecko platform layer. We want to wait for the `GeckoResult` to come back from there and have those results along with other results from the Android OS touch events update a Store. So this means that we now have at least three threads that provide results we need to sync before we can perform an action: Gecko Main, Store, Android UI (main).

Since the Android OS touch events happen on the main thread already, having a store that is also running on the same thread reduces that complexity.

Similarly, any of the fragment stores that we have also benefit from this simplification as well.

For historical context, the Store was meant to always run on the main thread, but the separate Store thread was possibly a premature optimization for the BrowserStore, which is fairly valuable today, but using the Store for other parts of the app was not something we considered then. In which case, running it on the main thread would be sufficient.

(this context is good for comments, but not useful for an RFC)

docs/rfcs/0011-customizable-store-dispatcher.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Motivation
+
+Android embedders use the main thread for UI, user-facing, or touch handling work. For example, notifying UI components when IO from storage layers have completed, an engine's tasks that can happen on a separate thread, or global-level state updates for different components to observe.
+
+When components dispatch actions to the store, these actions are performed on an independant single thread dispatcher in the `Store` to avoid overloading the main thread with heavy work that might be performed during the `reduce` or in a `Middleware`. In practice, these actions have been short and fast so they do not cause overhead. Work done in a `Middleware` that can be slow, like I/O,  is put onto separate Dispatchers. The performance optimization to switch to a `Store` thread, requires that components which are always run on the main thread, to ensure synchronisation is now kept between the main thread and the store thread for observers of the `State`.
+
+There are some advantages to this change:
+
+* Simplicity for main-thread `Store`s.
```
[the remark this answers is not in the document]
jonalmeida: See [this comment](https://github.com/mozilla-mobile/firefox-android/pull/5353#discussion_r1473200196) above for examples.

docs/rfcs/0011-customizable-store-dispatcher.md
```diff
@@ -0,0 +1,71 @@
...
+---
+
+* Start date: 2024-01-29
+* RFC PR: [#5353](https://github.com/mozilla-mobile/firefox-android/pull/5353)
+
+## Summary
+
+In most applications of the `Store`, it is preferable to have reducers perform work on the main thread. This helps to simplify the levels of async tasks that need to be synchronised with other async tasks that are already on the main thread.
+
+## Motivation
```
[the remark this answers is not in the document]
jonalmeida: I will update this section with a more concrete example. No, we are not trying to make the Store synchronous.

docs/rfcs/0011-customizable-store-dispatcher.md
```diff
@@ -0,0 +1,71 @@
...
+
+Android embedders use the main thread for UI, user-facing, or touch handling work. For example, notifying UI components when IO from storage layers have completed, an engine's tasks that can happen on a separate thread, or global-level state updates for different components to observe.
+
+When components dispatch actions to the store, these actions are performed on an independant single thread dispatcher in the `Store` to avoid overloading the main thread with heavy work that might be performed during the `reduce` or in a `Middleware`. In practice, these actions have been short and fast so they do not cause overhead. Work done in a `Middleware` that can be slow, like I/O,  is put onto separate Dispatchers. The performance optimization to switch to a `Store` thread, requires that components which are always run on the main thread, to ensure synchronisation is now kept between the main thread and the store thread for observers of the `State`.
+
+There are some advantages to this change:
+
+* Simplicity for main-thread `Store`s.
+* Unit testing can now occur on the test framework's thread.
+* Fewer resources needed for context shifting between threads.
```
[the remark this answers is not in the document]
jonalmeida: Which configuration are you saying is too complex? 

> the increased likelihood of performance issues appearing from forgetting to switch threads in a Middleware.

This wouldn't increase the likelihood of performance issues unless you choose to change the defaults and not understand the cost of the work you are performing. I would argue that this would help us catch bugs quicker because you would see the cost of doing non-performant work on the wrong dispatcher from the get-go. Today we are hiding this when we use the Store thread to do the wrong kind of work on it. 😄

docs/rfcs/0011-customizable-store-dispatcher.md
jonalmeida: > Is that then the goal of your proposal, and do we then still need the Dispatchers example as well?

Both examples were add for the sake of discussion to show that there will be some form of configuration added. It sounds like there is more preference towards the second option, so I can remove the one above if it is causing confusion.

docs/rfcs/0011-customizable-store-dispatcher.md
jonalmeida: This is not a problem we need to solve at the component-level, or at least not prematurely. @rahulsainani has recently added some helpful code to pass [in the scope used to create the store for middlewares](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/components/StoreProvider.kt#61-81), so this might be an area we can consider improving there first and we can go from there.

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
jonalmeida: Done.

docs/rfcs/0011-customizable-store-dispatcher.md
MatthewTighe: Rahul's patch is very useful for real usecases, but isn't usable in unit tests where we construct Stores without `StoreProvider.get()`. This proposal would allow us to more easily share the test scope between middleware and Stores, but I still think those could be automatically shared in the future. Agreed that this is tangential in regards to this proposal though

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
MatthewTighe: > From my understanding, joinBlocking will block the running of the test for that particular action alone and then continue with that test. With waitUntilIdle, we would wait until all work with that Store scope would be completed, so actions that are dispatched as side-effects of the original one.

Yeah, this is exactly right. So `joinBlocking` should suffice for any case where we are testing just the reducer, whereas `waitUntilIdle` is required if, for example, middlewares dispatch additional actions. However, `waitUntilIdle` doesn't account for coroutines launched by the middleware. 

> Fewer usages of `store.waitUntilIdle()` are required after every `dispatch` is called in a test, as a result.

Wouldn't we still need as many `waitUntilIdle`s (or `testScope.advanceUntilIdle`s) because the processing is still happening async, just on the test thread instead of a Store thread?

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
MatthewTighe: I will admit to not being very familiar with how `GeckoResult`s are handled, so would love to learn more about this specific use-case. Was there a specific issue with ordering in the example above that would've been solved by having the Store running on the UI thread?

> Since the Android OS touch events happen on the main thread already, having a store that is also running on the same thread reduces that complexity.

So in terms of simplicity, do we just mean thread switches? It seems like the complexity here should be hidden by coroutine primitives and Store abstractions anyway. That is, I would expect a `GeckoResult` handler to `dispatch` to a Store, and the eventual result of that `dispatch` to be consumed through observation by the View. None of these individual pieces should need to know what thread the others are executing on.

Regardless, I do think performance wins from avoiding context switches and simpler test setups are good things to have. Just trying to more fully understand the motivation 😄

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
MatthewTighe: > Which configuration are you saying is too complex?

I didn't mean to imply "too complex", but I do think generally that configurability == complexity. Yes, the complexity is currently hidden by abstractions, but that can be a useful tool for simplifying implementations.

This is a good point, but if we wanted we could pretty easily create profiling middleware that let us understand this with more rich detail rather than trying to catch it in production.

MatthewTighe:
👋 hey Jon! I think this is very close to approval from my perspective, just hoping to read through your response to some of my further comments.

Today our team had a discussion on how we want to handle RFCs moving forward (we want to solve for them recently being a little slow to be finished) and have decided to adopt the "stakeholder" pattern that is likely familiar from other internal technical documents. For now, our team agreement will be that RFCs should have at least 2 internal stakeholders and a 1 external stakeholder if the proposal is an external one. This should make it more clear who is responsible for helping things get over the finish line in a timely manner. In light of that, I have added @boek as the 2nd internal stakeholder and I'm assuming that @csadilek will act as the external.

There will be additional documentation of this decision forthcoming, but in the meantime I've at least added a line to the [first RFC](https://github.com/mozilla-mobile/firefox-android/pull/5408) to indicate our plans here.

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
jonalmeida: > Wouldn't we still need as many `waitUntilIdle`s (or `testScope.advanceUntilIdle`s) because the processing is still happening async, just on the test thread instead of a Store thread?

In our tests, we use the `MainCoroutineRule` so that all main tasks are done on the same test thread, so that would make all main work blocking in the sequence that they are executed:

```kotlin
@get:Rule
val coroutinesTestRule = MainCoroutineRule()
```

I will update the RFC to mention this because it isn't explicit.

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
jonalmeida: > So in terms of simplicity, do we just mean thread switches? It seems like the complexity here should be hidden by coroutine primitives and Store abstractions anyway. That is, I would expect a `GeckoResult` handler to `dispatch` to a Store, and the eventual result of that `dispatch` to be consumed through observation by the View. None of these individual pieces should need to know what thread the others are executing on.
> 
> Regardless, I do think performance wins from avoiding context switches and simpler test setups are good things to have. Just trying to more fully understand the motivation 😄

It's a bit more than that too: in the example above, it would make it easier for us to know that when I invoke two methods both of those will be scheduled in the same order too:

```kotlin
geckoView.asyncApi()
store.dispatch(UpdateAction)
```

docs/rfcs/0011-customizable-store-dispatcher.md
jonalmeida: This is only anecdotal evidence because it really depends on the specific store, so it's not worth adding to the RFC: I ran our performance tests against a prototype patch where I changed the Fenix AppStore to use the main dispatcher and there was only a small 33ms perf win (2199.0-2166.0) using a debug build.

If we start to switch over stores to use the main dispatchers we can run tests for those to see if they are quicker and no longer doing work on the wrong thread.

```
==> patch-pixel6a-store-mainthread-0.txt <==
{'max': 2256.0,
 'mean': 2104.9,
 'median': 2166.0,
 'min': 1514.0,
 'replicate_count': 10,
 'replicates': [2097.0, 1514.0, 2067.0, 2157.0, 2163.0, 2169.0, 2183.0, 2191.0,
                2252.0, 2256.0]}

==> patch-pixel6a-store-mainthread-pre-0.txt <==
{'max': 2929.0,
 'mean': 2220.9,
 'median': 2199.0,
 'min': 1613.0,
 'replicate_count': 10,
 'replicates': [2929.0, 1613.0, 2098.0, 2186.0, 2106.0, 2154.0, 2249.0, 2212.0,
                2317.0, 2345.0]}

```

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
jonalmeida: > Was there a specific issue with ordering in the example above that would've been solved by having the Store running on the UI thread?

There isn't any issue today, but @titooan and I will plan to write a design doc about how we would want to build on top of this work that will make APZ features (bottom toolbar and pull-to-refresh) easier to reason about by having a Store to collect data from various systems so that we can reason about them more easily. Specifically, a lot of the logic in [`NestedGeckoView.kt`](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/engine-gecko/src/main/java/mozilla/components/browser/engine/gecko/NestedGeckoView.kt) that is hard to understand and debug today, so requires a lot of time to understand before development can begin there.

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
jonalmeida: > This is a good point, but if we wanted we could pretty easily create profiling middleware that let us understand this with more rich detail rather than trying to catch it in production.

We have profile tooling today already. So we start changing Stores to use this optional behaviour, the developer should be testing their code before landing it.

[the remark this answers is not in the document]
MatthewTighe: Out of curiosity, do we have profile tooling specifically for Stores and Middleware? I was definitely thinking we could create a middleware that make use of the Firefox Profiler through markers etc, but if that already exists that's great to know

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
MatthewTighe: > In our tests, we use the MainCoroutineRule so that all main tasks are done on the same test thread, so that would make all main work blocking in the sequence that they are executed:

This is tangential, but just for clarity the problem I would be interested in solving is when not all tasks are done on the main thread. For example, I don't think this test will pass even after this proposal:

```kotlin
@Test
    fun `middleware does not finish executing`() = runTestOnMain {
        var captured = false
        val middleware = object : Middleware<TestState, TestAction> {
            override fun invoke(
                context: MiddlewareContext<TestState, TestAction>,
                next: (TestAction) -> Unit,
                action: TestAction
            ) {
                when (action) {
                    TestAction.IncrementAction -> this@runTestOnMain.launch(Dispatchers.IO) {
                        delay(5000)
                        captured = true
                    }
                    else -> Unit
                }
            }
        }
        val store = Store(
            TestState(0),
            ::reducer,
            listOf(middleware),
        )
        store.dispatch(TestAction.IncrementAction).joinBlocking()

        assertTrue(captured)
    }
```

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
jonalmeida: No, it's just kotlin-code, so any regular debugger or profiler would work. Out of curiosity, what would a Store/Middleware-specific profile provide?

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
jonalmeida: That's correct, we are not solving for that use case. If the developer is providing their own dispatcher for the side-effects, it's up to them to pass that in through the constructor ([example](https://searchfox.org/mozilla-mobile/rev/98a241402bf6ecfa7fdd302006604393e02b9d50/firefox-android/android-components/components/feature/downloads/src/test/java/mozilla/components/feature/downloads/DownloadMiddlewareTest.kt#56)) or other means for the test.

jonalmeida:
@MatthewTighe after speaking offline with Christian, he provided an alternative solution to the problem of passing in either a dispatcher or config flag:

We don't have any valid use case for having separate dispatchers and our aim to always run the stores on the UI thread and to make that very clear to the developer, we can create a `UiStore` that extends the current one that runs with `Dispatchers.Main.immediate`.

This solves two problems:
1. Making it clear that you are using a UiStore, so expect that you will block the UI thread if you do work on that.
2. The `Dispatchers.Main.immediate` will ensure we are always processing the actions immediately from when it was called. We do not end up in a state where the action has been put into the queue and not yet processed. It'll also completely remove the need for `joinBlocking` as well.

I will update the RFC on Monday to reflect this change. Happy weekend!

docs/rfcs/0011-customizable-store-dispatcher.md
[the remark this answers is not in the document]
jonalmeida: With the design changes in the updated RFC, the expectation of these changes should be more clear now so that reading the state (not only by flow observers) right after dispatching an action will produce the state change immediately.

> You will still get the results from observations. Maybe I need to rephrase this paragraph better.. 

Updated.

docs/rfcs/0012-introduce-ui-store.md
```diff
@@ -0,0 +1,145 @@
...
+    action: Action,
+  ) {
+    // 📝 perform side-effects that now happen on the main thread.
+  }
+}
+
+// main thread
+store.flowScoped { flow ->
+  flow.collect {
+    MainScope().launch { // 📝 no longer necessary to launch a new coroutine unless desired.
```
MatthewTighe, line 93: Should be noted that `flowScoped` [launches](https://searchfox.org/mozilla-mobile/rev/f6d5e81ad351bc641cdbc3b8bc9595cf7d5e9ed3/firefox-android/android-components/components/lib/state/src/main/java/mozilla/components/lib/state/ext/StoreExtensions.kt#217) a coroutine already

docs/rfcs/0012-introduce-ui-store.md
```diff
@@ -0,0 +1,145 @@
...
+    reducer: Reducer<S, A>,
+    middleware: List<Middleware<S, A>> = emptyList(),
+    threadNamePrefix: String? = null,
+) : Store<S, A>(
+    initialState,
+    reducer,
+    middleware,
+    threadNamePrefix,
+) {
+  override val dispatcher = Dispatchers.Main.immediate
```
MatthewTighe, line 117: For this `override` to work, we'd need to `open` the dispatcher on the `Store`, which would allow any `Store` subtype to have complete control over its threading model. If we're going that route, I'd still prefer moving the dispatcher to be an optional constructor parameter to the `Store` instead of using inheritance because I think it makes for cleaner code and easier overrides under test.

For example:
```
// this implementation:
val ioStore = object : Store<State, Action> {
    override val dispatcher = Dispatcher.IO
}

// vs constructor implementation
val ioStore = Store<State, Action>(dispatcher = Dispatcher.IO)
```

docs/rfcs/0012-introduce-ui-store.md
```diff
@@ -0,0 +1,145 @@
...
+title: Introduce a Store for UI components
+permalink: /rfc/0012-introduce-ui-store
+---
+
+* Start date: 2024-01-29
+* RFC PR: [#5353](https://github.com/mozilla-mobile/firefox-android/pull/5353)
+
+## Summary
+
+In most applications of the `Store`, it is preferable to have reducers perform work on the main thread. Having actions reduced immediately at the point of dispatch, simplifies the reasoning a developer would need to go through for most UI-based work that happens on the main thread.
```
MatthewTighe, line 12: > Having actions reduced immediately at the point of dispatch

I do agree with this, but I think this still misses the piece of observation. Even if we reduce immediately, if we rely on `Flow` collection to observe the results at the UI level then we still don't have guarantees about immediacy.

MatthewTighe (commented):
After reading through these latest changes, I'm starting to wonder if perhaps what we are looking for is a way to get immediate results from a Store? I don't think this proposal quite solves that problem because we still rely on `Flow`s which are `launch`ed to observe the results of the reducer chain, and may not receive updates immediately based on current thread/coroutine state.

If my understanding here is correct, perhaps we just need a new version of `dispatch` which is blocking and returns the resultant state? Something like:

`fun dispatchBlocking(val action: A) : S`

This last part of your comment leads me to believe that might be more what we're looking for:

> It'll also completely remove the need for joinBlocking as well.

With the current updates in the proposal, this still wouldn't be true as the signature of `dispatch` is still:

`fun dispatch(action: A): Job`

Introducing a blocking dispatch would beg the question of what changes would be needed to protect the reducer chain from long-running operations.

docs/rfcs/0012-introduce-ui-store.md
MatthewTighe: And we could still add a specific `UiStore` for clarity:

```
open class UiStore<S, A> : Store<S, A>(dispatcher = Dispatcher.Main.immediate)
```

jonalmeida:
> After reading through these latest changes, I'm starting to wonder if perhaps what we are looking for is a way to get immediate results from a Store? I don't think this proposal quite solves that problem because we still rely on `Flow`s which are `launch`ed to observe the results of the reducer chain, and may not receive updates immediately based on current thread/coroutine state.
> 
> If my understanding here is correct, perhaps we just need a new version of `dispatch` which is blocking and returns the resultant state? Something like:
> 
> `fun dispatchBlocking(val action: A) : S`
> 
> This last part of your comment leads me to believe that might be more what we're looking for:
> 
> > It'll also completely remove the need for joinBlocking as well.
> 
> With the current updates in the proposal, this still wouldn't be true as the signature of `dispatch` is still:
> 
> `fun dispatch(action: A): Job`
> 
> Introducing a blocking dispatch would beg the question of what changes would be needed to protect the reducer chain from long-running operations.

We spoke offline about this review, so summarizing here:
1. Flow observers are always async so that will not change. 
2. We do not want more store types that change the dispatchers to other ones because:
   1. We don't have any use cases for it currently.
   2. It increases complexity for developers to understand what is happening if different dispatchers are used.
4. In a `UiStore.dispatch` the action will be processed immediately. We discussed having a `dispatchBlocking` but we would need to `join` on another CoroutineScope which means:
    1. We are changing the `dispatch` API that returns a `Job`. Once the `dispatch` is run, we can already access the `store.state` immediately after to see the changes.
    2. Also increasing complexity for a new API change. While prototyping this change, we found that it made it more restrictive as we would have to change all `dispatch`s to `dispatchBlocking` in current code.
6. TODO: Update the example `UiStore` to show how we override the dispatcher while being restrictive.

docs/rfcs/0012-introduce-ui-store.md
jonalmeida: Addressed [here](https://github.com/mozilla-mobile/firefox-android/pull/5353#issuecomment-1974174872).

docs/rfcs/0012-introduce-ui-store.md
jonalmeida: Addressed [here](https://github.com/mozilla-mobile/firefox-android/pull/5353#issuecomment-1974174872).

docs/rfcs/0012-introduce-ui-store.md
```diff
@@ -0,0 +1,163 @@
...
+// main thread
+store.flowScoped { flow ->
+  flow.collect {
+      // perform work on the main thread.
+    }
+  }
+}
+```
+
+Additionally, from [performance investigations already done][2], we know that Fenix creates over a hundred threads within a few seconds of startup. Reducing the number of threads for Stores that do not have a strong requirement to run on a separate thread will lower the applications memory footprint.
```
jonalmeida, line 99: @kycn I noted this bit here as well as it's something that'll help us make incremental wins once this lands and if you're interested in working on this.

jonalmeida: There are some quick wins in this list of the Fenix/AC threads: https://github.com/mozilla-mobile/fenix/blob/main/docs/List-of-fenix-threads.md

These don't include the Store created threads which we probably have quite a few on startup. Let me know if you want help in working on reducing the Store threads!

docs/rfcs/0012-introduce-ui-store.md
```diff
@@ -0,0 +1,161 @@
...
+    initialState: S,
+    reducer: Reducer<S, A>,
+    middleware: List<Middleware<S, A>> = emptyList(),
+    threadNamePrefix: String? = null,
+) {
+  internal open fun getDispatcher(): CoroutineDispatcher {
+    return Executors.newSingleThreadExecutor(threadFactory).asCoroutineDispatcher()
+  }
+}
+```
```
MatthewTighe, line 132: Another alternative to this would be an internal constructor in which we can specify the dispatcher. Personally I would prefer that to an inheritance solution

```kotlin
open class Store<S : State, A : Action> internal constructor(
    initialState: S,
    reducer: Reducer<S, A>,
    middleware: List<Middleware<S, A>> = emptyList(),
    threadNamePrefix: String? = null,
    specifiedDispatcher: CoroutineDispatcher? = null,
) {
    
    constructor(
        initialState: S,
        reducer: Reducer<S, A>,
        middleware: List<Middleware<S, A>> = emptyList(),
        threadNamePrefix: String? = null,
    ) : this(initialState, reducer, middleware, threadNamePrefix, null)
    
    private val threadFactory = StoreThreadFactory(threadNamePrefix)
    private val dispatcher = specifiedDispatcher ?: Executors.newSingleThreadExecutor(threadFactory).asCoroutineDispatcher()
    private val reducerChainBuilder = ReducerChainBuilder(threadFactory, reducer, middleware)
    private val scope = CoroutineScope(dispatcher)

open class UiStore<S : State, A : Action>(
    initialState: S,
    reducer: Reducer<S, A>,
    middleware: List<Middleware<S, A>> = emptyList(),
    threadNamePrefix: String? = null,
) : Store<S, A>(
    initialState,
    reducer,
    middleware,
    threadNamePrefix,
    specifiedDispatcher = Dispatchers.Main.immediate
) {

```

docs/rfcs/0012-introduce-ui-store.md
MatthewTighe: I will just note tangentially that this specific concern would be eliminated if we moved to a full Redux, single-store model, which I would still like to investigate doing (at least in Fenix).

However, this will require quite a bit of prototyping and investigation to make sure we can create ephemeral state that goes out of scope based on the screen so we don't pollute users memory unnecessarily.

Even if we do go that route eventually, I don't foresee using `UiStore`s in the interim from adding an difficulty to that migration.

MatthewTighe (commented):
Could we get some additional detail about how `dispatch` will change for the `UiStore`? Based on our conversation offline and this comment:

> In a UiStore.dispatch the action will be processed immediately. We discussed having a dispatchBlocking but we would need to join on another CoroutineScope which means:
> We are changing the dispatch API that returns a Job. Once the dispatch is run, we can already access the store.state immediately after to see the changes.
> Also increasing complexity for a new API change. While prototyping this change, we found that it made it more restrictive as we would have to change all dispatchs to dispatchBlocking in current code.

Is the intent to change the underlying `Store.dispatch` function, or will we somehow override the return type in `UiStore` to be `Unit` instead of `Job`?

Finally - I think I understand pretty well the _what_ here: we want the ability to dispatch an action and immediately examine the updated State in the Store. For example:

```
uiStore.dispatch(ActionThatUpdatesFlag)
val result = store.state.flag // we have a guarantee that flag has been updated here
```

However, I think we could still use a little more detail as to _why_ we want this. It seems counter to the typical reactive UI style achieved through observation that is advised these days, even in our own docs. For example, we suggest [being reactive](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/docs/Architecture-Decisions.md#decision-1) and [observing the Store](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/docs/architecture-overview.md#store).

My assumption is there are corner cases I'm not aware of - like maybe wanting to update the State during a GV delegate callback and immediately inspect the result, but I would still appreciate a clear example here for my own understanding.

docs/rfcs/0012-introduce-ui-store.md
jonalmeida: Yeah, let's talk about a single store model separately. I have not personally seen any proposal that describes what the _concrete_ benefits our.

> > Even if we do go that route eventually, I don't foresee using UiStores in the interim from adding an difficulty to that migration.

I find this puzzling. Isn't it a bit quick to discount this while we are doing this work to make UI screen stores better? cc: @csadilek if you want to add more clarity that I might have missed in this doc or thread.

docs/rfcs/0012-introduce-ui-store.md
jonalmeida: Thanks for documenting our offline discussion here, I forgot about this!

We didn't want to use this approach because it allows for overriding the dispatcher in the constructor which is the footgun we want to avoid anyone from using and why the RFC was updated not to include that as an option.

jonalmeida: Matt clarified offline, and I missed this (sorry!), that the default constructor is `internal` that has the dispatcher. So yes, this would work.

Both directions solve the same problem, so I have no preference between the styles. I'll update the RFC with this one, thanks!

jonalmeida:
> Could we get some additional detail about how `dispatch` will change for the `UiStore`? Based on our conversation offline and this comment:
> 
> > In a UiStore.dispatch the action will be processed immediately. We discussed having a dispatchBlocking but we would need to join on another CoroutineScope which means:
> >    i. We are changing the dispatch API that returns a Job. Once the dispatch is run, we can already access the store.state immediately after to see the changes.
> >    ii. Also increasing complexity for a new API change. While prototyping this change, we found that it made it more restrictive as we would have to change all dispatchs to dispatchBlocking in current code.
> 
> Is the intent to change the underlying `Store.dispatch` function, or will we somehow override the return type in `UiStore` to be `Unit` instead of `Job`?

Sorry, this might be have been poor wording on my part: we don't want to change this API. The `Dispatchers.Main.immediate` will do all the work for us to ensure the work is completed before the function is done with the dispatch call. We don't have to change the return type or try to do any `join` call to get this for free.


> Finally - I think I understand pretty well the _what_ here: we want the ability to dispatch an action and immediately examine the updated State in the Store. For example:
> 
> ```
> uiStore.dispatch(ActionThatUpdatesFlag)
> val result = store.state.flag // we have a guarantee that flag has been updated here
> ```
> 
> However, I think we could still use a little more detail as to _why_ we want this. It seems counter to the typical reactive UI style achieved through observation that is advised these days, even in our own docs. For example, we suggest [being reactive](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/docs/Architecture-Decisions.md#decision-1) and [observing the Store](https://github.com/mozilla-mobile/firefox-android/blob/main/fenix/docs/architecture-overview.md#store).
> 
> My assumption is there are corner cases I'm not aware of - like maybe wanting to update the State during a GV delegate callback and immediately inspect the result, but I would still appreciate a clear example here for my own understanding.

This is also my bad wording again. In that sample code, the `result` is meant to indicate that the store.state will have already completed the reducing work by the next line. We are not trying to change the paradigm of observing a store and reacting to it. What we get is a simpler thread model to understand for stores that are typically made for each screen. So you can still observe changes and can reason immediately that a dispatch and an observation below it are happening in that order.

Maybe a better example to help understand this is:

```kotlin
uiStore.dispatch(UpdateState)
uiStore.flowScoped { flow ->
  // We know here that the dispatch would have completed whenever we have an observer here.
}

regularStore.dispatch(UpdateState)
regularStore.flowScoped { flow ->
  // UpdateState may not have been completed by the time we add this observer.
}
```

In the `regularStore` case, we could possibly get the initial state and then the update, and we need to then make sure we can handle these cases in our UI code to not have duplicate UI updates (debounces for example). In these UI stores, we don't need to give that complexity to the developers for those simple use cases.

docs/rfcs/0012-introduce-ui-store.md
```diff
@@ -0,0 +1,161 @@
...
+
+Additionally, from [performance investigations already done][2], we know that Fenix creates over a hundred threads within a few seconds of startup. Reducing the number of threads for Stores that do not have a strong requirement to run on a separate thread will lower the applications memory footprint.
+
+## Guide-level explanation
+
+Extending the existing `Store` class to use the `Dispatchers.Main.immediate` will ensure that UI stores will stay on the same UI thread and have that work done immediately. Using a distinct class named `UiStore` also makes it clear to the developer that this is work that will be done on the UI thread and its implications will be made a bit more clear when it's used.
+
+```kotlin
+@MainThread
+open class UiStore<S : State, A : Action>(
```
MatthewTighe, line 105: One other small implementation detail note while I'm here:

My preference would be for subtypes of UiStore to be typealiases, and leave `UiStore` as a closed class. This is again because I think inheritance makes code harder to trace and easier to introduce issues. For example, I think this is equivalent to the use cases you're imagining:

```
class UiStore<S : State, A : Action>(...) : Store<S, A>

typealias FragmentStore = UiStore<FragmentState, FragmentAction>
```

MatthewTighe (approved):
Ah, the piece I was missing in my understanding is that the stack trace of `dispatch` isn't suspendable, since the reducer chain is just a series of non-suspending functions. I had assumed that because it was running in a coroutine, any suspension point (like `launch`ing in a middleware) could prove problematic, and remove the guarantee you've laid out here:

```
uiStore.dispatch(UpdateState)
uiStore.flowScoped { flow ->
  // We know here that the dispatch would have completed whenever we have an observer here.
}
```

I had to figure out a way to test this, and added this code to `HomeActivity::onCreate`:

```
    // onCreate
    var count = 0
    for (i in 0..1000) {
        if (logTest()) count ++
    }
    Log.i("tighe", "count: $count")

    // elsewhere
    private fun logTest(): Boolean {
        var x = 0

        CoroutineScope(Main.immediate).launch {
            CoroutineScope(Dispatchers.Default).launch {
                delay(1)
            }
            x = 1
        }

        return x == 1
    }
```

I'm still not entirely sure what good use cases exist for observing after a call to dispatch (as opposed to just starting the observations along with the lifecycle of the UI), but I don't think we need to keep belaboring that point. I'll add my approval now, given that there are obvious benefits in reducing overall thread count and context switching, and that we can prevent footguns in overriding `dispatch` through clever visibility modifiers. 

Thanks for your patience in helping me understand all the details here!

rahulsainani (approved):
I think we could remove the `dispatcherWithExceptionHandler` for UIStore since we're already on the main dispatcher and an exception in the reducer would simply crash the app anyway.

I think removing the logic that creates a thread simply to convert it to a dispatcher makes a lot of sense for UI stores. Fenix can benefit from simply using UIStore for all the screen based stores, A quick search for `FragmentStore` in Fenix shows up 14 screen based stores. Off the top of my head, I think `ReviewQualityCheckStore` and `TabsTrayStore` can use this. 

As an iteration,  it would benefit to inject `CoroutineScope` instead of Dispatcher, the same scope can be passed to the middleware. Internally middlewares can use whatever dispatcher making the middleware function main safe.

jonalmeida:
> I think we could remove the `dispatcherWithExceptionHandler` for UIStore since we're already on the main dispatcher and an exception in the reducer would simply crash the app anyway.

That's a good idea! While it may not seem necessary to make this change because re-throwing on the main thread is fine as-is, this change fits in nicely with the other suggestion @MatthewTighe had added about using an internal constructor to pass in the dependant dispatcher.

> I think removing the logic that creates a thread simply to convert it to a dispatcher makes a lot of sense for UI stores. Fenix can benefit from simply using UIStore for all the screen based stores, A quick search for `FragmentStore` in Fenix shows up 14 screen based stores. Off the top of my head, I think `ReviewQualityCheckStore` and `TabsTrayStore` can use this.

💯 

While CPU performance is not a driver for this work, I've run our cold startup tests on a release build with a prototype that simply switches the Fenix `AppStore` to use a `UiStore` that gives a 6% perf gain. As mentioned in the RFC already, a developer should avoid doing non-UI work on the main thread, so the same rules apply when they use the `UiStore`.

<details>

<summary>Pixel 6a, Fenix Release against cc65693e4f</summary>

**Before:**
```
{'max': 1238.0,
 'mean': 1143.4,
 'median': 1149.5,
 'min': 1073.0,
 'replicate_count': 10,
 'replicates': [1154.0, 1073.0, 1075.0, 1084.0, 1114.0, 1215.0, 1145.0, 1238.0,
                1161.0, 1175.0]}

```

**After:** 
```
{'max': 1190.0,
 'mean': 1080.5,
 'median': 1073.0,
 'min': 1017.0,
 'replicate_count': 10,
 'replicates': [1190.0, 1091.0, 1067.0, 1055.0, 1017.0, 1064.0, 1079.0, 1036.0,
                1100.0, 1106.0]}

```

</details>


> As an iteration, it would benefit to inject `CoroutineScope` instead of Dispatcher, the same scope can be passed to the middleware. Internally middlewares can use whatever dispatcher making the middleware function main safe.

Thanks! I've included this bit in the RFC within a new section 'Future work'.

docs/rfcs/0012-introduce-ui-store.md
```diff
@@ -0,0 +1,177 @@
...
+    threadNamePrefix: String? = null,
+  ) : this(
+    initialState = initialState,
+    reducer = reducer,
+    middleware = middleware,
+    dispatcher = DefaultStoreDispatcher(threadNamePrefix),
+  )
+}
+
+interface StoreDispatcher {
```
jonalmeida, line 135: @rahulsainani @MatthewTighe this is the abstraction that I've added here which allows us to pull out the exception handler and solves for one of the 'Unresolved questions' around the `StoreThreadFactory`.

Note how `StoreDispatcher` has a `DefaultStoreDispatcher` that will hold all the existing code in the current Store, and the `UiStoreDispatcher` will be a simplified version that holds references to the main thread.

rahulsainani: We can continue to discuss this after the RFC lands as it's implementation. 

In this interface, we could only expose`scope` and the rest is implemented in `DefaultStoreDispatcher` and `UiStoreDispatcher`

Mugurell:
Being a bit late to the party - seems like the final plan has been set and approved I want to just ask if there are any recommendations for deciding which stores have to be a `UIStore`?

jonalmeida:
Hi @Mugurell ! Surprised the comments are still open here.

> Being a bit late to the party - seems like the final plan has been set and approved I want to just ask if there are any recommendations for deciding which stores have to be a `UIStore`?

The recommendations are to start with 'UI screens' like Rahul hinted [above](https://github.com/mozilla-mobile/firefox-android/pull/5353#pullrequestreview-1930487210): stores that end with "FragmentStore" are excellent examples of ones that we can switch.

I did try in a prototype to use it on the AppStore too [here](https://github.com/mozilla-mobile/firefox-android/pull/5353#issuecomment-1992654999), but I'd feel better if we validate that it works on smaller individual screens before we move over the AppStore since it's become a very core part of the app now.

EDIT: copying over this discussion to [the phabricator implementation review](https://phabricator.services.mozilla.com/D211800?vs=on&id=874700#7318267) so we can close the loop here and talk about it more in one place.

PR #5378, opened 2024-01-31
Bug 1874522 - Repalce coil-kt with AndroidSVG

jackyzy823:
1. Use AndriodSVG directly
    It is safe to use AndroidSVG , because it is the dependency of coil-svg

    Since android-components has its own cache mechanism , using coil-kt is just bloating.

2. Catch OOM error just like https://github.com/mozilla-mobile/firefox-android/blob/76213dc84f49a0400e009ea06e3d09fd3ad68ded/android-components/components/support/images/src/main/java/mozilla/components/support/images/decoder/AndroidImageDecoder.kt#L36
3. Set target bitmap size according to desiredSize to avoid svg too large which could also cause OOM

The code is **largely borrowed** from  https://github.com/coil-kt/coil/blob/2.4.0/coil-svg/src/main/java/coil/decode/SvgDecoder.kt with some fixed options.

( **However  I'm not sure if this patch will really fix the bug. and also i'm not sure if the bug was caused by using coil-svg**)

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1879116

### DIFF
--- a/android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/decoder/SvgIconDecoder.kt
+++ b/android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/decoder/SvgIconDecoder.kt
@@ -6,55 +6,70 @@ package mozilla.components.browser.icons.decoder
 
 import android.content.Context
 import android.graphics.Bitmap
-import androidx.core.graphics.drawable.toBitmap
-import coil.ImageLoader
-import coil.decode.SvgDecoder
-import coil.executeBlocking
-import coil.request.CachePolicy
-import coil.request.ImageRequest
+import android.graphics.Bitmap.Config.ARGB_8888
+import android.graphics.Canvas
+import android.graphics.RectF
+import androidx.core.graphics.createBitmap
+import com.caverock.androidsvg.SVG
+import com.caverock.androidsvg.SVGParseException
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.images.DesiredSize
 import mozilla.components.support.images.decoder.ImageDecoder
 
 /**
- * [ImageDecoder] that will use the Coil library [SvgDecoder.Factory] in order to decode the byte data.
+ * [ImageDecoder] that will use the AndroidSVG in order to decode the byte data.
  *
- * ⚠️ For guidance on use of the Coil library see comment for [ComponentsDependencies.thirdparty_coil_svg].
+ * The code is largely borrowed from [coil-svg](https://github.com/coil-kt/coil/blob/2.4.0/coil-svg/src/main/java/coil/decode/SvgDecoder.kt)
+ * with some fixed options.
  */
 class SvgIconDecoder(val context: Context) : ImageDecoder {
     private val logger = Logger("SvgIconDecoder")
 
     override fun decode(data: ByteArray, desiredSize: DesiredSize): Bitmap? =
         try {
-            val request = ImageRequest.Builder(context)
-                .size(desiredSize.targetSize)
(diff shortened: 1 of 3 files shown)


Roukanken42:
As I have managed to reproduce this bug consistently on any other build and both my android devices, I also tested build from this PR's CI artefacts and it fixes both ways I could reproduce the issue.

t-p-white:
@Roukanken42 🙌 Can I confirm [your comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1874522#c29) on Bugzilla that using the [patch here](https://bugzilla.mozilla.org/show_bug.cgi?id=1874522#c28) also resolves the bug?

Roukanken42:
@t-p-white Yes, the patch also resolves the bug.

t-p-white:
@Roukanken42 perfect, thank you for confirming 🙏 

@jackyzy823, nice work 👌  please see my latest comment on Bugzilla regarding this particular change - but essentially I think we should use your original patch for `SvgIconDecoder` for this bug and create a follow on if a library update is required 🙂

t-p-white (approved):

t-p-white:
Apologies, I got this PR mixed up with https://github.com/mozilla-mobile/firefox-android/pull/5443

gabrielluong:
@t-p-white I am wondering if you will have any capacity to review this?

t-p-white:
Thanks for the ping @gabrielluong. Yes, it's on my radar todo 🙂

android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/decoder/SvgIconDecoder.kt
```diff
@@ -6,55 +6,67 @@ package mozilla.components.browser.icons.decoder
...
+import androidx.core.graphics.createBitmap
+import com.caverock.androidsvg.SVG
+import com.caverock.androidsvg.SVGParseException
 import mozilla.components.support.base.log.logger.Logger
 import mozilla.components.support.images.DesiredSize
 import mozilla.components.support.images.decoder.ImageDecoder
 
 /**
- * [ImageDecoder] that will use the Coil library [SvgDecoder.Factory] in order to decode the byte data.
+ * [ImageDecoder] that will use the AndroidSVG in order to decode the byte data.
```
t-p-white, line 20: >The code is largely borrowed from https://github.com/coil-kt/coil/blob/2.4.0/coil-svg/src/main/java/coil/decode/SvgDecoder.kt with some fixed options.

Can you add a comment like this to the fun description please?

android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/decoder/SvgIconDecoder.kt
```diff
@@ -6,55 +6,67 @@ package mozilla.components.browser.icons.decoder
...
+            } else {
+                svgWidth = svg.documentWidth
+                svgHeight = svg.documentHeight
+            }
+
+            var bitmapWidth = desiredSize.targetSize
+            var bitmapHeight = desiredSize.targetSize
+
+            if (svgWidth > 0 && svgHeight > 0) {
+                val widthPercent = bitmapWidth / svgWidth.toDouble()
```
t-p-white, line 45: Nice 👌 
Perhaps a comment on this block will be useful for future maintainers? Maybe something like "Scale the bitmap to SVG maintaining the aspect ratio"

t-p-white (approved):
Thanks @jackyzy823!

PR #5380, opened 2024-01-31
Bug 1877639 - Add more test logs to FindInPageRobot

AndiAJ:
Summary:

- As part of the Android test hardening, to ease the debugging process, I've added test logs to the FindInPageRobot
- Removed some unused functions
- Both `verifyFindNextInPageResult` and `verifyFindPrevInPageResult` did the same assertions so I've left only one function and renamed it to `verifyFindInPageResult`

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1877639


fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/FindInPageRobot.kt
```diff
@@ -25,42 +27,39 @@ import org.mozilla.fenix.helpers.ext.waitNotNull
...
-
-    fun verifyFindInPageSearchBarItems() {
-        verifyFindInPageQuery()
-        verifyFindInPageNextButton()
-        verifyFindInPagePrevButton()
-        verifyFindInPageCloseButton()
+    fun verifyFindInPageNextButton() = assertFindInPageNextButton()
+    fun verifyFindInPagePrevButton() = assertFindInPagePrevButton()
+    fun verifyFindInPageCloseButton() = assertFindInPageCloseButton()
+    fun clickFindInPageNextButton() {
```
jjSDET, line 33: I'm approving this PR, but let's get a follow-up on this to avoid what looks like an unnecessary layer of abstraction here.

It looks like we're effectively just renaming `assertFindInPageNextButton`, when we could use it as-is.

If we want to instead have all of our assertions in methods named `verify`, then we should create a PR that renames this from `assert` to `verify` and update everywhere it's used.

With that said,  even `assertFindInPageNextButton` should probably be a lower-level method that takes a selector for the action being performed.

For example, what's used below in `findInPageNextButton().click()` could instead be `PageObject.mozClick(EspressoUIElement: inPageNextButton)` and `mozClick` would have our assertion in it like this:
```
fun mozClick(elementSelector: EspressoUIElement) {
    mozWaitForElement(elementSelector)
    Log.i("$elementSelector attempting to be clicked...")
    elementSelector.click()
    Log.i("$elementSelector click event successful! moving on...")
}
 fun mozWaitForElement(elementSelector: EspressoUIElement) {
    Log.i("waiting for $elementSelector to exist...")
    waitForElementToExist(elementSelector)
    Log.i("$elementSelector was found! moving on...")
}
suspend fun mozVerifyElements(shouldBePresent: Boolean, vararg elementSelectors: EspressoUIElement) {
    coroutineScope {
        elementSelectors.forEach { elementSelector ->
            launch {
                mozWaitForElement(elementSelector)
                if (shouldBePresent) {
                    assertTrue(elementSelect.exists(), "Element $elementSelector should be present but is not."
                } else {
                    assertFalse(elementSelector.exists(), "Element $elementSelector should not be present but is")
                }
            }
        }
    }
}
```

In the above example, if you wanted need simply to verify multiple elements at the same time, like if you're at the end of a test and want to verify the page/component state, you could use `mozVerifyElements` to pass an array of elements to assert they exist asyncronously.

However, if you're simply wanting to assert and element exists, then something like this could be used instead:

```
mozVerify(elementSelector: EspressoUIElement, condition: Boolean) {
    mozWaitForElement(elementSelector)
    if (shouldBePresent) {
                    assertTrue(elementSelector.exists(), "Element $elementSelector should be present but is not")
    } else {
        assertFalse(elementSelector.exists(), "Element $elementSelector should not be present but is")
    }
}
```

With that said, it's totally fine to not abstract your `assertTrue` and `assertFalse` in your tests at all, and is often preferred as it makes the expectations of your test more explicit. I'd only use the `mozVerifyElements` methods I recommended for necessary assertions along the way to the end of your test and keep the final `significant` assertions that are the last actions you're making in your test simply use `assertTrue` and `assertFalse`.

That way, all of the tests will be consistent with the way they are written and you will at at-a-glance if a `utility` assertion (one of the assertions you needed along the way in your test) failed or a `significant` assertion (the ultimate intent of what you're trying to test) failed in your test.

jjSDET (approved):
Please see comment to create a follow-up story to address instances of unnecessary abstractions we can condense to lower-level customer commands and utility functions.

PR #5403, opened 2024-02-01
Bug 1877939 - Remove redundant assertion functions from EnhancedTrackingProtectionRobot

AndiAJ:
Summary:

- As part of the Android test hardening, I've removed the redundant assertion functions

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1877939


fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/EnhancedTrackingProtectionRobot.kt
```diff
@@ -38,10 +38,29 @@ import org.mozilla.fenix.helpers.isChecked
...
+        mDevice.waitNotNull(Until.findObjects(By.text("Protections are $status for this site")))
+        onView(ViewMatchers.withResourceName("switch_widget")).check(
+            matches(
+                isChecked(
+                    state,
+                ),
+            ),
+        )
+        Log.i(TAG, "verifyEnhancedTrackingProtectionSheetStatus: Verified ETP toggle is checked: $state")
+    }
```
jjSDET, line 51: I haven't mentioned it up to this point, but I'd really like to see pairs of logs around the action/wait/assert we're doing.

This is especially helpful when determining when a race condition occurred; so we'll know with 100% confidence when we've entered into some action/wait/assert before we intended.

The general form of our logging that will help us catch these will be:
1. `Now Beginning $action...`
2. `$action was successful...`

These logs are typically more helpful in our `custom-commands`, `custom-matchers`, and `higher-order functions`, but can also be helpful at this scope.

We need to just make sure we're adding the logging in a way that we know could give us a specific insight into how a failure might be made more visible, or also how, when successful, this logging helps confirm there wasn't an issue here.

Because of that I'd like to see a `Log` indicating this verification has begun somewhere at the beginning of this.

The `Log` for this might be something like `Attempting to verify status: $status in state: $state`, or whatever we want here.

The specific purpose of that will be to help us know there never should be two `Log`s next to each other, one after another, that say `Attempting to verify...`, because if we see that we will know we've unintentionally begun two verifications at once and thus are looking at a race condition.

No need to address that here specifically, but we need to `norm` on what consistent implementation of wording we want to use for our `logging pairs`.

@oana let's make sure to sync on this together to find a pattern we can try out and see what will be most useful.

jjSDET (approved):
approving with comment on syncing to create a norm for us to follow for our logging pairs and general strategy.

fenix/app/src/androidTest/java/org/mozilla/fenix/ui/robots/EnhancedTrackingProtectionRobot.kt
AndiAJ: Pushed some changes today, based on the discussion we had in yesterday's meeting.

sv-ohorvath (approved):

PR #5406, opened 2024-02-01
Bug 1812843 - Maintain aspect ratio when resizing a bitmap for the 'add' button in CustomTabsToolbarFeature

t-p-white:
@MozillaNoah attempt no. 2

~~ℹ️  Removed the Bitmap scaling code added in [this patch](https://github.com/mozilla-mobile/android-components/pull/8536). I am unable to reproduce the conditions using the original STR however I've tested this current scaling method by manually setting the Bitmap size bigger than the icon height and there doesn't _appear_ to be any regression.
cc @mcarare~~ 

The added margin ensures that a wider icon (e.g. Twitter) has spacing from the URL.

🆕 I was wrong - I actually was able to reproduce this using a programmatically generated large Bitmap. We need to keep the original Bitmap scaling, which I've updated to maintain aspect ratios. Have not updated the screenshots, as the additional `resizeMaintainingAspectRatio` produces the same output.

Original:

Pocket OG

Share OG

X squash

With fix no margin:

Pocket no margin

Share button no margin

X no margin

With fix has margin:

Pocket with margin

Share button with margin

X with margin

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1812843


MozillaNoah (approved):
Round 2: FIGHT

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -301,8 +304,25 @@ class CustomTabsToolbarFeature(
         }
     }
 
+    private fun Bitmap.resizeMaintainingAspectRatio(): Size {
```
t-p-white, line 307: Ideally this would have been in [org.mozilla.fenix.ext.Bitmap](https://github.com/mozilla-mobile/firefox-android/blob/7d0c8cacb416c627eec2c1bb97b7d13964736130/fenix/app/src/main/java/org/mozilla/fenix/ext/Bitmap.kt#L4) but this would introduce a dependency on Fenix app

t-p-white:
Thanks for the review @MozillaNoah. I've updated based on some more testing (please see updated PR desc 🙂) - it's now essentially a combination of fix 1 & fix 2 with a pinch of aspect ratio

android-components/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeatureTest.kt
```diff
@@ -441,6 +441,84 @@ class CustomTabsToolbarFeatureTest {
...
+        verify(feature).addActionButton(anyInt(), any())
+        verify(toolbar).addBrowserAction(captor.capture())
+
+        val button = captor.value.createView(FrameLayout(testContext))
+        assertEquals(24, (button as ImageButton).drawable.intrinsicHeight)
+        assertEquals(48, button.drawable.intrinsicWidth)
+    }
+
+    @Test
+    fun `tall action button drawable is scaled to 12 width and 24 height`() {
```
MozillaNoah, line 484: let's get this in the "given, when, then" format. Maybe something like:

`GIVEN the action button is tall WHEN it's being drawn in the toolbar THEN scale it to 12 width and 24 height`

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
MozillaNoah: you could alternatively create a bitmap ext within `support-utils` if you feel this has value elsewhere

MozillaNoah (approved):

t-p-white:
Thanks @MozillaNoah for the re-review 🙏 I updated the test names as suggested, will leave the scaling function as private for now

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
t-p-white: Good suggestion 👍

PR #5389, opened 2024-01-31
Bug 1877205 - Translations Fetch Supported Languages Refactor

ohall-m:
This patch refactors a few things for fetching supported languages:

* `TranslateExpectedAction` is decoupled from fetching languages
* Now use `OperationRequestedAction` w/ `FETCH_SUPPORTED_LANGUAGES`
* Refactors the Fenix fragment to send `OperationRequestedAction` as a placeholder for data initialization
* `TranslateSetLanguagesAction` is now `SetSupportedLanguagesAction`

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1877205

PR #5078, opened 2024-01-08
Bug 1873511 - New `supportedLanguages` and `translationError` on TranslationsState, with `TranslateExceptionAction` Refactor

ohall-m:
This patch adds the additional state information of `supportedLanguages` and `translationError` to `TranslationsState`. The `supportedLanguages` are the languages the translation engine supports for translating for use in UI dropdowns. `translationError` is any error that occurred relating to translating.

`supportedLanguages` is currently set via an action that indicates the intention to translate on the new `EngineMiddleware` addition of `TranslationsMiddleware`. This patch also adds a new `TranslationOperation` of `FETCH_TO_AND_FROM_LANGUAGES` to indicate the process of getting `supportedLanguages`.

`translationError` is added to the `TranslationsState` to store errors as they arise. `TranslateExceptionAction` is also refactored from using the generic `Throwable` to the new `TranslationError` to accommodate `supportedLanguages` and better general error handling. 

### Pull Request checklist

- [X] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [X] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1873511

PR #5409, opened 2024-02-01
Bug 1877588 - TranslationsMiddleware Dispatcher Scope Issue

ohall-m:
This bug is to fix the `Must have a Handler` issue in the TranslationsMiddleware when attempting to call using `Dispatcher.IO`.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1877588


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -83,32 +83,30 @@ class TranslationsMiddleware(
...
-                            tabId = tabId,
-                            operation = TranslationOperation.FETCH_LANGUAGES,
-                            translationError = TranslationError.CouldNotLoadLanguagesError(it),
-                        ),
-                    )
-                    logger.error("Error requesting supported languages: ", it)
-                },
-            )
-        }
+    ) = withContext(MainScope().coroutineContext) {
```
ohall-m, line 86: Not sure of the recommended syntax here as `scope` also happens to be `MainScope` on this file.

ohall-m:
Bug background:
* Originally made this  `Dispatchers.IO` due to potential longer first query from this [conversation](https://github.com/mozilla-mobile/firefox-android/pull/5078#discussion_r1446243737)
* However, I had wrote 
```
= withContext(Dispatchers.IO) {
   scope.launch{
   // Stuff
  }
}
```

Where scope was Main scope, so it was launching in main.
* This double `launch` was spotted when adding another API [here](https://github.com/mozilla-mobile/firefox-android/pull/5348#discussion_r1470425716)
* I attempted to fix the double launch that in that PR (so we'd really be using the IO scope); however, we were seeing a `Must have a Handler` error in CI [failure](https://console.firebase.google.com/project/moz-fenix/testlab/histories/bh.66b7091e15d53d45/matrices/5104850940289471579/details?stepId=bs.e5ff33f6baef87f3&testCaseId=5)
* I left it alone to research more, and that leads to this bug.
* I was able to repeat that failure when manually testing and removing `scope.launch` and when refactoring #5389
* I have some research that points to not using GeckoResults outside of main scope:
  * ([Bug 1697248 comment 9](https://bugzilla.mozilla.org/show_bug.cgi?id=1697248#c9)
   * Patch for [that bug](https://github.com/mozilla-mobile/focus-android/pull/4968)
    * [This comment referenced on that patch](https://github.com/mozilla-mobile/fenix/issues/17045#issuecomment-745373480)) 
* Reverting back to main scope is the best lead I have. So, this patch puts it back in main scope. I think only `requestSupportedLanguages` has the potential to have a network call on first startup, but that is the only time it will be longer running. Please let me know if you know of other alternatives or know of some deeper context we can look at.

Other notes:
* [listSupportedLanguages() ](https://searchfox.org/mozilla-central/rev/5c4a45eb17423373ecb71aea9819d41a6231613e/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#167-168) is `static` and ` @AnyThread` on the GeckoView side. I added the `@AnyThread` annotation. I can't think of any reason it shouldn't be `@AnyThread` on that side. Same goes for the similar functions called in `requestTranslationPageSettings`.

ohall-m:
Pre-existing tests in [TranslationsMiddlewareTest.kt](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddlewareTest.kt). (Will be refactor some in #5389.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
gabrielluong: Coroutine and scopes are actually my greatest unknown and weakness. So, I will defer this to @MatthewTighe and @Amejia481, but happy to just follow along and learn something in the process.

Amejia481: I think the main idea of having a coroutines scope is defining until when your task should run until it can be automatically cancelled. For example, imagine that you have a task that fetches details of an item over the network when you are on a detailed screen, but what happens if the user leave the details screen, before the task completes? Ideally the task should be cancel as it's not needed anymore. That's the role of the coroutines scope, scoping the life of some coroutines.

To define the scope the question that should ask is until when this code should be run until it is cancelled,  
We have [different per-built scopes](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-coroutine-scope/) like  [LifecycleScope](https://developer.android.com/topic/libraries/architecture/coroutines#lifecyclescope) or we can create our [own custom scope depending](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-global-scope/) on our needs.

Amejia481: There are other middlewares that are also receiving scopes like the [LinkingMiddleware](https://github.com/mozilla-mobile/firefox-android/blob/main/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/LinkingMiddleware.kt#L27)

ohall-m: I think that might lend to using `= withContext(scope.coroutineContext)` then? So, it'll be tied to the scope of `TranslationsMiddleware.kt`?

Amejia481: When using `withContext` ideally we would like to be explicit about which [dispatcher](https://kotlinlang.org/docs/coroutine-context-and-dispatchers.html) the code should be executed like.

`withContext(Dispatchers.IO)` or `withContext(Dispatchers.Main)`, at the moment we are just extracting it from the scope.

Amejia481: it depends of the case, as you are already on suspend function, you don't need either `withContext` or `scope.launch` as you are already in the context of coroutine, see[ requestSupportedLanguages is getting from a coroutine](https://github.com/mozilla-mobile/firefox-android/blob/a83a0f49d3cb69718cf2663b12a43a59c37420c6/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt#L46). You use `withContext` or `scope.launch` when you want to specific a dispatcher that is different from the one called the suspend function.

Amejia481: So you can remove `withContext(MainScope().coroutineContext)` and the code should continue working.

ohall-m: Okay, that makes sense! Yes, I had tested without it at all and it works great. I guess I'm wondering if I need to be explicit about scopes so someone doesn't somehow end up [here](https://searchfox.org/mozilla-central/rev/5c4a45eb17423373ecb71aea9819d41a6231613e/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/GeckoResult.java#303) again down the road or if this is just a weird one-off odd scenario I landed myself in.

I'm also testing something with @kycn to try and keep a `Dispatcher.IO`. I think ideally we want to use that if possible.

Amejia481: You could pass the dispatcher directly to `scope.launch(Dispatchers.Main)` on [requestSupportedLanguages](https://github.com/mozilla-mobile/firefox-android/blob/a83a0f49d3cb69718cf2663b12a43a59c37420c6/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt#L46), I think gecko results are expecting to be run on main, but give it a try.

ohall-m: Sounds good, `scope.launch(Dispatchers.Main)` works. I set it to that since it is a bit more explicit. I set the patch to that for now and removed the `withContext`. (Open to whatever seems the best course of action you recommend here - I'm getting a bit bogged down in the nuanced differences here. Thanks, @Amejia481 for taking a look and giving all the background on the differences!)

@kycn and I had a [conversation over here](https://mozilla.slack.com/archives/G01EXNFFAVC/p1706798060455179) concurrently to this one and wanted to take a closer look tomorrow too. We're trying to salvage using `Dispatcher.IO` if it is possible too and not run into the GV issue.

Something like launching again might make sense out of `Dispatcher.IO` to give us the handler is what we were wondering:

This works as well and seems like it might be launching it correctly:
```
scope.launch{
    doSomething()
}

fun doSomething() = withContext(Dispatcher.IO){
    scope.launch {
      // Work
    }
  }
```

Thanks everyone for taking a look and the help!

ohall-m:
(Manual testing note, to see the  in action and the issues it causes, go to Fenix, turn on translations in secret settings, and open the sheet. If the dispatcher is set wrong, then it'll have the `Must have a Handler` crash. )

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -43,7 +42,7 @@ class TranslationsMiddleware(
         // Pre process actions
         when (action) {
             is TranslateExpectedAction -> {
-                scope.launch {
+                scope.launch(Dispatchers.Main) {
```
MatthewTighe, line 45: Since this scope already has its dispatcher set to Main in its context [here](https://searchfox.org/mozilla-mobile/rev/c3447daa62d50e8b05251a9610ea9611b66705c8/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/EngineMiddleware.kt#39), you shouldn't need to specify it explicitly

MatthewTighe (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Updated, also thanks for the conversation on Slack!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Based on the conversation on Slack, the consensus is the best option is to stay in main scope (`scope.launch{}`) here.

MatthewTighe (approved):

Amejia481:
@kycn  are you OK landing ? Your review is required.

ohall-m:
> @kycn are you OK landing ? Your review is required.

We talked in [Slack](https://mozilla.slack.com/archives/G01EXNFFAVC/p1707145996267499?thread_ts=1706798060.455179&cid=G01EXNFFAVC)! I filed [1878679](https://bugzilla.mozilla.org/show_bug.cgi?id=1878679) for us to look into using `async` as a precaution. GeckoView does seem to be handling this on the correct thread, but using `async` could be a good practice here.

PR #5416, opened 2024-02-01
Bug 1878185 - Add ChangeDetectionMiddleware

MatthewTighe:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1878185

### DIFF
--- a/android-components/components/lib/state/src/main/java/mozilla/components/lib/state/helpers/ChangeDetectionMiddleware.kt
+++ b/android-components/components/lib/state/src/main/java/mozilla/components/lib/state/helpers/ChangeDetectionMiddleware.kt
@@ -1,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.lib.state.helpers
+
+import mozilla.components.lib.state.Action
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.MiddlewareContext
+import mozilla.components.lib.state.State
+
+/**
+ * A Middleware for detecting changes to a state property, and offering a callback that captures the action that changed
+ * the property, the property before the change, and the property after the change.
+ *
+ * For example, this can be useful for debugging:
+ * ```
+ * val selectedTabChangedMiddleware = ChangeDetectionMiddleware<BrowserState, BrowserAction>(
+ *     val selector = { it.selectedTabId }
+ *     val onChange = { actionThatCausedResult, preResult, postResult ->
+ *         logger.debug("$actionThatCausedResult changed selectedTabId from $preResult to $postResult")
+ *     }
+ * ```
+ *
+ * @param S The state type of the [Store] this middleware is attached to.
+ * @param A The action subtype of the [Store] this middleware is attached to.
+ * @param T The type of the property to watch for changes in.
+ * @param selector A function to map from the State to the property that is being inspected.
+ * @param onChange A callback to react to changes to the property defined by [selector].
+ */
+class ChangeDetectionMiddleware<T, S : State, A : Action>(
+    private val selector: (S) -> T,
+    private val onChange: (A, pre: T, post: T) -> Unit,
+) : Middleware<S, A> {
+    override fun invoke(context: MiddlewareContext<S, A>, next: (A) -> Unit, action: A) {
+        val pre = selector(context.store.state)
+        next(action)
+        val post = selector(context.store.state)
+        if (pre != post) {
+            onChange(action, pre, post)
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


jonalmeida:
Based on [the conversation in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1878185#c3), this is desired during testing. We have a `CaptureActionsMiddleware` already, so if we want, we can include behaviour there to observe a state there. What do you think about reusing that, or does it's current abilities work for your use case?

MatthewTighe:
> Based on [the conversation in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1878185#c3), this is desired during testing. We have a `CaptureActionsMiddleware` already, so if we want, we can include behaviour there to observe a state there. What do you think about reusing that, or does it's current abilities work for your use case?

The `CaptureActionMiddleware` could be updated to include both, but I would prefer to keep them separate as they have different enough responsibilities. Additionally, `Capture` is in a test package, which is only included in the test [dependencies](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/build.gradle#713) for Fenix. We'd either need to move it, or import the dependency when we wanted to use it for debug logging.

TBH, when I demo'd the code in our show-n-tell it was also suggested that this would be a useful pattern to use more generally, for a few reasons:
1. Testing a middleware is easier than testing an observation in the `View`
2. Decouples business logic from Android framework components
3. Avoids difficulty with `Flow`s and keeps our async processing and queuing more encapsulated within the Store

Especially in light of our conversations the last few days, I don't want to rush this through without your approval given our intent to use it in more general cases eventually. Do you think it would make more sense to land this in Fenix for now, and consider upstreaming it with more robust documentation once it had had some time to soak?

jonalmeida:
@MatthewTighe 

> We'd either need to move it, or import the dependency when we wanted to use it for debug logging.

Based on what I understood from your use case, this should be well annotated that is for testing purposes only. The reason I'm concerned about this landing without making that clear, is because this middleware overrides the whole purpose of observing the state if this middleware acts as a shortcut to observe fields changing within.

>   1. Testing a middleware is easier than testing an observation in the `View`

What does easier mean? What are we specifically trying to do in `onChange`? If by debug logging you mean something like `Log.d`, can't we just add that to the reducer when we want to do testing? Example:

```kotlin
internal object Reducer {
  fun reduce(state: State, action: Action): State {
    Log.d("$state $action")
    return when (action) {
      is Action.UpdateAction -> { }
    }
  }
}
```

>   3. Avoids difficulty with `Flow`s and keeps our async processing and queuing more encapsulated within the Store

If we're testing state changes with a middleware, we are creating a layer that is different from what the `View` is observing, and what we test using this middleware will not be the same from what the `View` receives. Another way of saying this is that we would get false-positive results from our debugging that would not have been caught because the developer didn't consider that observes are always async.

A weak reason I'd also add, because a developer shouldn't be doing this in a store but you can't prevent this: The middleware solution might not always give you the results you're looking for as well. If there are two middlewares making state changes based on different actions (which is what you shouldn't be doing), you would only be getting the last action that made the state change and not any previous action before that in the list.

MatthewTighe:
Would still be happy to look into keeping this as a Fenix-only change until we have some live examples of whether it's useful, but wanted to respond to these points. 

> Based on what I understood from your use case, this should be well annotated that is for testing purposes only

It will not necessarily be only used for testing purposes. 

> What does easier mean? 

Observations of state typically happen within fragments or Composables, which require extensive test setup that run longer and more expensive tests (even just importing Robolectric will increase a standalone test runtime by about 20 seconds over JUnit IIRC). It introduces concurrency and lifecycle testing difficulties, but those are usually a well-understood  implementation pattern and don't need to be re-tested in every scenario. Generally, using fragments, activities, or Composables to handle business logic is less desirable than classes we have full control over and can execute without additional test runners.

> What are we specifically trying to do in onChange?

Whatever we want to, as is usually the case with middleware. They are the unrestricted, side-effectful portion of the architecture.

> If by debug logging you mean something like Log.d, can't we just add that to the reducer when we want to do testing?

This requires updating every action that can change the property you are interested in. Additionally, if we were ever interested in shipping something like opt-in debug telemetry to help handle incidences than the reducer would not be the appropriate place for side-effects.

> If we're testing state changes with a middleware, we are creating a layer that is different from what the View is observing, and what we test using this middleware will not be the same from what the View receives. Another way of saying this is that we would get false-positive results from our debugging that would not have been caught because the developer didn't consider that observes are always async.

This isn't intended to completely replace observations in the view, or tests for the view. The view should only need to observe state that is relevant to what it is displaying, though.

> A weak reason I'd also add, because a developer shouldn't be doing this in a store but you can't prevent this: The middleware solution might not always give you the results you're looking for as well. If there are two middlewares making state changes based on different actions (which is what you shouldn't be doing), you would only be getting the last action that made the state change and not any previous action before that in the list.

Middlewares don't directly make state changes, though they can dispatch additional actions to update that state. In that case, the callback would trigger for every action that completed the reducer chain. There are other edge cases where this wouldn't work perfectly, like if a middleware intercepts an action or rewrites an action. Both of those cases imply that a middleware has full knowledge of the system though, which may indicate a different problem anyway.

jonalmeida:
Yeah, let's keep this in fenix only for now.

PR #5432, opened 2024-02-02
Bug 1832357 - Fallback to loading new window requests in custom tab

rocketsroger:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1832357


android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -4,73 +4,27 @@
...
 
     private var scope: CoroutineScope? = null
 
-    /**
-     * Transform a [CustomTabConfig] into a [CustomTabsIntent] that creates a
-     * new custom tab with the same styling and layout
-     */
-    @Suppress("ComplexMethod")
-    @VisibleForTesting(otherwise = PRIVATE)
-    internal fun configToIntent(config: CustomTabConfig?): CustomTabsIntent {
```
MozillaNoah, line 47: Since this PR largely involves a big function deletion, I need some help understanding not the bug within the experience but rather within the code.

So was this function essentially trying to spin off another custom tab from the presently-open existing custom tab? So then putting it all together: 
- Users tries tapping on certain links
- Fenix tries to spin-off another custom tab when it's already in a custom tab
- Nothing happens

?

rocketsroger: Sure.  When a new window request happens, (could be because the site needed to open a new window, or the user taps on something that triggers a new window) we try to open that window in a new custom tab.  However, since custom tab doesn't have multiple tabs, the current way is we trigger a brand new separate custom tab instance.  Once the user is done with that custom tab, they close it and get back to the original custom tab (that sends the new window request).

But because we're triggering a new custom tab instance, a lot of things will fail.  In this particular case, a blob:// url is not something we support coming from another application (Not declared in Android Manifest), and since we consider the new custom tab request external (a new intent), we can't load it.  (similar if we get a javascript:// url etc).

rocketsroger: But all that said, the original behaviour of opening a new separate custom tab does not match what Chrome does.  If you try this with Chrome set as default browser ```adb shell am start -a android.intent.action.VIEW -d https://jsfiddle.net/actyp5gr/ -e android.support.customtabs.extra.SESSION true```, you'll see that Chrome opens the new window inside the same custom tab.

So in a way, this change aligns us with Chrome's behaviour and at the same time fixed the problem that if we're opening a new url in custom tab, we shouldn't be opening it in a new window because custom tab should always be one window.

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -83,18 +37,7 @@ class CustomTabWindowFeature(
...
-                        // and we are not registered to handle its schemes.
-                        // Let's log this to better asses how often this happens in real world and
-                        // if we need to add new schemes to properly support this workflow.
-                        // See Fenix #8412
-                        try {
-                            intent.launchUrl(activity, uri)
-                        } catch (e: ActivityNotFoundException) {
-                            onLaunchUrlFallback(uri)
-                        }
+                        state.engineState.engineSession?.loadUrl(windowRequest.url)
```
rocketsroger, line 40: This part is the main change.  The rest of the changes is just that we don't need `configToIntent` function anymore.

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
MozillaNoah: Gotcha gotcha. So we DO allow for, let's call it, "stacked" custom tabs, but because in this situation, there are URLs we don't support when coming from certain origins, causing the URL to just not open or do anything? Do we have a guarantee then that this behavior change will only be for custom tabs then and not the case you outlined above of one of the unsupported URLs from an external app? (The answer seems obvious but just making sure 😅)

Matching the behavior from Chrome for now makes sense to me. I'm admittedly still getting a handle on custom tabs as a whole, since there's no analog on iOS, so appreciate you walking me through all of this.

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
jonalmeida: @rocketsroger I've left a comment in [the original bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1832357#c15) that this change has the possibility of breaking a lot of web content use cases.

rocketsroger: I'll make sure I look into that so we won't break anything in custom tabs.

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
rocketsroger: This will only affect custom tabs.  However with Jon's comment below, I'll have to make sure we're not breaking anything that needs to have a separate tab.

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
jonalmeida: In case there is some misunderstanding, I am suggesting that we are going to break **web content** with this change, not the custom tabs API, implementation, or UI.

rocketsroger: Yes, what I meant was that this can only break web content in the custom tabs.  Since this code only handles the new window request on custom tabs.

That said, there is no reason for me to imitate what Chrome does in custom tab to fix this issue.  I'll update this change so we only open the link in the current tab if we know that the url scheme is not supported in a new instance of a custom tab.

rocketsroger:
Updated to only fallback to loading the URL in the same tab if we failed to load the new window.

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
```diff
@@ -87,13 +85,12 @@ class CustomTabWindowFeature(
...
-                        // Let's log this to better asses how often this happens in real world and
-                        // if we need to add new schemes to properly support this workflow.
-                        // See Fenix #8412
                         try {
                             intent.launchUrl(activity, uri)
                         } catch (e: ActivityNotFoundException) {
-                            onLaunchUrlFallback(uri)
+                            // Workaround for unsupported schemes
+                            // See https://bugzilla.mozilla.org/show_bug.cgi?id=1878704
+                            state.engineState.engineSession?.loadUrl(windowRequest.url)
```
MozillaNoah, line 93: since this fix is in AC, will this need a change log addition?

MozillaNoah (approved):

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabWindowFeature.kt
rocketsroger: Good idea.  I'll add this to changelog before landing.  Thanks

PR #5434, opened 2024-02-02
Bug 1878409 - Update Gradle to version 8.6

rvandermeulen:
Plan would be to land this along with bug 1878410 to minimize any mismatch between Fenix and GV.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1878409


mcarare (approved):
LGTM!

PR #5500, opened 2024-02-07
Bug 1870319 - Announce title in add-ons sub-menu

willdurand:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1870319

### DIFF
--- a/android-components/components/browser/menu/src/main/res/values/strings.xml
+++ b/android-components/components/browser/menu/src/main/res/values/strings.xml
@@ -13,4 +13,6 @@
     <string name="mozac_browser_menu_addons_manager">Add-ons Manager</string>
     <!-- Content description for the action bar "up" button -->
     <string name="action_bar_up_description">Navigate up</string>
+    <!-- Content description for the action bar "up" button of the add-ons sub menu item -->
+    <string name="mozac_browser_menu_addons_description">Add-ons, navigate up</string>
 </resources>
--- a/fenix/app/src/main/res/values/strings.xml
+++ b/fenix/app/src/main/res/values/strings.xml
@@ -2101,7 +2101,7 @@
     <string name="radio_preference_info_content_description">Click for more details</string>
 
     <!-- Content description for the action bar "up" button -->
-    <string name="action_bar_up_description">Navigate up</string>
+    <string name="action_bar_up_description" moz:removedIn="124" tools:ignore="UnusedResources">Navigate up</string>
 
     <!-- Content description for privacy content close button -->
     <string name="privacy_content_close_button_content_description">Close</string>
--- a/android-components/components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenuBuilder.kt
+++ b/android-components/components/browser/menu/src/main/java/mozilla/components/browser/menu/WebExtensionBrowserMenuBuilder.kt
@@ -91,7 +91,7 @@ class WebExtensionBrowserMenuBuilder(
     ): List<BrowserMenuItem> {
         val addonsMenuItem = if (filteredExtensionMenuItems.isNotEmpty()) {
             val backPressMenuItem = BackPressMenuItem(
-                contentDescription = context.getString(R.string.action_bar_up_description),
+                contentDescription = context.getString(R.string.mozac_browser_menu_addons_description),
                 label = context.getString(R.string.mozac_browser_menu_addons),
                 imageResource = style.backPressMenuItemDrawableRes,
                 iconTintColorResource = style.webExtIconTintColorResource,


Amejia481 (approved):

android-components/components/browser/menu/src/main/res/values/strings.xml
```diff
@@ -13,4 +13,6 @@
     <string name="mozac_browser_menu_addons_manager">Add-ons Manager</string>
     <!-- Content description for the action bar "up" button -->
     <string name="action_bar_up_description">Navigate up</string>
+    <!-- Content description for the action bar "up" button of the add-ons sub menu item -->
```
flodolo, line 16: I know you're copying from the button above, but is "up" the right choice here? To me, it's "back".

"Navigate up" implies more scrolling to the top of something.

flodolo (approved):

android-components/components/browser/menu/src/main/res/values/strings.xml
willdurand: I don't know why it is "up" here, maybe because we go one level up in the menu tree? (from sub-menu to main menu)

willdurand: Do you think we should change that? Ideally in all the strings defined in this file?

flodolo: I would ask accessibility folks. If that works for them, then we should keep it as is.

willdurand: I talked to @anna-yeddi and "back" is clearer so I updated.

android-components/components/browser/menu/src/main/res/values/strings.xml
```diff
@@ -12,5 +12,9 @@
     <!-- Label for add-ons sub menu item for add-ons manager -->
     <string name="mozac_browser_menu_addons_manager">Add-ons Manager</string>
     <!-- Content description for the action bar "up" button -->
-    <string name="action_bar_up_description">Navigate up</string>
+    <string name="action_bar_up_description" moz:removedIn="124" tools:ignore="UnusedResources">Navigate up</string>
+    <!-- Content description for the action bar "back" button -->
+    <string name="action_bar_back_description">Navigate back</string>
+    <!-- Content description for the action bar "back" button of the add-ons sub menu item -->
+    <string name="mozac_browser_menu_addons_description">Add-ons, Navigate back</string>
```
flodolo, line 19:
```suggestion
    <string name="mozac_browser_menu_addons_description">Add-ons, navigate back</string>
```

Nitpick that I missed (not very important): I assume lowercase is correct here?

flodolo (approved):

fenix/app/src/main/res/values/strings.xml
```diff
@@ -2101,7 +2101,7 @@
     <string name="radio_preference_info_content_description">Click for more details</string>
 
     <!-- Content description for the action bar "up" button -->
-    <string name="action_bar_up_description">Navigate up</string>
+    <string name="action_bar_up_description" moz:removedIn="124" tools:ignore="UnusedResources">Navigate up</string>
```
willdurand, line 2104: That's a bit odd, not sure why this string was duplicated.

flodolo: It's not duplicated, it only added the attributes so that it can be removed later?

willdurand: The same ID is referenced in two different `strings.xml` files

android-components/components/browser/menu/src/main/res/values/strings.xml
```diff
@@ -12,5 +12,9 @@
     <!-- Label for add-ons sub menu item for add-ons manager -->
     <string name="mozac_browser_menu_addons_manager">Add-ons Manager</string>
     <!-- Content description for the action bar "up" button -->
-    <string name="action_bar_up_description">Navigate up</string>
+    <string name="action_bar_up_description" moz:removedIn="124" tools:ignore="UnusedResources">Navigate up</string>
+    <!-- Content description for the action bar "back" button -->
+    <string name="action_bar_back_description">Navigate back</string>
```
Amejia481, line 17: nit: should we use the prefix `mozac_browser_menu`?

Amejia481 (approved):

android-components/components/browser/menu/src/main/res/values/strings.xml
willdurand: I'd rather not touch that for now

willdurand:
It looks like "Navigate up" is the right terminology in Android land: https://developer.android.com/guide/navigation/principles#up_and_back_are_identical_within_your_apps_task - I discovered that the hard way because changing "up" with "back" broke many UI tests and I realized some "Navigate up" descriptions were coming from the native UI components.

That is the reason why I reverted to the minimum amount of changes to fix the a11y bug.

PR #5502, opened 2024-02-07
Bug 1861459 - Update Home Screen to use BrowsingMode from AppStore

MatthewTighe:
This is largely unchanged code from https://github.com/mozilla-mobile/firefox-android/pull/4646, except with 2 major changes:

1. This re-orders the patches, so that this can land in isolation, since the Home Screen is the area most affected by changes to where PBM state is kept.
2. In order to achieve that, it adds a new `BrowsingModeManager` subtype called `AppStoreBrowsingModeManagerWrapper` that delegates calls to `HomeActivity.browsingModeManager` to the `AppStore`. This is so that all the usages in other areas of the app can continue existing, and we can address them in a follow-up(s).

This is still a little large, but I'm not sure there's a great way to break this down further without running into order-of-operations problems, since the workflow of "load Mode from intent or settings" is really finicky in how it plays with the `ThemeManager`, which fires an intent to set the mode after activity recreation due to mode changes. I will plan to update that at some point as well.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1861459

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -157,8 +157,6 @@ class HomeFragment : Fragment() {
         )
     }
 
-    private val browsingModeManager get() = (activity as HomeActivity).browsingModeManager
-
     private val collectionStorageObserver = object : TabCollectionStorage.Observer {
         @SuppressLint("NotifyDataSetChanged")
         override fun onCollectionRenamed(tabCollection: TabCollection, title: String) {
@@ -253,8 +251,6 @@ class HomeFragment : Fragment() {
             orientationChange = false,
             orientation = requireContext().resources.configuration.orientation,
         )
-
-        components.appStore.dispatch(AppAction.ModeChange(browsingModeManager.mode))
 
         lifecycleScope.launch(IO) {
             if (requireContext().settings().showPocketRecommendationsFeature) {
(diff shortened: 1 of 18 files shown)


fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -516,16 +512,8 @@ class HomeFragment : Fragment() {
      * doesn't get run right away which means that we won't draw on the first layout pass.
      */
     private fun updateSessionControlView() {
-        if (browsingModeManager.mode == BrowsingMode.Private) {
-            binding.root.consumeFrom(requireContext().components.appStore, viewLifecycleOwner) {
-                sessionControlView?.update(it)
-            }
-        } else {
-            sessionControlView?.update(requireContext().components.appStore.state)
```
boek, line 524: Do we no longer need this update outside of a `consumeFrom`?

MatthewTighe: Good catch! Looks like this was [indeed intentional](https://github.com/mozilla-mobile/firefox-android/commit/865eedddd3389480abddcde5c7d7200d33826096#diff-fbc1255e297417e1494bc57e3a8776e7e74fb16022729bfee4d76c10cc3e06fd), so I'll leave it for now with a small change to logic for readability. Hopefully this will go away shortly as we continue to Composify

boek (dismissed):

MatthewTighe:
I've discovered a bug with this while rebasing on Tom's patch https://github.com/mozilla-mobile/firefox-android/pull/5807. It looks like, when opening a custom tab in private mode, the selected tab is updated (to a normal tab, not the custom tab) because of a `TabListAction.RestoreAction`, which then overwrites the mode and theme [here](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt#1497). I'll request reviews again when I get around to fixing it.

PR #5509, opened 2024-02-08
Bug 1878827 - Add nav buttons click handling

mavduevskiy:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1878827

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
@@ -6,7 +6,6 @@ package org.mozilla.fenix.components.toolbar
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Arrangement
-import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
@@ -17,17 +16,15 @@ import androidx.compose.material.IconButton
 import androidx.compose.material.LocalContentColor
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.CompositionLocalProvider
-import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.viewinterop.AndroidView
-import mozilla.components.browser.menu.view.MenuButton
-import org.mozilla.fenix.R
-import org.mozilla.fenix.components.toolbar.ItemType.STANDARD
-import org.mozilla.fenix.components.toolbar.ItemType.TAB_COUNTER
+import org.mozilla.fenix.components.toolbar.navbar.DefaultButtons
+import org.mozilla.fenix.components.toolbar.navbar.NavBarButton
+import org.mozilla.fenix.compose.LongPressIconButton
 import org.mozilla.fenix.compose.TabCounter
 import org.mozilla.fenix.compose.annotation.LightDarkPreview
 import org.mozilla.fenix.theme.FirefoxTheme
@@ -36,64 +33,46 @@ import org.mozilla.fenix.theme.Theme
 /**
  * Top-level UI for displaying the navigation bar.
  *
- * @param actionItems A list of [ActionItem] used to populate the bar.
+ * @param buttons A list of [NavBarButton] used to populate the bar.
  * @param tabCount The number of opened tabs.
- * @param menuButton A [MenuButton] to be used for [ItemType.MENU].
  */
 @Composable
 fun NavigationBar(
-    actionItems: List<ActionItem>,
+    buttons: List<NavBarButton>,
     tabCount: Int,
-    menuButton: MenuButton? = null,
 ) {
-    Box(
+    Row(
         modifier = Modifier
             .background(FirefoxTheme.colors.layer1)
             .height(48.dp)
-            .fillMaxWidth(),
+            .fillMaxWidth()
+            .padding(horizontal = 16.dp),
+        horizontalArrangement = Arrangement.SpaceBetween,
     ) {
-        Row(
-            modifier = Modifier
-                .align(Alignment.Center)
-                .fillMaxWidth()
-                .padding(horizontal = 16.dp),
-            horizontalArrangement = Arrangement.SpaceBetween,
-        ) {
-            actionItems.forEach {
-                when (it.type) {
-                    ItemType.STANDARD -> {
-                        IconButton(onClick = {}) {
-                            Icon(
-                                painter = painterResource(it.iconId),
-                                stringResource(id = it.descriptionResourceId),
-                                tint = FirefoxTheme.colors.iconPrimary,
-                            )
-                        }
+        buttons.forEach { button ->
+            when (button) {
+                is NavBarButton.ActionButton -> {
+                    LongPressIconButton(
+                        onClick = { button.onClick() },
+                        onLongClick = { button.onLongPress() },
+                    ) {
+                        Icon(
+                            painter = painterResource(button.iconId),
+                            stringResource(id = button.descriptionResourceId),
+                            tint = FirefoxTheme.colors.iconPrimary,
+                        )
                     }
-                    ItemType.TAB_COUNTER -> {
-                        CompositionLocalProvider(LocalContentColor provides FirefoxTheme.colors.iconPrimary) {
-                            IconButton(onClick = {}) {
-                                TabCounter(tabCount = tabCount)
-                            }
-                        }
-                    }
-
-                    ItemType.MENU -> {
-                        // [ActionItem] will be refactored here:
-                        // https://bugzilla.mozilla.org/show_bug.cgi?id=1878827
-                        if (menuButton != null) {
-                            AndroidView(
-                                modifier = Modifier.size(48.dp),
-                                factory = { _ -> menuButton },
-                            )
-                        } else {
-                            IconButton(onClick = {}) {
-                                Icon(
-                                    painter = painterResource(it.iconId),
-                                    stringResource(id = it.descriptionResourceId),
-                                    tint = FirefoxTheme.colors.iconPrimary,
-                                )
-                            }
+                }
+                is NavBarButton.ThreeDotMenuButton -> {
+                    AndroidView(
+                        modifier = Modifier.size(48.dp),
+                        factory = { _ -> button.menuButton },
+                    )
+                }
+                is NavBarButton.TabCounterButton -> {
+                    CompositionLocalProvider(LocalContentColor provides FirefoxTheme.colors.iconPrimary) {
+                        IconButton(onClick = { button.onClick() }) {
+                            TabCounter(tabCount = tabCount)
                         }
                     }
                 }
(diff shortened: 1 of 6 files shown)


fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBar.kt
```diff
@@ -0,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components.toolbar.navbar
```
rocketsroger, line 5: I am assuming that this is a file move? or are there changes in NavigationBar.kt at the same time as the move?

mavduevskiy: now that I think of it, it made reviewing much harder :*) 

it's both a move and a change, sorry about that.

rocketsroger: I see, can we separate the move? Maybe in a follow up or before this review?

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
```diff
@@ -36,15 +35,13 @@ import org.mozilla.fenix.theme.Theme
...
  * @param tabCount The number of opened tabs.
- * @param menuButton A [MenuButton] to be used for [ItemType.MENU].
  */
 @Composable
 fun NavigationBar(
-    actionItems: List<ActionItem>,
+    buttons: List<NavBarButton>,
     tabCount: Int,
-    menuButton: MenuButton? = null,
 ) {
```
MozillaNoah, line 45: the Box + Row can likely be combined

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
```diff
@@ -59,112 +56,45 @@ fun NavigationBar(
...
             horizontalArrangement = Arrangement.SpaceBetween,
         ) {
-            actionItems.forEach {
-                when (it.type) {
-                    ItemType.STANDARD -> {
-                        IconButton(onClick = {}) {
+            buttons.forEach { button ->
+                when (button) {
+                    is NavBarButton.ActionButton -> {
+                        LongPressIconButton(
```
MozillaNoah, line 62: what was the motivation for having this UI creation be so dynamic? Was Product or UX wanting the flexibility of reordering the buttons or something?

Just wanting to understand what led you here versus other approaches. From I can glean here, there's only realy two variations of the UI.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/BottomToolbarContainerView.kt
```diff
@@ -55,6 +53,10 @@ class BottomToolbarContainerView(
                     }
                 }.value
 
+                // In future, this will also accept tab.content.canGoForward
+                // to manage active/inactive state of navigation buttons
+                val buttons = ActionItemListBuilder.build(location)
```
MozillaNoah, line 58: if we're passing in the complex `NavBarLocation` to only make the `buttons` list, what if we instead pass `buttons` into `BottomToolbarContainerView` instead? That way, the View doesn't even need to know about some of the complex dependencies and their transformations.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Constructs a list of button for a given navigation target.
+ *
+ * This function takes a [NavBarLocation] instance, identifying the current UI context,
+ * and returns a tailored list of [NavBarButton]s suitable for that location. Each [NavBarButton]
+ * is configured based on the specific requirements and characteristics of the navigation target,
+ * ensuring a context-appropriate user interface.
+ */
+object ActionItemListBuilder {
```
MozillaNoah, line 118: Reading through this file, it looks like we're trying to simplify some of the API of dynamically creating the navigation bar buttons.

What led you to the creation of `ActionItemListBuilder` + `DefaultButtons` objects? It looks like there's a ton of complexity between those objects, the `NavBarLocation`, and the two transform functions to generate a list of buttons. 

Essentially, I'm having a hard time understanding what problem we're ultimately trying to solve. Based on the code here and the figma, we need to have two versions of the navigation bar (one for browser and one for homescreen), but I have a concern we're coding for significantly beyond the scope of those two use cases.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
mavduevskiy: On a high level, I was thinking about the nav bar in the similar manner to the toolbar. We should probably someday make a AC component that would be customize and not tied to a particular design/implementation.

On a more practical level, the current design has four states and the dynamic way looked more appropriate.
Home screen (with a middle search button)
Browser screen (with a middle home button)
Browser screen, private mode (with a burn button)
Custom tab (with 4 buttons and a switch to the browser button)

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/BottomToolbarContainerView.kt
mavduevskiy: that's a good point, `BottomToolbarContainerView` could be _dummier_ and just receive buttons to display!

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
mavduevskiy: You got it right, I was trying to simplify things.

`DefaultButtons` allow us not create same classes with same resources every time we need to add buttons to a nav bar (we do that twice for regular tabs, twice for private and once for custom). 
`ActionItemListBuilder` just encapsulates the logic of handling the clicks. It's verbose (in case of BrowserFragment it is 50 of lines), and I didn't want to put into the Fragment itself.

It looks like I was trying to solve two problems:
1) not to create buttons with same resources multiple times
2) not putting click handling code into the fragment directly

mavduevskiy:
updated with @MozillaNoah suggestions

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
```diff
@@ -0,0 +1,206 @@
...
+        descriptionResourceId = R.string.browser_menu_forward,
+    )
+
+    val tabs = NavBarButton.TabCounterButton
+
+    val defaultItems = listOf(back, forward, home, tabs, menu)
+}
+
+/**
+ * Defines navigation bar locations with required dependencies for building the bav bar.
```
sarah541, line 85:
```suggestion
 * Defines navigation bar locations with required dependencies for building the nav bar.
```

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
```diff
@@ -0,0 +1,206 @@
...
+ */
+object DefaultButtons {
+    val home = NavBarButton.ActionButton(
+        iconId = R.drawable.mozac_ic_home_24,
+        descriptionResourceId = R.string.browser_toolbar_home,
+    )
+
+    val search = NavBarButton.ActionButton(
+        iconId = R.drawable.mozac_ic_search_24,
+        descriptionResourceId = R.string.browser_toolbar_home,
```
sarah541, line 61: We need a different description for search

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
```diff
@@ -0,0 +1,206 @@
...
+            },
+            NavBarButton.ThreeDotMenuButton(location.menuButton),
+        )
+    }
+
+    private fun buildNavButtonsForHomePage(location: NavBarLocation.HomeFragment): List<NavBarButton> {
+        return listOf(
+            DefaultButtons.back.copy(),
+            DefaultButtons.forward.copy(),
+            DefaultButtons.search.copy(),
```
sarah541, line 188: We are to add the search functionality?

sarah541:
Is adding navigation buttons for private mode and custom tab a separate task?

sarah541:
I am going to pull this down tomorrow and test a little .

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
mavduevskiy: the button will be focusing the address bar, bringing in the search dialog fragment.

my understanding is, keeping the home button on the home fragment didn't make sense, so it was replaces by a button that is essentially duplicating a click on the address bar. you can see it in the general flow [section](https://www.figma.com/file/8fsCvtnweBkWGZ9NfL907l/Toolbar-Redesign?type=design&node-id=5503-276255&mode=design&t=wwdfRA9QCvjSrWAB-4)

mavduevskiy:
> Is adding navigation buttons for private mode and custom tab a separate task?

yep, [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1883164)

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
MozillaNoah: Gotcha gotcha. Thanks for listing out the four UI variations. 

What are the advantages you're wanting to achieve from this dynamic approach? What I'm seeing is four use cases that could also easily be statically constructed in their own Composables, having lambdas for each button's click/long click. Is having the one `NavigationBar` Composable worth the trade-off of the setup complexity from `NavigationBarButtons`?

Is there already some desire to upstream this to AC and then be consumed in Focus?

I'm asking these exploratory questions to help me (and other readers) understand why this is a candidate for our best approach for this problem.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
MozillaNoah: Do you have any concerns with the maintenance of this code? If UX decides to change the layout down the road (whether that's to change the icons/functionality or add a non-icon action to the bar), will the proposed approach still hold up? Can we boost our confidence with the button generation with some tests?

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
```diff
@@ -0,0 +1,206 @@
...
+
+    val tabs = NavBarButton.TabCounterButton
+
+    val defaultItems = listOf(back, forward, home, tabs, menu)
+}
+
+/**
+ * Defines navigation bar locations with required dependencies for building the bav bar.
+ */
+sealed class NavBarLocation {
```
MozillaNoah, line 87: now that `NavBarLocation` is only used in this file for `ActionItemListBuilder.build()`, do we even need to manage this complex data type anymore? Can this be replaced with direct calls to builder functions from HomeFragment & BrowserFragment?

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
```diff
@@ -0,0 +1,206 @@
...
+    data class ThreeDotMenuButton(val menuButton: MenuButton) : NavBarButton()
+}
+
+/**
+ * Holds a collection of default buttons.
+ *
+ * Items could be copied and customized for usage across the app.
+ * Also used for building previews.
+ */
+object DefaultButtons {
```
MozillaNoah, line 53: If we stay with the dynamic approach, could we simplify it further? Could enum classes aid us in this situation?

MozillaNoah:
@sarah541 and @rocketsroger, I'm leaving for a long PTO Wednesday, 3/6, so I will leave feedback as I am able before then, but I am deferring final approval to y'all.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
mavduevskiy: Now that I think of it, part of the reason was I was just not very experienced with the `Composable` :) 

I tried tom implement the suggested [approach](https://github.com/mozilla-mobile/firefox-android/pull/5888), is it something you had in mind?

It ended up being a bit more verbose, especially because of the previews.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavigationBarButtons.kt
mavduevskiy: yeah, I think it's not a problem. it might be even easier to maintain since it's dynamic. 

but I am not sure we should go against the existing patters we already have, so right let's focus on non-dynamic approach.

mavduevskiy:
Following the conversation in the comments with @MozillaNoah , I opened up a [draft](https://github.com/mozilla-mobile/firefox-android/pull/5888) with an alternative approach. Might be closing this one in favour of non-dynamic approach.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/NavigationBar.kt
MozillaNoah: Maybe we can do a recorded session with @MatthewTighe when I'm back from PTO to go over our current state of fragments + Compose + redux.  In the meantime, my [walkthrough of the debug drawer code](https://bugzilla.mozilla.org/show_bug.cgi?id=1857080#c10) might shed some light on this.

mavduevskiy:
@MozillaNoah @rocketsroger @sarah541 
Thanks for the feedback! I will be closing it in favour of non dynamic approach, work will continue [here](https://github.com/mozilla-mobile/firefox-android/pull/5888)

PR #5542, opened 2024-02-09
Revert "Bug 1833666 - Update original RFC guidelines with stakeholders"

MatthewTighe:
This reverts commit e1e6e86bf6525a123518919bfc620108396ad3dc.

After discussion, it was agreed that updates to existing RFCs are not a pattern we want to accept. Any changes to process will be captured in additional documents.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1833666

### DIFF
--- a/docs/rfcs/0001-rfc-process.md
+++ b/docs/rfcs/0001-rfc-process.md
@@ -35,7 +35,6 @@ The high-level process of creating an RFC is:
 * Create an RFC document (like this one) using the template.
 * Open a pull request for the RFC document.
 * Ask for feedback on the pull request, via the [mailing list]() or in [chat](https://chat.mozilla.org/#/room/#android-components:mozilla.org).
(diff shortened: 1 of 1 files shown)


jonalmeida:
@MatthewTighe on Slack I had mentioned that we can still keep this here since that is what the team wants to do, but we should add that this is an update to original:

> * **(As of 09/02/2024):** Assign or receive volunteers to act as stakeholders for the RFC. They will be responsible for a final decision on the proposal. This should include at least 2 internal team members, and at least 1 external team member if the proposal has been suggested from an external team.

Something of that nature, I'm not tied to having the date there. Any kind of annotation that shows that it is a new addition.

MatthewTighe:
I do ultimately think that the point of RFCs not existing as living documentation but as previous decision records makes this change worth reverting. We don't want to end up in a situation where future information gathering has to be spread between a bunch of RFCs. I think instead it could be worth including with RFCs other artifacts that are intended to be living documents. 

For example, I am planning another RFC to introduce the stakeholders change and will also include artifacts such as an RFC template and an up-to-date guide.

jonalmeida (approved):
Sounds good! 👍

PR #5544, opened 2024-02-09
Bug 1877278 - AC Translations Check for if the Engine is Supported

ohall-m:
This patch supports the workflow for checking if the device architecture supports translations.
    
This patch adds:
* New `TranslationsBrowserState` to hold global translations engine state
*  New `SetEngineSupportedAction` to set the isEngineSupported value on `TranslationsBrowserState`
*  New `EngineExceptionAction` to set errors on `TranslationsBrowserState`

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1877278


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -39,8 +41,52 @@ class TranslationsMiddleware(
     ) {
         // Pre process actions
         when (action) {
+            // Initially adding the tab
+            is TabListAction.AddTabAction -> {
```
ohall-m, line 45: Between `AddTabAction` (new tab), `RestoreAction` (reopening after a shutdown), and `SelectTabAction` (reopening after a shutdown and changing tabs), all the possibilities seem to be covered.

This bug is just checking for if the engine is supported, since it'll be a very early requirement. (We won't show entryways into the feature if the engine isn't supported.) 

We had a conversation about this style flow over [here](https://github.com/mozilla-mobile/firefox-android/pull/5224).

[1873672](https://bugzilla.mozilla.org/show_bug.cgi?id=1873672) will probably use these same actions to initialize the rest of the translations data needs.  I'd like to get a stronger sense of how well it works for this case alone first.

fenix/app/src/test/java/org/mozilla/fenix/telemetry/TelemetryMiddlewareTest.kt
```diff
@@ -86,6 +86,8 @@ class TelemetryMiddlewareTest {
         val engine: Engine = mockk()
         every { engine.enableExtensionProcessSpawning() } just runs
         every { engine.disableExtensionProcessSpawning() } just runs
+        every { engine.isTranslationsEngineSupported(any(), any()) } just runs
```
ohall-m, line 89: In response to this [test-apk-fenix-debug ](https://firefoxci.taskcluster-artifacts.net/XYIsG1d5Rc-CiBRcZtdaxA/0/public/reports/test/testFenixDebugUnitTest/index.html) failure. Seems resolved with this correction. (It reports a different test in the report, but the logs helped me narrow down to this file.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TranslationsStateReducer.kt
```diff
@@ -83,6 +83,15 @@ internal object TranslationsStateReducer {
                     }
                 }
 
+                TranslationOperation.FETCH_IS_ENGINE_SUPPORTED -> {
+                    state.copyWithTranslationsState(action.tabId) {
+                        it.copy(
+                            isEngineSupported = true,
```
calumozilla, line 89: Just wondering if the fetch is successful why we should default to that the engine is supported

calumozilla (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TranslationsStateReducer.kt
ohall-m: Thanks, great catch! I updated to be more like `FETCH_SUPPORTED_LANGUAGES`, since we don't necessarily know it should be `true` and removed setting the engine state.

(I originally was thinking it was more like the `Translate` and `Restore` cases where we can infer the state. However, 100% spot on that success != true in this case.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Hey, @MatthewTighe and @gabrielluong, what do you think about this? Is there anything else I should consider or do you think it is good to start testing this option?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -948,6 +948,17 @@ sealed class TranslationsAction : BrowserAction() {
         val operation: TranslationOperation,
     ) : TranslationsAction(), ActionWithTab
 
+    /**
+     * Sets whether the device architecture supports translations or not.
+     *
+     * @property tabId The ID of the tab the [EngineSession] that requested the list.
+     * @property isEngineSupported If the engine supports translations on this device.
+     */
+    data class SetEngineSupportAction(
```
gabrielluong, line 957:
```suggestion
    data class SetEngineSupportedAction(
```
Maybe? This is more about just using the same wording in the properties.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -39,8 +41,52 @@ class TranslationsMiddleware(
     ) {
         // Pre process actions
         when (action) {
+            // Initially adding the tab
```
gabrielluong, line 44: Maybe include covering first session load which doesn't add a new tab.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
gabrielluong: Based on the amount of research you did, I feel like this is good enough. I don't really have enough context on these actions to provide a better suggestion. 

My understanding of what we are trying to achieve here is to make sure all the tabs have this engine supported property set. The only alternative I could see is calling this on-demand for the individual tab that is displayed instead of doing this for all the tabs. Do you have a sense of whether or not this operation is expensive? For instance, what happens if an user tries to add/restore 50, 100 or 1000 tabs?

gabrielluong: Do we need to check if the engine is supported per tab? 

I was re-reading the description for one of the properties below:
> @property isEngineSupported If the engine supports translations on this device.
and was wondering if we could just set this only once since this seem to indicate whether or not the engine supports translations on the device, which seems like a global state and not based on per tab.

MatthewTighe: +1 to this question about whether the state is global. though it is useful on the one hand to have all the translation state encapsulated in `TranslationState`, any global state could definitely move to the top-level browser state or even a `BrowserState.translationState`. If you went the latter route, the current `TranslationState` may make more sense as `SessionTranslationState`

ohall-m: Thanks for taking a look!

> Do we need to check if the engine is supported per tab?

Not necessarily, this won't change once found.

>Does the translation engine settings need to fetched per tab or do we only need to know if the engine is ready?

This `isEngineSupported` could be a global setting. 100% it is based on the users device and my understanding is it won't change, so once we know, we know.  (Only exception might be something like a possible firmware update, but I think that would be an extreme example an generally not easily changed - [more details](https://mozilla.slack.com/archives/C04DGR18D0F/p1701376615035419).)

The engine automatically becoming ready is fully managed by Gecko AFAIK currently. They are also managing the new single process [1815339](https://bugzilla.mozilla.org/show_bug.cgi?id=1815339).

LMK how we normally handle something like this and I can refactor for sure!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -39,8 +41,52 @@ class TranslationsMiddleware(
...
+                }
+            }
+
             is TranslationsAction.OperationRequestedAction -> {
                 when (action.operation) {
+                    TranslationOperation.FETCH_IS_ENGINE_SUPPORTED -> {
+                        scope.launch {
+                            requestEngineSupport(context, action.tabId)
+                        }
+                    }
```
MatthewTighe, line 88: I'm only seeing that this operation is only really dispatched from this middleware, and only really used in this middleware outside of resetting the error state. I'm wondering if we need the additional operation/action dispatch, or if we can just call `requestEngineSupport` directly from things like `AddTabAction`. 

If the `isEngineSupported` is global state as per the questions above, could we even maybe use `BrowserAction.Init`? I do seem to remember that you looked into that but can't recall the details

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: >If you went the latter route, the current TranslationState may make more sense as SessionTranslationState

That could be interesting because supported languages are also not directly tied to the current tab. I'd say download state would probably make sense there too. 

(I think there could also be an argument made for the never translate site and languages global settings to be on the global state, but they have a lot of coupling with the page settings, so they probably shouldn't be on a global state and stay tied to one for tabs/sessions.)

Do you have an example of something similar to this I could pattern off of?

MatthewTighe: Just looking through `BrowserState`, there is global extension state and then tab-specific extension state in `TabSessionState`. I'm not really familiar with extensions, but I would be happy to chat through any implementation details if you'd like

ohall-m: Awesome, thanks! I'll try following it and see where I end up. (Interestingly, I had to follow a lot of extensions patterns on the GeckoView side as well, so that makes sense too!)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
ohall-m: Thanks, updated!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Switched to a new pattern, but I'll remember this if I have to do something similar with initializing the tab states. (I'm going to take a deeper look at web extensions and around in general if I can find some other alternatives for the tabs too.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Thanks, updated the pattern! 

I can use `BrowserAction.Init` now with the tab independent state. The earlier issue was I needed a tabId to set on, if I set per tab. I still might have to figure out alternatives for tab initialization, but `BrowserAction.Init` seems much more stable for this patch.

fenix/app/src/test/java/org/mozilla/fenix/telemetry/TelemetryMiddlewareTest.kt
ohall-m: (I might be able to remove this with the new changes, I'll do one CI with and one CI without to see if it is still required.)

ohall-m: Appears to be required,[ failure without change](https://firefoxci.taskcluster-artifacts.net/A1R_M40fRVSG_JQaHct0sA/0/public/reports/test/testFenixDebugUnitTest/index.html).

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -54,4 +55,5 @@ data class BrowserState(
     val showExtensionsProcessDisabledPrompt: Boolean = false,
     val extensionsProcessDisabled: Boolean = false,
     val awesomeBarState: AwesomeBarState = AwesomeBarState(),
+    val translationEngine: TranslationsBrowserState = TranslationsBrowserState(),
```
ohall-m, line 58: Had trouble deciding with the naming here, please LMK if you have ideas!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -938,6 +938,16 @@ sealed class TranslationsAction : BrowserAction() {
         val translationError: TranslationError,
     ) : TranslationsAction(), ActionWithTab
 
+    /**
+     * Indicates an app level translations error occurred and to set the error on
+     * [BrowserState.translationEngine].
+     *
+     * @property error The error that occurred.
```
gabrielluong, line 945:
```suggestion
     * @property error The [TranslationError] that occurred.
```
We could reference the class where possible if the naming and wording just makes sense.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
```diff
@@ -35,6 +35,7 @@ import java.util.Locale
  * on application startup e.g. as an indicator that tabs have been restored.
  * @property locale The current locale of the app. Will be null when following the system default.
  * @property awesomeBarState Holds state for interactions with the [AwesomeBar].
+ * @property translationEngine Holds state for translations that apply to the entire browser.
```
gabrielluong, line 38:
```suggestion
 * @property translationEngine Holds translation state that applies to the browser.
```

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
gabrielluong, line 58:
```suggestion
    val translationsEngineState: TranslationsEngineState = TranslationsEngineState(),
```
wdyt?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsBrowserState.kt
```diff
@@ -0,0 +1,19 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.browser.state.state
+
+import mozilla.components.concept.engine.translate.TranslationError
+
+/**
+ * Value type that represents the state of the translations engine within a [BrowserState].
+ *
+ * @property isEngineSupported The translations engine supports the device architecture.
```
gabrielluong, line 12:
```suggestion
 * @property isEngineSupported Whether the translations engine supports the device architecture.
```
Just extra wording to indicate it is a boolean

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsBrowserState.kt
```diff
@@ -0,0 +1,19 @@
...
+ * @property engineError If the translations engine has a global error state.
```
gabrielluong, line 13:
```suggestion
 * @property engineError Holds the error state of the translations engine.
```

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
```diff
@@ -48,6 +48,14 @@ sealed class TranslationError(
     class EngineNotSupportedError(override val cause: Throwable?) :
         TranslationError(errorName = "engine-not-supported", displayError = false, cause = cause)
 
+    /**
+     * Could not determine if the translations engine works on the device architecture.
+     *
+     * @param cause The original throwable before it was converted into this error state.
```
gabrielluong, line 54:
```suggestion
     * @param cause The original [Throwable] before it was converted into this error state.
```

gabrielluong (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
ohall-m: Ohh, should have mentioned, the name of the class that the engine reports the tab state on is  `TranslationEngineState` (singular). 

Currently  have:
* [TranslationsState](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt) - This is the session one.
* [TranslationEngineState](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationEngineState.kt) - This is information GeckoView reports about the session that is set on the session state.

For this one, I really struggled to find a new name:

`TranslationsBrowserEngineState`? `TranslationsAppState`? `TranslationsEngineStatus`? I keep struggling on this one to avoid a naming collision or something that makes sense.

gabrielluong: Maybe `TranslationsState` should become `TabTanslationsState` or `TranslationsSessionState`. Otherwise, `TranslationsBrowState` is fine.

ohall-m: Okay, I'll leave `TranslationsBrowserState` for now, thanks for taking a look!

Agreed, I'll add a code health bug to refactor `TranslationsState` -> `TranslationsSessionState`. I think that'll help keep things more clear. Filed the session refactor naming to bug [1880352](https://bugzilla.mozilla.org/show_bug.cgi?id=1880352).

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsBrowserState.kt
ohall-m: Ty, nice wording

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
ohall-m: Thanks, good idea, updated!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/BrowserState.kt
ohall-m: Ty, updated

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsBrowserState.kt
ohall-m: Good call, ty!

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
ohall-m: Updated, ty!

rvandermeulen:
@mergifyio refresh

PR #4143, opened 2023-10-18
Bug 1858297 - Store FxSuggest db in cache directory

tiftran:
### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation

Used by GitHub Actions.

PR #5546, opened 2024-02-09
Bug 1879649 - Set the data path for the Suggest component

bendk:
This is not used yet, but we're planning on using it to store persistent data that shouldn't be cleared with the cache.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1879649

### DIFF
--- a/android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
+++ b/android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
@@ -13,6 +13,7 @@ import kotlinx.coroutines.withContext
 import mozilla.appservices.suggest.SuggestApiException
 import mozilla.appservices.suggest.SuggestIngestionConstraints
 import mozilla.appservices.suggest.SuggestStore
+import mozilla.appservices.suggest.SuggestStoreBuilder
 import mozilla.appservices.suggest.Suggestion
 import mozilla.appservices.suggest.SuggestionQuery
 import mozilla.components.concept.base.crash.CrashReporting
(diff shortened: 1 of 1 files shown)


android-components/components/feature/fxsuggest/src/main/java/mozilla/components/feature/fxsuggest/FxSuggestStorage.kt
```diff
@@ -31,7 +32,10 @@ class FxSuggestStorage(context: Context) {
     // does I/O, so `store.value` should only be accessed from the read or write scope.
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal val store: Lazy<SuggestStore> = lazy {
-        SuggestStore(File(context.cacheDir, DATABASE_NAME).absolutePath)
+        SuggestStoreBuilder()
+            .cachePath(File(context.cacheDir, DATABASE_NAME).absolutePath)
+            .dataPath(context.getDatabasePath(DATABASE_NAME).absolutePath)
```
bendk, line 37: I copied this from the logins component and I'm pretty sure this is the correct way to get a database path, but I'm no android engineer.  @linabutler How confident are you on this stuff?  Should we flag someone from the android team or do you feel good reviewing it?

linabutler: LGTM! I love this new builder API 🤩

`context.getDatabasePath(...).absolutePath` is what we were using before we switched to using the temporary cache (#4143), so I feel good about reviewing this!

One thing I am wondering about, though—should we use different names for the two databases, and rename the constants to match?

We can keep the cache one named `"suggest.sqlite"`, so that we don't trigger a re-ingest, but let's maybe change the name of the constant to something like `CACHE_DATABASE_NAME`, and add a new constant with a different name for the data database? I'm just thinking that it might be a little confusing to have two databases named `suggest.sqlite` in different directories and with different contents.

linabutler (approved):
Stamp!

PR #5577, opened 2024-02-12
Bug 1876594 - Refactor ThumbnailStorage to be an implicit dependency

MozillaNoah:
A change to `ThumbnailStorage` was breaking previews anywhere the `ThumbnailImage` composable was being used. 

Until we can figure out a long term strategy for properly handling our `components` singleton within previews (which do not have the application context), we'll simply have an `inComposePreview` check within `ThumbnailImage`

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1876594


fenix/app/src/main/java/org/mozilla/fenix/compose/ComposableFactory.kt
```diff
@@ -0,0 +1,75 @@
...
+import mozilla.components.concept.base.images.ImageLoadRequest
+import org.mozilla.fenix.components.components
+import org.mozilla.fenix.theme.FirefoxTheme
+import org.mozilla.fenix.compose.ThumbnailImage as ThumbnailImageContent
+
+/**
+ * Factory object used to instantiate Composables with preview-breaking dependencies, such as
+ * [components].
+ */
+object ComposableFactory {
```
MozillaNoah, line 23: Certainly open to workshopping the object name

fenix/app/src/main/java/org/mozilla/fenix/compose/ComposableFactory.kt
```diff
@@ -0,0 +1,75 @@
...
+     */
+    @Composable
+    fun ThumbnailImage(
+        request: ImageLoadRequest,
+        contentScale: ContentScale,
+        alignment: Alignment,
+        modifier: Modifier = Modifier,
+        fallbackContent: @Composable () -> Unit,
+    ) {
+        if (inComposePreview) {
```
MozillaNoah, line 43: we could optionally also create a compose wrapper for handling this logic check, if we felt like consolidating it further:

````
fun PreviewWrapper(
    previewContent: Lambda,
    actualContent: Lambda,
)

File: fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt [hunk 3]
```diff
@@ -20,74 +18,66 @@ import androidx.compose.ui.Modifier
...
- * @param storage [ThumbnailStorage] to obtain tab thumbnail bitmaps from.
- * @param modifier [Modifier] used to draw the image content.
  * @param contentScale [ContentScale] used to draw image content.
  * @param alignment [Alignment] used to draw the image content.
+ * @param modifier [Modifier] used to draw the image content.
  * @param fallbackContent The content to display with a thumbnail is unable to be loaded.
  */
 @Composable
-fun ThumbnailImage(
+internal fun ThumbnailImage(
```
MozillaNoah, line 35: making this `internal`, since this is only necessary for the Factory now

MozillaNoah:
Writing the info comments, I'm realizing I forgot to implement inside of [`ThumbnailCard`](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailCard.kt). I'll add this tomorrow

fenix/app/src/main/java/org/mozilla/fenix/compose/ComposableFactory.kt
```diff
@@ -0,0 +1,75 @@
...
+     * If this is in a preview, an empty box will be rendered instead.
+     *
+     * @param request [ImageLoadRequest] used to fetch the thumbnail bitmap.
+     * @param contentScale [ContentScale] used to draw image content.
+     * @param alignment [Alignment] used to draw the image content.
+     * @param modifier [Modifier] used to draw the image content.
+     * @param fallbackContent The content to display with a thumbnail is unable to be loaded.
+     */
+    @Composable
+    fun ThumbnailImage(
```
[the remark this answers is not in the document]
MatthewTighe: Noah and I were brainstorming about this yesterday. Here's some of the reasoning for something along this route:

1. Modularization. Modules won't have access to `components`, so they won't be able to have implicit, hidden dependencies on it like is in the `ThumbnailImage` child Composable below. This can result in "prop drilling", where we pass these dependencies down the entire Composable stack. 
2. Even before modularization, the `components` dependency is quite hidden in the stack here. I am not strictly against Composables that can retrieve their own dependencies, but they should then fully own those dependencies.

Using a factory pattern seemed like it might help reveal the hidden dependency a bit by requiring a step "outside" of the usual Compose world, but I don't think this is the only solution to these problems. In fact, there is probably quite a bit of prior art from the React and Redux communities that could inform our solution. @boek may have more insight there.

Ultimately, I would like to see us come to a team agreement on how to solve this problem as I think it would help the team iterate faster and avoid disparate novel solutions. That said, I am in no way married to anything that as presented in this patch.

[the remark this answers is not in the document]
MatthewTighe: Jeff and I spoke offline about this, and I have additional suggestions to share when I get some time tomorrow

fenix/app/src/main/java/org/mozilla/fenix/compose/ComposableFactory.kt
rahulsainani: ❓ I'm not really use if we need this object in the first place, what was the reason for this? From a caller perspectice, I'm more likely to search and write `ThumbnailImage` rather than `Object.ThumbnailImage`

fenix/app/src/main/java/org/mozilla/fenix/compose/ComposableFactory.kt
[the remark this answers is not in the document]
MatthewTighe: Apologies for coming back a little late to this.

My short-term suggestion to avoid the preview breakage would be to pass the dependency explicitly.

Longer-term, the strategy I think we should investigate but would require team buy-in

## FakeComponents

1. Create an abstract supertype of `Components`
2. Create fakes for each component type 
3. update the globally accessible `@Composable components` to be something like: `get() = if (inComposePreview) LocalContext.current.components else FakeComponents()`

I also think we should consider moving the `components` from the `LocalContext` to the `CompositionLocal` to distance their association from the application context. It would also give us more control over local overrides, which is risky but potentially useful.

Then, I think it is fine for some Composables to retrieve dependencies **for their children** from the local composition. This would be required for Composables defined in modules that won't have knowledge of `Components`, and also keeps dependencies more explicit as well as practicing a bit of inversion of control. We would still have implicit dependencies hidden in the tree in the Composables that retrieved components for their children, but I think that is an okay trade-off as long as we limit it to our normally globally-accessible types for now.

A fair amount of stuff is touched on in the [docs for CompositionLocal](https://developer.android.com/jetpack/compose/compositionlocal). For example, we could use the fake types as sane defaults. 

As an aside: an interesting case is that they suggest _not_ using `CompositionLocal` to hold `ViewModel`s, which makes sense as they are screen-specific. However, if we move to a full Redux model with a single store, I think that we could get away with injecting that through `CompositionLocal`, which means we could make more interesting decisions about what level of the tree we observed Store properties and define action dispatches at.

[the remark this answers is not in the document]
MatthewTighe: Ah, Noah reminded me that we could just do the `inComposePreview` check in `TabThumbnail` directly and access `components` there. I think that's fine for now, but that we'll want a better long-term solution that is well-defined and can be re-used across the team.

> Tangential on modularization in case we haven't thought of this already, compose/designsystem package should not contain any feature specific components like TabThumbnail to not have dependency on feature models.

`TabThumbnail` in particular looks to only have dependencies on lower-level types like `ThumbnailStorage` and `TabSessionState`. Given our domain, I imagine we could potentially include things like a `TabThumbnail` in our design system, or at the very least in a module between the design system and the app module. I admit it's been a while since I've looked over our design system tokens, so are we imagining a module that contains only components that are agnostic of the browser domain?

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
```diff
@@ -20,37 +21,36 @@ import androidx.compose.ui.Modifier
...
  * @param fallbackContent The content to display with a thumbnail is unable to be loaded.
  */
 @Composable
 fun ThumbnailImage(
     request: ImageLoadRequest,
-    storage: ThumbnailStorage,
-    modifier: Modifier,
     contentScale: ContentScale,
     alignment: Alignment,
+    modifier: Modifier = Modifier,
```
MatthewTighe, line 45: Is this just some unrelated bookkeeping? Fine with me if so, just checking

MatthewTighe (approved):
LGTM. I'll take a note to write a follow-up for investigating a more substantial strategy for these situations

fenix/app/src/main/java/org/mozilla/fenix/compose/ThumbnailImage.kt
MozillaNoah: Yes, Modifiers are supposed to be optional, so I opted to include that minor change now

gabrielluong (approved):

PR #5589, opened 2024-02-13
Bug 1880176 - Clean up some AndroidX dependency management

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1880176


Amejia481 (approved):
LGTM!

android-components/plugins/dependencies/src/main/java/DependenciesPlugin.kt
```diff
@@ -56,38 +56,39 @@ object Versions {
...
         const val autofill = "1.1.0"
-        const val browser = "1.7.0"
         const val biometric = "1.1.0"
+        const val browser = "1.7.0"
         const val cardview = "1.0.0"
         const val compose = compose_version
         const val constraintlayout = "2.1.4"
         const val coordinatorlayout = "1.2.0"
         const val core = "1.12.0"
+        const val datastore="1.0.0"
```
gabrielluong, line 71:
```suggestion
        const val datastore = "1.0.0"
```

fenix/plugins/fenixdependencies/src/main/java/FenixDependenciesPlugin.kt
```diff
@@ -54,14 +51,8 @@ object FenixDependencies {
...
     const val androidx_core_splashscreen = "androidx.core:core-splashscreen:${FenixVersions.androidx_splash_screen}"
     const val androidx_profileinstaller = "androidx.profileinstaller:profileinstaller:${FenixVersions.androidx_profileinstaller}"
-    const val androidx_activity_ktx = "androidx.activity:activity-ktx:${FenixVersions.androidx_activity}"
     const val androidx_legacy = "androidx.legacy:legacy-support-v4:${FenixVersions.androidx_legacy}"
-    const val androidx_lifecycle_common = "androidx.lifecycle:lifecycle-common:${FenixVersions.androidx_lifecycle}"
-    const val androidx_safeargs = "androidx.navigation:navigation-safe-args-gradle-plugin:${FenixVersions.androidx_navigation}"
-    const val androidx_navigation_fragment = "androidx.navigation:navigation-fragment-ktx:${FenixVersions.androidx_navigation}"
-    const val androidx_navigation_ui = "androidx.navigation:navigation-ui:${FenixVersions.androidx_navigation}"
     const val androidx_transition = "androidx.transition:transition:${FenixVersions.androidx_transition}"
-    const val androidx_datastore = "androidx.datastore:datastore:${FenixVersions.androidx_datastore}"
```
gabrielluong, line 64: We're doing something very subtle here, which is that we are moving some of these Fenix only dependencies over to the DependenciesPlugin which only should contain the AC dependencies. I don't believe we want our A-C dependencies to know about any dependencies other than the ones it needs to maintain the AC is its own independent library.

gabrielluong: I think we should absolutely avoid moving one-off dependencies into the AC DependenciesPlugin and instead we should make FenixDependenciesPlugin dependent on DependenciesPlugin and reuse the `Version.androidx_datastore`

gabrielluong (changes requested):
Let's avoid centralizing dependencies into the AC DependenciesPlugin

rvandermeulen:
I can't get the inheritance to work and am not interested in trying to figure it out. We'll just switch to version catalogs post-migration and drop the plugins entirely.

PR #5620, opened 2024-02-15
Bug 1876803 - Translation Download Size Action

ohall-m:
This patch adds a flow for determining the download size that would
occur for a given translation. Expected use case is for low-data modes.

* New `FetchTranslationDownloadSizeAction`
* New `SetTranslationDownloadSizeAction`
* New `CouldNotDetermineDownloadSizeError`
* New `TranslationDownloadSize` object and `translationDownloadSize` on the
session `TranslationsState`

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1876803


ohall-m:
Anticipating using this action for this [Figma scenario ](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7202-96855&mode=design&t=GLHQlykOU4AahtH1-0)on a low data translation request.

This is different from the other translations download APIs because it is the download size for the specific translation pairs requested. (Not the size of a whole translations language packet, like the core download flow.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -68,6 +71,18 @@ class TranslationsMiddleware(
                     -> Unit
                 }
             }
+
+            is TranslationsAction.FetchTranslationDownloadSizeAction -> {
+                scope.launch {
+                    requestTranslationSize(
+                        context = context,
+                        tabId = action.tabId,
```
calumozilla, line 79: wondering if the download size is correlated to a tab or if this should also be part of the app state TranslationsBrowserState

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
```diff
@@ -72,6 +72,14 @@ sealed class TranslationError(
     class CouldNotRestoreError(override val cause: Throwable?) :
         TranslationError(errorName = "could-not-restore", displayError = false, cause = cause)
 
+    /**
+     * Could not determine translation download size between two pairs.
+     *
+     * @param cause The original [Throwable] before it was converted into this error state.
+     */
+    class CouldNotDetermineSizeError(override val cause: Throwable?) :
```
calumozilla, line 80: Should this be `CouldNotDetermineDownloadSizeError`

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -912,6 +914,32 @@ sealed class TranslationsAction : BrowserAction() {
         override val tabId: String,
     ) : TranslationsAction(), ActionWithTab
 
+    /**
+     * Used to fetch the translation size for a [tabId]. Will use the specified
```
gabrielluong, line 918:
```suggestion
     * Fetch the translation download size for the given [tabId]. Will use the specified
```

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -912,6 +914,32 @@ sealed class TranslationsAction : BrowserAction() {
...
+     * @property toLanguage The to [Language] in the translation pair.
+     */
+    data class FetchTranslationDownloadSizeAction(
+        override val tabId: String,
+        val fromLanguage: Language,
+        val toLanguage: Language,
+    ) : TranslationsAction(), ActionWithTab
+
+    /**
+     * Used to set the translation size for a [tabId].
```
gabrielluong, line 932:
```suggestion
     * Set the [TranslationDownloadSize] for the given [tabId].
```

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
```diff
@@ -72,6 +72,14 @@ sealed class TranslationError(
     class CouldNotRestoreError(override val cause: Throwable?) :
         TranslationError(errorName = "could-not-restore", displayError = false, cause = cause)
 
+    /**
+     * Could not determine translation download size between two pairs.
```
gabrielluong, line 76:
```suggestion
     * Could not determine the translation download size between the translation pairs.
```
Could you also include what the "pairs" are

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationDownloadSize.kt
```diff
@@ -0,0 +1,21 @@
...
+package mozilla.components.concept.engine.translate
+
+/**
+ * A data class to contain information related to the downlad size required for a given translation pair.
+ *
+ * @property fromLanguage The [Language] to translate from on a given translation.
+ * @property toLanguage The [Language] to translate to on a given translation.
+ * @property size The size of the download to perform the translation in bytes. Null means the value has
+ * yet to be received or an error occurred. Zero means no download required or else a model does not exist.
+ * @property error If a [TranslationError] occurred while fetching the size.
```
gabrielluong, line 14:
```suggestion
 * @property error The [TranslationError] reported if an error occurred while fetching the size.
```

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddlewareTest.kt
```diff
@@ -386,6 +387,76 @@ class TranslationsMiddlewareTest {
...
+            error = null,
+        )
+
+        val action =
+            TranslationsAction.FetchTranslationDownloadSizeAction(
+                tabId = tab.id,
+                fromLanguage = translationSize.fromLanguage,
+                toLanguage = translationSize.toLanguage,
+            )
+        translationsMiddleware.invoke(context, {}, action)
```
gabrielluong, line 405: I would suggest adding the named parameters here and in general. It will make refactoring easier in the future when it's not obvious what a long list of parameters map to.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt
```diff
@@ -19,6 +20,7 @@ import mozilla.components.concept.engine.translate.TranslationSupport
  * @property isTranslated The page is currently translated.
  * @property isTranslateProcessing The page is currently attempting a translation.
  * @property isRestoreProcessing The page is currently attempting a restoration.
+ * @property translationSize The download size if a given to/from pair is translated.
```
gabrielluong, line 23:
```suggestion
 * @property translationSize The download size for the given to/from translation pair.
```

gabrielluong, line 23:
```suggestion
 * @property translationDownloadSize The download size if a given to/from pair is translated.
```
I think one common point that is coming up is about describing what the size represents? Just out of curiosity, this should represent the translation engine size of the translation pair?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -365,6 +380,57 @@ class TranslationsMiddleware(
         }
     }
 
+    /**
+     * Retrieves the download size and dispatches the result to the
+     * store via [TranslationsAction.SetTranslationDownloadSizeAction].
```
gabrielluong, line 385:
```suggestion
     *[BrowserStore] via [TranslationsAction.SetTranslationDownloadSizeAction].
```
I think we should be clear which store we are talking about given all the store we have.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Thanks for taking a look!

Great thing to mention, I was wondering that too. 

My reasoning is I think it makes sense being on the tab because the expected use case is calling this action whenever a translation is about to be requested, so it'll always follow a translate call when the device is in low data mode, so the data need is tab specific in that sense.

For example:
* User presses translate using from `fr` to `en` on tab 1 on low data mode
* Action will follow to request the size to show the user in the prompt for tab 1 and store it
* Prompt will show `fr` to `en` info
* User navigates to tab 2 without clearing prompt
* On tab 2 user presses translate using `es` to `en` on tab 2 in low data mode
* Action will follow to request the size to show the user in the prompt for tab 2 and store it
* Prompt will show `es` to `en` info

If the result was on the global object, then switching back to tab 1 would be required to re-request the data, since the global data would be storing `es` to `en` via tab 2.  So, I think keeping it on the tab might be more conceptually clear for the expected use case. I think we might also have to be more careful about potentially losing the requested translation state.

However, I think it could be valid on the global object and we would just need to expect to re-request and be more careful about managing state.  LMKWYT!

gabrielluong (commented):
I think this all looks good to me. Defer the final approval to @calumozilla

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt
ohall-m: Thanks for taking a look!

Had trouble naming, yes, I think `translationDownloadSize` is more descriptive. 

>I think one common point that is coming up is about describing what the size represents? 

It is the download size of the specific model or models necessary to complete a translation on a given pair. So, for example, `fr` -> `de` will need to download a few ML model files and then run the models on the device to complete the translation. (It'll download `fr` -> `en` and `en` -> `de`, at present.)

Once the user has the models on the device once, no other downloads are needed. For example, "es" -> "en" the first time will be X MB (required download), the next "es" -> "en" request will be 0 MB (no download needed). It would also be 0 MB, if the user had also downloaded the Spanish language "package".

The full "package" or "suite" of models that this [screen downloads](https://www.figma.com/file/GT6oZLFw32LiCwKl5GOMWD/Full-Page-Translation?type=design&node-id=7341-209489&mode=design&t=tiXPvZUQOJeZojlm-0) is conceptually made up of all the pairs for that language. 

Great call-out, I'll update the comments all around to try and make that more clear!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
calumozilla: Gotcha, thanks for explaining the ui flow

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
ohall-m: Thanks, updated, much clearer name!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
ohall-m: Updated, ty!

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt
ohall-m: Rephrased, ty!

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddlewareTest.kt
ohall-m: Updated, agreed!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsState.kt
ohall-m: Updated, ty

ohall-m: Added some more description here and in the object

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Updated, agreed, extremely important now!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
ohall-m: Updated, ty

calumozilla (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -365,6 +381,58 @@ class TranslationsMiddleware(
...
 
+    /**
+     * Retrieves the download size and dispatches the result to the
+     * [SessionState.translationsState] on the [BrowserStore]
+     * via [TranslationsAction.SetTranslationDownloadSizeAction].
+     *
+     * @param context Context to use to dispatch to the store.
+     * @param tabId Tab ID associated with the request.
+     * @param fromLanguage The from language to request the download size for.
+     * @param toLanguage The to language to request the download size for
```
gabrielluong, line 392:
```suggestion
     * @param fromLanguage The from language to request the translation download size for.
     * @param toLanguage The to language to request the translation download size for.
```

gabrielluong (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Great spotting, updated, ty!

PR #5621, opened 2024-02-15
Bug 1880511 - Update Kotlin Coroutines to version 1.8.0

rvandermeulen:
### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1880511


mcarare (approved):
LGTM!

For future reference, this update increases the default timeout for 'runTest' to the previous ( before 1.7.0) value of 60 seconds. We should keep this default value instead of manually setting it to 10 seconds which seemed to cause problems on a lot of other projects CI.

PR #5623, opened 2024-02-15
Bug 1877323 - Use target package name when avoiding fast duplicated app redirects

rocketsroger:
This is an update for a follow up for https://bugzilla.mozilla.org/show_bug.cgi?id=1877039.  Original fix was for a Amazon app bug that causes infinite loop.  This is so we only avoid fast app redirects if the package name matches.  

With the package name check, we can now increase the interval time (to 2 seconds), since we know we are preventing the same intent to the same external app.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1877323


android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt
```diff
@@ -190,9 +197,8 @@ class AppLinksInterceptor(
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
         internal var userDoNotInterceptCache: MutableMap<Int, Long> = mutableMapOf()
 
-        // This should be improved.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1877323
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-        internal var lastHasExternalAppTimestamp: Long = 0L
+        internal var lastApplinksTimestamp: Pair<String?, Long> = Pair(null, 0L)
```
titooan, line 201: I wonder if we shouldn't have a specific type to make it clearer? 

Or at least have a name reflect better its content, like `lastApplinksPackageWithTimestamp`?

titooan (approved):
I just left a small suggestion, but otherwise looks good to me!

android-components/components/feature/app-links/src/main/java/mozilla/components/feature/app/links/AppLinksInterceptor.kt
titooan: Sounds good to me!

PR #5652, opened 2024-02-19
Bug 1880476 — Remove redundant caching and destroy jexl helper after use

jhugman:
Relates to [Bug 1880476](https://bugzilla.mozilla.org/show_bug.cgi?id=1880476).

According to [The Big O of Code Reviews](https://www.egorand.dev/the-big-o-of-code-reviews/), this is a O(_n_) change.

This PR is the second part of a Nimbus improvement in the `NimbusMessagingHelper`, which adds JEXL caching to that class, rather than forcing clients to implement their own caching.

The NimbusMessagingHelper also grew a `destroy` method which calls the Rust drop methods on the native peers. This fixes a small, but long standing memory leak.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1880476

### DIFF
--- a/android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt
+++ b/android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt
@@ -53,8 +53,15 @@ class NimbusMessagingStorage(
     private val customAttributes: JSONObject
         get() = attributeProvider?.getCustomAttributes(context) ?: JSONObject()
 
-    val helper: NimbusMessagingHelperInterface
-        get() = gleanPlumb.createMessageHelper(customAttributes)
+    /**
+     * Returns a Nimbus message helper, for evaluating JEXL.
+     *
+     * The JEXL context is time-sensitive, so this should be created new for each set of evaluations.
+     *
+     * Since it has a native peer, it should be [destroy]ed after finishing the set of evaluations.
+     */
+    fun createMessagingHelper(): NimbusMessagingHelperInterface =
+        gleanPlumb.createMessageHelper(customAttributes)
 
     /**
      * Returns the [Message] for the given [key] or returns null if none found.
@@ -114,13 +121,14 @@ class NimbusMessagingStorage(
      * Returns the next higher priority message which all their triggers are true.
      */
     fun getNextMessage(surface: MessageSurfaceId, availableMessages: List<Message>): Message? =
-        getNextMessage(
-            surface,
-            availableMessages,
-            setOf(),
-            helper,
-            mutableMapOf(),
-        )
+        createMessagingHelper().use {
+            getNextMessage(
+                surface,
+                availableMessages,
+                setOf(),
+                it,
+            )
+        }
 
     @Suppress("ReturnCount")
     private fun getNextMessage(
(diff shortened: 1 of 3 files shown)


rahulsainani:
I think we should include an API like `NimbusMessagingHelperInterface.use { // caller code }` that allows the callers to simply use the helper and this API destroys  the helper internally.

The impl would look like:
```
fun NimbusMessagingHelperInterface.use(block: () -> Unit) {
  block()
  destroy()
}
```

It's not clear right now that destroy needs to be called and we'll hit these issues in the future 🤔

jhugman:
@rahulsainani this is a good idea! I've updated the PR. Please could you review?

fenix/app/src/main/java/org/mozilla/fenix/onboarding/OnboardingFragment.kt
```diff
@@ -249,12 +250,14 @@ class OnboardingFragment : Fragment() {
...
-            jexlConditions,
-        ) { condition -> jexlHelper.evalJexlSafe(condition) }
+        return jexlHelper.use {
+            FxNimbus.features.junoOnboarding.value().cards.values.toPageUiData(
+                privacyCaption,
+                showDefaultBrowserPage,
+                showNotificationPage,
+                showAddWidgetPage,
+                jexlConditions,
+            ) { condition -> jexlHelper.evalJexlSafe(condition) }
```
rahulsainani, line 260: Could we do this 👇  instead?

```
{ condition -> 
  jexlHelper.use {
    evalJexlSafe(condition) }
  }
``` 
So we don't have to wrap the whole mapper call inside the use 🤔

android-components/components/service/nimbus/src/main/java/mozilla/components/service/nimbus/messaging/NimbusMessagingStorage.kt
```diff
@@ -260,23 +266,19 @@ class NimbusMessagingStorage(
...
-                } catch (e: NimbusException.EvaluationException) {
-                    reportMalformedMessage(message.id)
-                    malFormedMap[condition] = message.id
-                    logger.info("Unable to evaluate $condition")
-                    false
+            try {
+                if (malFormedMap.containsKey(condition)) {
+                    return false
                 }
+                helper.evalJexl(condition)
```
rahulsainani, line 275: Similar to this [comment](https://github.com/mozilla-mobile/firefox-android/pull/5652/files#r1494791061), could we call `use` here so we don't wrap the parent functions in use?

fenix/app/src/main/java/org/mozilla/fenix/onboarding/OnboardingFragment.kt
jhugman: No; the jexl helper designed to be used for a collection of JEXL expressions, rather than one then done.

There are two FFI calls for each creation, and one for the destruction, then an internal cache to to memoize the `evalJexl` call.

rahulsainani: Gotcha! I had forgotten that `toPageUiData` is mapping a list to another list and is calling this function for each element, just double checked, this looks good, thanks for the clarification! 👍

rahulsainani (approved):
Looks good! 🚢

PR #5624, opened 2024-02-15
Bug 1876820 - Add translation fetch downloaded languages action

calumozilla:
### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1876820

PR #5051, opened 2024-01-06
Bug 1873302 - Update StoreThreadFactory exception message

jonalmeida:
I'm not particularly proud of this description update as it sounds a bit too similar to Clippy, so suggestions are welcome!

### Pull Request checklist

- [x] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [x] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [x] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1873302

PR #5680, opened 2024-02-21
Bug 1880490 - Moving Translations `supportedLanguages`

ohall-m:
This patch refactors Translations `supportedLanguages` from `[SessionState.translationsState]`
to `[BrowserState.translationEngine]`  in AC.

* Additionally adds `InitTranslationsBrowserState` to signal populating `[BrowserState.translationEngine]`.
* This patch also makes adjustments to migrates Fenix's `TranslationsDialogBinding` and `TranslationsBinding`
 to also monitor this new store location.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1880490


fenix/app/src/test/java/org/mozilla/fenix/telemetry/TelemetryMiddlewareTest.kt
```diff
@@ -86,6 +86,7 @@ class TelemetryMiddlewareTest {
         val engine: Engine = mockk()
         every { engine.enableExtensionProcessSpawning() } just runs
         every { engine.disableExtensionProcessSpawning() } just runs
+        every { engine.getSupportedTranslationLanguages(any(), any()) } just runs
```
ohall-m, line 89: In response to [CI issue here](https://firefoxci.taskcluster-artifacts.net/WIEwqhImSiaR2bmFoGw9mA/0/public/reports/test/testFenixDebugUnitTest/classes/org.mozilla.fenix.translations.TranslationsDialogBindingTest.html#WHEN%20translate%20action%20success%20is%20sent%20to%20the%20browserStore%20THEN%20update%20translation%20dialog%20store%20based%20on%20operation).

iorgamgabriel (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -48,6 +48,7 @@ class TranslationsMiddleware(
             is InitAction ->
                 scope.launch {
                     requestEngineSupport(context)
+                    requestSupportedLanguages(context)
```
gabrielluong, line 51: If the initial `requestEngineSupport()` fails, should we even bother calling `requestSupportedLanguages()`?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TranslationsStateReducer.kt
```diff
@@ -179,12 +178,12 @@ internal object TranslationsStateReducer {
...
         is TranslationsAction.SetSupportedLanguagesAction ->
-            state.copyWithTranslationsState(action.tabId) {
-                it.copy(
+            state.copy(
+                translationEngine = state.translationEngine.copy(
                     supportedLanguages = action.supportedLanguages,
-                    translationError = null,
-                )
-            }
+                    engineError = null,
```
gabrielluong, line 184: I was trying to understand the implications of this logic a bit more and verify the correctness of whether or not we needed to clear `engineError` here. Couple of things I found was that our list of potential [TranslationError](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationError.kt#17) is quite long. However, I think there is also a distinction between engine errors and the other errors.

The only errors that could be an `engineError` is https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt#163 and the newly introduced `CouldNotLoadLanguagesError` in this PR.

I think based on the sequence of events from my understanding of reading the code so far - (1) requestEngineSupport() and then (2) requestSupportedLanguages() in `InitAction` in https://github.com/mozilla-mobile/firefox-android/pull/5680/files#diff-a3c7bcd14d5c6f201fc9d9279918dc1b83ceba808076a42d0f3798ea000ffa62R49. There might not be a point of clearing this `engineError`. We would have an engine error to clear because requesting the engine support reported that we have an `UnknownEngineSupportError`  from `requestEngineSupport()`. We would then call `requestSupportedLanguages()`, but at that point, I don't know if there is a point if we don't have engine support.

Not sure if that made sense so let me know if this was confusing to follow.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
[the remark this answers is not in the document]
ohall-m: I changed how this is launched through a `Job` that the engine will run. I placed a separate `scope.launch` in that function. (Seems safer/more intentional, even if redundant.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -194,12 +205,19 @@ class TranslationsMiddleware(
 
             onError = {
                 context.store.dispatch(
+                    TranslationsAction.EngineExceptionAction(
+                        error = TranslationError.CouldNotLoadLanguagesError(it),
+                    ),
+                )
+
+                if (tabId != null) {
                     TranslationsAction.TranslateExceptionAction(
```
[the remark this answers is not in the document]
ohall-m: Thank you for catching this, updated!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/TranslationsStateReducer.kt
```diff
@@ -148,7 +148,6 @@ internal object TranslationsStateReducer {
                 TranslationOperation.FETCH_SUPPORTED_LANGUAGES -> {
                     state.copyWithTranslationsState(action.tabId) {
                         it.copy(
-                            supportedLanguages = null,
```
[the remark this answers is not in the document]
ohall-m: Good thing to mention! 

>another interpretation here would be that supportedLanguages is already null to begin with (since this is a global state now and we would only ever fetch the supported languages once)
* I would expect it to usually be null if we end up here
* If it wasn't null, I'd wonder why we were doing this operation and hesitant to null it out (a successful operation happened somewhere, so probably should be okay with the result

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Thanks, really good point! So, updated to make `requestEngineSupport` take a job of the remaining things to `init` `onSuccess`. (I have a feeling this area will need further refinement, but I think this should be good for now. No change in testing and I manually verified against Fenix.)

The fallback scenario is something I'm still thinking about. It is pretty clear what to do if we know we support the engine, more unclear if it is unknown.

ohall-m:
I had to make a few larger post review edits:

* `TranslationsBinding.kt` post-rebase, did it the same way as in `TranslationsDialogBinding.kt` and made `TranslationsFlowState` public for this purpose
* Added `requestEngineSupport` to take `() -> Unit` to initialize on successful completion

I plan on landing tomorrow after another look-through/testing. LMK, if there are any issues here on these changes!

cc: @gabrielluong and @iorgamgabriel and @calumozilla

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -143,12 +149,20 @@ class TranslationsMiddleware(
...
         context: MiddlewareContext<BrowserState, BrowserAction>,
+        initialization: () -> Unit,
     ) {
         engine.isTranslationsEngineSupported(
             onSuccess = { isEngineSupported ->
+                if (isEngineSupported) {
+                    scope.launch {
+                        initialization.invoke()
+                    }
+                }
```
gabrielluong, line 165: @MatthewTighe Could I get your thoughts on this? We basically have 2 async functions where the latter depends on the result of the former. I feel we could also do this by dispatching another action and then just hitting the middleware again.

fenix/app/src/main/java/org/mozilla/fenix/browser/TranslationsBinding.kt
```diff
@@ -36,37 +38,63 @@ class TranslationsBinding(
...
+        // Session level flows
+        val sessionFlow = flow.mapNotNull { state -> state.findTab(sessionId) }
+            .distinctUntilChangedBy {
+                it.translationsState
+            }
+
+        // Applying the flows together
+        sessionFlow
+            .combine(browserFlow) { sessionState, browserState ->
+                TranslationsFlowState(
```
gabrielluong, line 56: @MatthewTighe Would also like your thoughts on this?

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Just to mention, in case it changes the best choice, this will grow to include initialization of:
* Languages (this patch)
* Global Settings for Language Preferences #5726
* Global Setting for "Never translate this site" (Currently on session, realized it should be refactored to browser.)
* Model Download States #5624

ohall-m: (Just spotted  a [ test failure](https://firefoxci.taskcluster-artifacts.net/YFsPmM4-QAijMLuPgveFDA/0/public/reports/tests/testReleaseUnitTest/classes/mozilla.components.browser.state.engine.middleware.TranslationsMiddlewareTest.html#WHEN%20InitAction%20is%20dispatched%20AND%20succeeds%20THEN%20SetSupportedLanguagesAction%20is%20dispatched), probably related to how this is invoked. Will wait to fix, pending this conversation. Dispatching again might make testing easier, as a consideration.)

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBinding.kt
```diff
@@ -24,17 +27,60 @@ class TranslationsDialogBinding(
...
+                val browserTranslationsState = state.browserState.translationEngine
+                val translateFromLanguages =
+                    browserTranslationsState.supportedLanguages?.fromLanguages
+                translateFromLanguages?.let {
+                    translationsDialogStore.dispatch(
+                        TranslationsDialogAction.UpdateTranslateFromLanguages(
+                            translateFromLanguages,
+                        ),
+                    )
+                }
```
gabrielluong, line 60:
```suggestion
                val translateFromLanguages =
                    browserTranslationsState.supportedLanguages?.fromLanguages?.let {
                    translationsDialogStore.dispatch(
                        TranslationsDialogAction.UpdateTranslateFromLanguages(
                            it,
                        ),
                    )
                }
```
You can also do the suggested you wanted to.

gabrielluong (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Thanks for taking a look! Added:
* Engine check now calls `InitTranslationsBrowserState` on true success
    * `InitTranslationsBrowserState` calls `requestSupportedLanguages` (and will call other items in the future)

Thanks for mentioning this! Early when working on this, using coroutines was giving me a lot of trouble leaking middleware context via `context.dispatch` and I didn't know why. 😵‍💫 (So much so that @jonalmeida added to the error message in #5051). I for one would love more documentation in this area, especially when we have access to the wiki style docs since it is very subtle.

fenix/app/src/main/java/org/mozilla/fenix/browser/TranslationsBinding.kt
ohall-m: Thanks for taking a look! I'll keep it the same for now.

fenix/app/src/main/java/org/mozilla/fenix/browser/TranslationsBinding.kt
```diff
@@ -36,37 +38,63 @@ class TranslationsBinding(
 ) : AbstractBinding<BrowserState>(browserStore) {
 
     override suspend fun onState(flow: Flow<BrowserState>) {
-        flow.mapNotNull { state -> state.findTab(sessionId) }.distinctUntilChangedBy {
-            it.translationsState
-        }.collect { sessionState ->
-            val translationsState = sessionState.translationsState
+        // Browser level flows
+        val browserFlow = flow.mapNotNull { state -> state }
```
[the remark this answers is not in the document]
ohall-m: Updated, to `.map` in both places for `browserFlow`, thank you!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -143,12 +149,20 @@ class TranslationsMiddleware(
...
     private fun requestEngineSupport(
         context: MiddlewareContext<BrowserState, BrowserAction>,
+        initialization: () -> Unit,
     ) {
         engine.isTranslationsEngineSupported(
             onSuccess = { isEngineSupported ->
+                if (isEngineSupported) {
+                    scope.launch {
+                        initialization.invoke()
+                    }
```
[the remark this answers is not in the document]
ohall-m: Wasn't 100% sure at the time. I updated to the new action flow!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -47,7 +47,13 @@ class TranslationsMiddleware(
         when (action) {
             is InitAction ->
                 scope.launch {
-                    requestEngineSupport(context)
+                    // Once it is determined if the device supports the translation engine, then it
+                    // should continue initializing the rest of the data.
+                    val initialization = {
+                        requestSupportedLanguages(context)
+                    }
```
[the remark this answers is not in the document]
ohall-m: Moved to the action flow! Yes, good point too!

fenix/app/src/main/java/org/mozilla/fenix/browser/TranslationsBinding.kt
```diff
@@ -36,37 +38,63 @@ class TranslationsBinding(
...
+        sessionFlow
+            .combine(browserFlow) { sessionState, browserState ->
+                TranslationsFlowState(
+                    sessionState,
+                    browserState,
                 )
+            }
+            .collect { state ->
+
+                // Browser Translations State Behavior (Global)
```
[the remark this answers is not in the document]
ohall-m: Ty, updated!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBinding.kt
```diff
@@ -24,17 +27,60 @@ class TranslationsDialogBinding(
...
-            .collect { sessionState ->
-                val translationsState = sessionState.translationsState
 
+        // Applying the flows together
+        sessionFlow
+            .combine(browserFlow) { sessionState, browserState -> TranslationsFlowState(sessionState, browserState) }
+            .collect {
+                    state ->
+
+                // Browser Translations State Behavior (Global)
```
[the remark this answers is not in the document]
ohall-m: Updated!

fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationsDialogBinding.kt
ohall-m: Thanks! The value itself was also needed much further below, so left it the same

fenix/app/src/main/java/org/mozilla/fenix/browser/TranslationsBinding.kt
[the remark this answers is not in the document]
ohall-m: Had to revert back to `mapNotNull` to fix:

`FAILURE TO REDEFINE Unable to perform redefinition of 'Lorg/mozilla/fenix/translations/TranslationsDialogBinding$onState$$inlined$mapNotNull$1;': Field 'this$0' (sig: Lorg/mozilla/fenix/translations/TranslationsDialogBinding;) is missing!`

(Trace reference to `mapNotNull` is a little suspect to me in the trace, but it might be happening in the later session combination. The issue stopped reproducing with testing once `map` -> `mapNotNull`.)

The error test behavior was not fetching the languages when a pre-translated state.

Issue only occurred  on an auto-translated page when restoring app from background during manual testing, moving back to `mapNotNull` fixed the issue.

ohall-m:
(Keep getting random CI build intermittents on different components.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Do you mind if I go ahead and allow this to land as is? I immediately have a chance to refactor in #5726 or I might just land again on this bug or new bug.

Testing this patch between changes is very difficult with all of the Fenix dependencies considerations too and I keep hitting CI build intermittent failures on random components. (I can't wait to Phabracator, which allows patch chaining to guarantee landing together, so that Fenix part could have been moved out easier.)

Is this what you are thinking?:
* `InitAction` to call `InitTranslationsBrowserState` directly 
    * Move the engine call out of `InitAction` that is now calling `InitTranslationsBrowserState`
* `InitTranslationsBrowserState` will check the engine and call the remaining functions, all within the same `scope.launch`

Refactor:
`InitAction`:
```
scope.launch {
  context.store.dispatch(InitTranslationsBrowserState)
}

```

`InitTranslationsBrowserState`:
(I think there is a a chance we may need to call this from other places too for error recovery, so I assume a second `scope.launch` is correct?)
```
scope.launch {
   checkEngine();
     -> checkEngine will call finishTranslationsInit()
}

```

 `checkEngine()`
```
// Check engine
// onSuccess  and engine works is true
finishTranslationsInit()

```

`finishTranslationsInit`:

```
fetchLanguages()
fetchSettings()
etc.

```

ohall-m: Filed [1882403](https://bugzilla.mozilla.org/show_bug.cgi?id=1882403) for refactor, so it gets a new review, less entangled, etc.

MatthewTighe: Yep, that's what I was thinking! I don't mind if you land as is and refactor, though I'll leave the final call up to @gabrielluong.

One small note though, in the updating `InitAction` handling:
```
scope.launch {
  context.store.dispatch(InitTranslationsBrowserState)
}
```

The launch isn't needed here, because the `Store::dispatch` does it

gabrielluong: I think it's fine to land as-is and followup. I am still catching up on the conversation here.

ohall-m: Thanks all for the reviews! I'll immediately do the follow-up tomorrow, so the context doesn't get lost too.

ohall-m: >I felt that you did what I was originally imagining in the https://github.com/mozilla-mobile/firefox-android/pull/5680#discussion_r1503387910

Yeah, that is what I followed, I thought "In case that suggestion is followed" was pointing back to [comment 1](https://github.com/mozilla-mobile/firefox-android/pull/5680#discussion_r1503387910) and focused in on the second paragraph in [comment 4](https://github.com/mozilla-mobile/firefox-android/pull/5680#discussion_r1503456124). I somehow missed the first paragraph in comment 4 saying avoid.

I can see some benefit to having `InitTranslationsBrowserState` for re-initializing everything too if an error occurs. (It adds complexity, but now I'm wondering if it is justified?)

Open to whatever refactor, let me know what you all prefer! I'll tag you both on the refactor review too, so we can pick it up from there.

MatthewTighe: Sorry for muddying the waters! To clarify, my suggestion would be:

1. InitAction observed by Middleware
2. Middleware dispatches InitTranslationBrowserState through `context.store.dispatch`
3. Middleware observes `InitTranslationBrowserState`
4. scope.launch
    1.  checkEngine()
    2. finishTranslationsInit()

I think that will run the lowest risk of performance impact to BrowserStore startup and also atomize the initialization of the translation state.

PR #5681, opened 2024-02-21
Bug 1879663 - Add RFC proposing updates to the RFC process

MatthewTighe:
This patch introduces a number of small updates to the RFC process. I've selected a long initial deadline since this is just a process change and isn't blocking any work, plus things have been chaotic the last couple weeks. 

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


docs/rfcs/0013-rfc-process-updates.md
```diff
@@ -0,0 +1,39 @@
...
+This proposal suggests improving the RFC process by introducing the following:
+
+- an RFC template
+- a guide for when RFCs are appropriate and when they are not needed in the form of a README
+- explicit stakeholders for RFCs
+- an initial deadline for each RFC
+- a new RFC section, "artifacts" which contain links to any additional artifacts introduced alongside the RFC
+- updating our CODEOWNERS file to better facilitate ownership for new Firefox Android "groups" and external teams
+
+Each of these updates are handled as part of this RFC, with the exception of updating the CODEOWNERS file. That will need to undergo a discovery process and communication between Firefox Android groups and other teams.
```
MatthewTighe, line 34: I'm not sure if there's a better way to address this issue than reaching out to potentially CODEOWNERS directly but would be interested in a more public approach if anyone has suggestions.

docs/rfcs/0000-template.md
```diff
@@ -0,0 +1,49 @@
...
+## Drawbacks
+
+This section should include any drawbacks to the proposal. This section can be considered optional, but is highly encouraged in order to facilitate healthy debate.
+
+## Rationale and alternatives
+
+This section should include any alternative proposals considered, as well as the rationale for why
+they were not selected for the proposal. This section can be considered optional, but is highly encouraged in order to facilitate healthy debate.
+
+## Prior art
```
rahulsainani, line 39: I think this and Artifacts can be combined in one section called `Resources and Docs`, wdyt?

docs/rfcs/0013-rfc-process-updates.md
```diff
@@ -0,0 +1,39 @@
...
+## Guide-level explanation
+
+This proposal suggests improving the RFC process by introducing the following:
+
+- an RFC template
+- a guide for when RFCs are appropriate and when they are not needed in the form of a README
+- explicit stakeholders for RFCs
+- an initial deadline for each RFC
+- a new RFC section, "artifacts" which contain links to any additional artifacts introduced alongside the RFC
+- updating our CODEOWNERS file to better facilitate ownership for new Firefox Android "groups" and external teams
```
boek, line 32: I think using CODEOWNERS as part of the new process is great, but let's not couple updating CODEOWNERS as part of this RFC. With that we can also remove the explanation of why we're not actually updating the file with this change.

MatthewTighe: Yeah, fair enough. I filed [this follow-up to handle CODEOWNERS](https://bugzilla.mozilla.org/show_bug.cgi?id=1881373) and will include it in the artifacts for this RFC

docs/rfcs/0000-template.md
MatthewTighe: I think it seems useful to have separate sections for external resources and artifacts that are being added to the tree as part of the RFC , but I'm open to changing the name. "Prior art" is just borrowed from earlier RFC examples

docs/rfcs/0000-template.md
jonalmeida, line 1: What is the difference between this template and [the previous one](https://mozac.org/rfc/0001-rfc-process)? The prior one worked as a way to introduce the process and be used as a template.

Can we just amend that one with additional information that might be missing?

docs/rfcs/0013-rfc-process-updates.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Summary
+
+Clarify requirements for RFCs by introducing a README and a template.
+
+## Motivation
+
+- The scale of the Firefox Android team has grown substantially since the RFC process was first introduced. 
+- More external teams are using and contributing to Android Components. 
+- Approval or rejection of a proposal has required periods of time that can slow implementation.
```
jonalmeida, line 20: The approval or rejection time is set by the author already. The RFC is just a pull request that goes through the same process as a code review: if you have an approval, you can merge that work. Additional reviewers are welcome to add feedback that is up to the author to consider if they want to address it in the current PR, future work, or not applicable.

I'm not sure this is a valid motivation. What is the difference that we are adding here?

docs/rfcs/0013-rfc-process-updates.md
```diff
@@ -0,0 +1,37 @@
...
+- Stakeholders have either been implicit by sharing RFCs in various communication channels, or an author must be expected to correctly select reviewers that lack explicitly documented ownership of affected areas.
```
jonalmeida, line 21: It's fine to have secondary channels of communication for feedback - everyone has their own way. The missing step that I've seen in practice, is that developers forget to come back to the origin (their original doc or bug or PR), and document the communication and it's outcome.

If we want specific feedback from a person, we add them as a reviewer to the PR today. It's up to the author to decide then if they want to wait for that feedback or they are happy with the review of another reviewers.

Below it says, "explicit stakeholders for RFCs" which contradicts, "an author must be expected to correctly select reviewers.."

I'd recommend instead making this into something similar to: "New RFCs may not be visible to the entire team to have equal time to consider."

The proposal could then be: "Share the RFC to the mailing list/slack channel with the deadline date."

For prior art of what I'm suggesting, see the web standards process or the `services-rfcs@` mailing list.

docs/rfcs/0013-rfc-process-updates.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Guide-level explanation
+
+This proposal suggests improving the RFC process by introducing the following:
+
+- an RFC template
+- a guide for when RFCs are appropriate and when they are not needed in the form of a README
+- explicit stakeholders for RFCs
+- an initial deadline for each RFC
+- a new RFC section, "artifacts" which contain links to any additional artifacts introduced alongside the RFC
```
jonalmeida, line 31:
```suggestion
- An RFC template.
- A guide for when RFCs are appropriate and when they are not needed in the form of a README.
- Explicit stakeholders for RFCs.
- An initial deadline for each RFC.
- A new RFC section, "artifacts" which contain links to any additional artifacts introduced alongside the RFC.
```

nit: typos.

If you consider the comments above, this list would also be shorter, but this suggestion is just for the typos in the current revision.

docs/rfcs/0013-rfc-process-updates.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Artifacts
+
+[README](./README.md)
+[0000 RFC Template](./0000-template.md)
+[Follow-up bug for updating CODEOWNERS](https://bugzilla.mozilla.org/show_bug.cgi?id=1881373)
```
jonalmeida, line 37:
```suggestion
* [README](./README.md)
* [0000 RFC Template](./0000-template.md)
* [Follow-up bug for updating CODEOWNERS](https://bugzilla.mozilla.org/show_bug.cgi?id=1881373)
```

nit: formatting.

jonalmeida: This list without references to what it is refering to can be confusing. For example, what does this README link reference here? Consider adding anchor links or annotations.

That said, I'm also not sure what the difference between this section is to the 'Prior Art' one.

docs/rfcs/README.md
jonalmeida, line 1: As mentioned above, there is a lot of duplication here. Can we focus on the new additions to the process instead? e.g. formal deadlines in the doc.

jonalmeida (commented):
Thanks for doing this! Left some feedback below that revolves around focusing on the new additions. I'm slightly concerned that we making this more formal than it needs to be and having the inverse effect you are trying to avoid (i.e. slowing down development).

Let's also focus on making this change more about the new additions and not what already exists.

docs/rfcs/0000-template.md
MatthewTighe: The value add here would just be around clarity: the new template has detail about what each section is for, the file is clearly named, and has additional sections that have been pulled from other examples. How would we amend the original with additional templating information without diluting the proposal within it?

I also think there is value in treating an accepted RFC as finalized and not updating it in place after the fact, as I've come to learn 😅

docs/rfcs/0013-rfc-process-updates.md
MatthewTighe: The intent here is clarity and centralization of information. Setting a deadline in a different communication channel divorces that information from the proposal if the proposal is shared directly, for example.

My goal is to make these easier to write for people unfamiliar with the process, and codifying things like this I think will assist in knowledge transfer - future authors will have less reliance on offline (and usually hidden from public/history) mentorship.

docs/rfcs/0013-rfc-process-updates.md
MatthewTighe: > It's fine to have secondary channels of communication for feedback - everyone has their own way.

I am hoping this removes some of the guesswork in finding out what every reviewer's or author's individual way is, especially for people that are new to the process or working across teams/areas of ownership.

> Below it says, "explicit stakeholders for RFCs" which contradicts, "an author must be expected to correctly select reviewers.."

In the way that selecting explicit stakeholders and reviewers is basically the same thing? I do agree here, but I am hoping with a revised CODEOWNERS file that this is improved and that by offering the suggestion as part of the guide and template, it will give authors an explicit step in seeking feedback.

> The proposal could then be: "Share the RFC to the mailing list/slack channel with the deadline date."
> For prior art of what I'm suggesting, see the web standards process or the services-rfcs@ mailing list.

The services-rfcs mailing list looks great. I would definitely like to include something like that as part of the guide for distribution of these, but I don't feel like the firefox-android mailing list is all-inclusive enough especially as we talk about other teams potentially taking ownership of specific areas. Do you have any suggestions, or should we make a new one?

docs/rfcs/README.md
MatthewTighe: This file is intended as a guide for anyone looking to write an RFC, and is included in this patch to capture the context in which it was introduced. There is definitely duplication, but that is intentional

docs/rfcs/0013-rfc-process-updates.md
MatthewTighe: I tried to answer this a bit in [this comment](https://github.com/mozilla-mobile/firefox-android/pull/5681/files#r1498491586) as well, but I see "artifacts" as internal documentation or other resources that were generated as part of the proposal and "prior art" as reading into external sources that cover the same topic as the proposal

MatthewTighe (commented):
> I'm slightly concerned that we making this more formal than it needs to be and having the inverse effect you are trying to avoid (i.e. slowing down development).

I think this is a fair concern. My hope is that introducing a little more formality will make the process more accessible to people unfamiliar with it (or with the people who might engage in it). Longer-term Mozillians could probably just write proposals for each other all day because we better understand each other's work styles and expectations, but I'm hoping writing those down a bit will actually remove barriers to entry.

jonalmeida (approved):
I am still not convinced in the value of the duplication and some of the wording, but not enough to block on this work if it's intentional.

boek (approved):

docs/rfcs/0000-template.md
```diff
@@ -0,0 +1,49 @@
+---
+layout: page
+title: <your title>
+permalink: /docs/rfcs/<file-name>
+---
+
+* RFC PR: [<PR #>](https://github.com/mozilla-mobile/android-components/pull/<#>)
+* Start date: YYYY-MM-DD (day of posting)
+* End date: YYYY-MM-DD (initial deadline for feedback, subject to blocking changes required)
```
boek, line 9: Do we keep the PRs open until the end date even when approved for others to have the ability to comment?

MatthewTighe: Updated in a couple places to make clear that proposals can be merged early once they've been approved

docs/rfcs/0013-rfc-process-updates.md
MatthewTighe: > My impression is that we are solving for the wrong problem then. 🤷

Is there another problem you think we should be solving for? I am basing this proposal off my own and other's feedback that a lack of clarity has lowered engagement and speed of this process, but I am open to try and solve additional problems with this.

docs/rfcs/0000-template.md
MatthewTighe: > Since it is meant as a template, like our previous case, we can still amend it with annotations.

Is there an example of somewhere we've done that already? My hope is that the template and README become living, update-able documents that are clear entry points to the process. I think if we start amending RFCs, then it becomes less clear which are the most up-to-date, and begs the question of whether they serve as living documentation for the systems and changes they had proposed.

docs/rfcs/0000-template.md
```diff
@@ -0,0 +1,47 @@
+---
+layout: page
+title: <your title>
+permalink: /docs/rfcs/<file-name>
+---
+
+* RFC PR: [<PR #>](https://github.com/mozilla-mobile/android-components/pull/<#>)
+* Start date: YYYY-MM-DD (Day of proposal posting.)
+* End date: YYYY-MM-DD (Initial deadline for feedback, subject to blocking changes required. The proposal can be merged immediately after stakeholder approval.)
+* Stakeholders: <github-username>, <github-username>
```
t-p-white, line 10: <github-username> doesn't show up in the fie view, is that expected?
Screenshot 2024-02-28 at 10 21 48

docs/rfcs/0000-template.md
```diff
@@ -0,0 +1,47 @@
+---
+layout: page
+title: <your title>
```
t-p-white, line 3: I think generally we can remove the angled brackets in the file when used to reference 'your thing' as they're fairly self-explanatory? imo it just adds visual noise without much, if any benefit?

docs/rfcs/0000-template.md
```diff
@@ -0,0 +1,47 @@
...
+permalink: /docs/rfcs/<file-name>
+---
+
+* RFC PR: [<PR #>](https://github.com/mozilla-mobile/android-components/pull/<#>)
```
t-p-white, line 7: will this be updated to include a PR number?

docs/rfcs/0000-template.md
```diff
@@ -0,0 +1,47 @@
...
+* Start date: YYYY-MM-DD (Day of proposal posting.)
+* End date: YYYY-MM-DD (Initial deadline for feedback, subject to blocking changes required. The proposal can be merged immediately after stakeholder approval.)
```
t-p-white, line 9:
```suggestion
* End date: YYYY-MM-DD (Last day for general feedback, subject to blocking changes required. The proposal can be merged immediately after stakeholder approval.)
```

t-p-white, line 9: maybe we just call `End Date` `Deadline`?

t-p-white (commented):
There's a few 'optional' sections, perhaps we could hightlight that in the section title instead of in the section description? e.g. something like:

- Drawbacks (Optional)
- *Drawbacks

docs/rfcs/README.md
```diff
@@ -0,0 +1,47 @@
...
+### Stakeholders
+
+Stakeholders are required for each RFC. They will have the final say in acceptance and rejection. 
+Include at least 2 people as stakeholders: a CODEOWNER of the affected area and another (preferably a Firefox for Android team member).
+
+Stakeholders should be active in the RFC process - they should ask to be replaced if they do not have bandwidth to get the RFC finished in a short time span. This is to help the RFC process remain nimble and lightweight.
+
+### Deadlines
+
+An initial deadline should be included in each RFC. This should usually be at least a week, so plan accordingly.
```
t-p-white, line 44: Not sure on the use of `intial` here as it's a little fuzzy and implies there are other deadlines. I think ultimately what we're trying to convey here is this is the last day for _general_ feedback

docs/rfcs/README.md
```diff
@@ -0,0 +1,47 @@
...
+Stakeholders are required for each RFC. They will have the final say in acceptance and rejection. 
+Include at least 2 people as stakeholders: a CODEOWNER of the affected area and another (preferably a Firefox for Android team member).
+
+Stakeholders should be active in the RFC process - they should ask to be replaced if they do not have bandwidth to get the RFC finished in a short time span. This is to help the RFC process remain nimble and lightweight.
+
+### Deadlines
+
+An initial deadline should be included in each RFC. This should usually be at least a week, so plan accordingly. 
+For more substantial changes, it can be useful to push that to 2 or 3 weeks so that there is more opportunity for feedback from people that are not stakeholders.
+Blocking feedback can push these deadlines back, but they should be updated appropriately so that interested parties have a clear understanding of timelines.
```
t-p-white, line 46: Based on my comment above, again I'm not sure if we should say pushing deadlines back, it's more that this stage is dealing with any blockers, otherwise the deadline becomes a moving target which I don't think adds any value here?

t-p-white (commented):
Nice job @MatthewTighe, great that you took the initiative to put this together, should hopefully make RFCs a little more user-friendly going forward 🙂 just a few general comments from me

docs/rfcs/README.md
```diff
@@ -0,0 +1,47 @@
...
+3. Proposals for changes to areas of the codebase that are owned by CODEOWNERS outside the author's team.
+
+## What kind of other changes can an RFC be useful for?
+
+1. Announcing a rough plan for changes to a public API you own in order to solicit feedback.
+2. Soliciting feedback for architectural changes that affect the entire codebase, like [0009 Remove Interactors and Controllers](./0009-remove-interactors-and-controllers.md).
+
+## How to contribute an RFC
+
+There is a [template](./0000-template.md) that can be useful for structure. While drafting
```
t-p-white, line 24: > While drafting a proposal, consider the scope of your changes. Generally, the higher the impact the changes will have would be things like example use cases, prototypes, and considered alternatives.

on downstream consumers of APIs, other teams, or users the more detail the RFC should include. Detail
In an effort to keep this README concise, I wonder if this information is adding anything the new template isn't? Or perhaps this should move to the template?

docs/rfcs/README.md
MatthewTighe: Thanks this is useful feedback! Updated accordingly

docs/rfcs/README.md
MatthewTighe: I think this is useful general advice, and that it would probably need to be repeated in a few sections in the template so I've kept it here for now but tried to streamline it a bit. LMK if you feel strongly it should change

docs/rfcs/0013-rfc-process-updates.md
```diff
@@ -0,0 +1,37 @@
...
+## Summary
+
+Clarify requirements for RFCs by introducing a README and a template.
+
+## Motivation
+
+- The scale of the Firefox Android team has grown substantially since the RFC process was first introduced. 
+- More external teams are using and contributing to Android Components. 
+- Implicit deadlines have encouraged a lack of engagement with RFCs, slowing follow-up work.
+- There has been a low level of engagement with RFCs because of a lack of clarity around the process and when they are appropriate.
```
jonalmeida, line 21:
```suggestion
- There has been a low level of engagement with RFCs because of a lack of clarity around the process.
```

When they are appropriate is on the [first RFC](https://mozac.org/rfc/0001-rfc-process), so I don't think this makes sense to add. The template you are introducing does not solve the "when" problem either.

docs/rfcs/0000-template.md
```diff
@@ -0,0 +1,46 @@
...
+
+This section should include any alternative proposals considered, as well as the rationale for why
+they were not selected for the proposal.
+
+## Resources and Docs (optional)
+
+- Any (internal or external) similar proposals  or other documentation that shares concepts with the proposal.
+- Links to artifacts generated as part of the proposal, such as additional documentation or follow-up bugs.
+
+## Unresolved questions (optional)
```
jonalmeida, line 44:
```suggestion
## Unresolved questions
```

Let's not make this optional. If there are no unresolved questions, then this section would require the developer to be explicit that they considered it and couldn't find any. Otherwise, all optional sections will most likely always be ignored.

docs/rfcs/0000-template.md
```diff
@@ -0,0 +1,46 @@
...
+## Reference-level explanation (optional)
+
+This section should include a detailed walkthrough of technical steps or code changes that 
+will be required to implement the proposal. Code samples and prototypes are beneficial to include here.
+
+## Drawbacks (optional)
+
+This section should include any drawbacks to the proposal.
+
+## Rationale and alternatives (optional)
```
jonalmeida, line 34:
```suggestion
## Rationale and alternatives
```

Sorry, I missed this in my previous pass. For the same reason as the comment above on 'Unresolved questions', let's make it required that the developer explicitly says there are no alternatives to their design.

jonalmeida (commented):
Circling back to review the latest changes made and one I missed (sorry!).

docs/rfcs/0013-rfc-process-updates.md
MatthewTighe: Ah, I had hoped the README (as a slightly more abstract document which would serve as a landing place in the directory) would answer that question. Should the two documents just be combined?

jonalmeida: Sure. I have no strong preferences, so feel free to add it as a follow-up. 

Just make sure we align it with the 001 RFC to avoid confusion since we are two places now that offer recommendations.

MatthewTighe:
Landing this before the deadline since the monorepo merge is scheduled this week and the stakeholders have approved

PR #5708, opened 2024-02-22
Bug 1878185 - Add ChangeDetectionMiddleware

MatthewTighe:
### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


fenix/app/src/main/java/org/mozilla/fenix/components/ChangeDetectionMiddleware.kt
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
boek, line 1: Could we find a better place for this file to live than `org.mozilla.fenix.components`? Right now this namespace has turned into quite the junk drawer.

boek (changes requested):
The code looks good. I think that we can find a better place to put this before we land it

fenix/app/src/main/java/org/mozilla/fenix/components/ChangeDetectionMiddleware.kt
MatthewTighe: I'm definitely open to suggestions, but tbh I feel like many packages in the app are currently junk drawers and that the appropriate time to sort them will be if and when we start modularizing. I also think starting that process here by creating a common module would be a misstep and we should instead do some up-front planning and investigative work. Wdyt?

boek: Gonna try out this new RFC process

boek (approved):

PR #5726, opened 2024-02-22
Bug 1862291 - Operation for Fetching Global Language Settings

ohall-m:
This patch adds functionality for fetching global language settings for translations.

This patch adds:
* The Operation `FETCH_AUTOMATIC_LANGUAGE_SETTINGS` for fetching the global language settings from the translation engine.
* `SetLanguageSettingsAction` to set the `languageSettings` on `TranslationsBrowserState`.
* Setting up automatically completing this operation in `InitAction`
* Updates page settings to call `FETCH_AUTOMATIC_LANGUAGE_SETTINGS` after a page level language update. (These settings coordinate both globally and on a page level.)

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


ohall-m:
Related to helper added in #5727.
Followed the same pattern for a global/non-global as in #5680.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -62,6 +66,11 @@ class TranslationsMiddleware(
                             requestPageSettings(context, action.tabId)
                         }
                     }
+                    TranslationOperation.FETCH_AUTOMATIC_LANGUAGE_SETTINGS -> {
```
ohall-m, line 69: Manual action would only be used if something went wrong during `init` or to force a refresh. (Page settings and global language settings are intertwined.) 

I plan to move the current `neverTranslateSites` off of the session store and onto the global store too in [1881611](https://bugzilla.mozilla.org/show_bug.cgi?id=1881611). (Originally made sense because of the intertwining with the `pageSettings`, but I think long term, it would be better on the global object for an easier `init`.)

ohall-m:
Failing [intermittent](https://firefoxci.taskcluster-artifacts.net/GGZ_-5sTQJuGyHXdL64E_A/0/public/reports/test/testFenixDebugUnitTest/index.html) fixed in #5680 at this [line](https://github.com/mozilla-mobile/firefox-android/pull/5680/files#r1497671358)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
```diff
@@ -1033,6 +1034,16 @@ sealed class TranslationsAction : BrowserAction() {
         val setting: Boolean,
     ) : TranslationsAction(), ActionWithTab
 
+    /**
+     * Sets the map of BCP 47 laguage codes (key) and the [LanguageSetting] option (value).
```
MatthewTighe, line 1038: nit: 
```suggestion
     * Sets the map of BCP 47 language codes (key) and the [LanguageSetting] option (value).
```

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationOperation.kt
```diff
@@ -30,6 +30,11 @@ enum class TranslationOperation {
      */
     FETCH_PAGE_SETTINGS,
 
+    /**
+     * The list of never translate sites the translation engine should fetch.
+     */
+    FETCH_AUTOMATIC_LANGUAGE_SETTINGS,
```
MatthewTighe, line 36: Based on the comment, I'm wondering if this name should be something more like `FETCH_NEVER_TRANSLATE_SITES_SETTINGS`?

MatthewTighe (approved):
Couple small things but nothing blocking. LGTM!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsBrowserState.kt
```diff
@@ -4,16 +4,19 @@
...
 package mozilla.components.browser.state.state
 
+import mozilla.components.concept.engine.translate.LanguageSetting
 import mozilla.components.concept.engine.translate.TranslationError
 
 /**
  * Value type that represents the state of the translations engine within a [BrowserState].
  *
  * @property isEngineSupported Whether the translations engine supports the device architecture.
+ * @property languageSettings The automatic language settings of the translations engine.
```
gabrielluong, line 14: Might be worthwhile to describe the `Map` here similar to https://github.com/mozilla-mobile/firefox-android/pull/5726/files#diff-4dd92b761139e3e0ebeb8bcacaa1162d5708fc809f6b65dd2f88bcf9b5d208f1R1040 since it won't be immediately clear what the `String` key here represents.

android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationOperation.kt
gabrielluong: I suspect the comment used here should actually be closer to what you used in https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/concept/engine/src/main/java/mozilla/components/concept/engine/translate/TranslationsRuntime.kt#164.

```suggestion
    /**
     * Fetch the user preference on whether to offer, always translate, or never translate for all supported language settings.
     */
    FETCH_AUTOMATIC_LANGUAGE_SETTINGS,
```

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddlewareTest.kt
```diff
@@ -559,4 +560,150 @@ class TranslationsMiddlewareTest {
...
+        val languageSettingsCallback = argumentCaptor<((Map<String, LanguageSetting>) -> Unit)>()
+        // Checking twice, because InitAction is also implicitly called earlier
+        verify(engine, atLeast(2)).getLanguageSettings(
+            onSuccess = languageSettingsCallback.capture(),
+            onError = any(),
+        )
+    }
+
+    @Test
+    fun `WHEN an Operation to FETCH_AUTOMATIC_LANGUAGE_SETTINGS has an issue THEN TranslateExceptionAction is dispatched for language setting`() = runTest() {
```
gabrielluong, line 653:
```suggestion
    fun `WHEN an Operation to FETCH_AUTOMATIC_LANGUAGE_SETTINGS has an error THEN TranslateExceptionAction is dispatched for language setting`() = runTest() {
```
This might provide better clarity.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -361,6 +370,60 @@ class TranslationsMiddleware(
...
+                        error = TranslationError.CouldNotLoadLanguageSettingsError(it),
+                    ),
+                )
+
+                if (tabId != null) {
+                    TranslationsAction.TranslateExceptionAction(
+                        tabId = tabId,
+                        operation = TranslationOperation.FETCH_AUTOMATIC_LANGUAGE_SETTINGS,
+                        translationError = TranslationError.CouldNotLoadLanguageSettingsError(it),
+                    )
```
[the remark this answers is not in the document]
ohall-m: >I am kinda getting lost if not dizzy from just the many engines and operations we have.

I agree, I've been wondering about how best to manage error state too. There is something that definitely can be improved, but am not yet sure about the best way to go about it. 

Some of what is tricky is there is a lot of error handling, both globally and per-tab. I'm not sure the best way to consolidate and also keep everything in sync. The pattern I'm trying right now is `Init or global request = global error`, `tab request = session + global error`. A failed language request is also going to impact `Tab 2`, even if `Tab 1` requested. (But, I don't necessarily want to overwrite `Tab 2`s error.) I don't expect a tab to be calling the languages, unless something went wrong during `Init` too.

A good start to the refactor is `translationError`'s name should ideally be refactored to something like `generalError` or  `error`, `sessionError`. (I'm not totally sure `settingsError` is necessary, but  at the time was it seemed like it was a different path/non-core error and didn't want to overwrite a core error, which is another consideration.) I have a feeling there might be a nicer pattern to avoid overwriting errors too. 

One major concern outside of the intertwinedness/complexity I have right now is wondering how overwriting errors is going to impact error handling. (e.g., a. more recent error can overwrite an older error). The solution might be an error map that it is tied to operation or something like `TranslationOperation`  keys and `TranslationError` values. (There would be some redundancy, since implicitly certain errors are tied to certain operations, but it might be clearer in the views what to check for and since this has grown so much 😵‍💫.)

There is a lot to explore here, I'll file a bug to investigate!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/action/BrowserAction.kt
ohall-m: Thank you, updated!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -361,6 +370,60 @@ class TranslationsMiddleware(
...
+    /**
+     * Retrieves the list of languages and their settings and dispatches the result to the
+     * [BrowserState.translationEngine] via [TranslationsAction.SetLanguageSettingsAction] or
+     * else dispatches the failure.
+     *
+     * For failure dispatching:
+     * If a tab ID is not provided, then only [TranslationsAction.EngineExceptionAction] will be
+     * dispatched to set the error on the [BrowserState.translationEngine].
+     *
+     * If a tab ID is provided, then  [TranslationsAction.EngineExceptionAction]
```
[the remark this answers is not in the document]
ohall-m: Ty!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -361,6 +370,60 @@ class TranslationsMiddleware(
...
+     * AND [TranslationsAction.TranslateExceptionAction] will be dispatched
+     * to set the error both on the [BrowserState.translationEngine] and
+     * [SessionState.translationsState].
+     *
+     * @param context Context to use to dispatch to the store.
+     * @param tabId If a Tab ID is associated with the request for error handling.
+     * If null, this will only dispatch errors on the global translations browser state.
+     *
+     */
```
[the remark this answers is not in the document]
ohall-m: Updated

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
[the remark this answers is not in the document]
ohall-m: My thoughts were, if it was a request with a tab, then yes (for now, at least), so it sets the error on the tab. 

Related to this [discussion](https://github.com/mozilla-mobile/firefox-android/pull/5726#discussion_r1502075892) for sure. (LMK, if I missed the intent of the question too.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/state/TranslationsBrowserState.kt
ohall-m: Thanks, good catch!

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddlewareTest.kt
ohall-m: Updated, good idea!

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddlewareTest.kt
```diff
@@ -559,4 +560,150 @@ class TranslationsMiddlewareTest {
...
+            TranslationsAction.TranslateExceptionAction(
+                tabId = tab.id,
+                operation = TranslationOperation.FETCH_AUTOMATIC_LANGUAGE_SETTINGS,
+                translationError = TranslationError.CouldNotLoadLanguageSettingsError(any()),
+            ),
+        )
+    }
+
+    @Test
+    fun `WHEN InitAction is dispatched AND has an issue THEN TranslateExceptionAction is dispatched for language setting`() = runTest() {
```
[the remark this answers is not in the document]
ohall-m: Updated, ty

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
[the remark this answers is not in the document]
gabrielluong: I actually meant that we aren't doing a `context.store.dispatch()` of the `TranslationsAction.TranslateExceptionAction` here.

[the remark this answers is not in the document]
ohall-m: Ohh, yikes, I'll take a look and at the test too! Thanks for catching!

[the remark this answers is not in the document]
ohall-m: Added a `waitForIdle()` at the end of the tests. Had thought it was superfluous, but apparently was not. (I've had this issue earlier. I think the current tests block as expected, but have seen odd behavior especially with the exception dispatches verifying in particular.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -45,10 +45,14 @@ class TranslationsMiddleware(
     ) {
         // Pre process actions
         when (action) {
-            is InitAction ->
+            is InitAction -> {
                 scope.launch {
                     requestEngineSupport(context)
                 }
+                scope.launch {
```
ohall-m, line 52: Plan on refactoring this after #5680 lands to match new ordering accordingly. (e.g., posting after the engine support is requested and successfully true.)

ohall-m:
(Soft landing block until #5680 lands and [1882403](https://bugzilla.mozilla.org/show_bug.cgi?id=1882403), so this can be rebased to call from new ideal place.)

ohall-m:
Updated tests and adjusted how initialization occurred based on recent refactor.

PR #5768, opened 2024-02-26
Bug 1882151 - Always and Never Page Settings Bug

ohall-m:
This bug updates the logic for "never translate this language" and "always translate this language" to ensure they are mutually exclusive when one is true, but also allows them both to be false.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1882151


android-components/components/browser/state/src/test/java/mozilla/components/browser/state/action/TranslationsActionTest.kt
```diff
@@ -531,6 +531,48 @@ class TranslationsActionTest {
         assertTrue(tabState().translationsState.pageSettings?.neverTranslateSite!!)
     }
 
+    @Test
+    fun `WHEN an UpdatePageSettingAction is dispatched for UPDATE_ALWAYS_TRANSLATE_LANGUAGE AND UPDATE_ALWAYS_TRANSLATE_LANGUAGE THEN must be opposites of each other or both must be false `() {
+        // Action started to update the always offer setting to true
```
gabrielluong, line 536: Noticing we don't assert the initial state in some of these tests here. I think in general we should assert the initial state to ensure dispatching the action did indeed yield a change in the state.

iorgamgabriel (approved):

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/action/TranslationsActionTest.kt
ohall-m: Updated here! (Didn't update other areas to keep this a small and quick patch. Agree it is a good practice!)

PR #5773, opened 2024-02-26
Bug 1882196 - Adding a nav bar view

mavduevskiy:
This PR is moving the navigation bar [dev_branch](https://github.com/mozilla-mobile/firefox-android/tree/dev_navigation_bar) to main. Commits were originally separate PRs that were approved individually, with some added comments and smaller refactoring. Originally, we were discussing to patches separately, but after reviewing it, I discovered that it might end up more confusing and that the change so far isn't that big.

The navigation bar is only visible on debug builds.

### Screenshots
 Old | New
 --- | --- 
 |  
 |  

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
4. Click on `View task in Taskcluster` in the new `DETAILS` section.
5. The APK links should be on the right side of the screen, named for each CPU architecture.

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1867698
https://bugzilla.mozilla.org/show_bug.cgi?id=1868260
https://bugzilla.mozilla.org/show_bug.cgi?id=1868262
https://bugzilla.mozilla.org/show_bug.cgi?id=1873991
https://bugzilla.mozilla.org/show_bug.cgi?id=1882196

PR #5783, opened 2024-02-27
Bug 1879430 - Disable nav bar for horizontal mode and tablets

mavduevskiy:
Lands after nav view [patch](https://github.com/mozilla-mobile/firefox-android/pull/5773)
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1879430

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -112,6 +112,7 @@ import mozilla.components.support.ktx.android.view.hideKeyboard
 import mozilla.components.support.ktx.kotlin.getOrigin
 import mozilla.components.support.ktx.kotlinx.coroutines.flow.ifAnyChanged
 import mozilla.components.support.locale.ActivityContextWrapper
+import mozilla.components.support.utils.ext.isLandscape
 import mozilla.components.ui.widgets.behavior.EngineViewClippingBehavior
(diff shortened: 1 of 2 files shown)


rocketsroger (approved):

rocketsroger:
@mavduevskiy though the code looks clean,  there are some weird behaviour when I rotate in home screen vs in a tab.  Also once the Navigation bar is hidden it doesn't return when I rotate back.

rocketsroger (changes requested):
block my own approval since I see some weirdness when testing this change.

rocketsroger:
I found out how I can reproduce the issue I'm seeing.  If I build and install while the device is in landscape, then the navigation bar never shows even if I rotate back to portrait.  If I install/or fresh start on protrait, then the navigation bar will always show even when I rotate to landscape.

mavduevskiy:
okay, it might be a bigger problem and a bigger change required. 

I was assuming that the view gets recreated on the orientation change, but it doesn't. so utilizing ui initializastion doesn't work.

so we have to listen to the change and manually adjust the state of toolbars.

gabrielluong:
Let's move this over to phabricator and mozilla-central.

rahulsainani:
@rocketsroger @mavduevskiy The config issue should be resolved with this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1855519) and this should work as expected. 

Altough this should be part of the `isEnabled` logic so the same `wideDisplay` check is not repeated.

PR #5790, opened 2024-02-27
Bug 1882106 - Show undo snackbar when tab is closed via tab strip

rahulsainani:
Details in the bug

### How
– Extract common undo snackbar message utility
– Use that in existing places
– Add callbacks for `onCloseTab(private)` in `TabStrip`
– display snackbar using the callbacks and the utility function

There's a scenario where BrowserFragment navigates to Home when all tabs are closed. To "remember"/"know" to show the snackbar, `HomeScreenViewModel` is used as it's Activity scoped.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1882106

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
@@ -161,6 +161,7 @@ import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.runIfFragmentIsAttached
 import org.mozilla.fenix.ext.secure
 import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.ext.tabClosedUndoMessage
 import org.mozilla.fenix.home.HomeScreenViewModel
 import org.mozilla.fenix.home.SharedViewModel
 import org.mozilla.fenix.library.bookmarks.BookmarksSharedViewModel
@@ -399,23 +400,7 @@ abstract class BaseBrowserFragment :
             },
             onCloseTab = { closedSession ->
                 val closedTab = store.state.findTab(closedSession.id) ?: return@DefaultBrowserToolbarController
-
-                val snackbarMessage = if (closedTab.content.private) {
-                    requireContext().getString(R.string.snackbar_private_tab_closed)
-                } else {
-                    requireContext().getString(R.string.snackbar_tab_closed)
-                }
-
-                viewLifecycleOwner.lifecycleScope.allowUndo(
-                    binding.dynamicSnackbarContainer,
-                    snackbarMessage,
-                    requireContext().getString(R.string.snackbar_deleted_undo),
-                    {
-                        requireComponents.useCases.tabsUseCases.undo.invoke()
-                    },
-                    paddedForBottomToolbar = true,
-                    operation = { },
-                )
+                showUndoSnackbar(requireContext().tabClosedUndoMessage(closedTab.content.private))
             },
         )
         val browserToolbarMenuController = DefaultBrowserToolbarMenuController(
@@ -1015,6 +1000,19 @@ abstract class BaseBrowserFragment :
             view = view,
         )
         initializeEngineView(toolbarHeight)
+    }
+
+    protected fun showUndoSnackbar(message: String) {
+        viewLifecycleOwner.lifecycleScope.allowUndo(
+            binding.dynamicSnackbarContainer,
+            message,
+            requireContext().getString(R.string.snackbar_deleted_undo),
+            {
+                requireComponents.useCases.tabsUseCases.undo.invoke()
+            },
+            paddedForBottomToolbar = true,
+            operation = { },
+        )
     }
 
     /**
--- a/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
@@ -103,6 +103,7 @@ import org.mozilla.fenix.ext.nav
 import org.mozilla.fenix.ext.requireComponents
 import org.mozilla.fenix.ext.scaleToBottomOfView
 import org.mozilla.fenix.ext.settings
+import org.mozilla.fenix.ext.tabClosedUndoMessage
 import org.mozilla.fenix.home.pocket.DefaultPocketStoriesController
 import org.mozilla.fenix.home.pocket.PocketRecommendedStoriesCategory
 import org.mozilla.fenix.home.privatebrowsing.controller.DefaultPrivateBrowsingController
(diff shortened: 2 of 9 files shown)


fenix/app/src/main/java/org/mozilla/fenix/home/HomeScreenViewModel.kt
```diff
@@ -11,4 +11,10 @@ class HomeScreenViewModel : ViewModel() {
      * Used to delete a specific session once the home screen is resumed
      */
     var sessionToDelete: String? = null
+
+    /**
+     * Used to track if the last remaining tab closed was private. Only used by TabStrip to show
+     * undo snackbar on Home when navigating back from BrowserFragment to Home.
+     */
+    var wasLastTabsClosedPrivate: Boolean? = null
```
gabrielluong, line 19: Let's not add to this if possible. We should probably prefer our Stores.

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -977,6 +962,19 @@ abstract class BaseBrowserFragment :
...
+            binding.dynamicSnackbarContainer,
+            message,
+            requireContext().getString(R.string.snackbar_deleted_undo),
+            {
+                requireComponents.useCases.tabsUseCases.undo.invoke()
+            },
+            paddedForBottomToolbar = true,
+            operation = { },
+        )
+    }
```
MozillaNoah, line 976: I'd be curious having an optional follow-up (for anyone) to investigate getting `viewLifecycleOwner.lifecycleScope.allowUndo` to work in Compose previews or creating a Compose analog for us to use with future from-scratch features.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeScreenViewModel.kt
MozillaNoah: What led to needing this workaround? We're able to show the undo snackbar when closing the last remaining private tab via the Tabs Tray.

see: the usage of `dismissTabsTrayAndNavigateHome(...)` within [`TabsTrayController.deleteTab()`](
https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayController.kt#301-325)

MozillaNoah (commented):
changes to the viewmodel aside, this patch is looking good 👍 

Left a comment in the thread GL started to see how we can implement the undo without the viewmodel workaround.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeScreenViewModel.kt
rahulsainani: So this was needed as we navigate from `BrowserFragment` to `HomeFragment` when the last tab is closed, just passing the callback to show snackbar in BrowserFragment doesn't work for this last tab case. Tabs tray implementation also does something similar using `homeScreenViewModel.sessionToDelete` for the last tab case. In the link, `dismissTabsTrayAndNavigateHome` calls `navigateToHomeAndDeleteSession` which is a function  defined in `TabsTrayFragment`. [source](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/tabstray/TabsTrayFragment.kt#707). This sets `sessionToDelete` that's checked in `HomeFragment.onViewCreated` to show the snackbar and remove the tab.

❓ Or am i misunderstanding this?

💡 Could also simply do something similar but then the removal of tab for this case (last tab) will be handled differently than the rest of tabs, as the `sessionToDelete` logic in `HomeFragment` finds and removes the tab and then goes on to show the snackbar. 

🏗️ Ideally, I think we need something like a `SharedFlow` here for displaying snackbars which is good for nav events that can be set similar to set on dispatched actions. Simply observing state using the current abstractions will collect every time i.e show snackbar every time the screen collects the state. 

👂 I'm all ears for any ideas here.

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
rahulsainani: This would be nice and useful especially as we migrate the parents (Home, Fragment) to compose, maybe already in Tabs Tray? showing/hiding the snackbar would be a result of updating the state

fenix/app/src/main/java/org/mozilla/fenix/home/HomeScreenViewModel.kt
rahulsainani: I've update to use the AppStore. Note that this cannot be observed using the existing abstractions for the same reasons as above so we indeed need to do a check and then set it to null, similar to what was happening with the VM. LMKWYT!

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -636,6 +637,12 @@ class HomeFragment : Fragment() {
 
         homeViewModel.sessionToDelete = null
 
+        requireComponents.appStore.state.wasLastTabClosedPrivate?.also {
+            showUndoSnackbar(requireContext().tabClosedUndoMessage(it))
+        }
+
+        requireComponents.appStore.dispatch(AppAction.TabStripAction.UpdateLastTabClosedAction(null))
+
```
MozillaNoah, line 645: should the dispatch to reset the boolean is null only happen when the boolean is non-null?

```suggestion
        requireComponents.appStore.state.wasLastTabClosedPrivate?.also {
            showUndoSnackbar(requireContext().tabClosedUndoMessage(it))
            requireComponents.appStore.dispatch(AppAction.TabStripAction.UpdateLastTabClosedAction(null))
        }
```

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
```diff
@@ -636,6 +637,12 @@ class HomeFragment : Fragment() {
 
         homeViewModel.sessionToDelete = null
 
+        requireComponents.appStore.state.wasLastTabClosedPrivate?.also {
```
MozillaNoah, line 640: I'm not sure I 100% understood your last comment. Was it not possible to do an observation like this [other example in HomeFragment](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt#719)?

MozillaNoah (approved):
The observability is not a blocker for me, but it'd be my preference, if possible.

Don't forget to update the reference to ViewModel in the top comment.

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
rahulsainani: Good point, updated!

fenix/app/src/main/java/org/mozilla/fenix/home/HomeFragment.kt
MozillaNoah: ahhhh, right, the rest. Gotcha gotcha. In that case, I'm fine to proceed as-is

PR #5794, opened 2024-02-27
Bug 1882123 - Message middleware use store dispatch due to possibility of dispatching in IO thread

rocketsroger:
This is to fix an [crash](https://crash-stats.mozilla.org/report/index/94920999-291e-42f7-bd93-fa0130240227#tab-details) from https://github.com/mozilla-mobile/firefox-android/pull/5699.

Can't reproduce the issue but the crash stack shows that we were using `context.dispatch` in IO thread.
`context.dispatch` should only be used when we're on the same thread as the store and we are dispatching multiple actions and want to wait till all observers are notified before dispatching the next action.

Since the MessageMiddleware could uses IO thread in many areas, and it doesn't need to wait, the safe fix for this is to update all `context.dispatch` to `context.store.dispatch` which is safe on IO thread.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1882123


fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
```diff
@@ -69,15 +70,20 @@ class MessagingMiddleware(
         oldMessage: Message,
         context: AppStoreMiddlewareContext,
     ) {
+        val newMessage = runBlocking {
```
MozillaNoah, line 73: how come the addition of runBlocking? is it safe to thread-block here?

Would this be an ok alternative?

```
val newMessage
val newMessages
coroutineScope.launch {
            newMessage = controller.onMessageDisplayed(oldMessage)
            newMessages = if (!newMessage.isExpired) {
                updateMessage(context, oldMessage, newMessage)
            } else {
                removeMessage(context, oldMessage)
            }
        }

if (newMessage.isExpired) {
   consumeMessageToShowIfNeeded(context, oldMessage)
}

context.store.dispatch(UpdateMessages(newMessages))
```

rocketsroger: Unfortunately no, `newMessage.isExpired` could happen before `newMessage` is set.  This was the original code before Jame's change.   

```
        val newMessage = controller.updateMessageAsDisplayed(oldMessage)
        val newMessages = if (!newMessage.isExpired) {
            updateMessage(context, oldMessage, newMessage)
        } else {
            consumeMessageToShowIfNeeded(context, oldMessage)
            removeMessage(context, oldMessage)
        }
        context.dispatch(UpdateMessages(newMessages))
        coroutineScope.launch {
            controller.onMessageDisplayed(newMessage) 
```
I'm just trying to restore the way it was called.

MozillaNoah: Gotcha gotcha. I'm just having a hard time discerning whether this section can be a simple revert without knowing how the removal of `updateMessageAsDisplayed` can be 1 to 1 replaced with `onMessageDisplayed`

rocketsroger: Hmm you're right.  Maybe I should just put back what we had before the change.

rocketsroger: I'll look for a better solution.  This right now is risky.  Thanks @MozillaNoah

rocketsroger: The alternative solution that is safer is to only dispatch to the store using main thread.  [This](https://searchfox.org/mozilla-mobile/rev/21548505501d6cee5ce2aac3244821269bc7969d/firefox-android/android-components/components/lib/state/src/main/java/mozilla/components/lib/state/Middleware.kt#42) stated that we should only dispatch from the calling thread.  In this case it's the main thread.

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
```diff
@@ -37,7 +38,9 @@ class MessagingMiddleware(
             is Restore -> {
                 coroutineScope.launch {
                     val messages = controller.getMessages()
-                    context.store.dispatch(UpdateMessages(messages))
+                    MainScope().launch {
```
t-p-white, line 41: 🚗 drive-by:

🤔 I'm not sure what the fix is, but I don't think we should be doing this - it looks like a code smell to me. My interpretation of:

> Note that this method should only ever be called from a [Middleware] and the calling thread.
doesn't mean it has to be the main thread

t-p-white: Happy to Rubber Duck or pair if it helps 🙂

Amejia481: @rocketsroger could you clarify why w need the `MainScope`? 

Checking at stack trace,[ it looks like crash is coming form ](https://crash-stats.mozilla.org/report/index/d881258e-ca72-4f99-a2e7-7773a0240226#tab-details) `org.mozilla.fenix.messaging.state.MessagingMiddleware$onMessagedDisplayed$1`

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
```diff
@@ -77,7 +80,12 @@ class MessagingMiddleware(
                 consumeMessageToShowIfNeeded(context, oldMessage)
                 removeMessage(context, oldMessage)
             }
-            context.store.dispatch(UpdateMessages(newMessages))
+
+            MainScope().launch {
+                context.store.dispatch(UpdateMessages(newMessages))
+            }
```
Amejia481, line 86: Could you clarify why do we need to switch thread?

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
Amejia481: Happy to also pair on the solution :)

Amejia481: The crash is coming from our own [assertion](https://github.com/mozilla-mobile/firefox-android/blob/86bd17c82b227f3a7e538e7a469d82a57541082e/android-components/components/lib/state/src/main/java/mozilla/components/lib/state/internal/StoreThreadFactory.kt#L53-L56) 
``` 
`Expected `store` thread, but running on thread `DefaultDispatcher-worker-7`. Leaked MiddlewareContext or did you mean to use `MiddlewareContext.store.dispatch`?
```

 Indicating that we should be using the store thread.

rocketsroger: > Indicating that we should be using the store thread.

That's what I saw too.  The Store's thread is Main so that's why I try to dispatch everything on main.  But it does feel like a hack.  I don't know if we have a better solution unless we revert Jame's changes.

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
```diff
@@ -37,7 +38,9 @@ class MessagingMiddleware(
             is Restore -> {
                 coroutineScope.launch {
                     val messages = controller.getMessages()
-                    context.store.dispatch(UpdateMessages(messages))
+                    MainScope().launch {
+                        context.store.dispatch(UpdateMessages(messages))
```
Amejia481, line 42: Looking a the previous implementation we where just calling `context.dispatch(UpdateMessages(messages))
` instead of `context.store.dispatch`
https://github.com/mozilla-mobile/firefox-android/pull/5699/files#diff-05e0a602e09d4173a4b88f3854cedbb3d7cdf322902761aa827a337e88ccdceaL82

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
rocketsroger: > doesn't mean it has to be called from the main thread?

The store was created on main thread so the dispatch have to be called on the main thread.  Maybe I can let the caller pass in that.

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
rocketsroger: No, I tried that and it consistently crashes the app on startup.

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
rocketsroger: Store was created on the main thread so the dispatch [Middleware.kt](https://searchfox.org/mozilla-mobile/rev/21548505501d6cee5ce2aac3244821269bc7969d/firefox-android/android-components/components/lib/state/src/main/java/mozilla/components/lib/state/Middleware.kt#42) to be on main thread too.

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
Amejia481: Alternative we could also do:

```kotlin

                coroutineScope.launch {
                    val messages = controller.getMessages()
                    withContext(Dispatchers.Main) {
                        context.store.dispatch(UpdateMessages(messages))
                    }
                } 
```

fenix/app/src/main/java/org/mozilla/fenix/messaging/state/MessagingMiddleware.kt
Amejia481: 👍🏽

Amejia481: Thanks for the clarification!

rocketsroger: Actually, I think what I need to do is the opposite.  I think I have a much safer fix now.  I'll update.

Amejia481 (approved):

PR #5368, opened 2024-01-30
Bug 1813875 - Prevent the inherited theming implementation being applied to Custom Tabs views

t-p-white:
Based on my understanding that `ExternalAppBrowserActivity` should handle its own theming via the web manifest, I _think_ the issue that we are seeing (with the multiple bugs related to the status bar icons not being visible) is due to additional theming provided by the `HomeActivity` parent class (potentially another symptom of the coupled nature between Home and External App Browser). 

For now, I have gated these checks to make it explicit this is not required in External App browser usage. Ongoing refactoring of the `HomeActivity` will remove these, along with the other `ExternalAppBrowserActivity` gates.

I think there is a **separate issue** being [reported](https://bugzilla.mozilla.org/show_bug.cgi?id=1814646) in that Custom Tabs do not sync with the browser theme (light/dark) (see the Twitter app in the demo) - however, this may be by design - more investigation needed.

@MatthewTighe I believe this will have some crossover with https://bugzilla.mozilla.org/show_bug.cgi?id=1861459 - hopefully it should even be beneficial in the refactoring.

Original (Waffle & Twitter status icons not visible):

https://github.com/mozilla-mobile/firefox-android/assets/1833156/76faff1f-1ed7-43cc-90f3-65572f021419

Bug fix:

https://github.com/mozilla-mobile/firefox-android/assets/1833156/37779813-75e8-4121-8244-30a7749b355f

### Pull Request checklist

- [ ] **Quality**: This PR builds and passes detekt/ktlint checks (A pre-push hook is recommended)
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes [a changelog entry](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/changelog.md) or does not need one
- [ ] **Accessibility**: The code in this PR follows [accessibility best practices](https://github.com/mozilla-mobile/firefox-android/blob/main/docs/shared/android/accessibility_guide.md) or does not include any user facing features

### After merge
- **Breaking Changes**: If this is a breaking Android Components change, please push a draft PR on [Reference Browser](https://github.com/mozilla-mobile/reference-browser) to address the breaking issues.

### To download an APK when reviewing a PR (after all CI tasks finished running):
1. Click on `Checks` at the top of the PR page.
2. Click on the `firefoxci-taskcluster` group on the left to expand all tasks.
3. Click on the `build-apk-{fenix,focus,klar}-debug` task you're interested in.
[...]

PR #5807, opened 2024-02-28
Bug 1882308 - Updated custom tab to support private mode

t-p-white:
Fix non-private mode:

---

Fix private mode:

🍦 for those curious, the ice cream in the article is vanilla and chocolate chips. Arguably not really a 'flavour', it's vanilla ice cream with a topping...

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1882308

### DIFF
--- a/android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
+++ b/android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
@@ -11,6 +11,7 @@ import android.view.Window
 import androidx.annotation.ColorInt
 import androidx.annotation.VisibleForTesting
 import androidx.appcompat.content.res.AppCompatResources.getDrawable
+import androidx.core.content.ContextCompat.getColor
 import androidx.core.graphics.drawable.toDrawable
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.cancel
@@ -32,6 +33,7 @@ import mozilla.components.feature.tabs.CustomTabsUseCases
 import mozilla.components.lib.state.ext.flowScoped
 import mozilla.components.support.base.feature.LifecycleAwareFeature
 import mozilla.components.support.base.feature.UserInteractionHandler
+import mozilla.components.support.ktx.android.content.res.resolveAttribute
 import mozilla.components.support.ktx.android.content.share
 import mozilla.components.support.ktx.android.util.dpToPx
 import mozilla.components.support.ktx.android.view.setNavigationBarTheme
@@ -114,15 +116,24 @@ class CustomTabsToolbarFeature(
     internal fun init(config: CustomTabConfig) {
         // Don't allow clickable toolbar so a custom tab can't switch to edit mode.
         toolbar.display.onUrlClicked = { false }
-        val readableColor =
-            config.toolbarColor?.let { getReadableTextColor(it) } ?: toolbar.display.colors.menu
 
-        // Change the toolbar colour
-        updateToolbarColor(
-            config.toolbarColor,
-            config.navigationBarColor ?: config.toolbarColor,
-            readableColor,
-        )
+        val readableColor = if (updateToolbarBackground) {
+            config.toolbarColor?.let { getReadableTextColor(it) } ?: toolbar.display.colors.menu
+        } else {
+            // It's private mode, the readable color needs match the app.
+            // Note: The main app is configuring the private theme, Custom Tabs is adding the
+            // additional theming for the dynamic UI elements e.g. action & share buttons.
+            val colorResId = context.theme.resolveAttribute(android.R.attr.textColorPrimary)
+            getColor(context, colorResId)
+        }
+
+        if (updateToolbarBackground) {
+            updateToolbarColor(
+                config.toolbarColor,
+                config.navigationBarColor ?: config.toolbarColor,
+                readableColor,
+            )
+        }
 
         // Add navigation close action
         if (config.showCloseButton) {
@@ -149,8 +160,8 @@ class CustomTabsToolbarFeature(
         @ColorInt navigationBarColor: Int?,
         @ColorInt readableColor: Int,
     ) {
-        if (updateToolbarBackground && toolbarColor != null) {
-            toolbar.setBackgroundColor(toolbarColor)
+        toolbarColor?.let {
+            toolbar.setBackgroundColor(it)
 
             toolbar.display.colors = toolbar.display.colors.copy(
                 text = readableColor,
@@ -161,7 +172,7 @@ class CustomTabsToolbarFeature(
                 menu = readableColor,
             )
 
-            window?.setStatusBarTheme(toolbarColor)
+            window?.setStatusBarTheme(it)
         }
         navigationBarColor?.let { color ->
             window?.setNavigationBarTheme(color)
--- a/android-components/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeatureTest.kt
+++ b/android-components/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeatureTest.kt
@@ -13,6 +13,7 @@ import android.view.Window
 import android.widget.FrameLayout
 import android.widget.ImageButton
 import android.widget.TextView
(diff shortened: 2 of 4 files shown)


t-p-white:
@MatthewTighe [your original concern](https://github.com/mozilla-mobile/firefox-android/pull/5368/files#r1471657284) was indeed correct, apologies I misunderstood how the code was handling private mode. This update fixes it

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -114,15 +115,22 @@ class CustomTabsToolbarFeature(
...
+            Color.WHITE
+        }
 
-        // Change the toolbar colour
-        updateToolbarColor(
-            config.toolbarColor,
-            config.navigationBarColor ?: config.toolbarColor,
-            readableColor,
-        )
+        // Maybe update the toolbar color
```
MatthewTighe, line 126: I don't think we need this comment

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -149,8 +157,8 @@ class CustomTabsToolbarFeature(
         @ColorInt navigationBarColor: Int?,
         @ColorInt readableColor: Int,
     ) {
-        if (updateToolbarBackground && toolbarColor != null) {
```
MatthewTighe, line 152: I'm a little skeptical of removing `updateToolbarBackground` because it assumes this function is only called when it's supposed to be

MatthewTighe: I believe this change would also prevent the `navigationBarColor` update below from being called in cases where it previously wouldn't be prevented

fenix/app/src/main/java/org/mozilla/fenix/HomeActivity.kt
```diff
@@ -878,8 +878,8 @@ open class HomeActivity : LocaleAwareAppCompatActivity(), NavHostActivity {
 
     private fun setupTheme() {
         themeManager = createThemeManager()
-        // ExternalAppBrowserActivity handles it's own theming as it can be customized.
-        if (this !is ExternalAppBrowserActivity) {
+        // ExternalAppBrowserActivity exclusively handles it's own theming unless in private mode.
+        if (this !is ExternalAppBrowserActivity || browsingModeManager.mode.isPrivate) {
```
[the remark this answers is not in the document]
MatthewTighe: Hmmm, we might run into a race condition here with the updates in https://github.com/mozilla-mobile/firefox-android/pull/5502 because the mode gets set asynchronously. That's my problem to solve later though 😂

MatthewTighe (approved):
Couple small comments but nothing blocking. LGTM!

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -114,15 +115,22 @@ class CustomTabsToolbarFeature(
...
         // Don't allow clickable toolbar so a custom tab can't switch to edit mode.
         toolbar.display.onUrlClicked = { false }
-        val readableColor =
+
+        val readableColor = if (updateToolbarBackground) {
             config.toolbarColor?.let { getReadableTextColor(it) } ?: toolbar.display.colors.menu
+        } else {
+            // It's private mode, the foreground color needs to be white.
+            Color.WHITE
+        }
```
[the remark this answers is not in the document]
MozillaNoah: If we don't, we should at least create a top-level value for this, since I think we use `Color.WHITE` as the fallback in a few places in Custom Tabs

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
t-p-white: > I believe this change would also prevent the navigationBarColor update below from being called in cases where it previously wouldn't be prevented

Good spot, I should have mentioned this in my original comment. We should only be updating the navigation bar theme if not in private mode, which originally contributed to the theming bug which is why this was changed

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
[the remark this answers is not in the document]
t-p-white: Thanks both for the feedback, have updated to use a colour resource 👍

android-components/components/feature/customtabs/src/test/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeatureTest.kt
```diff
@@ -1094,6 +1095,43 @@ class CustomTabsToolbarFeatureTest {
         assertEquals(Color.WHITE, toolbar.display.colors.menu)
     }
 
+    @Test
```
t-p-white, line 1098: based on the similar tests above

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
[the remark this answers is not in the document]
MozillaNoah: I think I was thinking about something from `ColorUtils`. We're all good here 👍

android-components/components/feature/customtabs/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at http://mozilla.org/MPL/2.0/. -->
+<resources>
+    <color name="fx_mobile_private_text_color_primary">@color/photonLightGrey05</color>
```
MozillaNoah, line 6: this is a bit of a flag for me. We shouldn't be using firefox colors inside of AC. See the next comment for a suggested approach

android-components/components/feature/customtabs/src/main/java/mozilla/components/feature/customtabs/CustomTabsToolbarFeature.kt
```diff
@@ -114,15 +115,23 @@ class CustomTabsToolbarFeature(
...
         toolbar.display.onUrlClicked = { false }
-        val readableColor =
+
+        val readableColor = if (updateToolbarBackground) {
             config.toolbarColor?.let { getReadableTextColor(it) } ?: toolbar.display.colors.menu
+        } else {
+            // It's private mode, the readable color needs match the app.
+            // Note: The main app is configuring the private theme, Custom Tabs is adding the
+            // additional theming for the dynamic UI elements e.g. action & share buttons.
+            getColor(context, R.color.fx_mobile_private_text_color_primary)
```
MozillaNoah, line 125: so rather than pasting over `fx_mobile_private_text_color_primary`, let's instead use an attribute already available to us! 

`textColorPrimary` is an Android XML style attribute that we then override inside of our [styles file in Fenix](https://searchfox.org/mozilla-mobile/source/firefox-android/fenix/app/src/main/res/values/styles.xml#221), and this is a pattern used elsewhere in AC.

[Example of usage](https://searchfox.org/mozilla-mobile/source/firefox-android/android-components/components/feature/addons/src/main/java/mozilla/components/feature/addons/ui/Extensions.kt#135)

t-p-white: Awesome, great suggestion, thanks!

MozillaNoah (approved):
:shipit:

t-p-white:
@Mergifyio backport releases_v124

PR #5810, opened 2024-02-28
Bug 1882403 - Refactor Browser Store Initialization for Translations

ohall-m:
This bugs shifts around the initialization for the global store in translations.

Revised Flow:

* `InitAction` -> `InitTranslationsBrowserState`
* `InitTranslationsBrowserState`:
  * Checks the engine support, if everything is in order, then finishes initialization on `initializeBrowserStore`.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1882403


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -46,13 +46,16 @@ class TranslationsMiddleware(
...
             is InitAction ->
-                scope.launch {
-                    requestEngineSupportAndInit(context)
-                }
+                context.store.dispatch(TranslationsAction.InitTranslationsBrowserState)
 
             is TranslationsAction.InitTranslationsBrowserState -> {
                 scope.launch {
-                    requestSupportedLanguages(context)
+                    val engineIsSupported = requestEngineSupport(context = context)
```
MatthewTighe, line 53: nit: I don't think we need the named parameter `context = context` here. Usually that is reserved for when we need to have multi-line argument lists

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -46,13 +46,16 @@ class TranslationsMiddleware(
...
+                    // Only continue initializing if true.
+                    // If null or false, then take no action and don't continue with initialization.
```
MatthewTighe, line 55: nit: I think we could skip this comment

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -146,40 +149,54 @@ class TranslationsMiddleware(
...
+     * Use this to initialize translations data for [BrowserState.translationEngine]. If an
+     * issue occurs, the relevant error will be set on [BrowserState.translationEngine].
+     *
+     * This will populate:
+     * Language Support - [requestSupportedLanguages]
      *
      * @param context Context to use to dispatch to the store.
      */
-    private fun requestEngineSupportAndInit(
+    private fun initializeBrowserStore(
```
MatthewTighe, line 160: nit: I think we could probably just move the list of function invocations that will happen in this function inline, unless we might call this function from more than one place.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -146,40 +149,54 @@ class TranslationsMiddleware(
...
+     * [BrowserState.translationEngine].
+     *
+     * @param context Context to use to dispatch to the store.
+     * @return Whether the engine is supported or not, or null when the support cannot be
+     * determined.
+     */
+    private suspend fun requestEngineSupport(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+    ): Boolean? {
+        return suspendCoroutine { continuation ->
```
MatthewTighe, line 177: Is `suspendCoroutine` needed now?

MatthewTighe (approved):
Some small comments, but nothing that needs to block 🙂

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Updated, ty!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Updated, ty!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: I think it is possible we might need to call it for error handling. (We'd probably just recall `InitTranslationsBrowserState`, though as another possibility.) I also have a slight preference to keep it for clarity / self-documenting code too because we are also going to also have a parallel list for `initializeSessionStore`. (Trying to keep the concepts separate and clear as best I can.)

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Wasn't sure the best or recommended way to structure this. Thinking was that I needed to wait for the value to resolve for `InitTranslationsBrowserState` to see if I can continue once the engine reports back. LMKWYT!

MatthewTighe: That makes sense, just had to wrap my head around what was going on again. Thanks!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -146,40 +147,54 @@ class TranslationsMiddleware(
...
-            },
-        )
+                onError = { error ->
+                    context.store.dispatch(
+                        TranslationsAction.EngineExceptionAction(
+                            error = TranslationError.UnknownEngineSupportError(error),
+                        ),
+                    )
+                    logger.error("Error requesting engine support: ", error)
+                    continuation.resume(null)
```
MatthewTighe, line 194: I see this is landing already and I think we could handle any updates to this in a follow-up, but wondering if this could just be `false` instead of `null`? Is there a functional difference between the false case above and the exception case here?

ohall-m: Good thing to mention and thanks for taking a look! It is following a similar pattern as these [functions](https://searchfox.org/mozilla-mobile/rev/484082df2b810fe8ea7ec2167e63e063f907b0a7/firefox-android/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt#391,411) for parallelism. I think keeping the `true = yes, support` / `false = no support` / `null = unknown support` state distinction is helpful since we might want to retry on the unknown or null case. (Still thinking through how retry logic is going to work all around, but the distinction will show back up in some form.)

ohall-m: (Happened to notice in those linked functions that they are missing logging - added that to my list of things to refactor when in the area!)

PR #5814, opened 2024-02-29
Bug 1881571 - Add tooling-lint component to app projects

jonalmeida:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1881571


android-components/components/tooling/lint/src/main/java/mozilla/components/tooling/lint/ImageViewAndroidTintXmlDetector.kt
```diff
@@ -36,7 +35,6 @@ class ImageViewAndroidTintXmlDetector : ResourceXmlDetector() {
         const val ERROR_MESSAGE =
             "Using android:tint to tint ImageView instead of app:tint with AppCompatImageView"
 
-        @VisibleForTesting
```
jonalmeida, line 39: When doing a checksum diff between Fenix and the AC component, this was a difference and didn't need to be there. Removing it to ensure that when BZ-1872229 is worked on, its a tiny bit easier for the next person to de-duplicate and save them the trouble.

jonalmeida:
cc: @jackyzy823 - this should ensure the new linter is used in the app projects as well.

Amejia481 (approved):

jonalmeida:
> Will the lint-baseline.xml stops working if line number changes? 

This baseline file is for Android Lint, and we use a baseline in AC and Fenix today without issues. The concerns you are referring to were with ktlint baseline which isn't smart enough for some reason to not cause a problem.

I've also generated a baseline for Focus instead of fixing any of these issues because we don't actively maintain this project unless needed.

@rocketsroger do you think we should fix these issues in a fast follow-up regardless?

cc: @mcarare who might have experience here as well.

rocketsroger:
> @rocketsroger do you think we should fix these issues in a fast follow-up regardless?
> 
> cc: @mcarare who might have experience here as well.'

I think we should add the fix add the same time with this change so we don't need baseline.

mcarare:
Although the line numbers are mentioned in the baseline, for lint it should not matter if they are changed in the future.
I do think we should fix the issues, at least the errors added to the baseline.


I do see a lot of issues related to strings, where we have a string in one language but no value in the default string.xml. Is this something L10n team should look into?

rocketsroger:
> Although the line numbers are mentioned in the baseline, for lint it should not matter if they are changed in the future. I do think we should fix the issues, at least the errors added to the baseline.

I'm OK with merging this then.  Thanks

PR #5815, opened 2024-02-29
Bug 1882671 - Update buildconfig.yml helper doc

jonalmeida:
It seems that the artifact is now just a plain diff file that can be applied without needing to unzip it.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1882671


taskcluster/scripts/lint/is_buildconfig_yml_up_to_date.py
```diff
@@ -36,7 +36,7 @@ def _execute_taskcluster_steps(diff, task_id):
     artifact_url = f"{tc_root_url}/api/queue/v1/task/{task_id}/artifacts/public%2Fgit%2F{BUILDCONFIG_DIFF_FILE_NAME}"  # noqa E501
     message = f"""{BUILDCONFIG_FILE_NAME} file changed! Please update it by running:
 
-curl --location {artifact_url} | gunzip | git apply
+curl --location {artifact_url} | git apply
```
glandium, line 39: Use --compressed instead, it will be resilient to the url being compressed or not.

jonalmeida: TIL about `--compressed`!

In the curl man pages, it says: "This is a request, not an order; the server may or may not deliver data compressed."
So this works nicely if we request compressed, and we still get the uncompressed, then git apply will still succeed. 

It may fail.. "If this option is used and the server sends an unsupported encoding, curl reports an error." but gunzip was able to support some encoding, which seems like a small edge-case in practice, so I think this a pretty decent suggestion to add.

jonalmeida: Updated, thanks!

bhearsum: It looks like you didn't update the `curl` command with the new argument?

jonalmeida: Oh weird, sorry, give me a minute to fix that.

jonalmeida: @bhearsum fixed. Sorry, about that!

bhearsum (approved):

PR #5826, opened 2024-02-29
Bug 1882542 - Eagerly Calculate an Initial Translation Download Size

ohall-m:
This bug adds:
* When `TranslateExpectedAction` occurs, then calculate the expected translations model download size.

More background: `TranslateExpectedAction` indicates we have a default `from` (page language) and a default `to` (user preferred) language to initialize the translations dialog with expected values.

This queries how large of a download the translation pair will be eagerly, so we have an initial value for the default recommendation.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1882542

### DIFF
--- a/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
+++ b/android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
@@ -21,6 +21,7 @@ import mozilla.components.concept.engine.translate.TranslationError
 import mozilla.components.concept.engine.translate.TranslationOperation
 import mozilla.components.concept.engine.translate.TranslationPageSettingOperation
 import mozilla.components.concept.engine.translate.TranslationPageSettings
+import mozilla.components.concept.engine.translate.findLanguage
 import mozilla.components.lib.state.Middleware
 import mozilla.components.lib.state.MiddlewareContext
 import mozilla.components.support.base.log.logger.Logger
@@ -55,6 +56,10 @@ class TranslationsMiddleware(
                         initializeBrowserStore(context)
                     }
                 }
+            }
+
+            is TranslationsAction.TranslateExpectedAction -> {
+                requestDefaultModelDownloadSize(context, action.tabId)
             }
 
             is TranslationsAction.OperationRequestedAction -> {
@@ -541,6 +546,34 @@ class TranslationsMiddleware(
     }
 
     /**
+     * Fetches the expected translation model download size assuming the user intends to complete
+     * a translation using the detected default `from` (page language) and `to` (user preferred)
+     * languages.
+     *
+     * If the detected default languages are available, then this will fetch and set the
+     * corresponding model download size on [SessionState.translationsState].
+     *
+     * If no defaults are available, then no action will occur.
+     *
+     * @param context Context to use to dispatch to the store.
+     * @param tabId Tab ID associated with the request.
+     */
+    private fun requestDefaultModelDownloadSize(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        tabId: String,
+    ) {
+        val fromLanguage = getDefaultFromLanguage(context, tabId) ?: return
+        val toLanguage = getDefaultToLanguage(context, tabId) ?: return
+        context.store.dispatch(
+            TranslationsAction.FetchTranslationDownloadSizeAction(
+                tabId = tabId,
+                fromLanguage = fromLanguage,
+                toLanguage = toLanguage,
+            ),
+        )
+    }
+
+    /**
      * Updates the always offer popup setting with the [Engine].
      *
      * @param setting The value of the always offer setting to update.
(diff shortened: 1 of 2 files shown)


ohall-m:
Also, manually tested against Fenix watching the debugger.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -466,6 +471,47 @@ class TranslationsMiddleware(
         )
     }
 
+    /**
+     * Fetches the expected translation model download size assuming the user intends to complete
+     * a translation using detected defaults.
+     *
+     * If detected defaults are available, then this will sets the values on
+     * [SessionState.translationsState] on the [BrowserStore]
+     * via [TranslationsAction.FetchTranslationDownloadSizeAction].
```
gabrielluong, line 480: I am feeling that this piece of the comment is a bit forced? (possibility because it is copying similar comments from above) It's kinda repeating what the code is doing which from my experience usually means the comment is probably not adding much. The comment is also a bit vague about what values it is setting, which makes the comment itself feels somewhat unsatisfying in terms of being clear what it is doing. In reality, `FetchTranslationDownloadSizeAction` will attempt to fetch the download size and try and populate `translationDownloadSize`.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -466,6 +471,47 @@ class TranslationsMiddleware(
...
+    ) {
+        val detectedLanguages = context.store.state.findTab(tabId)
+            ?.translationsState?.translationEngineState?.detectedLanguages
+        val supportedLanguages = context.store.state.translationEngine.supportedLanguages
+        val fromLanguage = detectedLanguages?.documentLangTag?.let { fromTag ->
+            supportedLanguages?.findLanguage(fromTag)
+        }
+        val toLanguage = detectedLanguages?.userPreferredLangTag?.let { toTag ->
+            supportedLanguages?.findLanguage(toTag)
+        }
```
gabrielluong, line 500: I remember in another PR that you have added helper functions that feels similar to what I am seeing here. It is probably not the same as these, but I have a hunch that we might have to do this dance between getting the `detectedLanguages` and `supportedLanguages` in order to get the from/to language in more than one place. Would this be a worthy candidate for a helper function? To be specific, this helper function would get the from/to language from the detected languages.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -466,6 +471,47 @@ class TranslationsMiddleware(
...
+    private fun requestDefaultModelDownloadSize(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        tabId: String,
+    ) {
+        val detectedLanguages = context.store.state.findTab(tabId)
+            ?.translationsState?.translationEngineState?.detectedLanguages
+        val supportedLanguages = context.store.state.translationEngine.supportedLanguages
+        val fromLanguage = detectedLanguages?.documentLangTag?.let { fromTag ->
+            supportedLanguages?.findLanguage(fromTag)
+        }
```
gabrielluong, line 497:
```suggestion
        val fromLanguage = detectedLanguages?.documentLangTag?.let { fromTag ->
            supportedLanguages?.findLanguage(fromTag)
        } ?: return
```
If we hit null at any point, we might as well just escape instead of continuing to do more work.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -466,6 +471,47 @@ class TranslationsMiddleware(
...
+     * @param context Context to use to dispatch to the store.
+     * @param tabId Tab ID associated with the request.
+     *
+     */
+    private fun requestDefaultModelDownloadSize(
+        context: MiddlewareContext<BrowserState, BrowserAction>,
+        tabId: String,
+    ) {
+        val detectedLanguages = context.store.state.findTab(tabId)
+            ?.translationsState?.translationEngineState?.detectedLanguages
```
gabrielluong, line 493:
```suggestion
        val detectedLanguages = context.store.state.findTab(tabId)
            ?.translationsState?.translationEngineState?.detectedLanguages ?: return
```

gabrielluong (approved):

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -466,6 +471,47 @@ class TranslationsMiddleware(
         )
     }
 
+    /**
+     * Fetches the expected translation model download size assuming the user intends to complete
+     * a translation using detected defaults.
+     *
+     * If detected defaults are available, then this will sets the values on
```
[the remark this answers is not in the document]
ohall-m: Updated!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Yeah, agreed! Rephrased it

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -466,6 +471,47 @@ class TranslationsMiddleware(
...
+     * [SessionState.translationsState] on the [BrowserStore]
+     * via [TranslationsAction.FetchTranslationDownloadSizeAction].
+     *
+     * If no defaults are available, then no action will occur.
+     *
+     * @param context Context to use to dispatch to the store.
+     * @param tabId Tab ID associated with the request.
+     *
```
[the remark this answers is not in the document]
ohall-m: Updated, ty

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: Updated!

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
ohall-m: >Would this be a worthy candidate for a helper function? 

Sure, I think so too! Good chance it'll show up again somewhere. Not sure of a good place to globally put it, but I can put a private version of it on this file. One complication is part of the information is on the session store and part is in the browser store, for figuring out something more universal.

android-components/components/browser/state/src/main/java/mozilla/components/browser/state/engine/middleware/TranslationsMiddleware.kt
```diff
@@ -37,7 +38,7 @@ class TranslationsMiddleware(
 ) : Middleware<BrowserState, BrowserAction> {
     private val logger = Logger("TranslationsMiddleware")
 
-    @Suppress("LongMethod")
+    @Suppress("LongMethod","CyclomaticComplexMethod")
```
ohall-m, line 41: Something between updates triggered `CyclomaticComplexMethod` lint. Think it might be the `?:` operator [pushed it over the limit](https://detekt.dev/docs/rules/complexity/#cyclomaticcomplexmethod)? (Not much has changed besides that, unless a global rule changed on the rebase.)

gabrielluong: So weird that it could complain about the `invoke` method when the `?:` operators are all appearing in smaller methods.

ohall-m: Showed up in #5726 on rebase. Change there was just adjusting around how `Init` occurred. (Patch was written before the refactor to how init worked in translations middleware via #5810.) 🧐 Weird

PR #5845, opened 2024-03-01
Bug 1883127 - Clear `isOfferTranslate` Under Certain Conditions

ohall-m:
This bug clears `isOfferTranslate` similarly to `isExpectedTranslate` on a translations engine state change.

### Pull Request checklist

### After merge

### GitHub Automation

Used by GitHub Actions.


ohall-m:
Background:

* `isOfferTranslate` - popup and offer a translation dialog
* `isExpectedTranslate` - show translations icon on toolbar

Both should be cleared back to `false` on a similar lifecycle. (There might be a better way to clear, filed a code health bug as [1883142](https://bugzilla.mozilla.org/show_bug.cgi?id=1883142).)

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/action/TranslationsActionTest.kt
```diff
@@ -152,6 +152,45 @@ class TranslationsActionTest {
         assertFalse(tabState().translationsState.isExpectedTranslate)
     }
 
+    @Test
+    fun `WHEN a TranslateStateChangeAction is dispatched THEN the isOfferTranslate status updates accordingly`() {
```
gabrielluong, line 156: I am pretty sure this test is processing the same action using the same parameters as the previous test. I think it might be better to just add the `isOfferTranslate` state assertions to the previous test for `isExpectedTranslate` and just ensure we are checking that these states are updated accordingly instead of having 2 individuals tests that initializes the same state and runs the same actions with the same parameters, but only checks for individual state. 

The point here is to avoid repeating previous tests when we can just add extra assertions. Thus, reducing the maintenance burden of just having too many tests.

gabrielluong (approved):

android-components/components/browser/state/src/test/java/mozilla/components/browser/state/action/TranslationsActionTest.kt
ohall-m: Thanks for taking the look, the tests do have different corresponding actions in this case and aren't repetitive. (I noticed I had a comment that conflated something, which probably made them seem more similar than they were, so I updated that!)

They do have similar initialization states; however:
* `isOfferTranslate` requires a `TranslateOfferAction` - it occurs under a longer list of [conditions](https://searchfox.org/mozilla-central/rev/202c48686136360a23b73a49b611a19e64f3e1b8/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/TranslationsController.java#918-937)
* `isExpectedTranslate` requires a `TranslateExpectedAction` - generally occurs just on a supported page not in the user's preferred language

There is a situation like: "all offers to translate are expected translations, but not all expected translations are offers to translate".

gabrielluong: Gotcha, not a problem. I missed the difference between these 2 actions. All I saw were the same states and TranslateStateChangeAction(sp?).

ohall-m: `TranslateStateChangeAction` probably would be better as `TranslationsStateChangeAction` or `TranslationStateChangeAction`. (Was a very early addition before a lot of naming patterns were established, filed a quick [code health bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1883432) to look at it later.)

ohall-m: (Let me know if I misspelled `TranslateStateChangeAction` somewhere specifically too. I'm really bad at spotting spelling things and have trouble seeing them.)

PR #5846, opened 2024-03-01
Bug 1879377 - Hide nav bar when the toolbar becomes focused

mavduevskiy:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1879377


android-components/components/browser/state/src/main/java/mozilla/components/browser/state/reducer/AwesomeBarStateReducer.kt
```diff
@@ -25,6 +25,9 @@ internal object AwesomeBarStateReducer {
             is AwesomeBarAction.EngagementFinished -> {
                 state.copy(awesomeBarState = AwesomeBarState())
             }
+
+            AwesomeBarAction.EngagementStarted ->
+                state.copy(awesomeBarState = state.awesomeBarState.copy(isActive = true))
```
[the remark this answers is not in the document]
mavduevskiy: That's a good suggestion, I was also considering that, but after closer examination found out no way to achieve what we want based on the existing data. `VisibilityState` contains a list of providers and suggestions for the particular input, and until there is any input, it is empty. It makes the state where there is no search dialog fragment and where a search dialog fragment is present but has no suggestions indistinguishable. 

Another reason I would like to introduce a new parameter is transparency. `AwesomeBarState.visibilityState` was designed to hold the list of suggestions, not actually represent the visibility of the search bar itself. So, even if we could calculate it from `visibilityState`, any change to the behaviour of `visibilityState` would break the toolbar logic. But introducing a `AwesomebarState.isActive` in combination with `AwesomeBarAction.EngagementFinished` & `AwesomeBarAction.EngagementStarted` is very clear.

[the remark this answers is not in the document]
gabrielluong: Okay, I see there's no easy way to derive this with the existing properties. My initial thinking was that we don't usually expect to see a UI-specific state such as `isActive` in `BrowserState`. Hence, the attempt to try and derive this via extension functions instead. I think the best way to proceed here would be to add this state (`isAwesomeBarActive`) inside of our `AppState` (assuming this is needed globally) simply because we don't need this `isActive` within the AC `feature-` and `browser-` components therefore this state is specific to Fenix. @rocketsroger What do you think of this suggestion?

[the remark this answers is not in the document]
rocketsroger: @gabrielluong that sounds good to me.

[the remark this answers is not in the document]
mavduevskiy: thanks for the input, guys

I change the name to `isSearchDialogVisible`. The `AwesomeBar` isn't what we actually depend upon, it's the thing that is showing suggestions and it just so happens that it gets active with the `SearchDialogFragment`. I was considering other names like 'isToolbarFocused' but can't say which is the best fit. Open to suggestions if anybody feels strong about it, but for now decided to stick with the one that looked general and representative the most.

mavduevskiy:
@gabrielluong @rocketsroger 
thanks for the feedback, updated

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -209,6 +209,7 @@ abstract class BaseBrowserFragment :
         get() = _browserToolbarView!!
 
     internal lateinit var browserToolbar: BrowserToolbar
+    private var bottomToolbarContainerView: BottomToolbarContainerView? = null
```
gabrielluong, line 212: For this view, I think you may want to handle this similarly to `BrowserToolbarView` using the getters. I know from https://github.com/mozilla-mobile/fenix/pull/8619/files#diff-c05ae0c657bc8e08d523b61c5a6b7859d170b9fca19785593ea2ebeaefc72b66R102 that this was to prevent memory leaks (this was live-coded in https://www.youtube.com/watch?v=kHHOhPPRytc). 

I suspect `browserToolbar` may also be leaking memory since it's holding the view of `BrowserToolbarView`.

gabrielluong:
My major comment is around https://github.com/mozilla-mobile/firefox-android/pull/5846#discussion_r1515603939. Based on the historical context that I have linked, we want to be extra careful here since this is the **BaseBrowserFragment** about leaking memory when holding onto these view references. Everything else looks good to me so the final review will likely be deferred to @rocketsroger when I am away.

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
gabrielluong:
```suggestion
    internal var _bottomToolbarContainerView: BottomToolbarContainerView? = null
    
    @VisibleForTesting
    internal val bottomToolbarContainerView: BottomToolbarContainerView
        get() = _bottomToolbarContainerView!!
```
This is the change I am referring to. One thing to note is that we might want to add tests for this, otherwise, we should change `internal` to `private`

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -480,13 +481,20 @@ abstract class BaseBrowserFragment :
...
-            BottomToolbarContainerView(
+            bottomToolbarContainerView = BottomToolbarContainerView(
                 context = context,
                 container = binding.browserLayout,
                 androidToolbarView = if (isToolbarAtBottom) browserToolbar else null,
                 menuButton = menuButton,
                 browsingModeManager = activity.browsingModeManager,
             )
+
+            requireComponents.appStore.flowScoped { flow ->
```
gabrielluong, line 492: We will want to also do this via a `LifecycleAwareFeature` pattern (eg, `FindInPageIntegration`) so that we aren't calling `appStore.flowScoped` directly here.

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -210,7 +209,9 @@ abstract class BaseBrowserFragment :
     internal val browserToolbarView: BrowserToolbarView
         get() = _browserToolbarView!!
 
-    internal lateinit var browserToolbar: BrowserToolbar
```
mavduevskiy, line 213: removed it as it was causing a leak

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -467,19 +467,26 @@ abstract class BaseBrowserFragment :
...
-            ).also {
-                navbarIntegration.set(
-                    feature = it.navbarIntegration,
-                    owner = this,
-                    view = view,
-                )
-            }
+            )
+
+            navbarIntegration.set(
```
mavduevskiy, line 478: integration now is also responsible for showing/hiding the toolbar

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
mavduevskiy: addressed this, but it was more of a style problem because `bottomToolbarContainerView` reference was being cleared during onDestroy. but after some testing a found two leaks 🤦‍♂️ (`appStore.flowScope` and `browserToolbar`). addressed them

thanks for attaching the video, picked up a few things👍

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -210,7 +209,9 @@ abstract class BaseBrowserFragment :
     internal val browserToolbarView: BrowserToolbarView
         get() = _browserToolbarView!!
 
-    internal lateinit var browserToolbar: BrowserToolbar
+    private var _bottomToolbarContainerView: BottomToolbarContainerView? = null
```
rocketsroger, line 212: ❓ Not for this PR but maybe as a follow up we rename `BottomToolbarContainerView`?  Looks like it's more a `NavbarContainerView`?

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/NavbarIntegration.kt
```diff
@@ -18,17 +25,29 @@ import mozilla.components.support.base.feature.LifecycleAwareFeature
 class NavbarIntegration(
     val toolbar: ScrollableToolbar,
     val store: BrowserStore,
+    val appStore: AppStore,
+    val viewLifecycleOwner: LifecycleOwner,
+    val bottomToolbarContainerView: BottomToolbarContainerView,
```
rocketsroger, line 30: ❓ Should we remove `toolbar: ScrollableToolbar` now that we're passing in `BottomToolbarContainerView` which includes the `ToolbarContainerView`.  If so I suggest a follow up to do this.

fenix/app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserFragment.kt
```diff
@@ -66,7 +66,7 @@ class ExternalAppBrowserFragment : BaseBrowserFragment() {
             feature = CustomTabsIntegration(
                 store = requireComponents.core.store,
                 useCases = requireComponents.useCases.customTabsUseCases,
-                toolbar = browserToolbar,
+                toolbar = browserToolbarView.view,
```
rocketsroger, line 69: 🔍 nit: unrelated to this review.  Since `BrowserToolbar` is now in `BrowserToolbarView`.  Maybe we rename `browserToolbarView.view` to `browserToolbarView.toolbar`.

rocketsroger (approved):
Looks great! 🚢 
Only some naming comments that we can address is a followup if you agree.  Thanks

fenix/app/src/main/java/org/mozilla/fenix/customtabs/ExternalAppBrowserFragment.kt
mavduevskiy: filled a [follow up](https://bugzilla.mozilla.org/show_bug.cgi?id=1884455)

rvandermeulen:
@mergify refresh

PR #5895, opened 2024-03-05
Bug 1883621 - Add fade edge to long titles in tab strip

rahulsainani:
### Old - Ellipsis
ellipsis

### New - Fade
fade-edge

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1883621

### DIFF
--- a/fenix/app/src/main/java/org/mozilla/fenix/browser/tabstrip/TabStrip.kt
+++ b/fenix/app/src/main/java/org/mozilla/fenix/browser/tabstrip/TabStrip.kt
@@ -13,6 +13,7 @@ import androidx.compose.foundation.layout.BoxWithConstraints
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
@@ -32,19 +33,20 @@ import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.res.dimensionResource
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
-import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.tooling.preview.Devices
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.tooling.preview.PreviewParameter
 import androidx.compose.ui.tooling.preview.PreviewParameterProvider
 import androidx.compose.ui.unit.coerceIn
 import androidx.compose.ui.unit.dp
+import androidx.core.text.BidiFormatter
 import mozilla.components.browser.state.action.TabListAction
 import mozilla.components.browser.state.state.createTab
 import mozilla.components.browser.state.store.BrowserStore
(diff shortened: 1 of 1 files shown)


fenix/app/src/main/java/org/mozilla/fenix/browser/tabstrip/TabStrip.kt
```diff
@@ -279,22 +284,37 @@ private fun TabItem(
...
+                    backgroundColor = backgroundColor,
+                    isContentRtl = isTitleRtl,
+                ) {
+                    Text(
+                        text = state.title,
+                        color = FirefoxTheme.colors.textPrimary,
+                        maxLines = 1,
+                        softWrap = false,
+                        style = FirefoxTheme.typography.subtitle2,
+                        modifier = Modifier.align(Alignment.CenterStart),
```
gabrielluong, line 315: Could we reorder these parameters to follow the same order as they are defined https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material/material/src/commonMain/kotlin/androidx/compose/material/Text.kt;l=93-110?

rahulsainani: Sure, I've updated. I want to mention that if we want to do this, we should write a lint rule, this way won't scale. Personally I don't see the value in this as named params solve the readibility problem.

fenix/app/src/main/java/org/mozilla/fenix/browser/tabstrip/TabStrip.kt
```diff
@@ -279,22 +284,37 @@ private fun TabItem(
             horizontalArrangement = Arrangement.SpaceBetween,
         ) {
             Row(
-                modifier = Modifier.weight(1f, fill = false),
+                modifier = Modifier.weight(1f),
                 verticalAlignment = Alignment.CenterVertically,
             ) {
+                // This makes sure that isRtl is only calculated when the title changes.
```
MozillaNoah, line 290: this comment may not be super necessary after this review

fenix/app/src/main/java/org/mozilla/fenix/browser/tabstrip/TabStrip.kt
```diff
@@ -279,22 +284,37 @@ private fun TabItem(
...
+                    isContentRtl = isTitleRtl,
+                ) {
+                    Text(
+                        text = state.title,
+                        modifier = Modifier.align(Alignment.CenterStart),
+                        color = FirefoxTheme.colors.textPrimary,
+                        softWrap = false,
+                        maxLines = 1,
+                        style = FirefoxTheme.typography.subtitle2,
+                    )
```
MozillaNoah, line 316: do we need to apply a text direction here as well? I noticed we're doing it in `TabGridItem`, though it's been so long I can't remember why it was done that way.

MozillaNoah (approved):
nothing blocking here 👍. We might benefit from having an RTL preview, so we can easily validate text direction.

fenix/app/src/main/java/org/mozilla/fenix/browser/tabstrip/TabStrip.kt
rahulsainani: I tried with and without for RTL language website title and it's the same behaviour when the device locale is LTR or RTL.

Although, I think the way we show the fade edge at the end in `HorizontalFadingEdgeBox` could have edge cases where the fade is visible even when the text is actually smaller since fade is on the left side of RTL text. I'll investigate that as a separate issue.

fenix/app/src/main/java/org/mozilla/fenix/browser/tabstrip/TabStrip.kt
rahulsainani: Yeah added after to make it more understandable, guessing it was useful for the review so maybe it's useful for future readers 😄  seeing we don't have a lot of remember usages for this

fenix/app/src/main/java/org/mozilla/fenix/browser/tabstrip/TabStrip.kt
```diff
@@ -66,6 +69,7 @@ private val tabStripIconSize = 24.dp
 private val spaceBetweenTabs = 4.dp
 private val tabStripStartPadding = 8.dp
 private val addTabIconSize = 20.dp
+private val longerTitleFadeWidth = 16.dp
```
gabrielluong, line 72: Could this just be `titleFadeWidth`? I am just wondering what the 'longer' really refers to. Are we only applying a fade is longer than 16dp? I think we are just saying the fade has a width of 16dp here, and it's always just there regardless of whether or not the title overflows.

rahulsainani: Yeah, makes sense to omit "longer" 👍

PR #5926, opened 2024-03-07
Bug 1884034 - Add IllegalArgumentException for SvgIconDecoder

jackyzy823:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1884034


jackyzy823:
@t-p-white  Sorry that introduced a new bug :(

android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/decoder/SvgIconDecoder.kt
```diff
@@ -65,6 +65,8 @@ class SvgIconDecoder(val context: Context) : ImageDecoder {
             svg.renderToCanvas(Canvas(bitmap))
 
             bitmap
+        } catch (e: IllegalArgumentException) {
+            null
```
gabrielluong, line 69: It might be good to also add a logger error here

gabrielluong:
@t-p-white Could you prioritize this review and landing this during your day since it's also a crasher? Thanks!

t-p-white:
@jackyzy823 thanks for addressing so quickly! If you could add the ogger as suggested by @gabrielluong please we can get the CI running on this 🙏

android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/decoder/SvgIconDecoder.kt
rahulsainani: Instead of catching this, could we not pass the illegal argument in the affected function call?  That way we handle the edge case properly and prevent throwing an exception to catch it.

jackyzy823:
> @jackyzy823 thanks for addressing so quickly! If you could add the logger as suggested by @gabrielluong please we can get the CI running on this 🙏

Done.

android-components/components/browser/icons/src/main/java/mozilla/components/browser/icons/decoder/SvgIconDecoder.kt
jackyzy823: Coik-kt use https://github.com/coil-kt/coil/blob/e0411dd23c40dc05d69b8d05cb19b725f2d83a36/coil-svg/src/main/java/coil/decode/DecodeUtils.kt#L18-L20 to try to ensure it is a valid xml (but with no guarantee) 

https://github.com/coil-kt/coil/blob/e0411dd23c40dc05d69b8d05cb19b725f2d83a36/coil-svg/src/main/java/coil/decode/SvgDecoder.kt#L104

jackyzy823: It is so sad that kotlin don't have checked exceptions :(

t-p-white (approved):
Thanks again for fixing this so quickly @jackyzy823! 🛬

PR #5957, opened 2024-03-08
Bug 1853058 - Remove Translations Secret Settings and Use Nimbus

ohall-m:
* This patch removes the secret setting to turn on translations and removes the usage.
* Nimbus feature defaults for translations are all set to false.

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1853058


sv-ohorvath (approved):

gabrielluong:
If we remove the secret settings and rely on Nimbus, is there still a way to turn on the translation feature manually without having a running experiment (it would be acceptable to say that we will always have a running experiment if it already exists) or having to manually change the nimbus code to enable the feature? Just thinking about what options a regular user or QA would have to turn on the the translations right now without having to fiddle with the code.

ohall-m:
>I am thinking it might be too early to remove the secret setting since that is the easiest way to manually turn on/off the feature without interfering with Nimbus and making it harder to test or develop the feature.

Forgot to respond to this! There is a chance we will be running an experiment of some type on the 125 branch, so we need to remove secret settings to stop it from interfering with the experiment results. (If secret settings was a way to override the experiment, then we would have unenrolled experiment users accessing feature areas, which could skew counts.)

PR #5988, opened 2024-03-12
Bug 1879370 - Add a ClippingBehavior supporting multiple toolbars

mavduevskiy:
### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1879370


fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/RedesignToolbarFeature.kt
```diff
@@ -40,3 +43,37 @@ class IncompleteRedesignToolbarFeature(
...
         get() = settings.enableIncompleteToolbarRedesign
 }
+
+/**
+ * Returns the height of the bottom toolbar.
+ *
+ * The bottom toolbar can consist of a navigation bar,
+ * a combination of a navigation and address bar, or be absent.
+ */
+fun Context.getBottomToolbarHeight(): Int {
```
rahulsainani, line 53: 💡 Think we can move this and `getTopToolbarHeight` to `Settings.kt` 

Also, thoughts on naming this to`bottomBarsHeight` since there can be two bars at the bottom and not just toolbar?

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -1206,36 +1214,31 @@ abstract class BaseBrowserFragment :
...
-                    context,
-                    null,
-                    getSwipeRefreshLayout(),
-                    toolbarHeight,
-                    toolbarPosition,
+                    context = context,
+                    attrs = null,
+                    engineViewParent = getSwipeRefreshLayout(),
+                    topToolbarHeight = topToolbarHeight,
+                    hasTopToolbar = !requireComponents.settings.shouldUseBottomToolbar,
```
rahulsainani, line 1232: Just wondering if we can simply pass in the `bottomToolbarHeight` here instead of reading the pref. The check internally could be `topToobarHeight >0` ?

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
```diff
@@ -1206,36 +1214,31 @@ abstract class BaseBrowserFragment :
     }
 
     @VisibleForTesting
-    internal fun initializeEngineView(toolbarHeight: Int) {
+    internal fun initializeEngineView(
+        topToolbarHeight: Int,
+        bottomToolbarHeight: Int,
```
rahulsainani, line 1219: 💡 `bottomBarsHeight` Maybe we can add a doc for the function the describes the behavior and the params

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/EngineViewClippingBehavior.kt
```diff
@@ -0,0 +1,116 @@
...
+    // In such cases, as the top toolbar moves up, the bottom content clipping should be adjusted at twice the speed
+    // to compensate for the increased parent view height.
+    // However, once the top toolbar is completely hidden, the bottom content clipping should move at normal speed.
+    // Access to the latest top toolbar position allows solving this problem through a simple subtraction:
+    // val contentBottomClipping = recentTopToolbarTranslation - recentBottomToolbarTranslation
+    override fun onDependentViewChanged(parent: CoordinatorLayout, child: View, dependency: View): Boolean {
+        // I can't come up with a scenario where the translationY
+        // would return a NaN, but given that it was added to the
+        // previous version of the behaviour and there was a test
+        // written for it, I played it safe.
```
rahulsainani, line 76: Let's link that bug with the prospective fix -> https://bugzilla.mozilla.org/show_bug.cgi?id=1823306

Comment suggestion:

// Added NaN check for translationY as a precaution based on historical issues observed in [linked bug]. This check aims to prevent similar issues, as confirmed by the test. Further investigation might be needed to identify all possible causes of NaN values.

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/RedesignToolbarFeature.kt
mavduevskiy: I agree we could find a better place for the extension, but is `Settings.kt` a good fit? It doesn't have any extension functions.

It might be slightly confusing and lead readers to believe that there should be a different name for getting a single `bottomBarHeight`🤔 In context of toolbar translations, I believe it's better to think about toolbars as a top/bottom pair. From the `Behaviour` perspective, there are two `ScrollableToolbar` one at the top, one at the bottom, and sometimes they are there together. It doesn't really care about the contents, only about positioning. 
What do you think?

fenix/app/src/main/java/org/mozilla/fenix/browser/BaseBrowserFragment.kt
mavduevskiy: good point! we already are passing the `topToobarHeight`, so could be just a local parameter

fenix/app/src/test/java/org/mozilla/fenix/components/toolbar/EngineViewClippingBehaviorTest.kt
```diff
@@ -0,0 +1,284 @@
...
+
+    // Bottom toolbar position tests
+    @Test
+    fun `GIVEN the toolbar is at the bottom WHEN toolbar is being shifted THEN EngineView adjusts bottom clipping && EngineViewParent position doesn't change`() {
+        val engineView: EngineView = Mockito.spy(FakeEngineView(testContext))
+        val engineParentView: View = Mockito.spy(View(testContext))
+        val toolbar: BrowserToolbar = mock()
+        Mockito.doReturn(Y_DOWN_TRANSITION).`when`(toolbar).translationY
+
+        Assert.assertEquals(0f, engineParentView.translationY)
```
gabrielluong, line 31:
```suggestion
       doReturn(Y_DOWN_TRANSITION).`when`(toolbar).translationY

       assertEquals(0f, engineParentView.translationY)
```
Could you import the subclasses for these instead of relying on the top-level `Mockito`

fenix/app/src/test/java/org/mozilla/fenix/components/toolbar/EngineViewClippingBehaviorTest.kt
```diff
@@ -0,0 +1,284 @@
+package org.mozilla.fenix.components.toolbar
```
gabrielluong, line 1: License please

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/RedesignToolbarFeature.kt
rahulsainani: `Settings` is injected with `appContext` so we have prefs and can access resources too. In the long run, each feature should have it's own prefs but since we use Settings.kt as the only pref service, think it makes sense to put it there. 

That's a good point and perspective to look at the toolbars and is in sync with the rest of the implementation. Top/bottom pair 👍

fenix/app/src/test/java/org/mozilla/fenix/components/toolbar/EngineViewClippingBehaviorTest.kt
```diff
@@ -0,0 +1,292 @@
...
+        )
+
+        assertTrue(behavior.layoutDependsOn(mock(), mock(), BrowserToolbar(testContext)))
+        assertTrue(behavior.layoutDependsOn(mock(), mock(), ToolbarContainerView(testContext)))
+    }
+}
+
+const val TOOLBAR_HEIGHT = 100f
+const val Y_UP_TRANSITION = -42f
+const val Y_DOWN_TRANSITION = -42f
```
rahulsainani, line 292: 💡 Could be private

rahulsainani (approved):
Overall this is really good! 💯 Thanks for updating the docs!

A couple of comments about prefs and test contants!

fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/navbar/EngineViewClippingBehavior.kt
```diff
@@ -0,0 +1,114 @@
...
+
+        when (dependency) {
+            is BrowserToolbar -> {
+                if (hasTopToolbar) {
+                    recentTopToolbarTranslation = dependency.translationY
+                } else {
+                    recentBottomToolbarTranslation = dependency.translationY
+                }
+            }
+            is ToolbarContainerView -> recentBottomToolbarTranslation = dependency.translationY
```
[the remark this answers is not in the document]
mavduevskiy: That's an awesome idea! The future work should be exploring this idea in more detail✌️

rocketsroger (approved):
Looks good.  Confirmed working as advertised.  🚢

PR #6025, opened 2024-03-14
Bug 1883419 - Fix order of translation, reader mode and review quality icons in toolbar

sarah541:
For testing I was using scenarios 
1) When translation and reader mode both are available. Could be tested using [this website](https://www.mozilla.org/es)
2) When translation and review quality check both are available. Tested by changing the system language to something other than english. And then searching any product on amazon.us.

Note : Review Quality check and reader mode icons can never occur together.
To see Share and Reload icons which are always there, just enable the toolbar redesign feature in secret settings.
Attaching the screenshot from [designs](https://www.figma.com/file/8fsCvtnweBkWGZ9NfL907l/Toolbar-Redesign?type=design&node-id=5595-93037&mode=design&t=s8AHviESAWlR7oeB-0) herewith.

Screen Shot 2024-03-14 at 10 28 09 PM

### Pull Request checklist

### After merge

### GitHub Automation
https://bugzilla.mozilla.org/show_bug.cgi?id=1883419
https://bugzilla.mozilla.org/show_bug.cgi?id=1877526


android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/BrowserToolbar.kt
```diff
@@ -296,6 +297,16 @@ class BrowserToolbar @JvmOverloads constructor(
         display.addPageAction(action)
     }
 
+    /**
+     * Adds an action in order to be displayed on the right side of the URL in display mode.
```
gabrielluong, line 301:
```suggestion
     * Adds a page action in order to be displayed on the right side of the URL in display mode.
```

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -168,11 +169,11 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
                 listener = browserToolbarInteractor::onReaderModePressed,
             )
 
-        browserToolbarView.view.addPageAction(readerModeAction)
+        browserToolbarView.view.addPageActionInOrder(readerModeAction, ToolbarActionIdentifier.READER_MODE)
```
gabrielluong, line 172: Is there a way to implement this without using the identifiers by just making sure that the actions added in the correct order? For example, if we know the page action ordering should be:  
    READER_MODE,
    TRANSLATION,
    REVIEW_QUALITY_CHECK,
    SHARE,
    RELOAD,

We would just call addPageAction according to this ordering. The problem will adding these identifier into these internal AC components is that these are very generic actions and AC shouldn't have a concept of what the possible actions nor the ordering it should be. Any browser should be able to just say I have a new page action named "Blah" and be able to add it.

sarah541: Hmm, the thing is `addPageAction` isn't the only one deciding the order. The view only if visible gets added in order in the addPageAction function. However if it is not visible yet, it is just added to the list, without deciding it's ordering. The `invalidateActionActionsWithIdentifier` is the one one which gets called when a view becomes visible and hence puts the view at the right place.

ohall-m (commented):
Looks good to me from a functionality perspective and solves the translations ordering bug! Just one visual bug to take a look at!

**Visual bug I encountered, possibly a calculation issue?:**
* Have translations enabled
* Go to a Wikipedia article in another language that the system language so both the Translations and Reader Mode icon should appear
* Continually drill down to subsequent Wikipedia article links until the bug appears
   * I clicked links at a fairly quick pace, was able to reproduce consistently x3
   * The bug is the icons overlap and stay overlapped until navigation

Overlap of icon on patch

(On this patch)

**Question about transition animations:**

One question I have, that is probably out of scope for this bug, is if the redesign is going to also update the icon adding transitions? I sometimes think reader mode is performing a swoop from the right transition and sometimes it is not. Most of the other icons don't seem to have a transition and just appear. Seems like it sometimes causes brief unexpected overlaps between items.

Edit: 
Managed something similar on Main, using the same technique:

Overlap of icon on Main

(On Main)

Since it is on main too, I don't consider it a blocking issue.

Edit 2:
Adjusted the scope of [Bug 1883419](https://bugzilla.mozilla.org/show_bug.cgi?id=1883419) to just be fixing this issue, so no changes here!

fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
gabrielluong: @rocketsroger Could you take a look here and see if there is another path forward? I think we need to try to have a way where we are not including https://github.com/mozilla-mobile/firefox-android/pull/6025/files#diff-f854bc196fb764d678d6f74a529fd328a64052ff2c57bcfeeb8d480c8ba63c97R32 in the AC internals because not every browser is expected to have these page actions nor this particular ordering.

sarah541: I am going through another route to include a parameter called `weight`in the `Toolbar.Action` class. The concept behind that will be a weight will be passed every time a `pageAction` is created (which is on the consumer/fenix side). The left one will have the smallest weight and the right one the most. These will be positive integers and there will be a default weight of -1 which wouldn't have effect anything, i.e it will be added the way it is added today.

sarah541: I've updated the PR with above concept.

mavduevskiy (commented):
Looks good! some tests and refactoring suggestions!

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/internal/ActionContainer.kt
```diff
@@ -52,11 +52,39 @@ internal class ActionContainer @JvmOverloads constructor(
...
+                    addActionViewAt(it, insertionIndex)
+                }
             }
         }
 
         actions.add(wrapper)
+        actions.sortBy { it.actual.weight() }
+    }
+
+    private fun addActionViewAt(view: View, index: Int) {
```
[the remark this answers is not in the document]
sarah541: Sure I can do that. Just to note, addView calls `childCount` by default as the index which means it will it at the last.

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/internal/ActionContainer.kt
```diff
@@ -52,11 +52,39 @@ internal class ActionContainer @JvmOverloads constructor(
...
+        addView(view, index, LayoutParams(actionSize ?: 0, actionSize ?: 0))
+    }
+
+    private fun calculateInsertionIndex(newAction: Toolbar.Action): Int {
+        // If it's a default-weight action, add it at the beginning
+        if (newAction.weight() == -1) {
```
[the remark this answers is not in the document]
sarah541: Yeah I thought that too. The intention was more like to not disturb the current implementation but yeah I suppose we are already in the waters.

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/internal/ActionContainer.kt
```diff
@@ -52,11 +52,39 @@ internal class ActionContainer @JvmOverloads constructor(
...
+            return 0
```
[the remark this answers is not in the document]
sarah541: I've made it `childCount`, since it adds at the end by default. Also if it is -1, even then it would add at the end of all the children.

android-components/components/browser/toolbar/src/main/java/mozilla/components/browser/toolbar/internal/ActionContainer.kt
```diff
@@ -52,11 +52,39 @@ internal class ActionContainer @JvmOverloads constructor(
...
+        }
+        // Map existing actions to their views' indices and weights, but only if they are visible.
+        val visibleActionsWithIndices = actions.filter { it.actual.visible() }
```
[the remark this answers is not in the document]
sarah541: Possibly, yes. Just we'll have to loop through the views to get the view whose weight we found to be greater than `newAction` weight. Hence, believing my approach would be simpler.

gabrielluong:
Let's move this over to phabricator and mozilla-central for further review.

